Issue #9, opened 2021-05-17
[Tracking Issue] Support for cgroups v1


utam0k:
This issue is for tracking the implementation of [cgroups](https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt).
Since devices of cgoups has already been implemented, you can implement it while referring to it.
If you are interested, you can comment on this issue and I will assign it to you.

- [x] devices
- [x] network(assigned @tsturzl)
The integration test is this: https://github.com/opencontainers/runtime-tools/tree/master/validation/linux_cgroups_network
- [x] pids (assigned @0xdco) https://github.com/utam0k/youki/pull/26
- [x] hugepageLimits(assigned @Furisto) https://github.com/utam0k/youki/pull/15
- [x] memory (assigned @tsturzl) https://github.com/utam0k/youki/pull/16
- [x] cpu(assigned @Furisto) https://github.com/containers/youki/pull/63
The integration test is this: https://github.com/opencontainers/runtime-tools/tree/master/validation/linux_cgroups_cpus
- [x] blockIO(assigned @Furisto) https://github.com/utam0k/youki/pull/31
- [x] cpuacct (assigned @yjuba) https://github.com/containers/youki/pull/92
- [x] cpuset (assigned @Furisto) https://github.com/containers/youki/pull/63
- [x] perf_event (assigned @fbrv) https://github.com/containers/youki/pull/166
- [x] freezer (assigned @duduainankai ) https://github.com/containers/youki/pull/93

# Goal
Each of them will be targeted to pass integration tests. If possible, it is good to have unit tests as well.
It would be helpful to have a look at the GitHub Actions file to see how to run the integration tests.

# Reference
- https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#control-groups
- https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/index.html

sensaiankit:
I would like to work on this issue, but as I am a beginner i didn't understood the issue so can you please guide me

0xdco:
[...]

PR #15, opened 2021-05-21
Support hugetlb cgroup

Furisto:
Solves part of #9.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,3 +51,6 @@ jobs:
           if [ 0 -ne $(sudo RUNTIME=$GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/youki ./validation/default/default.t | grep "not ok" | wc -l) ]; then
             exit 1
           fi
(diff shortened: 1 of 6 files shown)


utam0k:
@Furisto 
`Support multiple subsystems` is good work! Thank you.
Could you please refer to this and add the integration test? https://github.com/utam0k/youki/pull/16

Furisto:
@utam0k Sure, will update soon.

tsturzl:
@Furisto I think he was more so talking about the github workflows in the path `.github/workflows/main.yml`. This runs the integration tests on a given runtime to check general compatibility with the OCI standard. However the tests you wrote are also useful!

Go look at the workflow file though, it should be as simple as copy and pasting one for a different subsystem and changing the path to reflect the hugepage subsystem. It might also be worth running the test locally to ensure.

The OCI integration tests are here: https://github.com/opencontainers/runtime-tools

The file to setup CI to run this test is here: https://github.com/utam0k/youki/blob/main/.github/workflows/main.yml

I missed it at first since it's in a hidden directory. Should be pretty simple to setup. You can also enable the tests to run on your fork, since I believe only repo members can execute tests.

Furisto:
@tsturzl Aaah, got it! You are right, I totally missed that.

utam0k (approved):
nice work!

PR #19, opened 2021-05-22
add information about the pronunciation and the etymology of youki

yuchiki:
youki is unquestionably a smart name that describes the very concept of the container.
However, it seems to be a little bit difficult for non-native Japanese speaker to guess its pronunciation and what 'youki' means.
This PR adds some explanation about the pronunciation of youki and the meanings term  'youki' carries in Japanese language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,13 @@
 
 > rusty container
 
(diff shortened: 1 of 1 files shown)


utam0k:
@yuchiki 
Thanks for you contibute!
I think it's very important to explain the origin and pronunciation of the name.
How about adding a section called `About the Name`?

yuchiki:
> How about adding a section called About the Name?
Sounds nice. I modified this PR as you said.

utam0k (approved):
nice work!!

PR #21, opened 2021-05-22
remove --experimental option from document

minakawa-daiki:
The experimental option throws up an error, so it has been fixed.

Screenshot from 2021-05-22 19-46-47


utam0k (approved):
Thasnks

utam0k (changes requested):
This logs gived you looks to be a conflict between your dockerd configuration file and the arguments.

PR #28, opened 2021-05-24
cgroups v1 cpus

reynn:
initial, will add tests shortly. pretty sure this is on the right track let me know if it looks wrong so far.

Solves part of #9.

### DIFF
--- a/src/cgroups/cpu.rs
+++ b/src/cgroups/cpu.rs
@@ -1,0 +1,88 @@
+//!
+//! Reference: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cpu
+
+use crate::{
+    cgroups::Controller,
+    rootfs::default_devices,
+    spec::{LinuxCpu, LinuxDeviceType, LinuxResources},
+};
+
+use anyhow::Result;
+use nix::unistd::Pid;
+use std::path::Path;
+
+pub struct Cpu {}
+
+impl Controller for Cpu {
+    fn apply(linux_resources: &LinuxResources, cgroup_root: &Path, pid: Pid) -> Result<()> {
+        fs::create_dir_all(cgroup_root)?;
+
+        if let Some(cpu) = &linux_resources.cpu {
+            Self::apply(linux_resources, cgroup_root, pid)
+        }
+
+        OpenOptions::new()
+            .create(false)
+            .write(true)
+            .truncate(false)
+            .open(cgroup_root.join("cgroup.procs"))?
+            .write_all(pid.to_string().as_bytes())?;
+
+        Ok(())
+    }
+}
+
+impl Cpu {
+    fn apply_cpu(cpu: &LinuxCpu, cgroup_root: &Path) -> Result<()> {
+        if let Some(cpus) = cpu.cpus {
+            // validate
+            // write to cpuset.cpus
+            Self::write_file(cgroup_root.join("cpuset.cpus"), cpus.to_string())?;
+        }
+        if let Some(mems) = cpu.mems {
+            // validate
+            // cpuset.mems
+            Self::write_file(cgroup_root.join("cpuset.mems"), &mems.to_string())?;
+        }
+        if let Some(quota) = cpu.quota {
+            // cpuset.quota
+            Self::write_file(cgroup_root.join("cpuset.quota"), &quota.to_string())?;
+        }
+        if let Some(period) = cpu.period {
+            // cpuset.period
+            Self::write_file(cgroup_root.join("cpuset.period"), &period.to_string())?;
+        }
+        if let Some(realtime_runtime) = cpu.realtime_runtime {
+            // cpuset.realtime_runtime
+            Self::write_file(
+                cgroup_root.join("cpuset.realtime_runtime"),
+                &realtime_runtime.to_string(),
+            )?;
+        }
+        if let Some(realtime_period) = cpu.realtime_period {
+            // cpuset.realtime_period
+            Self::write_file(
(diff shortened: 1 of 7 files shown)


src/cgroups/pids.rs
```diff
@@ -1,4 +1,4 @@
-use std::{
+euse std::{
```
tsturzl, line 1: I think you missed a small typo :)

reynn: thanks for that!

utam0k:
@reynn 
nice work!  Could you please refer to this PR and add the corresponding integration tests in the same way? This is optional, but it would be great if you could add unit tests as well.
https://github.com/utam0k/youki/pull/16

src/cgroups/cpu.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use anyhow::Result;
+use nix::unistd::Pid;
+use std::path::Path;
+
+pub struct Cpu {}
+
+impl Controller for Cpu {
+    fn apply(linux_resources: &LinuxResources, cgroup_root: &Path, pid: Pid) -> Result<()> {
+        fs::create_dir_all(cgroup_root)?;
```
tsturzl, line 18: CI is failing to build this. You're using `fs::...` but you never imported it `use std::fs;`.

src/cgroups/cpu.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        if let Some(cpu) = &linux_resources.cpu {
+            Self::apply(linux_resources, cgroup_root, pid)
```
tsturzl, line 21: Is this method declared? It looks like you have method `apply_cpu`. Otherwise it looks like you're trying to recursively declare the method that's being implemented, and I don't think this is allowed in a trait implementation. Inside a trait implementation calls to `Self::` methods won't consider the trait being implemented since it's not considered to be implemented yet as you're currently implementing it. Also this isn't really what you probably want to be doing, you should be invoking your `apply_cpu` method here.

src/cgroups/cpu.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        OpenOptions::new()
```
tsturzl, line 24: OpenOptions is also not imported, so this throws a compiler error.

src/cgroups/cpu.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl Cpu {
+    fn apply_cpu(cpu: &LinuxCpu, cgroup_root: &Path) -> Result<()> {
+        if let Some(cpus) = cpu.cpus {
+            // validate
+            // write to cpuset.cpus
+            Self::write_file(cgroup_root.join("cpuset.cpus"), cpus.to_string())?;
+        }
+        if let Some(mems) = cpu.mems {
```
tsturzl, line 42: This is also just a `String` not an `Option<String>`

src/cgroups/cpu.rs
```diff
@@ -0,0 +1,88 @@
...
+            // validate
+            // cpuset.mems
+            Self::write_file(cgroup_root.join("cpuset.mems"), &mems.to_string())?;
+        }
+        if let Some(quota) = cpu.quota {
+            // cpuset.quota
+            Self::write_file(cgroup_root.join("cpuset.quota"), &quota.to_string())?;
```
tsturzl, line 49: Same here

src/cgroups/cpu.rs
```diff
@@ -0,0 +1,88 @@
...
+            Self::write_file(
+                cgroup_root.join("cpuset.realtime_period"),
+                &realtime_period.to_string(),
+            )?;
+        }
+        Ok(())
+    }
+
+    fn write_file(file_path: &Path, data: &str) -> anyhow::Result<()> {
+        fs::OpenOptions::new()
```
tsturzl, line 73: You don't import `std::fs` at any point so this fails

src/cgroups/manager.rs
```diff
@@ -40,6 +41,7 @@ impl Manager {
                 "hugetlb" => Hugetlb::apply(linux_resources, &subsys.1, pid)?,
                 "memory" => Memory::apply(linux_resources, &subsys.1, pid)?,
                 "pids" => Pids::apply(linux_resources, &subsys.1, pid)?,
+                "cpu" => Cpu::apply(linux_resources, &subsys.1, pid)?,
```
tsturzl, line 44: You imported the enum used for dispatch, but you didn't actually import your controller. So this is not actually declared in scope.

tsturzl (changes requested):
Currently there are some errors which prevent this from compiling. It might be helpful to enable use of the rust language server(rls) within whatever editor or IDE you use. Also writing some basic unit tests, having them pass, as well as building the project and getting it to pass tests locally with the runtime-tools is important for verifying the work. I hope I'm not being to critical here, but this doesn't compile and `cargo check` should give you a pretty good outline of what's going wrong.

src/cgroups/cpu.rs
tsturzl: Also you use both `fs::OpenOptions` and `OpenOptions`, neither are currently imported, but it would be cleaner to choose one or the other. Personally I prefer just importing `std::fs::OpenOptions` and using it without the `fs::` module prefix, especially since you invoke it several times.

reynn:
Thanks @tsturzl this isn't yet complete, will work on it tonight and do the corrections and add testing.

reynn:
hey @utam0k and @tsturzl sorry but I dont have time to finish this for at least a few weeks, work has gotten a bit crazy.

PR #31, opened 2021-05-24
Blkio cgroup support

Furisto:
Implements blkio cgroup support #9. Not implemented at the moment are weight, leaf_weight, weight_device and leaf_weight-device as these have been removed by the kernel in version 5.0 (https://github.com/docker/cli/pull/2908) and will lead to an error trying to write these files. 
On the other hand the runtime tests have not been updated to reflect this yet and will fail because they expect these files. I have tested it for now with my own fork (https://github.com/Furisto/runtime-tools). 

How should we proceed here?

### DIFF
--- a/src/cgroups/blkio.rs
+++ b/src/cgroups/blkio.rs
@@ -1,0 +1,64 @@
+use std::{fs::{self, OpenOptions}, io::Write, path::Path};
+
+use crate::{cgroups::Controller, spec::{ LinuxBlockIo, LinuxResources}};
+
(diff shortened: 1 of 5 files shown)


utam0k:
I see. It's ok! 
It's great that you noticed that, and I recommend you to submit a PR to runntime-tools.
If runtime-tools supports it, it would be great to get a PR here as well.
>  I have tested it for now with my own fork (https://github.com/Furisto/runtime-tools).

utam0k (approved):
nice work!

PR #34, opened 2021-05-24
cgroup v1 networking

tsturzl:
This is a work in progress. I'm having problems with the CI integration tests, I seem to be getting different results locally than on CI. I might need some help digging in here. It's hard to know exactly what the environment of the CI looks like, perhaps it's running in some kind of container without network capabilities? It seems the CI failure is happening when trying to find the path for a given subsystem. I'm not completely sure which subsystem. See: https://github.com/tsturzl/youki/runs/2653607002

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
           test_cases=("default/default.t" "linux_cgroups_devices/linux_cgroups_devices.t" "linux_cgroups_hugetlb/linux_cgroups_hugetlb.t" "linux_cgroups_pids/linux_cgroups_pids.t", "linux_cgroups_memory/linux_cgroups_memory.t", "linux_cgroups_network/linux_cgroups_network.t")
           for case in "${test_cases[@]}"; do
             title="Running $case"
-            if [ 0 -ne $(sudo RUNTIME=$GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/youki ./validation/$case | grep "not ok" | wc -l) ]; then
+            if [ 0 -ne $(sudo RUST_BACKTRACE=1 RUNTIME=$GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/youki ./validation/$case | grep "not ok" | wc -l) ]; then
               exit 1
             fi
           done
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
           expect_err_num=8
           act_err_num=0
           cd $(go env GOPATH)/src/github.com/opencontainers/runtime-tools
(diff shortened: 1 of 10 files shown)


utam0k:
Thank you for your research, I know it was hard work.
How about trying to run CI with runc for example and see if the tests pass?
If it doesn't work, it's a problem with the CI environment, so you can skip the integration tests and only use unit tests.
> It's hard to know exactly what the environment of the CI looks like, perhaps it's running in some kind of container without network capabilities?

tsturzl:
I found the issue. It's going to take a bit of work. So some systems mount the net_prio and net_cls separately, which it what's happening on my machine, however more commonly they get mounted into the same directory that gets names "net_cls,net_prio" or "net_prio,net_cls". So we have to come up with a way of looking for those mount points. This same issue probably affects CPU subsystem too, the mount points could be separate directories or one big directory. I think for now I may commit an easy work around to check for all possibilities when either subsystem mount point is being searched for.

tsturzl:
@utam0k I squashed a lot of my commits down, and rebased off master to handle any merge conflicts. Give it a look at let me know what you think. The changes I made to `get_subsystem_path` aren't the most elegant, but it should handle the issue where systems are mounting subsystems at different paths without getting overly complicated. Those changes are found here https://github.com/utam0k/youki/pull/34/files#diff-ccab580a34cdb0bdfa6edc80e529ec9aa154936f614632ffda3b4d94bb452146R76

.github/workflows/main.yml
```diff
@@ -45,10 +45,10 @@ jobs:
           expect_err_num=8
           act_err_num=0
           cd $(go env GOPATH)/src/github.com/opencontainers/runtime-tools
-          test_cases=("default/default.t" "linux_cgroups_devices/linux_cgroups_devices.t" "linux_cgroups_hugetlb/linux_cgroups_hugetlb.t" "linux_cgroups_pids/linux_cgroups_pids.t", "linux_cgroups_memory/linux_cgroups_memory.t")
+          test_cases=("default/default.t" "linux_cgroups_devices/linux_cgroups_devices.t" "linux_cgroups_hugetlb/linux_cgroups_hugetlb.t" "linux_cgroups_pids/linux_cgroups_pids.t", "linux_cgroups_memory/linux_cgroups_memory.t", "linux_cgroups_network/linux_cgroups_network.t")
           for case in "${test_cases[@]}"; do
             title="Running $case"
-            if [ 0 -ne $(sudo RUNTIME=$GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/youki ./validation/$case | grep "not ok" | wc -l) ]; then
+            if [ 0 -ne $(sudo RUST_BACKTRACE=1 RUNTIME=$GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/youki ./validation/$case | grep "not ok" | wc -l) ]; then
```
utam0k, line 51: 👍

utam0k (approved):
nice work!

PR #36, opened 2021-05-26
fix a memory subsystem

utam0k:
I ran it with Docker and youki at local, and it didn't work because it tried to set 0, so I fixed it.
https://github.com/opencontainers/runc/blob/3f6594675675d4e88901c782462f56497260b1d2/libcontainer/cgroups/fs/memory.go#L40-L42

### DIFF
--- a/src/cgroups/blkio.rs
+++ b/src/cgroups/blkio.rs
@@ -140,7 +140,7 @@ mod tests {
     fn setup(testname: &str, throttle_type: &str) -> (PathBuf, PathBuf) {
         let tmp = create_temp_dir(testname).expect("create temp directory for test");
         let throttle_file = set_fixture(&tmp, throttle_type, "")
-            .expect(&format!("set fixture for {}", throttle_type));
+            .unwrap_or_else(|_| panic!("set fixture for {}", throttle_type));
 
         (tmp, throttle_file)
     }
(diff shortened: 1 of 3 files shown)


src/cgroups/blkio.rs
```diff
@@ -140,7 +140,7 @@ mod tests {
     fn setup(testname: &str, throttle_type: &str) -> (PathBuf, PathBuf) {
         let tmp = create_temp_dir(testname).expect("create temp directory for test");
         let throttle_file = set_fixture(&tmp, throttle_type, "")
-            .expect(&format!("set fixture for {}", throttle_type));
+            .unwrap_or_else(|_| panic!("set fixture for {}", throttle_type));
```
tsturzl, line 143: What is the reason for using and unwrap_or_else for panicking instead of using expect? Shouldn't these be basically the same thing?

src/cgroups/memory.rs
```diff
@@ -158,16 +161,16 @@ impl Memory {
                         Errno::EBUSY => {
                             let usage = Self::get_memory_usage(cgroup_root)?;
                             let max_usage = Self::get_memory_max_usage(cgroup_root)?;
-                            Err(anyhow!(
+                            bail!(
```
tsturzl, line 164: good catch, I haven't use anyhow, but this looks much cleaner.

tsturzl (commented):
Just curious about the unwrap_or_else instead of expect, otherwise looks good.

src/cgroups/blkio.rs
utam0k: @tsturzl According to cargo clippy, it's probably to prevent the format from being called every time.
https://rust-lang.github.io/rust-clippy/master/#expect_fun_call

utam0k: If you have no problem with this PR, please approve.

tsturzl (approved):

PR #37, opened 2021-05-26
extract the integration tests writen in the ci file as a script file.

utam0k:


utam0k:
It's not a problem in CI, but in my local, the memory of cgroups failed.
For now, I want to capture this PR in main.

tsturzl:
I'm wondering if it might be worthwhile to setup local tooling for testing against multiple different Linux distributions. This work might actually help with this. I'll outline that work in a new ticket.

.github/workflows/main.yml
```diff
@@ -33,22 +33,5 @@ jobs:
...
-          act_err_num=0
-          cd $(go env GOPATH)/src/github.com/opencontainers/runtime-tools
-          test_cases=("default/default.t" "linux_cgroups_devices/linux_cgroups_devices.t" "linux_cgroups_hugetlb/linux_cgroups_hugetlb.t" "linux_cgroups_pids/linux_cgroups_pids.t", "linux_cgroups_memory/linux_cgroups_memory.t", "linux_cgroups_network/linux_cgroups_network.t")
-          for case in "${test_cases[@]}"; do
-            title="Running $case"
-            if [ 0 -ne $(sudo RUST_BACKTRACE=1 RUNTIME=$GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/youki ./validation/$case | grep "not ok" | wc -l) ]; then
-              exit 1
-            fi
-          done
+        run: ./integration_test.sh
```
tsturzl, line 37: Where does the build happen for the integration tests now?

integration_test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
tsturzl, line 1: This script fails locally for me because the tests haven't been compiled. Also setting up testing might be worth a short addition to the README.md

tsturzl (changes requested):
Unless I'm missing something, I don't think this works since the integration tests are never build before executing. Probably adding a short instruction to the README.md would be a good idea. Also probably need to outline that in order to run the tests locally you need to have golang compiler and node-tap dependencies installed.

integration_test.sh
tsturzl: So the integration test passing was actually a false positive, if you look at it the script failed in CI also. I think the script itself must not be returning the exit code correctly. So maybe this is another thing to fix, making sure the integration tests fail if anything returns a non-zero code, perhaps this is as easy as adding `set -e` to the script? See the CI failure here: https://github.com/utam0k/youki/pull/37/checks?check_run_id=2673869152#step:6:18

utam0k:
@tsturzl 
Thanks for the advice. Your comments are very informative. Please wait a moment as I will respond after I finish my work today.

.github/workflows/main.yml
utam0k: @tsturzl The build of the integration test is included in this script.

utam0k:
@tsturzl 
Your advice has been reflected in this PR. Please review it again.

tsturzl (approved):
Tested, and it works locally

tsturzl (changes requested):
Actually, take a look at the CI. The CI is still failing, which might not be an issue concerning this ticket, however the CI on this PR says it succeeded when it really didn't. The exit code of the script is being set explicitly instead of returning the error code of the failed command. If you add `set -e` to the top of integration_test.sh` that should fix the issue and it should error out of the bash script with the proper exit code.

utam0k:
@tsturzl Thank you for being so cautious.
This part of the code is the `set -e` part. When I added the test cases that should actually fail, CI failed.
https://github.com/utam0k/youki/pull/37/files#diff-d2da4d075c191015bceebd47b3e6229755e65bed685c0cdd2b07a08343b9e49cR1
What probably looks like a failure in CI is the error log of the test that should fail (e.g., container creation fails with invalid value).
So it looks like this is probably working correctly, what do you think?

tsturzl:
@utam0k I agree, it looks like as far as this PR is concerned things look good. However I'm going to file a bug report concerning this failure hopefully tonight after dinner when I have time to read through it a little more.

tsturzl (approved):

utam0k:
@tsturzl 
Thank you for your review.
Bug reports are very helpful. I am looking forward to it. Have a good dinner time :)
> However I'm going to file a bug report concerning this failure hopefully tonight after dinner when I have time to read through it a little more.

tsturzl:
I think I totally misunderstood you before. I thought the error messages were failures and the exit codes were maybe getting caught somewhere. My mistake. I don't believe there to be a bug, I just missed what you said:

> What probably looks like a failure in CI is the error log of the test that should fail

utam0k:
@tsturzl In fact, your point wasn't wrong, since the CI went through even if the first commit failed.
Please keep up the nice reviews.

PR #40, opened 2021-05-27
Update Rust-Analyzer in Dockerfile

nalpine:
Hi all, 

I noticed while checking out Youki that Rust-Analyzer was failing to install. The project did not release an uncompressed binary with the latest release. Based on this [Pull Request](https://github.com/rust-analyzer/rust-analyzer/pull/8926) it seems the project will not be uploading uncompressed binaries going forward.

I did a quick edit to the docker file so it now points toward the latest  `rust-analyzer-x86_64-unknown-linux-gnu.gz`  and uses gzip to decompress the binary in `/usr/local/cargo/bin/rust-analyzer`.


tsturzl (approved):
Good catch!

PR #41, opened 2021-05-27
revert asynchronous devices mounting.

utam0k:
I had previously tried to mount asynchronously, but this made our integration tests unstable.
I reverted it once, but it was still there, so I deleted it.

### DIFF
--- a/src/create.rs
+++ b/src/create.rs
@@ -157,21 +157,19 @@ fn init_process(
     namespaces: Namespaces,
 ) -> Result<()> {
     let proc = spec.process.clone();
-    let clone_spec = std::sync::Arc::new(spec);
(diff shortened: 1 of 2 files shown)


utam0k:
@tsturzl You may not have read much of the code around here, but please give it a light review if you like.

tsturzl (approved):
Looks good. I'm curious what might have been causing the issues. Any ideas? I've noticed in the async work I'm doing that I have had to make sure I call sync on the writer. I wouldn't have thought I'd have to do this when writing to a virtual filesystem, so maybe there is something not so obvious happening?

utam0k:
@tsturzl 
When I ran it with docker in my local, it worked normally, but I could only find out that it failed in the integration test.
If you have any problems with cgroups asynchronous support, I should seriously look into it.

PR #42, opened 2021-05-28
get oci_spec in seperate crate

### LINKED ISSUE #33: Extract spec.rs as a crate
The [spec.rs](https://github.com/utam0k/youki/blob/92fa6bbda72137f0262283c5c8bd5c57c4dac5c4/src/spec.rs) that reads the oci-spec specification is a general-purpose code that can be used not only in this project.
Therefore, I would like to extract and provide it as a separate crate from youki in this repository.
Ideally, the youki itself should use the crate.
If anyone is interested in this, I would like to challenge you.

ferrell-code:
fixes #33 :)


utam0k (approved):
great work!

Issue #14, opened 2021-05-19
More comments and docs


utam0k:
Now, This project has too few comments and documentation.
I would like to improve the development experience by improving them.

utam0k:
@YJDoc2  
Can I assign this issue to you? Of course, I'm going to help.

YJDoc2:
Yes sure, I'd be happy to work on this :+1: I am a bit busy this week, so probably will start working on this weekend or next week.
Currently my plan is to start with main.rs, comment that and recursively work on the imports. I am also thinking to make a rough md file to describe the control flow as the program is run, so someone new can get a quick high level overview of things. Let me know if you have/get any suggestions or ideas. :)

utam0k:
I assigned this issue to you. Your plan looks great and I am looking forward to it!

I prefer to have them proceed slowly at their own pace. Can I ask you to send out PRs in small sections so I can see the progress if possible?
> I am a bit busy this week, so probably will start working on this weekend or next week.

This is a diagram I made to explain the process flow of youki to others. It is a wrong diagram because the process is different from now, but it may be helpful to see the big flow.
![image](https://user-images.githubusercontent.com/13010913/118834813-6ffcb080-b8fd-11eb-8fc4-7c42d5b5ac6d.png)

utam0k:
@YJDoc2 
I'd like to invite you to join this repository in a little while, with you in charge of maintaining the documentation. Maybe I'll ask for other ranges as well.
If you're up to the challenge, I'd love to work with you.

YJDoc2:
Hey, I think it'd be better if you see more of my commits/PRs and then decide, I have made only 1 PR till now :sweat_smile: 
But I would like to work on this for sure. Thank for giving the chance , I am getting to learn a lot about containers and Linux by doing this :)

YJDoc2:
I had a question about the code, in load_console_socket, what is exactly the console part? I understood the socket for the tty from given socket path; but in second half of function, fcntl open call is given only "console", what is it meant to open? I tried it in C, but it always returned -1.

If possible can you comment and give outerview for the spec file? I tried to go through it and understand, but it is a bit complex, so rather than commenting every struct, can you create a file-level comment describing what it is meant for and what various structs in it are meant to represent?

Also for the new files and PRs, it would be great if authors themselves comment it in PR, so that it'll best explain author's intent, and only original files will need to be commented, otherwise, the commenting work will keep on going as more and more files/changes are added.
Thank you!

[...]

PR #43, opened 2021-05-28
Add comments to create.rs

YJDoc2:
Part of #14 
Please review @utam0k 
I had a doubt that after fork_init in run_container, the Process::Child value of enum should be unreachable ,similar to Process::Parent ;  as fork_init only returns in child process with return value of Process::Init.
Thank you.


utam0k:
@YJDoc2 Thanks for your question!
When the fork(2) system call is invoked, the code execution is divided between the parent process and the child process.
So, after fork_init() is called, Child and Process will be executed in parallel.
fork(2) is called here. When this is called, it will split into child and parent. Therefore, fork(2) is unsafe.
https://github.com/utam0k/youki/blob/37243cd695cd27c4b252721f770d8718662ca914/src/process/fork.rs#L77

ref. https://docs.rs/nix/0.20.0/nix/unistd/fn.fork.html

YJDoc2:
Sorry, maybe my question was a little ambiguous, what I wanted to ask is: 
In fork init function, there are two outcomes after fork in child process or in parent process,
https://github.com/utam0k/youki/blob/37243cd695cd27c4b252721f770d8718662ca914/src/process/fork.rs#L77-L80
In case of child, it returns Process::Init, and in case of parent it waits till child changes state, and exits from there.
https://github.com/utam0k/youki/blob/37243cd695cd27c4b252721f770d8718662ca914/src/process/fork.rs#L83-L93

So when it is called from run_container in create.rs :
https://github.com/utam0k/youki/blob/37243cd695cd27c4b252721f770d8718662ca914/src/create.rs#L132
it can only return Process::Init value of enum, and the child value should be unreachable as well, like the parent value :
https://github.com/utam0k/youki/blob/37243cd695cd27c4b252721f770d8718662ca914/src/create.rs#L133-L146
In which case, the first line should also have an unreachable!() instead of returning ok()

Hope this makes it more clear.

utam0k:
@YJDoc2 
I understand your question. Since fork function calls `exit()`, so this code is unreachable.
You are right, the fork function is supposed to just create the process, but now it is doing a lot of processing. I think you should refactor the process you are doing in the fork function part to move it to `create.rs`.
Am I able to answer your questions?

YJDoc2:
@utam0k 
Hey, I have updated the code in create.rs, I'll try refactoring the code in fork, as I will start commenting it. I think it will not provide substantial benefit from refactoring the code from fork_init into create.rs, as the snippet is small, and with that removed, it will be just a normal fork call. 
Please check the comments and merge if ok, so I'll start working on process files.
Thanks.

src/create.rs
```diff
@@ -41,16 +58,20 @@ impl Create {
...
 
+        // change directory to the bundle directory, and load configuration,
+        // copy that to the container's directory
         unistd::chdir(&self.bundle)?;
 
         let spec = oci_spec::Spec::load("config.json")?;
         fs::copy("config.json", container_dir.join("config.json"))?;
         log::debug!("spec: {:?}", spec);
 
+        // convert path to absolute path
```
utam0k, line 69: It might be better to indicate the reason here.
The reason for converting to an absolute path is that a path relative to where the youki command is executed will be difficult to manipulate using that relative path when chdir is used.

src/create.rs
```diff
@@ -125,22 +161,29 @@ fn run_container<P: AsRef<Path>>(
...
                 tty::ready(csocketfd)?;
             }
 
+            // set namespaces
             namespaces.apply_setns()?;
 
+            // fork second time, which will later create container
             match fork::fork_init(child)? {
-                Process::Child(child) => Ok(Process::Child(child)),
+                Process::Child(child) => unreachable!(),
```
utam0k, line 174: The variable `child` is not used in this line. Therefore, the `child` variable needs to be renamed to `_child`.

utam0k (commented):
@YJDoc2 
I added some comments. Please look at them.

YJDoc2:
Hey, I have updated the requested changes. Thanks.

utam0k (approved):
awesome!

PR #48, opened 2021-05-29
Initial support for cgroups v2

Furisto:
This is the initial support for cgroups v2. Currently either cgroups v1 or cgroups v2 are supported (no hybrid mode of v1 and v2).  If you want to test this you need cgroups v2 which you can enable by adding systemd.unified_cgroup_hierarchy=1 to your kernel cmdline if you do not already have it enabled. 

Cpu and CpuSet controller are already implemented, rest can be implemented by someone else if there is interest. This is not yet completely finished, I have created some common functions that are used in the new stuff but the v1 cgroups need still be migrated to it. I am also investigating some weird behavior where child processes escape the cgroup.

Still I would like to merge this now and the rest of the work will be done in other PRs.


utam0k:
@Furisto 
This is a cool PR. When I looked at the CI, it seemed that the integration test was failing.
Is this dependent on the CI environment? Is there any way to fix it?
> Error: cannot get available controllers. "/sys/fs/cgroup/cgroup.controllers" does not exist

utam0k:
Could you apply the format?
https://github.com/rust-lang/rustfmt

Furisto:
I think what is happening is that cgroup v1 and cgroup v2 are mounted on the CI machine and that it returns CgroupV2 as filesystem type when it stats /sys/fs/cgroup. Thats why it tries to use /sys/fs/cgroup with v2 but it is actually located in /sys/fs/cgroup/unified. 
Anyway the runtime-tools tests support only v1 as test target so I changed it to use v1 for the moment.

utam0k:
@Furisto 
I think that in the future Youki should have our own integration test for this. For now, there aren't  integration tests for v2, so there isn't need to add integration tests.
> Anyway the runtime-tools tests support only v1 as test target so I changed it to use v1 for the moment.

How about having markers (environment variables, etc.) to use only v1?

Furisto:
@utam0k 
Sure, I have introduced an environment variable to control the behavior when both v1 and v2 are present. It will default to v1 as long as the v2 support is experimental. Later I will invert it to prefer v2.

utam0k:
@Furisto 👍 
> Later I will invert it to prefer v2.
Do I look okay to start a review?

Furisto:
@utam0k 
Yes, it should be fine now to review.

src/cgroups/v2/io.rs
```diff
@@ -0,0 +1,12 @@
+use anyhow::Result;
+
+use super::controller::Controller;
+use oci_spec::LinuxResources;
+
+pub struct Io {}
+
+impl Controller for Io {
+    fn apply(linux_resources: &LinuxResources, cgroup_path: &std::path::Path) -> Result<()> {
```
utam0k, line 9: As in the other cases, you can get rid of this warning by starting unused variables with `_`.

utam0k (approved):
I made one small comment of correction, but if you could fix that, it would be perfect.

PR #52, opened 2021-05-31
cargo clippy.

utam0k:


Furisto:
@utam0k 
What is the advantage of unwrap_or_else over expect?

utam0k:
@ferrell-code 
 According to the documents of cargo clippy, this improvement probably to prevent the format! from being called every time.
https://rust-lang.github.io/rust-clippy/master/#expect_fun_call

Furisto (approved):

PR #54, opened 2021-06-01
Align cgroup controller implementations

Furisto:
There are multiple identical write file helper methods throughout the cgroup controller implementations. This PR replaces them with the write file helper method from the common module. Sometimes files were truncated before being written to, but this should not be necessary for cgroup files.

It also ensures that the container pid is written to the cgroup.procs file regardless if the controller has to apply resource restrictions or not. Some controllers did this, some did not. That way the cgroup can be used to track the resource consumption which some monitoring tools might rely on. It is also consistent with the behavior of runc.

### DIFF
--- a/src/cgroups/common.rs
+++ b/src/cgroups/common.rs
@@ -38,18 +38,8 @@ impl Display for Cgroup {
     }
 }
 
-pub fn write_cgroup_file_truncate(path: &Path, data: &str) -> Result<()> {
-    fs::OpenOptions::new()
-        .create(false)
-        .write(true)
-        .truncate(true)
-        .open(path)?
-        .write_all(data.as_bytes())?;
-
-    Ok(())
-}
-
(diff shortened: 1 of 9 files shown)


utam0k:
How did you notice that? Excellent.
> It also ensures that the container pid is written to the cgroup.procs file regardless if the controller has to apply resource restrictions or not. Some controllers did this, some did not. That way the cgroup can be used to track the resource consumption which some monitoring tools might rely on. It is also consistent with the behavior of runc.

utam0k (approved):
lgtm

tsturzl:
Sorry, I've been out of town for the holiday here in the US. I was going to suggest that perhaps the most reusable thing would be to make the write function take any type that implements ToString `<T: ToString>`. It would be mostly aesthetic, and not an important change. I believe the truncate option was added because of a caveat with how were running tests. Tests might benefit from a temp dir type that deletes the directory when the temp dir drops out of scope using the drop trait, but that's outside the context of this PR.

PR #55, opened 2021-06-02
Adds spec cli command

ferrell-code:
thought it would be useful to be able to generate the config.json natively like runc :)

Also might be useful for unit tests


utam0k:
@ferrell-code 
Thanks for your PR. I think this is a good addition. For example, I felt it would be nice to use this in the tutorials in the README.
If you'd like, you can put a simple test on it.

ferrell-code:
I wanted to get Spec to fully serialize and de-serialize, unfortunately I'm not sure how to manually impl it for the case of LinuxCapabilityType https://github.com/containers/youki/blob/adb43102e022d05c4730fa95e431b30050dec7d8/oci_spec/src/lib.rs#L94-L105

ferrell-code:
other than that let me know what you think

utam0k:
@ferrell-code 
This deserializing code could be written more concisely by using rename_all. Like this.
https://github.com/containers/youki/blob/adb43102e022d05c4730fa95e431b30050dec7d8/oci_spec/src/lib.rs#L73
Wouldn't that also make serialization easier?

Also, you can reference the railcar's code.
https://github.com/oracle/railcar/blob/ef5918e21e7ad9ffd25c1a507df96458ec4e0c24/oci/src/lib.rs#L86-L127

ferrell-code:
I changed LinuxCapabilityType to enum. Thanks for the railcar reference, I think that it is the most sensible solution, unless there is a reason it is a struct I'm unaware of

oci_spec/src/caps_migrations.rs
```diff
@@ -0,0 +1,92 @@
...
+        LinuxCapabilityType::CAP_LEASE => Capability::CAP_LEASE,
+        LinuxCapabilityType::CAP_AUDIT_WRITE => Capability::CAP_AUDIT_WRITE,
+        LinuxCapabilityType::CAP_AUDIT_CONTROL => Capability::CAP_AUDIT_CONTROL,
+        LinuxCapabilityType::CAP_AUDIT_READ => Capability::CAP_AUDIT_READ,
+        LinuxCapabilityType::CAP_SETFCAP => Capability::CAP_SETFCAP,
+        LinuxCapabilityType::CAP_MAC_OVERRIDE => Capability::CAP_MAC_OVERRIDE,
+        LinuxCapabilityType::CAP_MAC_ADMIN => Capability::CAP_MAC_ADMIN,
+        LinuxCapabilityType::CAP_WAKE_ALARM => Capability::CAP_WAKE_ALARM,
+        LinuxCapabilityType::CAP_BLOCK_SUSPEND => Capability::CAP_BLOCK_SUSPEND,
+        _ => panic!("not a Linux Capability Type"),
```
Furisto, line 45: You can use unreachable! here instead of panic! to communicate the intent a bit better. I am wondering though, is this even necessary? If you exhaustively check for all variants of the enum, a wildcard pattern should not be needed.

oci_spec/src/caps_migrations.rs
```diff
@@ -0,0 +1,92 @@
...
+        Capability::CAP_LEASE => LinuxCapabilityType::CAP_LEASE,
+        Capability::CAP_AUDIT_WRITE => LinuxCapabilityType::CAP_AUDIT_WRITE,
+        Capability::CAP_AUDIT_CONTROL => LinuxCapabilityType::CAP_AUDIT_CONTROL,
+        Capability::CAP_AUDIT_READ => LinuxCapabilityType::CAP_AUDIT_READ,
+        Capability::CAP_SETFCAP => LinuxCapabilityType::CAP_SETFCAP,
+        Capability::CAP_MAC_OVERRIDE => LinuxCapabilityType::CAP_MAC_OVERRIDE,
+        Capability::CAP_MAC_ADMIN => LinuxCapabilityType::CAP_MAC_ADMIN,
+        Capability::CAP_WAKE_ALARM => LinuxCapabilityType::CAP_WAKE_ALARM,
+        Capability::CAP_BLOCK_SUSPEND => LinuxCapabilityType::CAP_BLOCK_SUSPEND,
+        _ => panic!("not a Linux Capability type"),
```
Furisto, line 90: See above

oci_spec/src/caps_migrations.rs
```diff
@@ -0,0 +1,92 @@
+use caps::Capability;
+use crate::LinuxCapabilityType;
+
+// utility function to switch between the types LinuxCapabilityType and Capabality
+pub fn to_caps(local_cap: &LinuxCapabilityType) -> Capability {
```
utam0k, line 5: It is better to use the `Into` trait.
https://doc.rust-lang.org/std/convert/trait.Into.html

oci_spec/src/caps_migrations.rs
ferrell-code: Altered it a bit to be fully explicit :+1: . the caps dependency has a wildcard though

ferrell-code:
I think it would be better to fork the caps dependency and host it in its own crate in youki to avoid the migrations back and forth and modify it as we see fit

oci_spec/src/caps_migrations.rs
ferrell-code: I agree but we would need to be able to impl the `From` trait for `caps::Capability` which cannot be done with an external crate. 

See idea below to fix this :)

utam0k:
Ok, do you want to try it?
> Perhaps that should be done in a different PR

oci_spec/src/caps_migrations.rs
ferrell-code: I must of been sleepy when I looked at it last night you can totally `impl From` for an external crate! updated :+1:

ferrell-code:
I'd be happy to do it if you think it is necessary, but looking at it again that might be overkill. Let me know what you think of the current PR :)

utam0k:
@ferrell-code 
I also feel a little overwhelmed. I think it's not too late to deal with it when it becomes more necessary.
I'm sorry I can't provide the opportunity. If you are up for some challenge, please look at the cgroups v2 support and so on.

I'm planning to do a careful review again after Draft comes off, but for now, I think it's good overall.
I'm going to do a careful review again when the Draft is off, but for now, I think it's good overall. However, it would be nice to comment on why Default is needed (perhaps to be used in the spec command).
Also, although it doesn't have to be in this PR, `lib.rs` is too big, and It is better to split it.

ferrell-code:
I think its a great project for us to learn more about containers and linux internals! Take as much time as you need, I know its not easy to have time for opensource work. We appreciate your effort!

My thought process behind Default was to have a basic boilerplate for `Spec`, but if you think it is more idiomatic for `Spec` to be a function instead of the Default trait i can easily copy paste the Default into that. 

I will split the lib.rs file too but probably after this PR would be easier :sake:

utam0k:
@ferrell-code 
It encourages me to hear you say that. If you want, you can continue to contribute to the project.
> I think its a great project for us to learn more about containers and linux internals! Take as much time as you need, I know its not easy to have time for opensource work. We appreciate your effort!

I think your idea is a good one.
In any case, in my opinion, it would be better to have some comments on the background.
> My thought process behind Default was to have a basic boilerplate for Spec, but if you think it is more idiomatic for Spec to be initialized with some fn init_spec() instead of the Default trait i can easily copy paste the Default into that.

👍 
> I will split the lib.rs file too but probably after this PR would be easier 🍶

ferrell-code:
Ok I added a lot of documentation to it, let me know what you think

oci_spec/src/lib.rs
```diff
@@ -22,265 +37,491 @@ pub struct Box {
...
 pub enum LinuxDeviceType {
+    // block (buffered)
     B,
+    // character (unbuffered)
     C,
+    // character (unbufferd)
     U,
+    // FIFO
     P,
+    // ??
```
ferrell-code, line 524: Do you know what A is for? and why does it default to A (It already had a default impl)

oci_spec/src/lib.rs
```diff
@@ -22,265 +37,491 @@ pub struct Box {
...
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq)]
 #[serde(rename_all = "lowercase")]
 pub enum LinuxDeviceType {
+    // block (buffered)
     B,
+    // character (unbuffered)
     C,
+    // character (unbufferd)
     U,
+    // FIFO
```
ferrell-code, line 522: Do you know what A is for? It defaults to A (previous to this pr)

oci_spec/src/lib.rs
```diff
@@ -22,265 +37,491 @@ pub struct Box {
...
     P,
+    // ??
     A,
 }
```
ferrell-code, line 526: Do you know what A is for? It defaults to A (previous to this pr)

utam0k: Sorry It's my mistake. Could you like to remove it?
https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices

Furisto:
@utam0k @ferrell-code This looks fine to me. What is us preventing from merging this? (apart from the minor merge conflicts that have accumulated in the mean time)

utam0k:
@ferrell-code Do you have any implementation left? Other than conflicts, I think this is fine.

ferrell-code:
Ah yes sorry for leaving this hanging. It was all good to go, I resolved the conflicts and split lib.rs into multiple files. Unfortunately right now it breaks something related to LinuxCapabilities... so do not merge, I am investigating

Furisto:
@ferrell-code Can you squash some of the commits? I am talking about the commits from June 10 to June 14 which are basically all "Comments & Formatting" and from June 23 to today which are mostly "Resolve conflicts & compile"

utam0k:
Could you remove the `WIP` in the title of this PR if you're ready to merge?

ferrell-code:
I squashed it all to one commit, because rebase would make me resolve all the conflicts. I think this is the most reasonable solution.

In any future prs I will keep in mind to squash before pushing :+1:

Furisto (approved):

Furisto:
Looks very good to me. If @utam0k also agrees we can do the merge.

utam0k (approved):
Perfect!

utam0k:
@Furisto  thanks for your review

PR #58, opened 2021-06-02
Update README.md

aoki:
- Change comment out sign `//` to `#` in shell script
- Add new line at end of file

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ For other platforms, please use the devcontainer that we prepared.
 $ git clone git@github.com:utam0k/youki.git
 $ cd youki
 $ ./build.sh
-$ ./youki -h // you can get information about youki command
+$ ./youki -h # you can get information about youki command
 ```
 
 ## Tutorial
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,14 +54,14 @@ $ mkdir tutorial
 $ cd tutorial
 $ mkdir rootfs
 $ docker export $(docker create busybox) | tar -C rootfs -xvf -
-// Prepare a configuration file for the container that will run `sleep 5`.
+$ # Prepare a configuration file for the container that will run `sleep 5`.
```
utam0k, line 57: There are only comments in this line, have you any reason to put `$` at the beginning?

utam0k:
@aoki 
Hi, aoki.
Thank you for your PR. This is nice work. I have added some comments if you would like to see them.

README.md
aoki: I think sometimes confuse `stdout` when not add `$` mark ahead (or execute on root `#`).

Another approach is to remove new line and join before command.
Like this, 
```sh
$ docker export $(docker create busybox) | tar -C rootfs -xvf -  # Prepare a configuration file for the container that will run `sleep 5`.
```
Which do you like? I think both okay.

utam0k: That's certainly a good perspective.
> I think sometimes confuse stdout when not add $ mark ahead (or execute on root #).

I think it's fine as it is now.  If you don't have time, this PR will be very good to merge as it is.
If you like, how about dividing some commands into meaningful units (e.g. build/prepare for container execution / execute container)? This way you could add a light description to each separated section to erase this commented-out description.

aoki: Thank you for your suggestion. Okay, I change to separate some section. Please wait a moment.

utam0k (approved):
lgtm

PR #63, opened 2021-06-04
Support for cgroup v1 cpu and cpuset subsystem

Furisto:
This implements the cpu and cpuset controller for cgroup v1 #9 . I also implemented @tsturzl suggestion from a previous PR to accept all types that implement ToString for write_cgroup_file. If you have already a String or &str you can use write_cgroup_file_str instead and avoid the String allocation. Lastly I noticed that cgroups are sometimes not cleaned up properly, so this hopefully improves the situation a bit.

### DIFF
--- a/src/cgroups/v1/controller_type.rs
+++ b/src/cgroups/v1/controller_type.rs
@@ -2,6 +2,7 @@ use std::string::ToString;
 
 pub enum ControllerType {
     Cpu,
+    CpuSet,
     Devices,
     HugeTlb,
     Pids,
@@ -15,6 +16,7 @@ impl ToString for ControllerType {
     fn to_string(&self) -> String {
         match self {
             Self::Cpu => "cpu".into(),
+            Self::CpuSet => "cpuset".into(),
             Self::Devices => "devices".into(),
             Self::HugeTlb => "hugetlb".into(),
             Self::Pids => "pids".into(),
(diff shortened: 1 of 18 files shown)


tsturzl:
I will review within the hour

src/cgroups/v1/controller_type.rs
```diff
@@ -13,6 +15,8 @@ pub enum ControllerType {
 impl ToString for ControllerType {
     fn to_string(&self) -> String {
         match self {
+            Self::Cpu => "cpu".into(),
```
tsturzl, line 18: Confusingly some systems have the CPU subsystem and CPU accounting subsystem mounted at the same path, or what's often called comounting. I may be missing something, but I don't believe we account for that. See the section **Mounting v1 Controllers** here: https://man7.org/linux/man-pages/man7/cgroups.7.html

tsturzl (changes requested):
I believe we need to account for the fact that certain cgroup controllers can be comounted into the same path for certain subsystems. In this case cpu can have a path ending either in "cpu" or "cpu,cpuacct", possibly even in some cases "cpuacct,cpu". This is similar for the network controller, and I had thought we had added in a case to check for comounted network subsystems. Admittedly it was sloppy, I was just hard coding additional checks when we searched the cgroup mounts. Maybe we need to figure out a better way to handle these comounted subsystems.

src/cgroups/v1/controller_type.rs
Furisto: I check for this in the cgroup manager when we are constructing the subsystem paths.

tsturzl: I totally missed that somehow. Looks good.

tsturzl (approved):

Issue #39, opened 2021-05-26
Multisystem testing


tsturzl:
The integration tests can pass or fail depending on the system it's actually running on, one example of this is the network cgroups. On some systems the network cgroups mount into 2 different directories "net_prio" and "net_cls", on others they mount into one directory "net_cls,net_prio" or "net_prio,net_cls". I'd assume this is probably specific to how distributions are setup. So it might be worth while to setup some kind of testing in 3-4 of the most popular distributions. These tests would be setup locally only, since our current CI cannot handle it at the moment.

My thoughts on which systems to test is:

- Redhat/Fedora
- Ubuntu LTS
- Latest stable Debian
- Alpine Linux, this is a good candidate to also test a system that doesn't use systemd

This test tooling will likely use a VM, something like vagrant might be most useful here. I don't believe testing container runtimes in a container will work all that well. Another choice might be something like Amazon's Firecracker.

In addition to providing multiple test environments this could also allow Windows and Mac users to more easily test changes.

YJDoc2:
Hey @tsturzl , can you tell me what direction you have in mind for this? I was working on #82 and when trying to run tests, the cgroup memory limit tests failed, as the kernel I was using was not compiled with that option. But aside from just failing the test, I am now not much confident in making changes required for #82, as I cannot test on my local machine, I might introduce changes that might break the youki in a place I didn't expect, and I would have no way of testing that other than pushing to github and running the action on my fork. I tested https://github.com/nektos/act which lets you test github actions locally, but that didn't work out for me either as somehow the configuration selected by it for the run does not match the one used by github and it gives issues for running rust and go commands.

If you have any idea how we can not only extend the testing to multiple systems, but also somehow allow to test integration locally, I would like to check that out and see if I can make any progress, as resolving/improving this issue will give all contributors confidence that the changes they have made does not break the system.

One though I had was to convert the OCI test in rust, and use those as cargo tests, but I think that would be a huge project by itself, and also we will need to keep up with the changes that are made in the OCI test, which will be extra work. If you can suggest some other ways, maybe making container configuration using docker compose etc to test this which can be used locally as well as on multiple systems, please let me know.

Thanks.

tsturzl:
@YJDoc2 I did get the base systems mostly setup with the exception of alpine Linux on my fork. Unfortunately I haven't had the time to finish this yet. I'm hoping maybe tomorrow evening I'll have some focus time for this.

Issue #17, opened 2021-05-22
Asynchronous cgroup controller configuration


tsturzl:
We should be able to concurrently configure each of the  cgroup subsystems to hopefully see some performance improvements. One thing to note is write order within a controller is important to keep track of since values can be validated against each other by the kernel. Initially it may be best to write configs in each controller in series and have each controller be executing concurrently with the next, and eventually move into more granular concurrency.

utam0k:
I think this is a wonderful thing to be able to achieve. I would love to verify and implement this.

tsturzl:
I'll probably be building out some kind of profiling and benchmarking on this to see what kind of improvements this feature brings. I've got a lot of the initial work done for making the manager async. I need to do some work to get the unit tests working again. Hoping to have the integration tests running on this by tonight.

utam0k:
I was trying it out at my local computer, and I am very aware that cgroups is making things run slower than before.
Therefore, I still think that improving performance by async/await will be quite an interesting challenge.
For benchmarking, here may be helpful.
https://github.com/TaKO8Ki/frum/pull/16

tsturzl:
I'm going to reattempt this. I found out that the actual implementation of almost all async runtimes for handling file operations is to just move file IO off into a thread pool. I naively assumed it chose between different kernel APIs depending on the build targets. I think in our case this isn't awesome. The cost of starting a threadpool diminishes and possibly outright negates the potential performance improvements of concurrent cgroups configuration. I think what I'd like to do is approach this from a lower level. It would be an interesting experiment to make a small program to see if the AIO or epoll work well for interacting with a cgroup controller, otherwise lib_uring might address all of the possible caveats and dead ends in the other async IO kernel features. As far as I'm aware AIO and epoll are both incapable of doing the more typical buffered IO, which some virtual filesystems support exclusively. If the cgroups vfs handles direct IO well then AIO or epoll might be preferable simply for the fact that older kernel version support them, while io uring is relatively new and only available in kernel version 5.1 or greater.

[...]

PR #65, opened 2021-06-06
[WIP] async cgroups

tsturzl:
I'm putting this up as a _work in progress_ since I'm having trouble understanding some of the test failures. I'm running Void Linux(glibc kernel 5.10.15) on my machine and locally I'm seeing integration tests pass, however on CI I'm seeing a lot of them fail. If I revert one of the subsystems to using non-async filesystem interactions I don't see the failures on CI, but it seems to be affecting many of the controllers, if not all of them. Some interesting details here is that when I was working on this locally the integration tests were failing prior to adding in an explicit `sync_data` call on the `File`. This makes me assume that perhaps there's some caveats I'm not aware of in the virtual filesystem in the kernel. I hope to try to read some of the source code in the kernel to better understand what's going on here, but decided to create this PR to hopefully get some more eyes on this. Maybe someone has a better idea of what's going on here? Or perhaps someone who can get these failures to happen on their machine can provide some more insight? For now I'm planning to hopefully start work on ticket #39 which could provide the tooling I need to dig deeper into this, and hopefully spending some of my free time learning more about how the cgroups vfs works under the hood, since I'm aware that current generation of async filesystem interactions have some caveats such as that most async IO is actually direct not buffered and some virtual filesystems do not support this well.

Also please forgive the mess of commits here. I thought I could get CI working with a couple of quick changes, but alas I'm stumped. This feature branch is also trailing behind main branch quite a ways now, and I will rebase main in soon, but I'm hoping to wait until the underlying issue is resolved. I'm inclined to think this problem is addressable since this passing tests for me locally. Let me know what you think.


Furisto:
I will test this on my machine this week. Have you benchmarked this @tsturzl ? What is the performance improvement?

tsturzl:
@Furisto I have not setup any kind of profiling or benchmarking yet as it's not currently passing CI tests. I'm actually now aware that most async frameworks, including the one I'm using(`smol`), are actually pushing blocking IO off into a threadpool. I naively assumed most of them relied on something like IOCP or AIO, but this isn't the case. So my assumption is that the threadpool startup costs will likely outweigh the performance gains of concurrent IO in our case. I think I'm going to revisit this and possibly have to do some of the low level work myself. Even looking at Tokio, many of it's file operations get pushed off into a background threadpool.

I've done some work with [rio](https://github.com/spacejam/rio) which relies on the io_uring features in kernel 5.1, and then I'm familiar with poll, aio, epoll, etc but they come with a lot of caveats, mainly that they only support direct IO and not buffered IO which I'm uncertain if that will work with the cgroups vfs.

tsturzl:
Closing the PR. I'm going to reattempt this with some more experimental approaches. Will comment details in #17

PR #67, opened 2021-06-06
Fixed spelling mistake in src/rootfs.rs

PeterYordanov:


utam0k (approved):
I'm not good at English, so this kind of correction is very helpful.

PR #69, opened 2021-06-06
Fix issues with cgroup v1 and v2

Furisto:
- The child pid was used instead of the init pid during create for the apply operation of the cgroup manager. This meant that resource restrictions were not properly applied and it was possible to consume more resources than specified. This affected cgroup v1 and v2.
- Cgroup controllers were not written to subtree_control at cgroup root level, which caused 'file does not exist' errors. This affected only v2.

### DIFF
--- a/src/process/fork.rs
+++ b/src/process/fork.rs
@@ -11,6 +11,7 @@ use init::InitProcess;
 use nix::sched;
 use nix::sys::wait::{waitpid, WaitStatus};
 use nix::unistd;
+use nix::unistd::Pid;
 
 use crate::cgroups::common::CgroupManager;
 use crate::container::ContainerStatus;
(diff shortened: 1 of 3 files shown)


utam0k:
I'll do a review of this after work. Please wait for a while.

src/process/fork.rs
```diff
@@ -62,11 +63,11 @@ pub fn fork_first<P: AsRef<Path>>(
             unistd::ForkResult::Parent { child } => {
                 ccond.wait()?;
 
-                // apply the control group to the child process
-                cmanager.apply(&linux.resources.as_ref().unwrap(), child)?;
-
                 // wait for child to fork init process and report back its pid
                 let init_pid = parent.wait_for_child_ready()?;
+                log::debug!("init pid is {:?}", init_pid);
+                cmanager.apply(&linux.resources.as_ref().unwrap(), Pid::from_raw(init_pid))?;
```
utam0k, line 69: 👍

utam0k (approved):
This is a great bug fix.

Issue #56, opened 2021-06-02
impl: original integration test.


utam0k:
Youki is currently using the integration tests from rutime-tools, but there are some areas where this isn't sufficient (e.g. cgv2), so we will implement our own.

minakawa-daiki:
@utam0k I'd like to try it.

utam0k:
Sure! I assigned you to it. I'm looking forward your PR.

minakawa-daiki:
@utam0k 

I would like to report that I was able to reproduce the container lifecycle of the tutorial on rust's `cargo test`, although it is still only in my repository.

And I have a few concerns that I would like to discuss with you.

- I'm planning to run the test by using `sudo` to run the binary generated by `cargo test --no-run`, but is that OK?
- I think the container created by `youki create` will be created in `/run/youki`, is there any problem to create it there when running the test?
- I'm wondering how to delete the file created in `/run/youki` when the test fails.

As for the future roadmap, I would like to proceed as follows.

1. start with `youki create`, write a test to see if `youki start`, `youki state`, `youki kill`, `youki delete`, etc. work properly, and I'll submit it as a pull request.

2. I will implement the integration test with reference to https://github.com/opencontainers/runtime-tools/tree/master/validation

3. I will add tests that runtime-tools does not seem to be sufficient.

Also, I would like to know if there are any high priority tests.

YJDoc2:
Hey @minakawa-daiki ,sorry for delay in reply. If it is possible, can you try implementing the validation tests first? As once we have those, we can test the changes made on local system, currently the runtime-tools validation works in github action only, see discussion in #82 .

@utam0k can you take a look at the concerns?  I think it would be fine to allow root requirements for running the tests, but it would be better to leave docket export part out of the tests, as it would force us to depend on docker being correctly configured and also the network status. It might be better to check if the rootfs folder exists in the root dir of project, and fail the test if it does not.

utam0k:
@YJDoc2 Thanks for your mention.

utam0k:
I think it is inevitable that you will need root privileges. I agree with the assumption that root privileges are required, although it is possible that rootless could be used to get around this.

As for rootfs, why not mimic runtime-tools and zip it up and put it in the repository first?
> I'm planning to run the test by using sudo to run the binary generated by cargo test --no-run, but is that OK?
The tutorial uses docker export $(docker create busybox) | tar -C rootfs -xvf - to create a rootfs, what should I do here in my test

[...]

PR #71, opened 2021-06-06
Added Integration test template

minakawa-daiki:
I have added a template for future implementation of integration tests (#56) in rust.
To run it, use `cargo test --test integration`.

Screenshot from 2021-06-06 22-36-07

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,14 +1,16 @@
 use std::env;
 use std::process::{Command, Stdio};
+use std::path::PathBuf;
 
 #[test]
 fn main() {
-    let path_result = env::current_dir();
-    let path = match path_result {
-        Ok(path) => path.display().to_string(),
-        Err(_) => panic!("Path is not found"),
+    let current_dir_path_result = env::current_dir();
+    let current_dir_path = match current_dir_path_result {
+        Ok(path_buf) => path_buf,
+        Err(_) => panic!("directory is not found"),
     };
-    let status = Command::new(path + "/target/x86_64-unknown-linux-gnu/debug/youki")
+    let youki_path = current_dir_path.join(PathBuf::from("youki"));
+    let status = Command::new(youki_path)
         .stdout(Stdio::null())
         .arg("-h")
         .status()
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,0 +1,17 @@
+use std::env;
+use std::process::{Command, Stdio};
+
+#[test]
+fn main() {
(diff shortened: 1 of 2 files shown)


tests/integration.rs
```diff
@@ -0,0 +1,17 @@
+use std::env;
+use std::process::{Command, Stdio};
+
+#[test]
+fn main() {
+    let path_result = env::current_dir();
+    let path = match path_result {
+        Ok(path) => path.display().to_string(),
```
utam0k, line 8: Can you use PathBuf?

tests/integration.rs
```diff
@@ -0,0 +1,17 @@
...
+        Err(_) => panic!("Path is not found"),
+    };
+    let status = Command::new(path + "/target/x86_64-unknown-linux-gnu/debug/youki")
```
utam0k, line 11: If you want to use the pre-built one, why not simply merge the paths with youki?

utam0k (changes requested):

tests/integration.rs
minakawa-daiki: https://github.com/containers/youki/blob/adb43102e022d05c4730fa95e431b30050dec7d8/integration_test.sh#L9

I was just following your direct specification here, but if you want to just merge the youki paths, I'll do that!

minakawa-daiki:
@utam0k How about a fix based on your review?

tests/integration.rs
utam0k: I think you just need to connect it to the `youki` path. That code is a bit out of date.

utam0k:
@minakawa-daiki Ok, source codes look good. Why don't you add a README? How about stating there that it is a preview implementation and how to run it?

minakawa-daiki:
@utam0k I added it to the README.md

utam0k (approved):
nice work!

utam0k (changes requested):
sorry, the ci is failed. can you fix it?

minakawa-daiki:
@utam0k 
https://github.com/containers/youki/pull/71#discussion_r646139293
It seems that youki Path pointed out here does not exist and CI has been dropped. I think the solution is to add `"/target/x86_64-unknown-linux-gnu/debug/youki"` to the path before you pointed it out, what do you think?

minakawa-daiki:
@utam0k The CI has been passed, please check.

utam0k (approved):
It's a good first step.

utam0k (approved):

PR #74, opened 2021-06-07
Handle relative cgroup paths

Furisto:
This handles relative cgroup paths. Relative cgroup paths will be appended to the cgroup root, similar to absolute cgroup paths. Also handles a problem where cpuset parent values were not set, which lead to a permission denied error in the tests.


Furisto:
I am looking into why the CI is failing

tsturzl (approved):
Looks good! It would be nice to have some testing around these changes, but seems out of scope for this given we don't have any testing for managers yet.

PR #79, opened 2021-06-10
Document Container and Command modules

YJDoc2:
This adds comments to container and command modules, for #14 , does not have major changes in container, as most of container module was already commented.


src/container/container.rs
```diff
@@ -36,10 +39,12 @@ impl Container {
     pub fn status(&self) -> ContainerStatus {
         self.state.status
     }
-
     pub fn refresh_status(&self) -> Result<Self> {
         let new_status = match self.pid() {
             Some(pid) => {
+                // Note that Process::new does not spawn a new process
+                // but instead creates a new Process structure, and fill
+                // it with information about the process with given pid
```
utam0k, line 47: This is a good comment. I, the author, didn't think anything of it, but it certainly could be misunderstood.

src/command/linux.rs
```diff
@@ -22,36 +23,60 @@ use oci_spec::LinuxRlimit;
...
+        // open the path as directory and read only
         let newroot = open(path, OFlag::O_DIRECTORY | OFlag::O_RDONLY, Mode::empty())?;
 
+        // make the given path as the root directory for the container
+        // see https://man7.org/linux/man-pages/man2/pivot_root.2.html, specially the notes
+        // pivot root usually changes the root directory to first argument, and then mounts the original root
+        // directory at second argument. Giving same path for both stacks mapping of the original root directory
+        // above the new directory at the same path, then the call to umount unmounts the original root directory from
+        // this path. This is done, as otherwise, we will need to create a separate temporary directory under the new root path
+        // so we can move the original root there, and then unmount that. This way saves the creation of the temporary
```
utam0k, line 48: :+1:

utam0k (approved):
lgtm

src/container/container.rs
YJDoc2: Yes, the first time I was looking at this while commenting create.rs, I was wondering why we create a new process here, then I checked the docs of process and found out it fills new structure. So I kept that in mind when writing this :)

PR #84, opened 2021-06-11
Provide better error messages

Furisto:
This adds the file name to the error message when we cannot write to a cgroup file.
Before: 
`No such file or directory (os error 2)`

After:
```
failed to open "/sys/fs/cgroup/blkio/docker/9d96321423f09dffdf2a9559971408d29c984bfb17746589ee83670f5da06f70/foo"

Caused by:
```

### DIFF
--- a/src/cgroups/common.rs
+++ b/src/cgroups/common.rs
@@ -6,7 +6,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{anyhow, bail, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use nix::unistd::Pid;
 use oci_spec::LinuxResources;
 use procfs::process::Process;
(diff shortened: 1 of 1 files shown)


src/cgroups/common.rs
```diff
@@ -45,8 +45,10 @@ pub fn write_cgroup_file_str<P: AsRef<Path>>(path: P, data: &str) -> Result<()>
         .create(false)
         .write(true)
         .truncate(false)
-        .open(path)?
-        .write_all(data.as_bytes())?;
+        .open(path.as_ref())
+        .with_context(|| format!("failed to open {:?}", path.as_ref()))?
+        .write_all(data.as_bytes())
+        .with_context(|| format!("failed to write to {:?}", path.as_ref()))?;
```
tsturzl, line 51: Sometimes the OS error is indicative of cgroups errors, for example if you write an invalid value to a cgroup you'll get an "Invalid Argument" error. So it might be nice not to loose that context entirely. Perhaps you can just print the underlying error out in the formatting?

tsturzl (changes requested):
Overall I like this. I had similar thoughts, but I had more so thought about trying to make sense of the OS error. Like the example I provided earlier, if you write "asdf" to the `memory.limit_in_bytes` file you'll get a "Invalid argument" error, which we can reasonably deduce is an kernel validation error on the cgroup configuration. There may be more like this, and perhaps addressing all the possible OS error to cgroup error translations is beyond the scope of this PR, but I think it might still be nice to get the underlying error included in the error you're producing.

src/cgroups/common.rs
tsturzl: I just realized in reading the anyhow docs that this will still print out the original error! Always read the docs!

tsturzl (approved):

tsturzl:
I read the anyhow docs and realized that this still prints the underlying error and only provides additional context. The PR looks good to me.

PR #85, opened 2021-06-11
Clean up use of unsafe

tsturzl:
I primarily did 2 things here. First in `src/process/fork.rs` I limited the scope of the unsafe to only wrap the call to `fork`, since this is the only actually unsafe thing happening, that way we don't lose compile-time safety on anything happening in the larger scope. Secondly I pulled out some of the libc stuff used to set the environment variables. While the format of the environment variables is already in a format better suited to libc, the rust standard library has all of these without the need to use `unsafe`. This effectively reduces our unsafe usage from 6 to 4 occurrences, and removes ~70 lines of code from within an unsafe block.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,12 +1,13 @@
 //! Utility functionality
 
+use std::env;
 use std::ffi::CString;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
 use anyhow::{bail, Result};
-use nix::{env::clearenv, errno::Errno, unistd};
+use nix::unistd;
 
 pub trait PathBufExt {
     fn as_in_container(&self) -> Result<PathBuf>;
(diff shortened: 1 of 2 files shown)


src/utils.rs
```diff
@@ -40,29 +41,25 @@ pub fn do_exec(path: impl AsRef<Path>, args: &[String], envs: &[String]) -> Resu
...
-        .collect();
 
-    unsafe {
-        clearenv()?;
-    }
-    for e in envs {
-        putenv(&e)?
-    }
+    // clear env vars
+    env::vars().for_each(|(key, _value)| std::env::remove_var(key));
```
utam0k, line 46: I'm questioning myself why I was using libc. I must have been sleepy.

utam0k (approved):
I was planning to make these improvements myself, little by little, but it is very helpful to have someone else's eyes to see and fix them.

src/utils.rs
tsturzl: Libc seemed obvious probably because you were already using it. Honestly the code was a little more concise too since libc takes "key=value" as one string and rust takes key and value as separate strings so I had to split on "=".

PR #89, opened 2021-06-14
clean up around the tty.

utam0k:
I think the tty area was probably a point that the developers were wondering about, so I commented on the organization and difficulties.

### DIFF
--- a/src/cgroups/v2/manager.rs
+++ b/src/cgroups/v2/manager.rs
@@ -102,7 +102,7 @@ impl Manager {
         Ok(controllers)
     }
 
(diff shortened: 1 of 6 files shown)


constreference:
Thank you for this, had been wanting to ask you regarding the socket creation logic, this clean up makes it a lot better

Furisto (approved):

Furisto:
Thanks for this! I liked that you included the reference to containerd. I had wondered about how it is handled from containerd side.

Furisto:
lgtm

PR #91, opened 2021-06-16
make sure to log any unimplemented controllers.

utam0k:
It's a small improvement.

### DIFF
--- a/src/cgroups/v2/manager.rs
+++ b/src/cgroups/v2/manager.rs
@@ -4,7 +4,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{bail, Result};
 
 use nix::unistd::Pid;
 use oci_spec::LinuxResources;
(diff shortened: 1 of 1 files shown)


src/cgroups/v2/manager.rs
```diff
@@ -123,6 +123,10 @@ impl CgroupManager for Manager {
                 ControllerType::Io => Io::apply(linux_resources, &&full_cgroup_path)?,
                 ControllerType::Memory => Memory::apply(linux_resources, &full_cgroup_path)?,
                 ControllerType::Pids => Pids::apply(linux_resources, &&full_cgroup_path)?,
+                ControllerType::Unknown(tpe) => {
+                    log::warn!("Controller {} is not yet implemented.", tpe);
+                    continue;
```
Furisto, line 128: The 'continue' is not necessary

Furisto: Actually this never gets called, because CONTROLLER_TYPE is a constant list of values.

utam0k: You're right! I changed it to a simpler way of implementation.

Furisto (approved):

PR #98, opened 2021-06-18
Experimental support for rootless containers

Furisto:
This introduces experimental support for rootless containers. To test this:
```
// create spec for rootless containers
runc spec --rootless
``` 

```
// provide console socket, 
recvtty tty.sock
```

```
// in another shell, note the missing sudo
./youki create --bundle <bundlePath> --console-socket <SocketPath>/tty.sock 2309pasf
./youki start 2309pasf
```

```
// in container shell
id
uid=0(root) gid=0(root) groups=65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),65534(nobody),0(root)
```

```
// on the host, note the user, it is not root
furisto    41875    1562  0 21:48 pts/8    00:00:00  \_ ./youki create --bundle...
furisto    41876   41875  0 21:48 pts/8    00:00:00      \_ /bin/sh
```

Known limitations for now:
- Resource restrictions will be ignored. This requires systemd support
- Only single id mapping fully supported


utam0k:
I'll review it later today.

src/rootless.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+/// Checks if rootless mode should be used
+pub fn should_use_rootless() -> Result<bool> {
+    if !nix::unistd::geteuid().is_root() {
+        return Ok(true);
+    }
+
+    if let Ok("true") = std::env::var("YOUKI_USE_ROOTLESS").as_deref() {
```
utam0k, line 38: Since rootless works for now, let's output a warn log that there is an unimplemented feature when this function is `true`.

utam0k: Isn't it also necessary to have `uidMapping` and `gidMapping` in the config?

src/rootless.rs
```diff
@@ -0,0 +1,121 @@
...
+fn lookup_map_binary(binary: &str) -> Result<Option<PathBuf>> {
+    let paths = env::var("PATH")?;
+    for p in paths.split_terminator(':') {
+        let binary_path = PathBuf::from(p).join(binary);
+        if binary_path.exists() {
+            return Ok(Some(binary_path));
+        }
+    }
+
+    Ok(None)
```
utam0k, line 120:
```suggestion
    Ok(paths
        .split_terminator(':')
        .find(|p| PathBuf::from(p).join(binary).exists())
        .map(PathBuf::from))
```

utam0k (commented):
@Furisto 
I made a couple of comments, but it worked on my enviroment. It's amazing!
The `fork/parent.rs` file is too big and may need to be refactored soon, even if it's not this PR. This is an early design mistake on my part.

src/rootless.rs
Furisto: Yes, this is checked in the validate() function (L50-56)

Furisto:
@utam0k Addressed your comments

src/rootless.rs
utam0k: @Furisto Is it persistent to switch rootless with environment variables? Or is it temporary?

Furisto: I used it during development. I can remove it if you want.

utam0k (approved):

utam0k:
🎉

PR #100, opened 2021-06-19
prepare Vagrant instead of devcontainer for platforms other than linux.

utam0k:
Use vagrant because container in container is difficult for security reasons.


YJDoc2:
Hey, I think it would be good idea to include information that something like virtualbox needs to be also installed for vagrant to work, I was trying to test this on linux, and at least on it, vagrant needs virtualbox already installed.

That said, this is meant for non-linux OSs so someone running such OSs might give a better opinion.

utam0k:
@YJDoc2 I wasn't sure if I should include it in the README, but I didn't think it was necessary to support it in the README. Do you think about?

Vagrantfile
```diff
@@ -0,0 +1,30 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+    config.vm.box = "fedora/33-cloud-base"
+    config.vm.provider :virtualbox do |v|
```
Furisto, line 6: On Windows Hyper-V is the most likely virtualization solution to be available

tsturzl:
I had worked on something similar, mostly with the intention of addressing #39. I found that it was difficult to get the integration tests working in these VMs. Often times the swap accounting features were disabled in the kernel which causes the runtime-tools to outright fail even when testing runc. I hadn't figured out a way to address this yet without doing a lot of setup steps to modify kernel boot params, which quickly became cumbersome. I had fooled around on this branch: https://github.com/tsturzl/youki/tree/ts-multi-system-testing

I ultimately put this work aside awaiting a response on https://github.com/opencontainers/runtime-tools/issues/721

Vagrantfile
```diff
@@ -0,0 +1,30 @@
...
+      yum install -y git gcc docker
+      grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
+      service docker start
+    SHELL
+
+    config.vm.provision "shell", privileged: false, inline: <<-SHELL
+      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+      echo "export PATH=$PATH:$HOME/.cargo/bin" >> ~/.bashrc
+
+      git clone https://github.com/containers/youki
```
tsturzl, line 28: Why clone youki in rather than use the synced folder? I get that you might end up sharing build artifacts between platforms, but if you're using vagrant as your dev environment that shouldn't be too common of a problem. I think with this it just make it so you have to do all of your work in the VM rather than just the compilation steps.

tsturzl (changes requested):

Vagrantfile
utam0k: @tsturzl 
I believe that Vagrant is just for people who want to try youki out. That's why I wanted to leave out as some things as possible that might make that errors due to differences between platforms, such as sync.

utam0k:
@tsturzl @Furisto @YJDoc2 PTAL

utam0k:
> I had worked on something similar, mostly with the intention of addressing #39. I found that it was difficult to get the integration tests working in these VMs. Often times the swap accounting features were disabled in the kernel which causes the runtime-tools to outright fail even when testing runc. I hadn't figured out a way to address this yet without doing a lot of setup steps to modify kernel boot params, which quickly became cumbersome. I had fooled around on this branch: https://github.com/tsturzl/youki/tree/ts-multi-system-testing
> 
> I ultimately put this work aside awaiting a response on [opencontainers/runtime-tools#721](https://github.com/opencontainers/runtime-tools/issues/721)

I have seen your branch. It is great. It would be better if runtime-tools could solve this problem, but they don't seem to be very active. I guess we'll just have to deal with it on youki side.

tsturzl (approved):

PR #102, opened 2021-06-20
Add unit tests for tty module

constreference:
This PR addresses part of https://github.com/containers/youki/issues/8, by adding unit tests for the tty module.
It adds two dev-dependencies, tempfile for easing the creation of temporary directories and serial_test to force the tests to execute in serial, since we require setting the current working directory to the testdir for each test, since console-socket is expected to be in this current working directory

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,4 @@ oci_spec = { version = "0.1.0", path = "./oci_spec" }
 [dev-dependencies]
 oci_spec = { version = "0.1.0", path = "./oci_spec", features = ["proptests"] }
 quickcheck = "1"
+serial_test = "0.5.1"
--- a/src/tty.rs
+++ b/src/tty.rs
@@ -68,3 +68,79 @@ pub fn setup_console(console_fd: FileDescriptor) -> Result<()> {
     close(console_fd.as_raw_fd())?;
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    
+    use std::env;
+    use std::fs::{self, File};
+    use std::os::unix::net::UnixListener;
+    use std::path::PathBuf;
+
+    use serial_test::serial;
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -26,3 +26,5 @@ oci_spec = { version = "0.1.0", path = "./oci_spec" }
 [dev-dependencies]
 oci_spec = { version = "0.1.0", path = "./oci_spec", features = ["proptests"] }
 quickcheck = "1"
+serial_test = "0.5.1"
+tempfile = "3.2.0"
```
utam0k, line 30: How about moving the one used in the cgroups test to a more generic location and using it?
https://github.com/containers/youki/blob/1afde706fbfb94715f554f7fe7a31bb0eb86206c/src/cgroups/test.rs#L13

src/tty.rs
```diff
@@ -68,3 +68,92 @@ pub fn setup_console(console_fd: FileDescriptor) -> Result<()> {
...
+        let cleanup = teardown(paths.0);
+        assert!(cleanup.is_ok());
+    }
+
+    #[test]
+    #[serial]
+    fn test_setup_console_socket_empty() {
+        let init =  setup();
+        assert!(init.is_ok());
+        let paths = init.unwrap();
```
utam0k, line 122: Other codes, too, but how about something like the following?
```suggestion
        let (testdir_path, rundir_path, socket_path) = init.unwrap();
```

Cargo.toml
constreference: Moved the TempDir implmentation onto utils

src/tty.rs
constreference: Changed it to the following

constreference: Thank you for reviewing

utam0k (approved):
lgtm

PR #107, opened 2021-06-23
Fix README.md Fedora & Centos instructions

nimrodshn:
related issue: https://github.com/containers/youki/issues/106

cc: @Furisto  @utam0k


utam0k (approved):
nice work!

PR #110, opened 2021-06-26
split the subcommands into their own files.

utam0k:
`main.rs` is so large that I split it into separate files for each subcommand.


Furisto (approved):

Furisto:
lgtm

Issue #20, opened 2021-05-22
impl: exec sub command


minakawa-daiki:
The exec command is not listed in the [runtime-spec](https://github.com/opencontainers/runtime-spec), but I consider it one of the required features. For example, when I run `docker exec -it [container id] bash`, I currently get the following error.

![Screenshot from 2021-05-22 19-39-11](https://user-images.githubusercontent.com/22878067/119223639-842af280-bb35-11eb-84c1-c9a3b96a9df2.png)

minakawa-daiki:
@utamoku Can I have you assign me to solve this issue?

utam0k:
@minakawa-daiki 
Thanks for your interest. Oh... That was definitely not implemented there.
I'd love to have it for you! I've assigned you to it.

minakawa-daiki:
@utam0k May I remove the assignment because there may be someone else who wants to work on this issue?

utam0k:
@minakawa-daiki 
I understand. Please challenge this again whenever you have time.

Furisto:
@utam0k 
I can take a look at this

utam0k:
@Furisto I'm looking forward to your PR.

PR #111, opened 2021-06-26
Seperate adding tasks to cgroups and applying resource restrictions

Furisto:
This is part of #20. We need to be able to move a task into a cgroup without having to apply resource restrictions again.

### DIFF
--- a/src/cgroups/common.rs
+++ b/src/cgroups/common.rs
@@ -19,7 +19,8 @@ pub const CGROUP_PROCS: &str = "cgroup.procs";
 pub const DEFAULT_CGROUP_ROOT: &str = "/sys/fs/cgroup";
 
 pub trait CgroupManager {
-    fn apply(&self, linux_resources: &LinuxResources, pid: Pid) -> Result<()>;
+    fn add_task(&self, pid: Pid) -> Result<()>;
+    fn apply(&self, linux_resources: &LinuxResources) -> Result<()>;
     fn remove(&self) -> Result<()>;
 }
 
--- a/src/cgroups/common.rs
+++ b/src/cgroups/common.rs
@@ -19,8 +19,11 @@ pub const CGROUP_PROCS: &str = "cgroup.procs";
 pub const DEFAULT_CGROUP_ROOT: &str = "/sys/fs/cgroup";
 
 pub trait CgroupManager {
+    /// Adds a task specified by its pid to the cgroup
     fn add_task(&self, pid: Pid) -> Result<()>;
+    /// Applies resource restrictions to the cgroup
     fn apply(&self, linux_resources: &LinuxResources) -> Result<()>;
+    /// Removes the cgroup
     fn remove(&self) -> Result<()>;
 }
 
--- a/src/cgroups/v1/blkio.rs
+++ b/src/cgroups/v1/blkio.rs
@@ -1,12 +1,7 @@
-use std::{
-    fs::{self},
-    path::Path,
-};
+use std::path::Path;
 
-use crate::cgroups::{
-    common::{self, CGROUP_PROCS},
-    v1::Controller,
-};
+use crate::cgroups::{common, v1::Controller};
+use anyhow::Result;
 use oci_spec::{LinuxBlockIo, LinuxResources};
 
 const CGROUP_BLKIO_THROTTLE_READ_BPS: &str = "blkio.throttle.read_bps_device";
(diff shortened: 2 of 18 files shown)


src/cgroups/common.rs
```diff
@@ -19,7 +19,8 @@ pub const CGROUP_PROCS: &str = "cgroup.procs";
 pub const DEFAULT_CGROUP_ROOT: &str = "/sys/fs/cgroup";
 
 pub trait CgroupManager {
-    fn apply(&self, linux_resources: &LinuxResources, pid: Pid) -> Result<()>;
+    fn add_task(&self, pid: Pid) -> Result<()>;
```
utam0k, line 22: This is slightly out of the scope of this PR, However Could you provide comments on each function?

src/cgroups/v1/blkio.rs
```diff
@@ -17,25 +12,19 @@ const CGROUP_BLKIO_THROTTLE_WRITE_IOPS: &str = "blkio.throttle.write_iops_device
...
         }
 
-        common::write_cgroup_file(cgroup_root.join(CGROUP_PROCS), pid)?;
         Ok(())
     }
 }
 
 impl Blkio {
-    fn apply(root_path: &Path, blkio: &LinuxBlockIo) -> anyhow::Result<()> {
+    fn apply(root_path: &Path, blkio: &LinuxBlockIo) -> Result<()> {
```
utam0k, line 27: :+1:

src/cgroups/v2/systemd_manager.rs
```diff
@@ -45,6 +46,19 @@ struct CgroupsPath {
 
 impl SystemDCGroupManager {
     pub fn new(root_path: PathBuf, cgroups_path: PathBuf) -> Result<Self> {
+        // TODO: create the systemd unit using a dbus client.
+        let cgroups_path = Self::new_cgroups_path(cgroups_path)?;
+        let cgroups_path = Self::get_cgroups_path(cgroups_path)?;
```
nimrodshn, line 51: @Furisto This API is slightly confusing i.e. `new_cgroups_path` vs. `get_cgroups_path`. I realize naming is hard but perhaps we can come up with a better name here?

The corresponding method is called `initPath` in [runc](https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/v2.go#L354-L376) - perhaps we call it `init_path` or simply `generate_path`? :shrug: 

@utam0k  WDYT?

src/cgroups/v2/systemd_manager.rs
```diff
@@ -172,8 +176,8 @@ impl SystemDCGroupManager {
             }
         }
 
-        common::write_cgroup_file(full_path.join(CGROUP_PROCS), &pid)?;
-        Ok(full_path)
+        common::write_cgroup_file(self.full_path.join(CGROUP_PROCS), &pid)?;
+        Ok(())
```
nimrodshn, line 180: I wonder if this method can just end with `common::write_cgroup_file(self.full_path.join(CGROUP_PROCS), &pid)`

nimrodshn:
@Furisto Just to make sure I follow: `add_task` is used to generate the `cgroups` tree structure while `apply` is the method used to apply the resource restrictions on the corresponding `cgroup`?

nimrodshn:
BTW do you mind squashing commits?

src/cgroups/common.rs
Furisto: Done

src/cgroups/v2/systemd_manager.rs
Furisto: Was not super happy with the naming either. Reworked it. Let me know what you think.

src/cgroups/v2/systemd_manager.rs
Furisto: Done

Furisto:
@nimrodshn Yes, correct. Add_task will move a process into the cgroup and construct the path if it doesn't exist. Apply will apply the resource restrictions. The create command would use add_task and apply, while exec would only use add_task because here we are moving in a cgroup that has already been set up.

I folded the last commit, which was indeed a bit trivial into my new commit. I would prefer to keep the other ones though because they deal with separate things.

src/cgroups/v2/systemd_manager.rs
nimrodshn: Very nice, Much better.

nimrodshn:
> while exec would only use add_task because here we are moving in a cgroup that has already been set up.

@Furisto Just out of curiosity how does the `exec` command work in theory (or practice) ?

nimrodshn:
LGTM :100:

nimrodshn (approved):
LGTM :100:

PR #113, opened 2021-06-27
Add integration tests for life cycle

minakawa-daiki:
Implemented integration tests.
First of all, I implemented tests for the container lifecycle and the `create` command, referring to `opencontainers/runtime-tools`.

See `tests/README.md` for how to run the integration tests.

The flow of test execution is not to run `cargo test` directly, but to first generate a binary and then run it via `sudo`.

I am currently trying not to put it on the CI, but will do so if I can.

ref: #56


utam0k:
@minakawa-daiki 
Thank you for the PR. I'll review it later. I'm a little busy so it may be later.

minakawa-daiki:
I'd like to ask about binary generation for tests. Currently, the binary obtained by `cargo test --no-run` has a hash value and cannot be uniquely identified.
The binary obtained by `cargo test --no-run` currently has a hash value and cannot be uniquely identified, so as discussed in the following issue, we parse the json with `--message-format=json` to get the binary filename.
I can't think of a better idea at the moment, but what do you think?

https://github.com/rust-lang/cargo/issues/1924
https://github.com/rust-lang/cargo/issues/3670

utam0k:
@minakawa-daiki 
I think it would be better to make it independent from youki's main code, like oci-spec. And how about writing it in main instead of the test folder as a test tool instead of test? What do you think?

tsturzl:
I agree with @utam0k. It would be best to split this out into another module. I would be really interested in contributing in this area, as I think we'll want to start building out our own test suite for cgroups v2. Maybe we could even make this not a cargo test at all, and so something similar to the oci runtime-tools and use the test anything protocol or similar. This way we can essentially just run the built tests which can invoke a new build of the project and test it. Food for thought.

minakawa-daiki:
I apologize for the delay in my response. It's certainly better to separate it as a testing tool, and I've reminded myself that there's no need to force it to be included in `cargo test`. I'm a little busy with my personal life and will deal with this a little later, but I'll try.

tsturzl:
@minakawa-daiki Let me know if you need any help with this. I think if the effort to break this out into it's own tool is too high that perhaps we allow this change in, and when I have time I can move this into it's own crate which will serve as a place to build out our own integration testing tool. What do you think @utam0k?

utam0k:
> @minakawa-daiki Let me know if you need any help with this. I think if the effort to break this out into it's own tool is too high that perhaps we allow this change in, and when I have time I can move this into it's own crate which will serve as a place to build out our own integration testing tool. What do you think @utam0k?

@tsturzl 
I would like to hear the opinion of the author of this PR and make a decision. If he is busy, I think that's fine.

YJDoc2:
Hey, I was checking the test anything protocol which is used by oci-spec validation tests, and rust has [testanything](crates.io/crates/testanything) crate which provides producer of data for tests as per TAP protocol. I'm not sure how much maintained and active it is, but that might be useful. We can use that as the part of our main program, and take path of youki executable and bundle etc.as cmd options. Then our binary will run all the tests that are in oci-validation tests, written in rust, and produce output using testanything. Maybe that way we won't need to create a complete tool for testing right now, instead just port validation suite, which will be used by us to validate the changes. @minakawa-daiki @tsturzl ?
Also, If you decide to port the tests to rust in some way, let me know , I would like to help.

minakawa-daiki:
@tsturzl @utam0k 
Thank you for your concern!
I'm currently writing the code to migrate using `oci_spec` as a reference. I'll be able to submit a pull request after this, and I hope you'll take a look at that code and decide for yourself!

minakawa-daiki:
@YJDoc2 
I think `testanything` is a very good idea!
I think we can use this to implement it once, and if `testanything` is not active, we can make it a policy to replace the implementation ourselves later. Thanks!

minakawa-daiki:
@utam0k @tsturzl @YJDoc2 
I wrote out the integration test as a test tool.
Since the `integration-test` directory already exists, it is temporarily named `youki-integration-test`.
As an implementation, I used `testanything` that @YJDoc2 taught me. As the first test, I've added tests for the container lifecycle and the `create` command.
Please refer to the `Readme.md` for more information on how to use the tests.
I hope you will review if this is a good policy for now.

youki_integration_test/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "youki_integration_test"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+uuid = "0.8"
+rand = "0.8.0"
+tar = "0.4"
+flate2 = "1.0"
+testanything = "0.2.1"
```
utam0k, line 11: How about error handling [anyhow](https://docs.rs/anyhow/1.0.42/anyhow/)?

youki_integration_test/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+        panic!("Can not initilize test.");
+    }
+    container_create_test(&project_path);
+    if cleanup_test(&project_path).is_err() {
+        panic!("Can not cleanup test.");
+    }
+}
+
+// This tests the entire lifecycle of the container.
+fn life_cycle_test(project_path: &PathBuf) {
```
utam0k, line 35:
```suggestion
fn life_cycle_test(project_path: &Path) {
```
https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

youki_integration_test/src/support.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::io;
+use std::{env, fs, path::PathBuf};
+use tar::Archive;
+use testanything::tap_suite_builder::TapSuiteBuilder;
+use testanything::{tap_test::TapTest, tap_test_builder::TapTestBuilder};
+use uuid::Uuid;
+
+pub fn initialize_test(project_path: &PathBuf) -> Result<(), std::io::Error> {
+    let result = prepare_test_workspace(project_path);
+    return result;
```
utam0k, line 13: The `return` is unnecessary.

youki_integration_test/src/support.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+pub fn cleanup_test(project_path: &PathBuf) -> Result<(), std::io::Error> {
+    let result = delete_test_workspace(project_path);
+    return result;
```
utam0k, line 18: The `return` is unnecessary.

youki_integration_test/src/support.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+pub fn create_project_path() -> PathBuf {
+    let current_dir_path_result = env::current_dir();
+    return match current_dir_path_result {
```
utam0k, line 23: The `return` is unnecessary.

youki_integration_test/src/support.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub fn test_builder(status: bool, name: &str, diagnostic: &str) -> TapTest {
+    TapTestBuilder::new()
+        .name(name)
+        .passed(status)
+        .diagnostics(&vec![diagnostic])
+        .finalize()
+}
+
+pub fn print_test_results(test_name: &str, tests: Vec<TapTest>) -> () {
```
utam0k, line 55: The `()` is unnecessary.

youki_integration_test/src/command/create.rs
```diff
@@ -0,0 +1,19 @@
+use std::path::PathBuf;
+use std::process::{Command, Stdio};
+
+// TODO Allow to receive arguments.
+// TODO Wrapping up the results
+pub fn exec(project_path: &PathBuf, id: &str) -> bool {
+    let status = Command::new(project_path.join(PathBuf::from("youki")))
```
utam0k, line 7: I think `"youki"` is too hard coding, so how about using an environment, e.g. `RUNTIME`
I think the default value should be youki.

youki_integration_test/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+    let youki = Container::new(project_path);
```
utam0k, line 36: I felt that the variable name `container_runtime` was a better level of abstraction.

utam0k (changes requested):
Nice!
However, I commented on a few points.
But basically, I think that once the commented parts are fixed, we can merge them and discuss them with the developers to improve the whole` thing. It is good enough for a first prototype.

utam0k:
@minakawa-daiki I ran the test according to the README, and it failed. Is there something wrong with this?
```
 sudo ./youki_integration_test
# Start Create comand test suite
1..6
ok 1 Create a new container test
# This operation must create a new container.
ok 2 Execute state test
# This operation must state the container.
ok 3 Execute start test
# This operation must start the container.
ok 4 Execute state test
# This operation must state the container.
ok 5 Execute kill test
# This operation must kill the container.
not ok 6 Execute delete test
# This operation must delete the container.
# End Create comand test suite
# Start Create comand test suite
1..3
ok 1 create with no ID test
# This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
ok 2 create with ID test
# This operation MUST create a new container.
ok 3 create with an already existing ID test
# If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error and a new container MUST NOT be created.
# End Create comand test suite

```

youki_integration_test/src/command/create.rs
YJDoc2: What would be your thoughts on using cmd args instead of env vars? That way we user won't have to define env vars specifically for this purpose, and it might make more sense to pass those as args to the command, rather than to set them as general env vars.

YJDoc2:
Would it be possible , that instead of relying on panic as indication of test failure, we take the result of spawned process, along with its stderr, and accordingly set status and error message? That way we can have more information on why the test failed.
The process ExitStatus implements fmt , so it can convert the exit code of the process to readable string, and we might be able to use it as an error message.

tsturzl:
I agree with @YJDoc2. I think that would provide more context. And in one of @utam0k's review comments he mentions using the `anyhow` crate that we use heavily in youki already. I think that could give you a lot of useful tools for dealing with errors, and should be pretty simple to add.

YJDoc2:
@utam0k I pulled this PR and ran tests as stated, several times, but all tests give ok for me consistently. :thinking: :thinking:

minakawa-daiki:
As for the error messages, I'll change the implementation to use `anyhow`.
Please wait for a while as we will respond to other reviews as they come in.
The reason why the test fails is not yet known, so we are investigating.

youki_integration_test/src/command/create.rs
minakawa-daiki: @utamoku What do you think about this?
My feeling is that when comparing `env vars` and `cmd args`, `env vars` is basically unchanged and users are expected to use default values, while `cmd args` is assumed to be specified by users.
In this case, I think `env vars` is fine because it is basically used with `youki`, but `cmd args` may be an option if you are considering using a runtime other than `youki` in the future.

minakawa-daiki:
@utamoku Corrected the reviews we received except for the hard coding of the `youki` command.
I have commented on the hard coding of the `youki` command separately, so please check it.

utam0k:
Sorry, It had nothing to do with this PR, so I sent out a PR to fix it.
https://github.com/containers/youki/pull/165
> @utam0k I pulled this PR and ran tests as stated, several times, but all tests give ok for me consistently. thinking thinking

utam0k (approved):
This is a great PR move. And it is one of the most needed features in youki right now. However, I think there are still many things to be developed in this PR. Let's merge it once and improve it together.

PR #116, opened 2021-06-30
add comments in intergration_test.sh about test case that runc no paas

TinySong:
add some comments for this issue. https://github.com/containers/youki/issues/27#issuecomment-859488845
then https://github.com/containers/youki/issues/27 can be closed.


utam0k (approved):
thanks!

PR #117, opened 2021-06-30
remove unnecessary clone() in create.rs

utam0k:
I will try to make create.rs easier to understand little by little.

### DIFF
--- a/src/create.rs
+++ b/src/create.rs
@@ -120,6 +120,7 @@ impl Create {
         Ok(())
     }
 }
+
 /// Fork the process and actually start the container process
 fn run_container<P: AsRef<Path>>(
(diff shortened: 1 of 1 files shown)


utam0k:
The create.rs should be split up more, but I don't have a good idea how to split it up, so if you have any suggestions, please let me know.

Furisto (approved):

Furisto:
@utam0k I plan to move the create.rs code to a builder, so that we can reuse it in exec and in tests. This should give me more ideas on how to organize it.

utam0k:
@Furisto Your idea sounds good to me. Since I think it's fine to proceed in that direction, but I'd really like to hear other developer's opinions, could you create an issue about it?

Issue #24, opened 2021-05-23
Support for systemd cgroup driver


utam0k:
I'm considering supporting podman, but I don't know the details of podman, so I'm looking for opinions.
I don't know what features youki lacks to support podman yet.
I'm also looking for people to support this issue.

stappersg:
If I understand correct  opens `podman` the path to _IPv6_ support.

utam0k:
@stappersg 
Thank you for your valuable information! I don't understand much about it yet, but what kind of implemental do you need to accomplish this?
> If I understand correct opens podman the path to IPv6 support.

tsturzl:
Has anyone tried dropping Youki into Podman as a runtime? I've only used Podman a handful of times, but I don't believe there is a whole lot different from docker. It seems like a good starting off point would be to see if how they work together.

tsturzl:
@stappersg I'm actually working on the network cgroup controller right now. So far as I can tell the network subsystem of cgroups doesn't really seem to be concerned with the transport layer you're mostly configuring the priority of each interface irrespective to the IP layer. So I wouldn't assume anything special needs to be done to support IPv6, but I could be wrong.

utam0k:
I've rarely tried the podman, and when I tried it this past weekend, it seemed to lack support for some of the arguments passing to the youki.
I haven't looked deeper into the details yet.

utam0k:
Unlike docker, the podman allows you to specify a container runtime when creating a container, which is useful when using container runtimes other than the runc. For this reason, I think it is good to support it as much as possible.

stappersg:
Elaborating my humble
> If I understand correct  opens `podman` the path to _IPv6_ support.

I do like the idea of an OCI runtime implemented in Rust.
Please do things different (and possible better) as Docker. One such thing is supporting _IPv6_.

My websearch on _"Podman IPv6"_ learnt me that it would make sense to tell about a possible road to success on _IPv6_ support in `youki`.

utam0k:
@stappersg 
Thanks for the advice! Let me clarify the situation a bit.
It is the same container layer software as youki and runc. This means that youki is the software used by Docker, and podman is software in the same layer as Docker. This issue tries to support podman.
Is it correct that if youki will support podman, then podman has IPv6 support and therefore youki has IPv6 support?

nimrodshn:
@utam0k It looks like `podman` is using [conmon](https://github.com/containers/conmon) to run the container runtime[1][2]

[...]

PR #120, opened 2021-07-02
[WIP] Initial implementation of resource-control via systemd and dbus

nimrodshn:
## What this does?
Adds easy API for interacting with systemd via dbus (namely creating units), using this we can apply [resource control](https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html) via systemd.

Please note: I try to align with [runc's](https://github.com/opencontainers/runc/blob/6023d635d725a74c6eaa11ab7f3c870c073badd2/docs/systemd.md#systemd-cgroup-driver) approach in this line of work to walk in the steps of giants.

TODO: 

- [ ] Implement the "controllers" abstraction for each resource supported by `systemd` properties. (Traits and respective implementation).
- [ ] Finish implementing `start_transient_unit_for_container` which eventually applies the properties to the newly created unit associated with the container (the last `for` loop doesn't compile in the current vesion :cry: ). The bindings generated to the dbus interfaces are used using the following command: ```> dbus-codegen-rust -s -g -m None -d org.freedesktop.systemd1 -p /org/freedesktop/systemd1``` (See https://github.com/diwic/dbus-rs)

Related issue: #24 

cc: @utam0k @Furisto


src/dbus/client.rs
```diff
@@ -15,19 +18,90 @@ impl Client {
...
+        } else {
+            properties.push(PropSlice::new(slice).into());
+        }
+
+        proxy.start_transient_unit(unit_name, "replace", properties, vec![])?;
         Ok(())
     }
 }
+
+type Property = (&'static str, Variant<Box<dyn RefArg>>);
```
nimrodshn, line 56: I'm considering using a procedural macro to derive all these helper structs and their impls.

nimrodshn: Or I can keep it simple and just write the tuples directly.

utam0k:
@nimrodshn Is there anything I can do to help?

nimrodshn:
@utam0k Sorry I was busy with other projects :cry: I will try and get some work on this soon..

utam0k:
@nimrodshn I understand. You can do it at your own pace. I was a little worried because you've been contributing to youki quite a bit.

nimrodshn:
@utam0k  Sorry I was busy with life :sweat_smile: . I will try to advance this asap.

@Furisto suggested I make the resource control part modular (similar to how we provide `Controller` impls for each controller type, e.g. memory, io etc.) How do you envision that API work here? Should we pass a list of controllers to `start_transient_unit_for_container` in addition to `LinuxResources`? As These are provided statically in each cgroups manager (For example [here](https://github.com/containers/youki/blob/main/src/cgroups/v2/manager.rs#L28-L36).) - I think that might be a simple solution that way this dbus client could be used both for `v1` and `v2` systemd cgroup managers.

Furisto:
@nimrodshn My idea was that the dbus client should not know anything about resources or controllers. The responsibility of the dbus client is to communicate with systemd and nothing else. The responsibility of the controllers is to translate the resource restrictions described in the spec to systemd properties. 
That means they get passed in a `Resource` and return back properties. We collect these properties from the controllers (and maybe additional ones that are created by the cgroup manager directly like for pids) and hand them over do the dbus client. Should systemd support further resources in the future we can just add another controller without having to touch the surrounding code. 

You do not need to implement it exactly like this though, if you think you have a better idea, go for it.

Furisto:
@nimrodshn Can I support you in any way?

nimrodshn:
@Furisto No I'm good thanks! I just came back to this :smile:

utam0k:
@nimrodshn Hi! Can I ask you to write out the remaining work?

nimrodshn:
@utam0k  Wrote the work to be done in the PRs description :arrow_up:

utam0k:
@nimrodshn Thanks! Let me know if you are busy or need any help. This isn't going to be included in the first release of youki, so taking it slow is fine. Please don't feel pressured. We need your help, I'll be waiting for you to come back ;)

nimrodshn:
Thanks @utam0k currently finding it hard to work on open source :crying_cat_face: as I'm in between jobs so taking it one step at a time :smile_cat:

utam0k:
> Thanks @utam0k currently finding it hard to work on open source  as I'm in between jobs so taking it one step at a time

Of course, there are times like that. No problem. I'm waiting for this PR to be open, so let me know when you have more time. If you have plans, please let me know.

utam0k:
We'll continue here. @nimrodshn Thanks a lot.
#451

Issue #118, opened 2021-07-01
Redesign create.rs


utam0k:
The create.rs should be split up more, but I don't have a good idea how to split it up, so if you have any suggestions, please let me know.

# Ideas
- builder pattern(https://github.com/containers/youki/pull/117#issuecomment-871488278)
https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

nimrodshn:
> builder pattern

Furisto:
This is what the preliminary api looks like: 

```
 // required values (must be specified in new...)
        // - create
        //   - id
        //   - bundle
        // - exec
        //   - id
        //
        // use with_... methods to specify
        // optional values
        // - console-socket
        // - pid-file
        //
        // overwritable values
        // - systemd (default true)
        // - root_path (default /run/youki)
        //
        // overwritable values (for exec only?)
        // - env
        // - cwd
        // - container command
        //
        // calculated in build()
        // computed values
        // - rootless
        // - container_dir
        // - spec
        // - notify_socket
        // - container 

        // create
        let container = ContainerBuilder::new_init(id, bundle)?
        .with_pid_file(pid_file) // optional
        .with_console_socket(console_socket) //optional
        .with_systemd(false) // overwrite default
        .with_root_path(root_path) // overwrite default
        .build()?;

        // exec
        let container = ContainerBuilder::new_tenant(id)
        .with_pid_file(pid_file)
        .with_console_socket(console_socket)
        .with_cwd(cwd)
        .with_env(env)
        .with_container_command(vec!["sleep".to_owned(), "9001".to_owned()])
        .build()?;    
```

PR #121, opened 2021-07-02
Modularize container creation

Furisto:
This attempts to make the container creation code more modular. Related to #20 and #118.


Furisto:
Will check the test failure. Works fine for me locally.

utam0k:
As for the overall direction, I agree with you. It's so cool. I'll review this when WIP comes off.

Furisto:
@utam0k This is now in a state where it can be reviewed

.github/workflows/main.yml
```diff
@@ -31,6 +31,8 @@ jobs:
         run: ./build.sh
       - name: Run tests
         run: cargo test
+      - name: Run doc tests
+        run: cargo test --doc
```
utam0k, line 35: :+1:

src/capabilities.rs
```diff
@@ -1,4 +1,4 @@
-use crate::command::Command;
+use crate::command::{Syscall};
```
utam0k, line 1: The `{}` is probably unnecessary.

src/container/builder.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use youki::container::builder::ContainerBuilder;
+    ///
+    /// let builder = ContainerBuilder::new("74f1a4cb3801".to_owned());
+    /// ```
+    pub fn new(container_id: String) -> Self {
+        let root_path = PathBuf::from("/run/youki");
```
utam0k, line 44: It may be better to use Default Trait.

src/utils.rs
```diff
@@ -96,6 +96,11 @@ pub fn write_file<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Resul
     Ok(())
 }
 
+pub fn create_dir_all<P: AsRef<Path>>(path: P) -> Result<()> {
```
utam0k, line 99: I thought there might be no need to bother wrapping it in a function, is there any reason for this?

src/container/builder.rs
utam0k: Also, it's better to use Default for other builders as well.

src/utils.rs
Furisto: In the case of an error I want that the error message contains the path that should have been created, In the past it had been difficult to find out where the error is happening without this information. I could add with_context to every call, but having it in one place made more sense to me.

src/container/builder.rs
Furisto: The default trait does not allow to specify arguments. I could do
```
ContainerBuilder::default()
.with_id(id)
```

instead of with new(id) and generate an id automatically if none is specified in with_id if you like. The TenantContainerBuilder and InitContainerBuilder are not intended to be instantiated directly but by calling as_tenant/is_tenant on the ContainerBuilder, so you do not really have a default value for them because you always have a builder that will be provided in new(...)

utam0k: I was planning to use default() in new, and I thought default would be better to use in case there are more constructor functions of different types.
However, it's a small thing, so I think it's fine to prepare it when it's needed.
What do you think?

Furisto: We only have one constructor function at the moment, so I think we should do a default implementation when we start to get multiple.

Furisto:
There still seems to be some sporadic issue where the container is stopped during the kill integration test. Looking into this...

utam0k:
> There still seems to be some sporadic issue where the container is stopped during the kill integration test. Looking into this...

This may take a little time to find the cause. I've been trying to find time to do it, but haven't been able to. If you find out anything, please let me know. In the worst case, just creating an issue is fine for this PR.

Furisto:
@utam0k I have created #126 for the CI issues. If you have no objections we can merge this.

utam0k (approved):
awesome

PR #122, opened 2021-07-04
make String to signal conversion more simplify by using a Trait.

utam0k:


src/signal.rs
```diff
@@ -3,40 +3,46 @@
...
-        "25" | "XFSZ" | "SIGXFSZ" => SIGXFSZ,
-        "26" | "VTALRM" | "SIGVTALRM" => SIGVTALRM,
-        "27" | "PROF" | "SIGPROF" => SIGPROF,
-        "28" | "WINCH" | "SIGWINCH" => SIGWINCH,
-        "29" | "IO" | "SIGIO" => SIGIO,
-        "30" | "PWR" | "SIGPWR" => SIGPWR,
-        "31" | "SYS" | "SIGSYS" => SIGSYS,
-        _ => bail! {"{} is not a valid signal", signal},
-    })
+pub trait ToSignal<From = Self> {
```
Furisto, line 6: Couldn't we use TryInto instead? Or do you want to avoid consuming the String?

nimrodshn: +1 for implementing `From` for `String` - it can even be for `&String` if you do not wish to consume the String.

nimrodshn:
Also. @utam0k please squash commits if possible.

src/signal.rs
utam0k: I also think that might be better in this case, but I can't do it because of the following limitations.
https://doc.rust-lang.org/stable/error-index.html#E0117

Furisto: Hmm, you are right. Then this seems to be the best way.

Furisto (approved):

PR #123, opened 2021-07-07
Add cgroup v2 freezer controller

duduainankai:
Add support for freezer in cgroup. #78

### DIFF
--- a/src/cgroups/v2/freezer.rs
+++ b/src/cgroups/v2/freezer.rs
@@ -1,0 +1,193 @@
+use anyhow::{bail, Result};
+use std::{
+    fs::OpenOptions,
+    io::{BufRead, BufReader, Read, Seek, SeekFrom, Write},
+    path::Path,
+    str, thread,
+    time::Duration,
+};
+
+use oci_spec::{FreezerState, LinuxResources};
+
+use super::controller::Controller;
+
+const CGROUP_FREEZE: &str = "cgroup.freeze";
+const CGROUP_EVENTS: &str = "cgroup.events";
+
+pub struct Freezer {}
+
+impl Controller for Freezer {
+    fn apply(linux_resources: &LinuxResources, cgroup_path: &Path) -> Result<()> {
+        if let Some(freezer_state) = linux_resources.freezer {
+            Self::apply(freezer_state, cgroup_path)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl Freezer {
+    fn apply(freezer_state: FreezerState, path: &Path) -> Result<()> {
+        let state_str = match freezer_state {
+            FreezerState::Undefined => return Ok(()),
+            FreezerState::Frozen => "1",
+            FreezerState::Thawed => "0",
+        };
+
+        match OpenOptions::new()
+            .create(false)
+            .write(true)
+            .open(path.join(CGROUP_FREEZE))
+        {
+            Err(e) => {
+                if let FreezerState::Frozen = freezer_state {
+                    bail!("freezer not supported {}", e);
+                }
+                return Ok(());
+            }
+            Ok(mut file) => file.write_all(state_str.as_bytes())?,
+        };
+
+        // confirm that the cgroup did actually change states.
+        let actual_state = Self::read_freezer_state(path)?;
+        if !actual_state.eq(&freezer_state) {
+            bail!(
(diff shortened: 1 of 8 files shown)


duduainankai:
@utam0k 

I found that some files disobey `cargo fmt`, maybe we should add a check in CI.

If you agree with that, maybe I can create an issue and finish it.

src/cgroups/v2/freezer.rs
```diff
@@ -0,0 +1,206 @@
+use anyhow::{Result, *};
```
utam0k, line 1: I feel the following code is more clear since this file only uses  `Result` and `bail!`.
```suggestion
use anyhow::{Result, bail};
```

src/cgroups/v2/freezer.rs
```diff
@@ -0,0 +1,206 @@
...
+        let tmp =
+            create_temp_dir("test_set_freezer_state").expect("create temp directory for test");
+        set_fixture(&tmp, CGROUP_FREEZE, "").expect("Set fixure for freezer state");
+        set_fixture(&tmp, CGROUP_EVENTS, "").expect("Set fixure for freezer state");
+
+        // events file does not contain "frozen 1"
+        {
+            let freezer_state = FreezerState::Frozen;
+            let r = Freezer::apply(freezer_state, &tmp);
+            assert_eq!(true, r.is_err());
```
utam0k, line 203: For comparing boolean values, is more simpler to use `assert!`
```suggestion
            assert!(true, r.is_err());
```

src/cgroups/v2/freezer.rs
```diff
@@ -0,0 +1,206 @@
...
+            .open(path.join(CGROUP_FREEZE));
+        if r.is_err() {
+            // ignore error if it is not a freeze op.
+            if let FreezerState::Frozen = freezer_state {
+                bail!("freezer not support {}", r.unwrap_err());
+            }
+            return Ok(());
+        }
+
+        r.unwrap().write_all(state_str.as_bytes())?;
```
utam0k, line 56: How about using the match syntax?
```suggestion
        match OpenOptions::new()
            .create(false)
            .write(true)
            .open(path.join(CGROUP_FREEZE))
        {
            Err(err) => {
                // ignore error if it is not a freeze op.
                if let FreezerState::Frozen = freezer_state {
                    bail!("freezer not support {}", err);
                }
                return Ok(());
            }
            Ok(mut file) => file.write_all(state_str.as_bytes())?,
        }
```

src/cgroups/v2/freezer.rs
```diff
@@ -0,0 +1,206 @@
...
+        match state {
+            "0" => {
+                return Ok(FreezerState::Thawed);
+            }
+            "1" => {
+                return Self::wait_frozen(path);
+            }
+            _ => {
+                bail!("unknown \"cgroup.freeze\" state: {}", state);
+            }
```
utam0k, line 89:
```suggestion
        return match state {
            "0" => Ok(FreezerState::Thawed),
            "1" => Self::wait_frozen(path),
            _ => bail!("unknown \"cgroup.freeze\" state: {}", state),
```

duduainankai:
Cool advice. 

Already updated. PTAL.

src/cgroups/v2/freezer.rs
```diff
@@ -0,0 +1,193 @@
...
+        };
+
+        match OpenOptions::new()
+            .create(false)
+            .write(true)
+            .open(path.join(CGROUP_FREEZE))
+        {
+            Err(e) => {
+                if let FreezerState::Frozen = freezer_state {
+                    bail!("freezer not support {}", e);
```
Furisto, line 44: Nit: freezer not support**ed**

Furisto:
Otherwise looks great!

utam0k:
@duduainankai Sure. Can I ask you to do this?
> I found that some files disobey cargo fmt, maybe we should add a check in CI.
> If you agree with that, maybe I can create an issue and finish it.

src/cgroups/v2/freezer.rs
duduainankai: :stuck_out_tongue:
--


<br class="Apple-interchange-newline">

utam0k (approved):
lgtm

PR #124, opened 2021-07-07
Reduce size of binary

Furisto:
This reduces the size of the youki release binary by 51% from 7019736 bytes to 3449632 bytes. This is achieved by using link time optimization (reduction of ~2MB) and pruning the dependencies (removal of regex crate and restricting the features from clap to what we actually use). 
I also tried codegen-units=1 but the size reduction is negligible. Stripping the symbols from the binary on the other hand cuts the size in half again to about 1.6 MB, but this is not included in this PR as we would lose readable stack traces.


utam0k (approved):
Wow! It's great the work.

nimrodshn:
Very nice.

PR #129, opened 2021-07-11
Add format check ci

### LINKED ISSUE #125: Add code format check in CI.
cc @utam0k

duduainankai:
Close #125

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,22 @@ on:
       - main
 
 jobs:
+  rustfmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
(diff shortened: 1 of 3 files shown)


utam0k (approved):
lgtm

duduainankai:
Do you think we should add dependency on these steps ? 

Such like do not run test if fail to check format.

utam0k:
@duduainankai Thanks for your advice. I don't think we need to add the dependencies. Are there any advantages or disadvantages?

duduainankai:
pros: simplify CI pipline procedure, since code format check failed means CI failed and cannot be merged.

cons: test and format check steps cannot run in parallel, and maybe another CI pipline to find test failed.

I think both are acceptable, so just a simple ask, it's up to you.

@utam0k

utam0k:
@duduainankai Thanks for your simplified summary. I understood. This time we will adopt not to create dependencies.

PR #130, opened 2021-07-11
Fix spec path in delete

duduainankai:
image


Furisto:
@duduainankai Did you specify a relative bundle path during creation? I am wondering if this is related to missing canonicalization of the bundle path during create.

duduainankai:
Oh I think I used a container created by stale youki binary and try to delete it by newly binary which it's not canonicalized in `state.json`.

@Furisto sorry about that.

But I do not understand why not use config.json in container.root and avoid these `canonicalization` problem ?

Furisto:
We can. I think your change makes sense. I just wanted to understand how we ended up in this situation.

Furisto (approved):

PR #131, opened 2021-07-11
Document capabilities rs and refactor its drop_privileges function

YJDoc2:
This PR is part of #14 , it documents capabilities.rs
This also refactors the drop_privileges function in it :
Before it had
```
let all = caps::all();
    log::debug!("dropping bounding capabilities to {:?}", cs.bounding);
    for c in all.difference(&to_set(&cs.bounding)) {
        match c {
            Capability::CAP_PERFMON | Capability::CAP_CHECKPOINT_RESTORE | Capability::CAP_BPF => {
                log::warn!("{:?} is not supported.", c);
                continue;
            }
            _ => caps::drop(None, CapSet::Bounding, *c)?,
        }
    }
```
as part of it, but in my opinion, this should not be responsibility of drop_privileges function, but the Syscall structure abstraction should take care of it in its implementation. Thus I have moved this part to LinuxSyscall::set_capabilities.

### DIFF
--- a/docs/doc-draft.md
+++ b/docs/doc-draft.md
@@ -92,3 +92,7 @@ This also provides implementation for Linux syscalls for the trait.
 
 [oci runtime specification]: https://github.com/opencontainers/runtime-spec/blob/master/runtime.md
 [runc man pages]: (https://github.com/opencontainers/runc/blob/master/man/runc.8.md)
+
+## Capabilities
+
+- [Simple explanation of capabilities](https://blog.container-solutions.com/linux-capabilities-in-practice)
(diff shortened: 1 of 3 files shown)


YJDoc2:
@utam0k , please check this once

nimrodshn:
LGTM :100: 

@YJDoc2 Please squash commits :pray:

utam0k:
> @YJDoc2 Please squash commits 🙏

@nimrodshn 
Thanks for the advice! You can squash when merging, but the problem is that the squash commit hash will change. I'm not that particular about squashing. Rather, it's very nice to have commits separated into meaningful units.

utam0k (approved):

PR #134, opened 2021-07-12
Document Info module, and change info.exec -> Info::print

YJDoc2:
Part of #14 
This documents Info module
Also originally Info structure had instance method as 
```
#[derive(Clap, Debug)]
pub struct Info {}

impl Info {
    pub fn exec(&self) -> Result<()> {
        print_youki();
        print_kernel();
        print_os();
        print_hardware();
        print_cgroups();

        Ok(())
    }
}
```
But Info structure does not store any data, and all things we print are taken from the underlying system itself, so I don't think we will store anything in future as well. Also I think naming it `print` rather than `exec` would be more apt. So I have changed it to
```
impl Info {
    pub fn print() -> Result<()> {
        print_youki();
        print_kernel();
        print_os();
        print_hardware();
        print_cgroups();

        Ok(())
    }
}
```

Please take a look, @utam0k  and @Furisto

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -53,7 +53,7 @@ enum SubCommand {
     #[clap(version = "0.0.1", author = "utam0k <k0ma@utam0k.jp>")]
     State(state::State),
     #[clap(version = "0.0.1", author = "utam0k <k0ma@utam0k.jp>")]
-    Info(info::Info),
+    Info,
     #[clap(version = "0.0.1", author = "utam0k <k0ma@utam0k.jp>")]
     List(list::List),
 }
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -82,7 +82,7 @@ fn main() -> Result<()> {
         SubCommand::Kill(kill) => kill.exec(root_path),
         SubCommand::Delete(delete) => delete.exec(root_path, systemd_cgroup),
         SubCommand::State(state) => state.exec(root_path),
-        SubCommand::Info(info) => info.exec(),
+        SubCommand::Info => info::Info::print(),
```
Furisto, line 85: We should keep this consistent with all the other commands.

YJDoc2: Yes, that makes sense :sweat_smile:
I'll close this one and open new one with only comments.

YJDoc2:
Closing in favor of #136 , as the changes made to info module in this PR are not required.

PR #137, opened 2021-07-15
Document list and logger modules

YJDoc2:
Part of #14 
This documents list module, and logger module,
Also changes default log level in logger module :
Originally it was Warn , now it Conditionally compiles to Debug for debug builds and Warn for other builds
@utam0k


utam0k (approved):
lgtm

PR #139, opened 2021-07-16
Add pause and resume command

### LINKED ISSUE #99: Implement the commands `pause` and `resume`

duduainankai:
Close #99 

I have done a manual test. Also works with docker pause and unpause.

image
image
image


src/pause.rs
```diff
@@ -0,0 +1,98 @@
+//! Contains functionality of pause and resume container command
```
Furisto, line 1: I would either split this up into two files (pause and resume) or rename the file to freeze.rs. Pause.rs does not reflect the complete content of the file.

src/pause.rs
```diff
@@ -0,0 +1,98 @@
...
+        let cgroups_path =
+            utils::get_cgroup_path(&spec.linux.unwrap().cgroups_path, &self.container_id);
+        let cmanager = cgroups::common::create_cgroup_manager(cgroups_path, systemd_cgroup)?;
+        let linux_resources = LinuxResources {
+            freezer: Some(FreezerState::Frozen),
+            ..Default::default()
+        };
+
+        log::debug!("cgroup manager applying");
+        cmanager.apply(&linux_resources)?;
```
Furisto, line 52: Pause and resume is not like the other resource restrictions. It would be better to add a freeze method to the cgroup manager.

src/pause.rs
duduainankai: I chose the former.

Furisto (approved):

Furisto:
Great!

utam0k:
🎉

PR #141, opened 2021-07-16
memory cgv2 subsystem implemented

tsturzl:
After a lot of fumbling around with my weird local distro I finally was able to get cgroups v2 working properly to test this. I tested this by creating the container with youki directly(following the youki tutorial in the README) where I set the memory limits in the config.json directly. I was able to observe the values being set appropriately in the cgroup files.

**Notice 1:** I do not have systemd installed locally and I cannot get the systemd crate to build properly as I cannot get libsystemd installed locally without a lot of hassle. What I did was I made the systemd create an optional dependency and included it as a default feature, so for default builds things should work exactly the same for everyone else but this gives me the option of building and testing with the `--no-default-options` cargo flag. I implement the `booted` function from the crate inline with the only place it is used, which is the only use we have for this dependecy currently. The function will simply return an error indicating you are using a build that does not have this dependency. I think this will also be useful for cases where we want to create builds for platforms which are NOT using systemd, this would include Alpine Linux which is a very popular distro for small embedded devices and I think it's useful for youki to support these distros as systemd does not currently compile against alternative C libraries(eg musl) which are commonly used for these smaller embedded distros.

**Notice 2:** I had been away for a while and the history of my branch had lagged far behind that of the main branch. With that I decided to just reapply my changes over to a new branch which is directly branched off the main branch as of making this PR. I will therefore close PR #95 in favor of this PR.

### DIFF
--- a/src/cgroups/v2/memory.rs
+++ b/src/cgroups/v2/memory.rs
@@ -1,4 +1,4 @@
-use anyhow::{Result, *};
+use anyhow::{bail, Result};
 use std::path::Path;
 
 use oci_spec::{LinuxMemory, LinuxResources};
@@ -86,7 +86,7 @@ mod tests {
     use std::fs::read_to_string;
 
     #[test]
-    fn test_set_memory_v2() {
+    fn test_set_memory() {
         let tmp = create_temp_dir("test_set_memory_v2").expect("create temp directory for test");
         set_fixture(&tmp, CGROUP_MEMORY_MAX, "0").expect("set fixture for memory limit");
         set_fixture(&tmp, CGROUP_MEMORY_LOW, "0").expect("set fixture for memory reservation");
@@ -117,7 +117,7 @@ mod tests {
     }
 
     #[test]
-    fn test_set_memory_unlimited_v2() {
+    fn test_set_memory_unlimited() {
         let tmp = create_temp_dir("test_set_memory_unlimited_v2")
             .expect("create temp directory for test");
         set_fixture(&tmp, CGROUP_MEMORY_MAX, "0").expect("set fixture for memory limit");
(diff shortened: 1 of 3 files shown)


src/cgroups/v2/memory.rs
```diff
@@ -1,10 +1,16 @@
-use anyhow::Result;
+use anyhow::{Result, *};
```
Furisto, line 1: Why the *? Shouldn't 
```suggestion
use anyhow::{bail, Result};
```
be enough?

src/cgroups/v2/memory.rs
```diff
@@ -18,7 +24,253 @@ impl Controller for Memory {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::cgroups::test::set_fixture;
+    use crate::utils::create_temp_dir;
+    use oci_spec::LinuxMemory;
+    use std::fs::read_to_string;
+
+    #[test]
+    fn test_set_memory_v2() {
```
Furisto, line 89: I don't think there is a need to specify v2 in the name. Cargo test shows the complete name of the test e.g. cgroups::v2::memory::test::test_set_memory, so you already know it's v2.

src/cgroups/v2/memory.rs
tsturzl: This is probably left behind from something I intended to clean up.

src/cgroups/v2/memory.rs
tsturzl: makes sense

tsturzl:
requested changes have been made

nimrodshn:
LGTM :100

Furisto (approved):

PR #138, opened 2021-07-15
Implement exec command

Furisto:
Solves #20

PR #143, opened 2021-07-17
Refactor the container creation to use `clone(2)` instead of fork.

### LINKED ISSUE #87: Explore the use of clone to reduce the number of forks youki does in creating containers
With the release of `clone` as a new systemcall in kernel version 5.7^, we can now use it to hopefully reduce the number of forks `youki` does when creating a container (See [here](https://github.com/containers/youki/blob/main/src/create.rs#L140-L195)).

cc @utam0k @Furisto @flxo

yihuaf:
Fix #87

Use `clone(2)` instead of `fork` to create the container process. Using `clone(2)` allows us to create new namespaces in a single function call. As a result, we create one process instead of two processes. The code also ended up cleaner and easier to follow.

Currently, the integration test on my local machine is breaking with errors in cgroups that are mentioned in other issues already. This PR didn't touch cgroup logic so unlikely this PR is causing it.

@utam0k I didn't change the comments and the docs in this PR yet. I'd like to get feedback from you first on the change. I tried to keep the original code as much as possible, but it was difficult. I ended up re-write more parts of the code than I originally imagined. If anyone else is touching this area, it will sure create merge conflicts. Let me know what you think.


yihuaf:
There seems to be a large change that went in a few hours ago... I will resolve this.

utam0k:
> There seems to be a large change that went in a few hours ago... I will resolve this.

Sorry... 😢

utam0k:
This is a big and great change.  I was prepared for the changes to be that big.
If it is no different from the existing behavior, merge it once. We will refactor it from merged. However, I need comments to help others understand it, so please do that.

@Furisto @tsturzl 
What do you think?

Furisto:
> If it is no different from the existing behavior, merge it once. We will refactor it from merged.
>
I am not 100% sure what you mean by that. Can you clarify?

tsturzl:
This will take a while to review. I'll see if I can finish a review by tomorrow night.

tsturzl:
This will take a while to review. I'll see if I can finish a review by tomorrow night.

utam0k:
> > If it is no different from the existing behavior, merge it once. We will refactor it from merged.
> 
> I am not 100% sure what you mean by that. Can you clarify?

@Furisto 
I'm sorry 😢   I was trying to say that refactoring should be done in another PR if the behavior itself is fine in this PR. I expected that a major refactoring in this PR would put a lot of work on the author of this PR and would take a long time to merge this one. Of course, simple fixes should be done, but why not do major rewrites in a separate PR?

yihuaf:
@utam0k I resolved the conflict and rebased it to the latest master branch. There should not be any behavior changes, other than now we create the container process using one clone instead of two fork calls. Hopefully, the code is also cleaner this way. Please let me know if you need anything.

@Furisto Most of the conflicts were with #138. I believe I resolved it correctly, but please double-check just in case.

utam0k (changes requested):
Could you change the control flow diagram in `docs/doc-draft`?

yihuaf:
> Could you change the control flow diagram in docs/doc-draft.md?

Done

Furisto:
Sorry, I am busy with RL stuff today. Will take a look at this tomorrow.

utam0k (approved):
It looks good except for the conflicts, but please review the other @Furisto  and @tsturzl  as well.

src/container/builder_impl.rs
```diff
@@ -42,104 +43,167 @@ pub(super) struct ContainerBuilderImpl {
...
+    rootfs: PathBuf,
+    console_socket: Option<FileDescriptor>,
+    notify_name: PathBuf,
+    child: &mut child::ChildProcess,
+) -> Result<()> {
+    let linux = spec.linux.as_ref().unwrap();
+    let namespaces: Namespaces = linux.namespaces.clone().into();
+    // need to create the notify socket before we pivot root, since the unix
+    // domain socket used here is outside of the rootfs of container
+    let mut notify_socket: NotifyListener = NotifyListener::new(&notify_name)?;
```
Furisto, line 129: This is the line where it fails for me. Notify_name is the full path, which is longer than SUN_LEN.

src/container/init_builder.rs
```diff
@@ -45,12 +42,11 @@ impl InitContainerBuilder {
         let container_dir = self.create_container_dir()?;
         let spec = self.load_and_safeguard_spec(&container_dir)?;
 
-        unistd::chdir(&*container_dir)?;
```
Furisto, line 48: The removal of this line causes EBADF when trying to send the file descriptor of the pseudo terminal. The following setup_console_socket call fails to find the socket because the working directory is not the container directory.

Furisto:
This does not work for me. Tried 
```
sudo docker run --rm --runtime youki busybox printenv
sudo docker run --rm -it --runtime youki busybox sh
```
both hang the shell. I highlighted the problematic areas. The first error is `path must be shorter than SUN_LEN`which comes from the NotifyListener, specifically UnixListener::bind because the path is too long. 
After I fixed this, I encountered another error which was that setup_console fails with "Bad file number" because during setup_console_socket it fails to find the socket file and creates an invalid file descriptor.  After that I was able to create a container, but could not exec into it, same problem I think. 

The quick fix is to add back changing the working directory and using only the last component of the notfiy_name. I have made a mental note to rework how the setup of the NotifyListener and the console-socket is done, this is too easy to get wrong, not your fault.

Furisto:
Also sorry, didn't mean to commit that suggestion.

yihuaf:
> The first error is path must be shorter than SUN_LENwhich comes from the NotifyListener, specifically UnixListener::bind because the path is too long.

I was wondering why taking the extra steps to chdir... Maybe the solution is to chdir before bind and chdir back after, since this limitation is imposed on unix domain socket only? I will make a quick fix and leave the proper fix in a future PR.

Thanks for the review. Don't worry about the extra commit. I will clean it up with review and rebase.

oci_spec/src/lib.rs
```diff
@@ -465,7 +466,7 @@ pub struct LinuxResources {
     pub freezer: Option<FreezerState>,
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone, Copy)]
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
tsturzl, line 469: Why do you need to implement the Hash? I'm not seeing this used anywhere. I'm also unsure where you need Eq or PartialEq.

src/stdio.rs
```diff
@@ -3,7 +3,7 @@ use std::os::unix::io::{AsRawFd, RawFd};
 use anyhow::Result;
 use nix::unistd::dup2;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
tsturzl, line 6: It doesn't seem FileDescriptor is used in any kind of new way. Is there a reason for it that I'm not seeing? I'm sure a clone of a RawFd isn't an issue since it's just an i32, but I'm wondering if this trait is ever actually used.

tsturzl (commented):
Mostly looks good to me. I'm not sure what we'll want to do in regards to stack size, but I believe 1MB is the default stack on windows whereas Linux the default stack is 8MB. There are a couple trait implementations that I don't see being used, and if so they should probably be removed.

src/process/fork.rs
```diff
@@ -1,129 +1,60 @@
...
 
-use crate::cgroups::common::CgroupManager;
-use crate::container::Container;
-use crate::container::ContainerStatus;
-use crate::process::{child, init, parent, Process};
-use crate::rootless::Rootless;
+pub fn clone(cb: sched::CloneCb, clone_flags: sched::CloneFlags) -> Result<Pid> {
+    // unlike fork, clone requires the caller to allocate the stack. here, we use the default
+    // 1MB for stack size.
+    const STACK_SIZE: usize = 1024 * 1024;
```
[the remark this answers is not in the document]
yihuaf: I've since changed the stack size allocated here to 4k (4096). This is the same value used in `runc` implementation, so I think we should be OK.

src/stdio.rs
yihuaf: This is needed to clone the `console_socket`, which is `Option<FileDescriptor>`. I need to pass this into the `container_init` function, which runs inside the new process. Unlike fork where everything is duplicated behind scenes, here we need to clone the memory explicitly.

oci_spec/src/lib.rs
yihuaf: You are right. This was leftover from something I was experimenting with. Will clean up.

tsturzl:
So this is less so related to any changes in this PR, but I noticed while reviewing that we pass around a type we call `FileDescriptor` which is really just a raw file descriptor. I wonder if this is the right way to handle this, even if we do end up only needing the file descriptor we lose some guard rails at runtime over it. Acting on RawFd type isn't always safe or can cause EBADF. File and file-like objects generally are closed implicitly by the Drop trait, so passing around a File object instead of just it's file descriptor will usually give you some guarantee that the lifetime of the file descriptor and the lifetime of the File object are the same. I think in passing around a RawFd we might be holding a file descriptor to a file that is no longer open by the process. This could possibly be the reason why Furisto saw a EBADF because using a previously closed file descriptor means it no longer is valid.

tsturzl:
In conclusion it might be worth the effort to pass around the file like object rather than the RawFd, and we could do this by allowing FileDescriptor to contain any type implementing AsRawFd. Ownership may become an issue then, but this would probably give us better runtime sanity

utam0k:
FileDescriptor was something I introduced in the early days of youki to create a useful function. However, it's currently in a half-baked state, and I think it's worth fixing.

yihuaf:
> The quick fix is to add back changing the working directory and using only the last component of the notfiy_name. I have made a mental note to rework how the setup of the NotifyListener and the console-socket is done, this is too easy to get wrong, not your fault.

@Furisto I fixed the NotifyListener and isolated the chdir logic around bind call. I also left a comment explaining the need to use chdir and the limit of unix domain socket path length. Hopefully this make it clear for future readers why chdir is needed, since I was really confused about it. I left the chdir logic for console-socket unchanged and put back the `chdir` call. I agree there should be improved in a different PR.

I tested with the 2 commands you mentioned as well. They seem to work correctly. However, the exec seems to not be working at the moment, even in master branch, so I suspect it is not this PR that's causing it. I will dig a bit more, but want to to see the fix to NotifyListener.

tsturzl (approved):

src/container/builder_impl.rs
```diff
@@ -42,104 +43,167 @@ pub(super) struct ContainerBuilderImpl {
...
+
+        let cb = Box::new(|| {
+            if let Err(error) = container_init(
+                self.rootless.clone(),
+                self.spec.clone(),
+                self.syscall.clone(),
+                self.rootfs.clone(),
+                self.console_socket.clone(),
+                self.notify_path.clone(),
+                &mut child,
```
Furisto, line 79: Add init here

```suggestion
          if let Err(error) = container_init(
                self.init.clone(),
                self.rootless.clone(),
                self.spec.clone(),
                self.syscall.clone(),
                self.rootfs.clone(),
                self.console_socket.clone(),
                self.notify_path.clone(),
                &mut child,
```

src/container/builder_impl.rs
```diff
@@ -42,104 +43,167 @@ pub(super) struct ContainerBuilderImpl {
...
+    rootfs::prepare_rootfs(
+        &spec,
+        &rootfs,
+        namespaces
+            .clone_flags
+            .contains(sched::CloneFlags::CLONE_NEWUSER),
+    )?;
+
+    // change the root of filesystem of the process to the rootfs
+    command.pivot_rootfs(&rootfs)?;
```
Furisto, line 188:
```suggestion
   if init {
        rootfs::prepare_rootfs(
            &spec,
            &rootfs,
            namespaces
                .clone_flags
                .contains(sched::CloneFlags::CLONE_NEWUSER),
        )?;
    
        // change the root of filesystem of the process to the rootfs
        command.pivot_rootfs(&rootfs)?;
    }
```

Furisto (commented):
The tenant process shares the mount namespace of the init process, so pivot_rooting the rootfs again in the tenant will fail.

yihuaf:
@Furisto I fixed it as you suggested. Combining with #153 I can successfully run `docker run` and `docker exec`.

Furisto (approved):

Furisto:
lgtm

utam0k:
@yihuaf I tried to benchmark youki with a release build, but it seemed to fail to run because of this PR. Do you have any idea what it might be?
```
I tried to benchmark youki with a release build, but it seemed to fail to run because of this PR. Do you have any idea what it might be?
```

utam0k:
I expect that something is wrong around NotifySockt.
```
$ sudo YOUKI_LOG_LEVEL=debug ./youki create -b tutorial tutorial_container
[DEBUG src/container/init_builder.rs:89] 2021-07-26T19:16:11.038939918+09:00 container directory will be "/run/youki/tutorial_container"
[DEBUG src/container/container.rs:84] 2021-07-26T19:16:11.041038794+09:00 Save container status: Container { state: State { oci_version: "v1.0.2", id: "tutorial_container", status: Creating, pid: None, bundle: "tutorial", annotations: {}, created: None, creator: None, use_systemd: None }, root: "/run/youki/tutorial_container" } in "/run/youki/tutorial_container"
RUN CONTAINER
[INFO src/cgroups/common.rs:125] 2021-07-26T19:16:11.042030660+09:00 cgroup manager V2 will be used
[DEBUG src/container/builder_impl.rs:91] 2021-07-26T19:16:11.042572219+09:00 init pid is Pid(101195)
[DEBUG src/rootfs.rs:34] 2021-07-26T19:16:11.042734143+09:00 mount root fs "/home/utam0k/ghq/github.com/utam0k/youki/tutorial/rootfs"
[WARN src/rootfs.rs:48] 2021-07-26T19:16:11.043335406+09:00 A feature of cgroup is unimplemented.
[DEBUG src/capabilities.rs:21] 2021-07-26T19:16:11.043791182+09:00 reset all caps
[DEBUG src/capabilities.rs:28] 2021-07-26T19:16:11.043800345+09:00 dropping bounding capabilities to [CAP_AUDIT_WRITE, CAP_KILL, CAP_NET_BIND_SERVICE]
[WARN src/syscall/linux.rs:132] 2021-07-26T19:16:11.043809634+09:00 CAP_PERFMON is not supported.
[WARN src/syscall/linux.rs:132] 2021-07-26T19:16:11.043824355+09:00 CAP_CHECKPOINT_RESTORE is not supported.
[WARN src/syscall/linux.rs:132] 2021-07-26T19:16:11.043832108+09:00 CAP_BPF is not supported.
[DEBUG src/process/parent.rs:78] 2021-07-26T19:16:11.043919536+09:00 [child to parent] sending child ready
```

yihuaf:
@utam0k I can't reproduce the behavior on my end. I tried `docker run -it --rm --runtime youki busybox echo hello ` with the current master branch. Do you have the benchmark command for me to try? Also, when you say failed, what is the behavior?

yihuaf:
The log also looks correct.

yihuaf:
As discussed offline, the issue happens when building the released version of the binary. #163 filed.

PR #151, opened 2021-07-20
add serde_support to caps

ferrell-code:
I added serde_support upstream to the caps-rs dependency. This allows us to remove the LinuxCapabilityType and use caps::Capability directly which now implements Serialize Deserialize traits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,8 @@ features = ["std", "suggestions", "derive"]
 [dependencies]
 nix = "0.19.1"
 procfs = "0.9.1"
-caps = "0.5.1"
+# Waiting for new caps release, replace git with version on release
+caps = { git = "https://github.com/lucab/caps-rs", rev = "cb54844", features = ["serde_support"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 prctl = "1.0.0"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -13,7 +13,7 @@ features = ["std", "suggestions", "derive"]
 [dependencies]
 nix = "0.19.1"
 procfs = "0.9.1"
-caps = "0.5.1"
+caps = { git = "https://github.com/lucab/caps-rs", features = ["serde_support"] }
```
utam0k, line 16: Can you add a `rev` or `tag` and comment that you are waiting for caps to be released?

oci_spec/Cargo.toml
```diff
@@ -12,5 +12,5 @@ serde = { version = "1.0", features = ["derive"] }
 nix = "0.19.1"
 anyhow = "1.0"
 serde_json = "1.0"
-caps = "0.5.1"
+caps = { git = "https://github.com/lucab/caps-rs", features = ["serde_support"] }
```
utam0k, line 15: Can you add a rev or tag and comment that you are waiting for caps to be released?

utam0k (changes requested):

utam0k (approved):
nice!

PR #153, opened 2021-07-21
Correctly handle the rootfs path with bundle

### LINKED ISSUE #152: spec.root.path is not correctly handled
Currently, the spec.root.fs is not correctly handled. When `init_builder` calls `load_and_safeguard_spec`, the copied version of the spec.json doesn't contain the change to canonicalize the rootfs path. As a result, when `youki exec` tries to load the safeguarded spec in `tenent_builder`, the spec.root.path can be incorrect missing the bundle path. As a result, currently, `youki exec` is also broken.

There are two solutions. One is to save the canonicalized rootfs path with bundle path as part of the spec when safeguard the spec into the `/run/youki/<id>`. The other option is to take care to canonicalize the path with bundle path again when loading the path in `tenent_builder`.

yihuaf:
The rootfs path is not correctly handled. The safe guarded version of
the spec doesn't contain the canonicalized rootfs with the bundle path.
The solution here is to save the rootfs path with bundle path into the
safe guarded version of the spec.

fix #152

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -565,6 +565,7 @@ dependencies = [
  "quickcheck",
  "serde",
  "serde_json",
+ "tempfile",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


yihuaf:
> @yihuaf
> Can I ask you to add unit tests for `save` and `canonicalize_rootfs`?

Will do.

utam0k (approved):

utam0k (changes requested):
The CI is down, please fix it.

utam0k (approved):

PR #155, opened 2021-07-22
Move commands into dedicated module

Furisto:
With the number of commands youki now supports, we should move them to a dedicated module.


utam0k (approved):
I knew I had to do this too.

nimrodshn:
Very nice. :+1:

PR #156, opened 2021-07-22
Document Pause and Resume

YJDoc2:
Documents pause and resume commands, and add references for freezer cgroups in doc-draft.
@duduainankai I have changed an unwrap statement to a temp variable with match statement. It compiles correctly, can you verify that it does not do anything incorrect? I changed it to prevent potential panic due to unwrap. Also please verify the comments as well.

### DIFF
--- a/src/commands/pause.rs
+++ b/src/commands/pause.rs
@@ -11,12 +11,19 @@ use crate::container::ContainerStatus;
 use crate::utils;
 use oci_spec::FreezerState;
 
+/// Structure to implement pause command
 #[derive(Clap, Debug)]
 pub struct Pause {
     pub container_id: String,
 }
 
+// Pausing a container indicates suspending all processes in given container
+// This uses Freezer cgroup to suspend and resume processes
+// For more information see :
+// https://man7.org/linux/man-pages/man7/cgroups.7.html
+// https://www.kernel.org/doc/Documentation/cgroup-v1/freezer-subsystem.txt
 impl Pause {
+    /// Suspend the running container
     pub fn exec(&self, root_path: PathBuf, systemd_cgroup: bool) -> Result<()> {
         log::debug!("start pausing container {}", self.container_id);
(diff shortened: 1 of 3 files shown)


src/commands/pause.rs
```diff
@@ -35,9 +45,15 @@ impl Pause {
         }
 
         let spec = container.spec()?;
-        let cgroups_path =
-            utils::get_cgroup_path(&spec.linux.unwrap().cgroups_path, &self.container_id);
+        // get cgroup path defined in spec
+        let path_in_spec = match spec.linux {
```
duduainankai, line 49: Since when create container we  guarantee `spec.linux` must not be `None` also by `unwrap` it directly (https://github.com/containers/youki/blob/main/src/container/builder_impl.rs#L62)
I think it's unnecessary and we can keep it simple here. 
(Actually we unwrap it directly in all commands)

YJDoc2: Ohk, I'll revert these chagnes back. 
I had a question though, if we can guarantee that it won't be None, then why we have kept it as Option type? In case it is for the case when loading spec from JSON it might be None, should we try to make some workaround so that it will never be None? 
The point I am trying to make is that if we are calling unwrap at all places, then Youki will not work when it is None, and we may as well as move that error to the place where we load the spec, and convert Option\<T\> to the T .

I might be missing something of the big picture :sweat_smile: , but can you help me understanding this? @utam0k can you take a look as well?
Thanks

duduainankai: Yeah, I agree with you.

It seems the `default()` for `Linux` has been implemented, I think we can remove Option for Linux.

Or at least we should add a validate function when create container to check input meets our expectation(like `Linux` is not `None` and so on), otherwise it would panic and unreadable for users.

WDYT? @utam0k

utam0k: I agree with you.

utam0k (approved):

PR #158, opened 2021-07-24
Converted linux in spec from Option<Linux> to Linux

YJDoc2:
As Linux has Deafult trait, all occurrences of spec.linux would always unwrap it, and not having linux present is a fetal error in youki, there is no need to keep it in an Option wrapper.
This was discussed in https://github.com/containers/youki/pull/156#discussion_r675475868

I have checked that it successfully builds youki (No type errors), passes all unit tests, and ran it with docker in busybox container with shell. In the shell I ran ls,cd and ps, and they seemed to work fine.

I have ran this on Linux, it would be great if someone can verify that it also works on windows.

@utam0k @Furisto  please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,10 +121,30 @@ Starting the docker daemon.
 $ dockerd --experimental --add-runtime="youki=$(pwd)/target/x86_64-unknown-linux-gnu/debug/youki"
 ```
 
+In case you get an error like :
+
+```
+failed to start daemon: pid file found, ensure docker is not running or delete /var/run/docker.pid
+```
+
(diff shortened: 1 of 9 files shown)


Furisto (approved):

Furisto:
Looks good to me. Can you please squash the formatting commit @YJDoc2 . I would like to avoid have all these formatting commits in the history.

utam0k (approved):

YJDoc2:
Hey, I have squashed the format commit, and will do the same in case of such issues in future as well. If everything else is fine, please merge.

PR #161, opened 2021-07-26
Cleanup state file path construction

saschagrunert:
#### Introduction

Hey folks, I started looking into this project, because I didn't even know that it exists since a couple of days. I'm Sascha and I'm already working on a couple of other container related topics. I can probably spent some time contributing to this project during my hack days if you're still looking for contributors. :upside_down_face: 

#### The change

This small cleanup creates a new static method `State::file_path()` which can be used to retrieve the state JSON file path. This makes the state module public and allows its usage from the `commands` module.

### DIFF
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -10,7 +10,7 @@ use chrono::{DateTime, Local};
 use clap::Clap;
 use tabwriter::TabWriter;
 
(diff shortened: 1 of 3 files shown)


Furisto (approved):

Furisto:
Thanks @saschagrunert

Issue #163, opened 2021-07-27
Using clone(2) to create the init container process segfault when build with release version of the youki binary


yihuaf:
When building the released version of youki binary, the clone(2) used to create the container process will fail with the following error from dmesg:
```
[176643.245728] youki[340675]: segfault at 529 ip 00005581f9b434a0 sp 00007fff14905090 error 4 in youki[5581f9a20000+157000]
[176643.245741] Code: 00 00 ff d5 4c 89 bc 24 00 1c 00 00 48 89 9c 24 08 1c 00 00 4c 89 b4 24 10 1c 00 00 48 8b 84 24 18 03 00 00 48 8b 00 48 8b 00 <4c> 8b b8 28 05 00 00 48 8b 98 38 05 00 00 41 be 01 00 00 00 bd 01
```

Preliminary investigation points to the child stack pointer passed to clone(2) call. The debug version of the binary is not affected. Changing the allocated child stack size to 1MB or 2MB seems to make the issue go away, but I'd like a proper fix. Using this issue to track.

yihuaf:
@utam0k I suspect in addition to PR #167, we would need to fix the issue mentioned here: https://github.com/nix-rust/nix/issues/919. The whole discussion on boxed closure in https://github.com/nix-rust/nix/issues/919 is a bit hard to wrap my head around, but I think it means the closure (the cb function in the clone call) context (the arguments to the closure) and its ownership belongs to the parent process instead of the child process callbacks.

PR #164, opened 2021-07-27
bump up to nix-0.22.0

utam0k:
I tried to see if  I could possibly solve #163, but no luck. The upgrade is not bad and I sent out a PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,9 +403,9 @@ checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 
 [[package]]
 name = "libc"
-version = "0.2.95"
+version = "0.2.98"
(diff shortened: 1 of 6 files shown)


src/process/parent.rs
```diff
@@ -147,6 +147,7 @@ impl ChildChannel {
     pub fn wait_for_child_ready(&mut self, child_pid: Pid) -> Result<()> {
         // Create collection with capacity to store up to MAX_EVENTS events
         let mut events = Events::with_capacity(MAX_EVENTS);
+        log::error!("wait for child start");
```
Furisto, line 150: Should this be an error?

utam0k: Oops, I forgot to clear this log while debugging. I fixed it.

Furisto (approved):

PR #165, opened 2021-07-28
Use `remove_dir` instead of `remove_dir_all` because youki doesn't have permission to delete contents in cgroup directory.

utam0k:
I ran the delete in the tutorial and got an error.

https://doc.rust-lang.org/std/fs/fn.remove_dir_all.html
>  Removes a directory at this path, **after removing all its contents**. Use carefully!

https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html
> A cgroup which doesn’t have any children or live processes can be destroyed by removing the directory. Note that a cgroup which doesn’t have any children and is associated only with zombie processes is considered empty and can be removed:
> rmdir $CGROUP_NAME

### DIFF
--- a/src/cgroups/v2/manager.rs
+++ b/src/cgroups/v2/manager.rs
@@ -142,7 +142,7 @@ impl CgroupManager for Manager {
 
     fn remove(&self) -> Result<()> {
         log::debug!("remove cgroup {:?}", self.full_path);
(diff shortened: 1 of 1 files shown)


YJDoc2 (approved):
lgtm :+1:

PR #166, opened 2021-07-28
add perf_event to cgroups v1

fbrv:


src/cgroups/v1/perf_event.rs
```diff
@@ -0,0 +1,40 @@
...
+use crate::cgroups::v1::Controller;
+use oci_spec::{LinuxResources};
+use std::path::Path;
+
+pub struct PerfEvent {}
+
+impl Controller for PerfEvent {
+    type Resource = ();
+
+    fn apply(_linux_resources: &LinuxResources, _cgroup_root: &Path) -> anyhow::Result<()> {
```
Furisto, line 11: Please replace with
```suggestion
     use anyhow::Result;

    fn apply(_linux_resources: &LinuxResources, _cgroup_root: &Path) -> Result<()> {
```

src/cgroups/v1/perf_event.rs
```diff
@@ -0,0 +1,40 @@
...
+    use std::fs;
+
+    use nix::unistd::Pid;
+
+    use super::*;
+    use crate::cgroups::{common::CGROUP_PROCS, test::setup};
+
+    #[test]
+    fn test_add_task() {
+        let (tmp, procs) = setup("test_perf_event_apply", CGROUP_PROCS);
```
Furisto, line 31:
```suggestion
        let (tmp, procs) = setup("test_perf_event_add_task", CGROUP_PROCS);
```

Furisto (changes requested):

src/cgroups/v1/perf_event.rs
```diff
@@ -0,0 +1,42 @@
+
+use crate::cgroups::v1::Controller;
+use oci_spec::{LinuxResources};
+use std::path::Path;
+
```
Furisto, line 5:
```suggestion
 use anyhow::Result;
```

Furisto:
@fbrv Sorry for the confusion. With that suggestion I meant that you should add the use "above", not literally in the trait implementation.

fbrv:
> @fbrv Sorry for the confusion. With that suggestion I meant that you should add the use "above", not literally in the trait implementation.

My fault, I haven't properly checked the suggestion, still learning rust :) hopefully in the next contributions will not happens anymore

Furisto:
@fbrv You need to run cargo fmt, otherwise the check will not succeed.

fbrv:
> @fbrv You need to run cargo fmt, otherwise the check will not succeed.

Run multiple times but for some reason it did not catch any error. Reinstalled rust and now seems working

Furisto (approved):

Furisto:
🎉

PR #167, opened 2021-07-29
Refactor clone(2) child stack creation.

### LINKED ISSUE #163: Using clone(2) to create the init container process segfault when build with release version of the youki binary

yihuaf:
Fix #163 

After looking into more about how pthread_create and fork sets up child stack, refactored/fix how the child stack was created. Before this PR, as #163 indicates, LLVM with high optimization level will make memory behave not as we originally imagined. Therefore, we delved into lower APIs using `mmap` to create the child stack memory. This also gives us the benefit of reserving the max process/thread stack size on the system, without committing to actual physical memory. As a result, we don't have to hardcode a child's stack size or guess how much we need for the child process.

### DIFF
--- a/src/process/fork.rs
+++ b/src/process/fork.rs
@@ -76,6 +76,13 @@ pub fn clone(mut cb: sched::CloneCb, clone_flags: sched::CloneFlags) -> Result<P
         // the top of the stack address.
         let child_stack_top = child_stack.add(default_stack_size);
 
+        // Using the clone syscall is one of the rare cases where we don't want
+        // rust to manage the child stack memory.  Instead, we want to use
+        // c_void directly here. The nix::sched::clone wrapper doesn't provide
+        // the right interface and its interface can't be changed. Therefore,
+        // here we are using libc::clone syscall directly for better control.
+        // The child stack will be cleaned when exec is called or the child
+        // process terminates.
         libc::clone(
             mem::transmute(callback as extern "C" fn(*mut Box<dyn FnMut() -> isize>) -> i32),
             child_stack_top,
--- a/src/process/fork.rs
+++ b/src/process/fork.rs
@@ -47,10 +47,10 @@ pub fn clone(mut cb: sched::CloneCb, clone_flags: sched::CloneFlags) -> Result<P
     // physical memory upfront.  The stack will grow as needed, up to the size
     // researved, so no wasted memory here. Lastly, the child stack only needs
     // to support the container init process set up code in Youki. When Youki
-    // calls exec into the container payload, exec will reset the stack.
+    // calls exec into the container payload, exec will reset the stack.  Note,
+    // do not use MAP_GROWSDOWN since it is not well supported.
     // Ref: https://man7.org/linux/man-pages/man2/mmap.2.html
     let child_stack = unsafe {
-        // Note, do not use MAP_GROWSDOWN since it is not well supported.
         mman::mmap(
             ptr::null_mut(),
             default_stack_size,
@@ -61,6 +61,9 @@ pub fn clone(mut cb: sched::CloneCb, clone_flags: sched::CloneFlags) -> Result<P
         )?
     };
 
+    // Adds SIGCHLD flag to mimic the same behavior as fork.
+    let signal = sys::signal::Signal::SIGCHLD;
+    let combined = clone_flags.bits() | signal as c_int;
     let res = unsafe {
         // Consistant with how pthread_create sets up the stack, we create a
         // guard page of 1 page, to protect the child stack collision. Note, for
@@ -73,8 +76,6 @@ pub fn clone(mut cb: sched::CloneCb, clone_flags: sched::CloneFlags) -> Result<P
         // the top of the stack address.
         let child_stack_top = child_stack.add(default_stack_size);
 
-        let signal = sys::signal::Signal::SIGCHLD;
-        let combined = clone_flags.bits() | signal as c_int;
         libc::clone(
             mem::transmute(callback as extern "C" fn(*mut Box<dyn FnMut() -> isize>) -> i32),
             child_stack_top,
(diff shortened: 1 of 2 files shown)


yihuaf:
@utam0k @Furisto I think this is a proper fix to #163. The current code is definitely wrong to allocate the child stack memory on the stack of the current process. Let me know what you think.

src/process/fork.rs
```diff
@@ -1,15 +1,84 @@
...
+    // Use sysconf to find the page size. If there is an error, we assume
+    // the default 4K page size.
+    let page_size: usize = unsafe {
+        match libc::sysconf(libc::_SC_PAGE_SIZE) {
+            -1 => 4 * 1024, // default to 4K page size
+            x => x as usize,
+        }
+    };
+
+    // Find out the default stack max size through getrlimit.
```
utam0k, line 30: Could you comment that we are waiting for this PR?
https://github.com/nix-rust/nix/pull/1302
I'd like to avoid direct use of libc if possible.

src/process/fork.rs
```diff
@@ -1,15 +1,84 @@
...
+    // which is 8MB on most of the linux system today. This is OK since mmap
+    // will only researve the address space upfront, instead of allocating
+    // physical memory upfront.  The stack will grow as needed, up to the size
+    // researved, so no wasted memory here. Lastly, the child stack only needs
+    // to support the container init process set up code in Youki. When Youki
+    // calls exec into the container payload, exec will reset the stack.
+    // Ref: https://man7.org/linux/man-pages/man2/mmap.2.html
+    let child_stack = unsafe {
+        // Note, do not use MAP_GROWSDOWN since it is not well supported.
+        libc::mmap(
```
utam0k, line 51: Could we use this one?
https://docs.rs/nix/0.22.0/nix/sys/mman/fn.mprotect.html

src/process/fork.rs
```diff
@@ -1,15 +1,84 @@
...
+            0,
+        )
+    };
+
+    let res = unsafe {
+        // Consistant with how pthread_create sets up the stack, we create a
+        // guard page of 1 page, to protect the child stack collision. Note, for
+        // clone call, the child stack will grow downward, so the bottom of the
+        // child stack is in the beginning.
+        Errno::result(libc::mprotect(child_stack, page_size, libc::PROT_NONE))?;
```
utam0k, line 66: Could we use this one?
https://docs.rs/nix/0.22.0/nix/sys/mman/fn.mprotect.html

src/process/fork.rs
Nugine: There is another crate for getrlimit.
Its design is different from nix.
https://docs.rs/rlimit/0.6.2/rlimit/

yihuaf: @utam0k Yes, that PR would provide the nix wrapper for `getrlimit`, which is not there yet. Another way is we use the rlimit crate mentioned here just for this. Your call.

utam0k: > There is another crate for getrlimit.
> Its design is different from nix.
> https://docs.rs/rlimit/0.6.2/rlimit/

Thank you for the wonderful crate introduction. This time, nix is starting to support rlimit, so I'll wait for that.

utam0k: > @utam0k Yes, that PR would provide the nix wrapper for `getrlimit`, which is not there yet. Another way is we use the rlimit crate mentioned here just for this. Your call.

This implementation is fine for this time, but please comment in your code that you are waiting for this PR.

yihuaf: So let's leave the `libc::getrlimit` in this PR and create another issue to track when nix supports `getrlimit`?

src/process/fork.rs
```diff
@@ -1,15 +1,88 @@
...
+        mman::mprotect(child_stack, page_size, mman::ProtFlags::PROT_NONE)
+            .with_context(|| "Failed to create guard page")?;
+
+        // Since the child stack for clone grows downward, we need to pass in
+        // the top of the stack address.
+        let child_stack_top = child_stack.add(default_stack_size);
+
+        let signal = sys::signal::Signal::SIGCHLD;
+        let combined = clone_flags.bits() | signal as c_int;
+        libc::clone(
```
utam0k, line 78: Can I ask you to write down the reason why we aren't using nix's clone?

src/process/fork.rs
utam0k: That's a good idea. Let's do that.

yihuaf: #168 filed.

src/process/fork.rs
yihuaf: I think this is one of the rare cases where we don't want any rust memory management automatically done for us. I wanted to pass in `*c_void` directly, but the nix wrapper has a defined APIs that can't be changed.

utam0k: Sorry, I did not explain myself well. Could you please add a comment to the code about it? At first glance, this code probably seems strange.

utam0k (approved):
superb!

PR #170, opened 2021-07-29
reduce the number of unnecessary clones and arguments by changing the order in which NotifyListeners are generated.

utam0k:
fix lint warning
https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments

### DIFF
--- a/src/commands/start.rs
+++ b/src/commands/start.rs
@@ -16,8 +16,9 @@ pub struct Start {
 
 impl Start {
     pub fn new(container_id: String) -> Self {
-        return Self { container_id };
+        Self { container_id }
     }
+
     pub fn exec(&self, root_path: PathBuf) -> Result<()> {
         let container_root = root_path.join(&self.container_id);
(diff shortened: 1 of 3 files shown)


yihuaf:
Just a thought, maybe it is better to create an arg struct and have `container_init` takes in the arg? The `container_init` function needs to be passed to the child process, where these args need to be cloned, so the memory can be owned by the child process. The only concern is this arg struct would be 70% similar member to `ContainerBuilderImpl`. In addition, likely we will have to pass extra args into the container_init function in the future as we support more features. "preserve-fds" is already one more that I am working on. So it sure will be pass the 7 threshold again and again.

yihuaf:
I just created PR #173 and I did exactly what I suggested above, grouping the args into a struct. Let me know what you think about the pros and cons to both approaches :)

utam0k:
https://github.com/containers/youki/pull/173#pullrequestreview-719145441 is better.

PR #173, opened 2021-07-30
Fix how closure is transferred to the clone call.

### LINKED ISSUE #163: Using clone(2) to create the init container process segfault when build with release version of the youki binary

yihuaf:
The main objective of this PR is to fix how the closure is passed into the clone call. Because the new process created by `libc::clone` will execute the closure, and because of how rust manages the lifetime of memory, special care is needed to make sure the ownership of the memory associated with the closure is correctly transferred to the new process. Otherwise, the new process will only get a reference of the memory and LLVM can decide to free the memory right after the clone call returns in the parent process.

https://github.com/nix-rust/nix/issues/919 is an explanation of why `nix::sched::clone` wrapper has the exact same issue and is not implemented correctly. While there is a fix, https://github.com/nix-rust/nix/pull/920, but it seems to be abandoned since 2018. I took the hint from that PR as the basis of this PR.

Note: this patch is a bit complicated than I originally intended. So please double check if the comments around how the closure is passed to clone in fork.rs are easy to understand. This is a case where Rust makes this more painful but also illustrates that rust does force you to reason through how your code allocates memory correctly.

A follow up to fix #163

### DIFF
--- a/src/container/builder_impl.rs
+++ b/src/container/builder_impl.rs
@@ -68,15 +68,23 @@ impl ContainerBuilderImpl {
         let (mut parent, parent_channel) = parent::ParentProcess::new(self.rootless.clone())?;
         let mut child = child::ChildProcess::new(parent_channel)?;
 
-        let cb = Box::new(|| {
+        let init = self.init;
+        let rootless = self.rootless.clone();
+        let spec = self.spec.clone();
+        let syscall = self.syscall.clone();
+        let rootfs = self.rootfs.clone();
+        let console_socket = self.console_socket.clone();
+        let notify_path = self.notify_path.clone();
+
+        let cb = Box::new(move || {
             if let Err(error) = container_init(
-                self.init,
-                self.rootless.clone(),
-                self.spec.clone(),
-                self.syscall.clone(),
-                self.rootfs.clone(),
-                self.console_socket.clone(),
-                self.notify_path.clone(),
+                init,
+                rootless,
+                spec,
+                syscall,
+                rootfs,
+                console_socket,
+                notify_path,
                 &mut child,
             ) {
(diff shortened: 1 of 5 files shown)


src/container/builder_impl.rs
```diff
@@ -21,6 +21,24 @@ use crate::{
 
 use super::{Container, ContainerStatus};
 
+struct ContainerInitArgs {
```
utam0k, line 24: Could you have the code placed a little closer to container_init() to make the relationship easier to understand? Or cutting it out into a file would be good too.

utam0k (changes requested):
I commented on this, and I agree with you overall. I'll close the PR on my end.

src/container/builder_impl.rs
yihuaf: I moved it closer to container_init().

utam0k (approved):
awesome!

PR #174, opened 2021-07-30
Generalize OCI spec root

saschagrunert:
We now generalize and document the OCI `Spec` root structure. This means
that some fields have been added and other are now optional.

All corresponding usages of the new spec format have been changed and
tests have been adapted.

### DIFF
--- a/src/container/builder_impl.rs
+++ b/src/container/builder_impl.rs
@@ -59,7 +59,7 @@ impl ContainerBuilderImpl {
     fn run_container(&mut self) -> Result<()> {
         prctl::set_dumpable(false).unwrap();
 
-        let linux = &self.spec.linux;
+        let linux = self.spec.linux.as_ref().context("no linux in spec")?;
         let cgroups_path = utils::get_cgroup_path(&linux.cgroups_path, &self.container_id);
         let cmanager = cgroups::common::create_cgroup_manager(&cgroups_path, self.use_systemd)?;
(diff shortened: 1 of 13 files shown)


src/container/builder_impl.rs
```diff
@@ -124,12 +124,12 @@ fn container_init(
...
     child: &mut child::ChildProcess,
 ) -> Result<()> {
-    let linux = &spec.linux;
+    let linux = &spec.linux.as_ref().context("no linux spec")?;
     let namespaces: Namespaces = linux.namespaces.clone().into();
     // need to create the notify socket before we pivot root, since the unix
     // domain socket used here is outside of the rootfs of container
     let mut notify_socket: NotifyListener = NotifyListener::new(&notify_name)?;
-    let proc = &spec.process;
+    let proc = &spec.process.as_ref().context("no process")?;
```
Furisto, line 132: Can you change this to "no process spec" here and everywhere else? My first reaction to "no process" in an error message would be to assume that the container process has exited, not that some part of the spec is missing.

saschagrunert: Yep, changed as suggested.

Furisto (approved):

Furisto:
@saschagrunert Looks good to me. As @utam0k has taken the lead on this, I will wait for his review before merge.

utam0k (approved):
great!

Issue #185, opened 2021-08-05
`youki exec` did not join the correct pid namespace


yihuaf:
~~This will lead `youki exec` to fail at the moment.~~
```
[DEBUG src/container/builder_impl.rs:83] 2021-08-05T08:28:56.915335338+02:00 failed to run container_init: Failed to clean up extra fds

Caused by:
    0: Failed to obtain opened fds
    1: /proc/self/fd is not the actual procfs
    2: No such file or directory (os error 2)
```

Edit: this is just the side effect, where `/proc/self` is not set correctly. The root cause is that the `youki exec` process didn't enter the correct pid namespace. The `setns` for pid namespace require an additional fork/clone to enter into the right namepsace.

yihuaf:
Looks like we didn't set the `/proc` correctly when doing `youki exec`. As a result, there is no `/proc/self` for the process that executes the `youki exec`. I may have to spend some time trying to go through the prepre_rootfs code.

yihuaf:
@utam0k @Furisto @tsturzl The root cause of this is that the `youki exec` process didn't join the correct pid namespaces. While `clone` allows us to create new user and pid namespaces by creating only 1 process, the `setns` for pid namespace has the same limitation. When calling `setns` for pid namespace, only the future children will join the pid namespace, instead of the calling process itself.

I am not sure what is the right approach to resolve this yet. Perhaps we need to revert the change and clone/fork two processes after all.

tsturzl:
I will take a look and see if I have any ideas, otherwise perhaps the double fork is the right solution for now. I'm not entirely sure what runc and crun are doing, and I'd like to look deeper at what they're doing. I think doing what runc is doing is a pretty reliable baseline for us to work off of, because at worst we are only as efficient as what's currently the standard.

yihuaf:
> I will take a look and see if I have any ideas, otherwise perhaps the double fork is the right solution for now. I'm not entirely sure what runc and crun are doing, and I'd like to look deeper at what they're doing. I think doing what runc is doing is a pretty reliable baseline for us to work off of, because at worst we are only as efficient as what's currently the standard.

Thanks. I have a feeling that we may have to revert to creating two processes but want to take a bit of time to think it over. As far as I know, `crun` uses double fork similar to what we used to do. `runc` internally does two `clone` calls to solve the user and pid namespace issue in a similar way. `runc` actually does one more fork because of the limitation of golang.

tsturzl:
[...]

PR #187, opened 2021-08-05
Implemented hooks

### LINKED ISSUE #13: impl: hooks
Hooks are a vital aspect when mapping complex input devices i.e. gpu or accelerator devices into the container namespace. The most common use case is using machine learning workloads in a container workload that requires cuda access i.e. for CI jobs of [juice](https://github.com/spearow/juice)

Spec ref: https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks

### LINKED ISSUE #201: Handle annotations correctly for the container state
Currently, Youki doesn't store annotations into the container state. This also cause the `hook_stdin` test to fail.

yihuaf:
Fix #13 Fix #201
~~Note: I left `PreStart` hooks out since the spec mentioned that it is deprecated. Adding it takes no effort in the future.~~
Edit: I have to add this since integration and docker seems to use this.
Note2: I disabled `preserve-fds` for `exec` due to #185, so at least this code can be tested manually. This will be fixed immediatly afterwards.
Note3: We may have to do a refactor to re-org the code a little bit, but probably should do it in a different PR.


utam0k:
@yihuaf Could you add the hooks-related integration tests for runtime-tools?

yihuaf:
@utam0k PTAL

Note: hook_stdin is failing at the moment because annotation is not handled correctly in Youki. Since this PR is already large by itself, the annotation issue should be another PR. Otherwise, the hook related tests are passing.

Cargo.toml
```diff
@@ -35,6 +35,7 @@ systemd = { version = "0.8", default-features = false, optional = true }
 dbus = "0.9.2"
 tabwriter = "1"
 fastrand = "1.4.1"
+crossbeam-channel = "0.5"
```
utam0k, line 38: Is there any way to narrow down the features?
https://github.com/crossbeam-rs/crossbeam/blob/02e74146ca0b871974fccda579911bf57a9c0ea9/Cargo.toml#L18-L41

src/process/init.rs
```diff
@@ -249,18 +260,34 @@ pub fn container_init(args: ContainerInitArgs) -> Result<()> {
...
+            hooks::run_hooks(hooks.start_container.as_ref(), container)?
+        }
+    }
+
     if let Some(args) = proc.args.as_ref() {
-        utils::do_exec(&args[0], args, &envs)?;
+        utils::do_exec(&args[0], args)?;
     } else {
-        log::warn!("The command to be executed isn't set")
+        bail!("The command to be executed isn't set")
```
utam0k, line 290: Can I have the error message include the fact that args are required for non-Windows?
https://github.com/opencontainers/runtime-spec/blob/master/config.md#process

Cargo.toml
yihuaf: I think we already do. If I understand how this crate works, there is crossbeam with feature ("crossbeam-channel/std" for example), or use `crossbeam-channel` directly. crossbeam with feature "crossbeam-channel/std" seems to point to `crossbeam-channel` underneath.

For context, to be honest, I only used crossbeam because ` std::sync::mpsc` documentation [here](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv_timeout) mentioned [this unfixed issue](https://github.com/rust-lang/rust/issues/39364#issuecomment-880136715) where it mentioned `crossbeam` being a potential replacement for the ` std::sync::mpsc`,

src/hooks.rs
```diff
@@ -0,0 +1,196 @@
...
+            };
+            let hooks = Some(vec![hook]);
+            run_hooks(hooks.as_ref(), Some(&default_container))?;
+        }
+
+        {
+            // Use `printenv` to make sure the environment is set correctly.
+            let default_container: Container = Default::default();
+            let hook = Hook {
+                path: PathBuf::from("/bin/printenv"),
```
utam0k, line 150: In my environment, the test failed because of this different path.
```sh
$ which printenv
/usr/bin/printenv
```

src/hooks.rs
```diff
@@ -0,0 +1,196 @@
...
+                        let _ = signal::kill(hook_command_pid, signal::Signal::SIGKILL);
+                        return Err(HookTimeoutError.into());
+                    }
+                    Err(_) => {
+                        unreachable!();
+                    }
+                }
+            } else {
+                hook_command.wait()
+            };
```
utam0k, line 96: I've refactored it a bit. What do you think?
```suggestion
            let mut hook_command = process::Command::new(&hook.path);
            if let Some((arg0, args)) = hook.args.as_ref().map(|a| a.split_first()).flatten() {
                log::debug!("run_hooks arg0: {:?}, args: {:?}", arg0, args);
                hook_command.arg0(arg0).args(args)
            } else {
                hook_command.arg0(&hook.path.as_path().display().to_string())
            };

            let mut hook_process = hook_command
                .env_clear()
                .envs(envs)
                .stdin(process::Stdio::piped())
                .spawn()
                .with_context(|| "Failed to execute hook")?;
            let hook_process_pid = Pid::from_raw(hook_process.id() as i32);
            // Based on the OCI spec, we need to pipe the container state into
            // the hook command through stdin.
            if let Some(stdin) = hook_process.stdin.as_ref() {
                serde_json::to_writer(stdin, state)?;
            }

            let res = if let Some(timeout_sec) = hook.timeout {
                // Rust does not make it easy to handle executing a command and
                // timeout. Here we decided to wait for the command in a
                // different thread, so the main thread is not blocked. We use a
                // channel shared between main thread and the wait thread, since
                // the channel has timeout functions out of the box. Rust won't
                // let us copy the Command structure, so we can't share it
                // between the wait thread and main thread. Therefore, we will
                // use pid to identify the process and send a kill signal. This
                // is what the Command.kill() does under the hood anyway. When
                // timeout, we have to kill the process and clean up properly.
                let (s, r) = crossbeam_channel::unbounded();
                thread::spawn(move || {
                    let res = hook_process.wait();
                    let _ = s.send(res);
                });
                match r.recv_timeout(time::Duration::from_secs(timeout_sec as u64)) {
                    Ok(res) => res,
                    Err(crossbeam_channel::RecvTimeoutError::Timeout) => {
                        // Kill the process. There is no need to further clean
                        // up because we will be error out.
                        let _ = signal::kill(hook_process_pid, signal::Signal::SIGKILL);
                        return Err(HookTimeoutError.into());
                    }
                    Err(_) => {
                        unreachable!();
                    }
                }
            } else {
                hook_process.wait()
            };
```

yihuaf: I like this.

utam0k (approved):

utam0k:
@yihuaf Thanks!

PR #193, opened 2021-08-07
Improve looking up the root directory

Furisto:
This improves upon the initial solution for looking up the root directory that I introduced in the PR for rootless containers. It checks
- XDG_RUNTIME_DIR
- the location XDG_RUNTIME_DIR usually points to
- the home directory if the user has one
- tmp


utam0k (approved):
lgtm

PR #194, opened 2021-08-08
reduce the number of clones by introducing lifetime to rootless.

utam0k:
I'm not sure if I should name the lifetime parameters.

### DIFF
--- a/src/container/builder_impl.rs
+++ b/src/container/builder_impl.rs
@@ -70,7 +70,7 @@ impl ContainerBuilderImpl {
             spec: self.spec.clone(),
             rootfs: self.rootfs.clone(),
             console_socket: self.console_socket.clone(),
-            rootless: self.rootless.clone(),
+            is_rootless: self.rootless.is_some(),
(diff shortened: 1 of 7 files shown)


Furisto:
I am not sure about this change. This would trade ergonomics (no need to ensure that lifetimes match up) for performance (less clones). Is the performance gain significant though?

utam0k:
@Furisto 
Thank you for your review. Performance gains are currently small. However, heavy use of `clone()` can be a performance problem. It is possible to introduce a lifetime after this problem surfaces, but that could be a major code change. I thought that we should introduce something simple. Of course, the code will be more or less complex, so there is a trade-off there. I'd love to hear your opinions on this.
This PR is a small step at the moment, but it is a step towards introducing order in the use of lifetime in youki's code.

utam0k:
Of course, this is not to prohibit the use of clones, but to reduce the use of clones in simple situations. It is useful and should be used in situations where the code becomes unreasonably complex.

Furisto (approved):

PR #196, opened 2021-08-11
support readonly path

### LINKED ISSUE #144: implementation of readonly paths
ref: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#readonly-paths

duduainankai:
Close #144


src/process/init.rs
```diff
@@ -257,6 +266,29 @@ pub fn container_init(args: ContainerInitArgs) -> Result<()> {
...
+        MsFlags::MS_BIND
+            | MsFlags::MS_REC
+            | MsFlags::MS_NOSUID
+            | MsFlags::MS_NODEV
+            | MsFlags::MS_NOEXEC
+            | MsFlags::MS_BIND
+            | MsFlags::MS_RDONLY,
+        None::<&str>,
+    ) {
+        // ignore error if path is not exist.
```
utam0k, line 283: Could you add a warn log?

utam0k (changes requested):
@duduainankai 
Thanks for the PR!
It seems a bit detailed, but I commented on it and would appreciate a correction.

utam0k (approved):
Thanks!

PR #198, opened 2021-08-11
Move cgroups into own crate

Furisto:
This moves the cgroup code into its own crate.


Furisto:
@saschagrunert This could be interesting to you as well.

saschagrunert:
Good idea! Unfortunately this needs a rebase.

utam0k (approved):
One part(e.g. `default_devices`) is still dependent on youki, on it may be too early to provide it as a crate, but this pr is a good first step.

PR #202, opened 2021-08-15
Reflected that oci_spec has been moved to a separate repository

utam0k:
ref https://github.com/containers/youki/issues/179


Cargo.toml
```diff
@@ -29,15 +29,15 @@ mio = { version = "0.7.13", features = ["os-ext", "os-poll"] }
 chrono = { version="0.4", features = ["serde"] }
 once_cell = "1.6.0"
 futures = { version = "0.3", features = ["thread-pool"] }
-oci_spec = { version = "0.1.0", path = "./oci_spec" }
+oci_spec = { git = "https://github.com/containers/oci-spec-rs", rev = "e0de21b89dc1e65f69a5f45a08bbe426787c7fa1"}
```
yihuaf, line 32: Should we publish the `oci-spec-rs` and/or use tag, instead of commit?

utam0k: It should, but I can't use it tag because I haven't published it to crates.io yet. Or does it indicate a git tag?
In any case, I fear that if we don't isolate this soon, I'll accidentally make changes to the oci_spec directory.

yihuaf: That's a good reason.

yihuaf (approved):

yihuaf:
LGTM, I will let you hit merge.

PR #203, opened 2021-08-16
Support unified resource section

Furisto:
Adds support for the [unified](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#unified) section of the oci runtime spec. This does not map to a controller, so I decided not to implement it as one either. Once we have another one of these "pseudo" controllers (like devices) it probably makes sense to introduce a PseudoControllerType but for now I wanted to keep it simple.

While I was at it, I also added some additional context in case of errors occurring while applying resource restrictions.

### DIFF
--- a/cgroups/src/common.rs
+++ b/cgroups/src/common.rs
@@ -227,27 +227,3 @@ impl PathBufExt for PathBuf {
         Ok(PathBuf::from(format!("{}{}", self.display(), p.display())))
     }
 }
-
-pub(crate) trait StringExt {
-    // this should be replaced by split_once, when we switch to Rust 2021
-    fn split_one(&self, delimiter: &str) -> Option<(&str, &str)>;
-}
-
-impl StringExt for str {
-    fn split_one(&self, delimiter: &str) -> Option<(&str, &str)> {
-        let mut splits = self.splitn(2, delimiter);
-        let first = splits.next()?;
-        let second = splits.next()?;
-
-        Some((first, second))
-    }
-}
-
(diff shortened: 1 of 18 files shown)


utam0k:
@tsturzl Can I ask for a review?

cgroups/src/common.rs
```diff
@@ -227,3 +227,27 @@ impl PathBufExt for PathBuf {
...
+        let mut splits = self.splitn(2, delimiter);
+        let first = splits.next()?;
+        let second = splits.next()?;
+
+        Some((first, second))
+    }
+}
+
+#[macro_export]
+macro_rules! hashmap {
```
utam0k, line 247: What about this one, available since 1.51?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4a06f285da06c18cab7164b9e83bff1f

Furisto:
@utam0k PTAL

cgroups/src/v1/cpu.rs
```diff
@@ -26,7 +26,7 @@ impl Controller for Cpu {
         log::debug!("Apply Cpu cgroup config");
 
         if let Some(cpu) = Self::needs_to_handle(linux_resources) {
-            Self::apply(cgroup_root, cpu)?;
+            Self::apply(cgroup_root, cpu).context("failed to apply cpu resource restrictions")?;
```
utam0k, line 29: :+1:

utam0k (approved):
Perfect!

PR #206, opened 2021-08-19
make sure integration tests complete in ubuntu 20.04 enviroment.

utam0k:


integration_test.sh
```diff
@@ -65,6 +65,17 @@ test_cases=(
   # "start/start.t"
   "state/state.t"
 )
+
+check_enviroment() {
+  test_case=$1
+  if [[ $test_case =~ linux_cgroups* ]]; then
```
Furisto, line 71: Could we check for *memory instead? Or is that not possible because the method that checks the test results checks this file even if it is not a memory related test case? I think I remember something like that...

utam0k: The only test cases that did not pass were hugetlb and memory, so I made sure to skip those test cases.

Furisto (approved):

Furisto:
lgtm

PR #208, opened 2021-08-20
[WIP] cgroups v2: PoC of devices controller

MoZhonghua:
This a PoC of cgroup v2 BPF devices controller, based on the implementation of runc.

It works but with many FIXMEs and TODOs:
1. build BPF bytecode instruction by instruction using ebpf
2. attach/detach using libbpf-sys,  always attach new program and detach old programs, no BF_F_REPLACE
3. use naive algorithm to check rules, runc's device emulator is not implemented
4. there is a simple binary in cgroups/examples/bpf.rs, used to query/attach/detach BPF program, should help testing
5. toggled on by feature "cgroupsv2_devices", which is enabled by default in Cargo.toml

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,7 +56,7 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
       - run: sudo apt-get -y update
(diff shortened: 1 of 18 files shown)


Furisto:
lgtm

Furisto (approved):

PR #210, opened 2021-08-21
Fail fast to create a container if bundle path is illegal

tiqwab:
After `youki create` fails with an illegal bundle path, the container directory remains but it cannot be deleted by youki commands because `state.json` doesn't exist.

```
$ sudo youki create -b foo mycontainer
[DEBUG src/container/init_builder.rs:89] 2021-08-21T04:11:06.571670587+00:00 container directory will be "/run/youki/mycontainer"
Error: load spec: failed to open "foo/config.json"

Caused by:

$ sudo youki list
ID  PID  STATUS  BUNDLE  CREATED  CREATOR

$ ls /var/run/youki/
mycontainer

$ ls -al /var/run/youki/mycontainer/
total 0
drwxr-xr-x. 2 root root 40 Aug 21 04:11 .
drwxr-xr-x. 3 root root 60 Aug 21 04:11 ..

$ sudo youki delete mycontainer
[DEBUG src/commands/delete.rs:23] 2021-08-21T04:11:35.678810300+00:00 start deleting mycontainer
[DEBUG src/commands/delete.rs:32] 2021-08-21T04:11:35.678928184+00:00 load the container from "/run/youki/mycontainer"
Error: failed to open container state file "/run/youki/mycontainer/state.json"

Caused by:
```

This pull request fixes it by loading spec in bundle before creating container directory.

### DIFF
--- a/src/container/init_builder.rs
+++ b/src/container/init_builder.rs
@@ -39,8 +39,9 @@ impl InitContainerBuilder {
 
     /// Creates a new container
     pub fn build(self) -> Result<()> {
+        let spec = self.load_spec()?;
         let container_dir = self.create_container_dir()?;
-        let spec = self.load_and_safeguard_spec(&container_dir)?;
+        self.save_spec(&spec, &container_dir)?;
 
         let container_state = self
             .create_container_state(&container_dir)?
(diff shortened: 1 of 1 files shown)


utam0k (approved):
Perfect. Thanks for your first contribution!

PR #212, opened 2021-08-21
ensure theat read only paths work properly.

utam0k:
In the case of read-only, the flags are ignored on the first bind mount, so it has to be mounted twice.

https://man7.org/linux/man-pages/man2/mount.2.html
> Creating a bind mount
> ....
>       The filesystemtype and data arguments are ignored.
>       The remaining bits (other than MS_REC, described below) in the
>       mountflags argument are also ignored.  (The bind mount has the
>       same mount options as the underlying mount point.)  However, see
>       the discussion of remounting above, for a method of making an
>       existing bind mount read-only.


Furisto (approved):

Furisto:
lgtm

Issue #207, opened 2021-08-19
Pass the process integration test 


utam0k:
# Overview
The test is failing because youki is not looking at the `cwd` settings. This PR fixes this problem
https://github.com/opencontainers/runtime-spec/blob/master/config.md#process

# Goal
Pass the [process](https://github.com/opencontainers/runtime-tools/blob/master/validation/process/process.go) integration test in runtime-tools. 
If you comment in this part and the test passes, you are all set.
https://github.com/containers/youki/blob/4a0262a3ddc10b81d2cdfc0780036cca416442dd/integration_test.sh#L56

utam0k:
I'd like to request this issue to @guni1192 who is interested in youki on Twitter.

PR #215, opened 2021-08-22
Chdir to process.cwd before starting container to pass integration test

guni1192:
ref: #207

Integration Test case: `process/process.t`
This test case is failed because youki does not change directory to `proecss.cwd` before starting container.
I tried fix this.

### DIFF
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -213,18 +213,17 @@ pub fn container_init(args: ContainerInitArgs) -> Result<()> {
         }
     }
 
-    let mut do_chdir = !proc.cwd.is_empty();
-    // change directory to process.cwd if process.cwd is not empty
-    if do_chdir {
+   let do_chdir = if proc.cwd.is_empty() {
+        false
+    } else {
         // This chdir must run before setting up the user.
         // This may allow the user running youki to access directories
         // that the container user cannot access.
         match unistd::chdir(&*proc.cwd) {
-            Ok(_) => do_chdir = false,
-            Err(nix::Error::EPERM) => {}
-            Err(e) => return Err(anyhow::anyhow!("Failed to chdir: {}", e)),
-        };
-    }
+            Ok(_) => false,
+            Err(nix::Error::EPERM) => true,
+            Err(e) => bail!("Failed to chdir: {}", e),
+    };
 
     command.set_id(Uid::from_raw(proc.user.uid), Gid::from_raw(proc.user.gid))?;
     capabilities::reset_effective(command)?;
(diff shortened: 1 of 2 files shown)


src/process/init.rs
```diff
@@ -213,6 +213,20 @@ pub fn container_init(args: ContainerInitArgs) -> Result<()> {
         }
     }
 
+    let mut do_chdir = !proc.cwd.is_empty();
+    // change directory to process.cwd if process.cwd is not empty
+    if do_chdir {
+        do_chdir = false;
+        // This chdir must run before setting up the user.
```
utam0k, line 220: I hadn't thought of this. Great comment :+1:

src/process/init.rs
```diff
@@ -264,6 +278,11 @@ pub fn container_init(args: ContainerInitArgs) -> Result<()> {
         cleanup_file_descriptors(preserve_fds).with_context(|| "Failed to clean up extra fds")?;
     }
 
+    // change directory to process.cwd if process.cwd is not empty
+    if do_chdir {
```
utam0k, line 282: I felt like it would always be false here, but in what cases would it be executed?

guni1192: Oh, sorry, I forgot to delete line 219.
I'll fix it right away.
This chdir will be executed when the first chdir fails by EPERM.

src/process/init.rs
```diff
@@ -213,6 +213,19 @@ pub fn container_init(args: ContainerInitArgs) -> Result<()> {
...
+        // This chdir must run before setting up the user.
+        // This may allow the user running youki to access directories
+        // that the container user cannot access.
+        match unistd::chdir(&*proc.cwd) {
+            Ok(_) => do_chdir = false,
+            Err(nix::Error::EPERM) => {}
+            Err(e) => return Err(anyhow::anyhow!("Failed to chdir: {}", e)),
+        };
+    }
```
utam0k, line 227: @guni1192  It's fine, but how about this code to remove a `mut`?
```suggestion
   let do_chdir = if proc.cwd.is_empty() {
        false
    } else {
        // This chdir must run before setting up the user.
        // This may allow the user running youki to access directories
        // that the container user cannot access.
        match unistd::chdir(&*proc.cwd) {
            Ok(_) => false,
            Err(nix::Error::EPERM) => true,
            Err(e) => bail!("Failed to chdir: {}", e),
    };
```

guni1192: That sounds better!

utam0k:
@guni1192 Sorry, the code I suggested failed fmt, can you fix it?

guni1192:
@utam0k OK. I fixed.

utam0k (approved):
perfect!

Issue #150, opened 2021-07-20
Make failure message great again


lizhemingi:
Right now if we had a test failed when run `cargo test`, the failure message is missing and we can not locate the assert that failed directly.  

**And if comment the tests code in `tty.rs`, the failure message show up again. Just like the failure message is eaten by the tty.**

Failure test without commenting tests in tty.rs.
![image](https://user-images.githubusercontent.com/9639449/125227109-d6ea8480-e304-11eb-91e9-865f2ae397f0.png)

Failure test with commenting tests in tty.rs.
![image](https://user-images.githubusercontent.com/9639449/125227181-fbdef780-e304-11eb-8100-f26a29903d74.png)

It would be a little annoying for debugging when CI failed. So maybe we should find a way to make failure message come back.

cc @utam0k

yihuaf:
@utam0k I wasn't sure if you want to add `--nocapture` to `cargo test` on CI as a solution? This issue is asking for that.

utam0k:
> @utam0k I wasn't sure if you want to add `--nocapture` to `cargo test` on CI as a solution? This issue is asking for that.

That may be fine for CI, but what about local execution unit tests...

yihuaf:
Sorry I don't follow. For local unit test, you also run `cargo test -- --nocapture` to see the error output in case test fails when developing. It works exactly the same.

duduainankai:
@utam0k @yihuaf 

I think `nocapture` can not meet the requirement of this issue.

https://doc.rust-lang.org/cargo/commands/cargo-test.html#display-options

It seems that the `nocapture` option means `do not capture the output of our tested binary`, something like `println` in our code.

But this issue is looking for the output information by `rust test`, something like which assert in which file is failed (The difference between two screenshots in the description.)

But maybe add `nocapture` to CI is also a good advice.

utam0k:
I agree with this. However, I don't think this is the root solution.
> But maybe add nocapture to CI is also a good advice.

PR #217, opened 2021-08-22
Fix clone(2) with double fork

### LINKED ISSUE #185: `youki exec` did not join the correct pid namespace

yihuaf:
Fix #185

This PR replaces clone(2) with double fork to create the container process. Double fork was the original implementation, and we wisely thought that we could improve it by using a single clone call. Turns out jokes on us that we can't use a single clone call. With clone, if we enter into a new pid namespace, we are OK. However, if we need to set_ns into an existing pid namespace, we are out of luck. So the right path forward needs to be double fork. For a more detailed discussion, see #185 why it doesn't work.

Also, turns out, this is a bigger change than I intended. So I apologize up front for the big change.


docs/doc-draft.md
```diff
@@ -23,24 +23,26 @@ This is diagram as given in #14, which is not actually how this works, but helpf
 sequenceDiagram
 participant U as User
 participant D as Docker
-participant YP as Youki(Parent Process)
-participant YI as Youki(Init Process)
+participant Y_Main as Youki(Main Process)
+participant Y_Intermediate as Youki(Intermeidate Process)
```
utam0k, line 27: I've been thinking about the name a little bit, how about `Joint Process`? This is quite difficult and I'm not sure about it, so I'd like to hear your opinion.

.github/workflows/main.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       - name: Build
         run: ./build.sh --release
       - name: Run tests
-        run: cargo test
+        run: cargo test -- --nocapture
```
utam0k, line 63: Is this necessary?

utam0k:
@yihuaf 
Could you give me an example command that I can use to chuck that this has been fixed correctly?

utam0k:
Maybe, but if we don't use rootless, do we need to do double-fork?

.github/workflows/main.yml
yihuaf: I can remove this. I was debugging an error that I couldn't reproduce on the local machine, so I turned `--nocapture` on. Otherwise, cargo test will not print failed error message.

yihuaf:
> @yihuaf
> Could you give me an example command that I can use to check that this has been fixed correctly?

Try 
```
docker run -it --rm --runtime youki --name youki busybox
```
and then 
```
docker exec -it youki sh
```
In the shell using `docker exec`, check `/proc/self` and `/proc/2` are set up correctly. Before, because `exec` didn't enter the correct pid namespace, `/proc/<pid>` was not configured correctly. In the main shell by `docker run`, you should also be able to see `/proc/2` for the `docker exec` shell.
Pid 1 -> docker run shell
Pid 2 -> docker exec shell

yihuaf:
> Maybe, but if we don't use rootless, do we need to do double-fork?

There are 2 cases where we don't need to double fork.
1. If we don't use rootless and/or don't enter into an existing user namespace, we can enter into Pid namespace on the Youki main process and then fork the init process.
2. If we don't need to enter an existing Pid namespace through `set_ns`, we can use clone to enter user and pid namespace together and create the init process directly.

docs/doc-draft.md
yihuaf: I think `intermediate` is likely the closest description to what the middle/first fork/intermediate process does, tbh. I think init process is probably the right name for the second/init process. I think `main` can be used to describe the main process.

I think the parent/child process was confusing because Youki main process and intermediate process can be parent/child, but also intermediate and init process can be parent/child. So we shouldn't call the `first fork` the child process. This is part of the reason why I refactored the parent and child process structure to just use a channel.

I know this is hard LOL. What is your thinking behind `Joint Process`?

.github/workflows/main.yml
yihuaf: Actually, this seems to be the solution to #150 LOL. Without `--nocapture`, cargo test hides everything.

utam0k:
> > @yihuaf
> > Could you give me an example command that I can use to check that this has been fixed correctly?
> 
> Try
> 
> ```
> docker run -it --rm --runtime youki --name youki busybox
> ```
> 
> and then
> 
> ```
> docker exec -it youki sh
> ```
> 
> In the shell using `docker exec`, check `/proc/self` and `/proc/2` are set up correctly. Before, because `exec` didn't enter the correct pid namespace, `/proc/<pid>` was not configured correctly. In the main shell by `docker run`, you should also be able to see `/proc/2` for the `docker exec` shell.
> Pid 1 -> docker run shell
> Pid 2 -> docker exec shell

I was able to confirm the change. Thank you very much.

main
```
$ docker exec -it youki sh
/ # ps
PID   USER     TIME  COMMAND
    1 root      0:00 sh
    4 root      0:00 ps
```
this pr
```
$ docker exec -it youki sh
/ # ps
PID   USER     TIME  COMMAND
    1 root      0:00 sh
    2 root      0:00 sh
    3 root      0:00 ps
/ #

utam0k:
> There are 2 cases where we don't need to double fork.
> 
> 1. If we don't use rootless and/or don't enter into an existing user namespace, we can enter into Pid namespace on the Youki main process and then fork the init process.
> 2. If we don't need to enter an existing Pid namespace through `set_ns`, we can use clone to enter user and pid namespace together and create the init process directly.

Does this mean that if uid_mapping and gid_mapping are not needed, double-fork is not needed?
I thought it might be a good idea to improve the case where double-fork is not needed. However, once this PR is done, it's good.

docs/doc-draft.md
utam0k: I wanted to emphasize that it is a process of highlighting and joining main and init. This is honestly difficult to name. I think `intermediate` is at least a better name than before, so let's go with that for now.

utam0k:
I was hoping these PRs would pass with this PR fix, but it looks like not.
https://github.com/containers/youki/blob/9acc899239ee0b9d58ea1a81b69884d13ba561fe/integration_test.sh#L37-L40

utam0k:
I will not point out any details in this PR. We will not point out details in this PR, because this PR is very important and will be very hard to merge because of the large number of changes. Those corrections will be made in the PR little by little. I am looking forward to your reviews.

yihuaf:
> I was hoping these PRs would pass with this PR fix, but it looks like not.
> https://github.com/containers/youki/blob/9acc899239ee0b9d58ea1a81b69884d13ba561fe/integration_test.sh#L37-L40

I will take a look at these tests right after this. #221 filed.

yihuaf:
> > There are 2 cases where we don't need to double fork.
> > 
> > 1. If we don't use rootless and/or don't enter into an existing user namespace, we can enter into Pid namespace on the Youki main process and then fork the init process.
> > 2. If we don't need to enter an existing Pid namespace through `set_ns`, we can use clone to enter user and pid namespace together and create the init process directly.
> 
> Does this mean that if uid_mapping and gid_mapping are not needed, double-fork is not needed?
> I thought it might be a good idea to improve the case where double-fork is not needed. However, once this PR is done, it's good.

Actually no. Without uid/gid mapping, we would still require double fork unless we are in one of two cases mentioned above. I think we can potentially improve this by differentiating which case to use, but it would take some effort to get all the corner cases right. The double fork implementation would cover all the cases and should be good enough for the moment.

yihuaf:
> I will not point out any details in this PR. We will not point out details in this PR, because this PR is very important and will be very hard to merge because of the large number of changes. Those corrections will be made in the PR little by little. I am looking forward to your reviews.
> If you respond to some of my comments, I'll try to merge them.
> Finally, it would be helpful if you could leave a brief summary of this issue on issue or on this PR.

That's probably a good idea, and I am happy to continue to fix all the details. I will summarize the issue here again, and #185 contains a history of how we got here.

Before this PR, the implementation was to use `clone` to create the container init process, joining new user and new pid namespaces together. However, in the case of `docker exec`, the container process will join existing namespaces user and pid namespaces. The `set_ns` to pid namespaces requires an extra fork to enter into the correct pid namespace. As a result, the container process didn't join the right pid namespaces. Many of the pid related operations were failing, such as looking at `/proc/self` and `/proc/<pid>`. The solution in this PR is to always double fork, one to enter user namespace and another fork to enter into pid namespace. We can't enter user and pid namespace together because entering into pid namespace requires CAP_SYSADMIN so we have to enter into user namespace first.

utam0k (approved):
Superb!

PR #219, opened 2021-08-23
Add necessary libraries to build youki in Vagrant provision

tiqwab:
Install `systemd-devel` and `dbus-devel` in provisioning step to build youki in VM.


utam0k (approved):
I had completely forgotten about it. Thank you very much.

PR #223, opened 2021-08-24
Show error log only when error happens

chenyukang:


integration_test.sh
```diff
@@ -90,7 +90,11 @@ for case in "${test_cases[@]}"; do
   fi
 
   echo "Running $case"
-  if [ 0 -ne $(sudo RUST_BACKTRACE=1 YOUKI_LOG_LEVEL=debug RUNTIME=${RUNTIME} ${ROOT}/integration_test/src/github.com/opencontainers/runtime-tools/validation/$case | grep "not ok" | wc -l) ]; then
+  logfile="./log/$case.log"
+  mkdir -p "$(dirname $logfile)"
+  sudo RUST_BACKTRACE=1 RUNTIME=${RUNTIME} ${ROOT}/integration_test/src/github.com/opencontainers/runtime-tools/validation/$case >$logfile 2>&1
+  if [ 0 -ne $(cat $logfile | grep "not ok" | wc -l ) ]; then
```
utam0k, line 96: How about this?
```suggestion
  if [ 0 -ne $(grep "not ok" $logfile | wc -l ) ]; then
```

chenyukang: yes, it's better.

utam0k (approved):
Cool!

PR #224, opened 2021-08-24
Fix #209, pass root-readonly

### LINKED ISSUE #209: Pass the `root_read_only` integration test
# Overview
Support read-only root directory.
https://github.com/opencontainers/runtime-spec/blob/master/config.md#process

# Goal
Pass the [root_read_only](https://github.com/opencontainers/runtime-tools/blob/master/validation/root_readonly_true/root_readonly_true.go) integration test in runtime-tools. 
If you comment in this part and the test passes, you are all set.
https://github.com/containers/youki/blob/a8427617836f98e904b504aa183b012d3c440b66/integration_test.sh#L42

# Reference
- [implementation of runc](https://github.com/opencontainers/runc/blob/4d26c4a0a14e55fb71a3f7be2d3cd9f228523fff/libcontainer/rootfs_linux.go#L794-L807)

chenyukang:
close #209

### DIFF
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -311,6 +311,11 @@ pub fn container_init(
         }
     }
 
+    if let Some(true) = spec.root.as_ref().map(|r| r.readonly.unwrap_or(false)) {
+        let flags = MsFlags::MS_RDONLY | MsFlags::MS_REMOUNT | MsFlags::MS_BIND;
+        nix_mount(None::<&str>, "/", None::<&str>, flags, None::<&str>)?
+    }
+
     if let Some(paths) = &linux.readonly_paths {
         // mount readonly path
         for path in paths {
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -312,8 +312,13 @@ pub fn container_init(
     }
 
     if let Some(true) = spec.root.as_ref().map(|r| r.readonly.unwrap_or(false)) {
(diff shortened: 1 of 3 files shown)


chenyukang:
Btw, add pattern matching for running specific tests, it's useful when trying to debug specific test:

```bash 
yukang@mango:~/youki$ ./integration_test.sh readonly
Running linux_readonly_paths/linux_readonly_paths.t
Running root_readonly_true/root_readonly_true.t
```

src/process/init.rs
```diff
@@ -311,6 +311,16 @@ pub fn container_init(
         }
     }
 
+    if let Some(true) = spec.root.as_ref().map(|r| r.readonly.unwrap_or(false)) {
```
utam0k, line 314: 💯

chenyukang:
> > Btw, add pattern matching for running specific tests, it's useful when trying to debug specific test:
> > ```shell
> > yukang@mango:~/youki$ ./integration_test.sh readonly
> > Running linux_readonly_paths/linux_readonly_paths.t
> > Running root_readonly_true/root_readonly_true.t
> > ```
> 
> This is a great change because I used to change `integration.sh` every time! If you don't mind, could you put an example of how to use it somewhere in the README or `integraiton.sh`?

I added an example in README.

utam0k (approved):
cool

Issue #221, opened 2021-08-24
Pass the namespace related integration tests


yihuaf:
Pass these tests:
 "linux_ns_itype/linux_ns_itype.t" 
 "linux_ns_nopath/linux_ns_nopath.t" 
 "linux_ns_path/linux_ns_path.t" 
 "linux_ns_path_type/linux_ns_path_type.t"

yihuaf:
"linux_ns_itype/linux_ns_itype.t" won't fix because runtime-tools has a bug causing the test to hang on Github Action. Ref: https://github.com/opencontainers/runtime-tools/issues/698

PR #238, opened 2021-08-27
Fix graceful shutdown when intermediate or init process errors or panic

### LINKED ISSUE #237: Graceful shutdown Youki when forked process returns an error
Currently, if either the intermediate or the init process errors out, the Youki main process stucks on waiting for child ready message and cause the main process to hang. Some integration tests are deliberately testing for error conditions and requires Youki to shutdown gracefully. Also this is the right thing to do.

yihuaf:
Fix #237 Tracking #221 

Gracefully shutdown by closing down unused fds in the channel. In this case, when the child process exits, the parent waiting (read_exact) is exit blocking and error out. When fork, the fds in the channel are duplicated, so there are two receivers and two senders. Since we force the channel to become uni-directional, we can safely close the additional fds.

Also in the PR, we should only set host name when entering into a new UTS namespace. One of the integration test doesn't create any new uts namespace, so calling set_hostname is not a good idea.

Now pass one more integration test. The test essentially asks Youki to enter into the wrong type of namespace and see if Youki will error out.


src/process/channel.rs
```diff
@@ -107,6 +109,18 @@ impl Channel {
...
+    pub fn close_receiver(&self) -> Result<()> {
+        unistd::close(self.receiver.as_raw_fd())?;
+
+        Ok(())
+    }
+
+    pub fn close_sender(&self) -> Result<()> {
+        unistd::close(self.sender.as_raw_fd())?;
+
+        Ok(())
```
utam0k, line 121: How about this code?
```suggestion
        unistd::close(self.receiver.as_raw_fd())
    }

    pub fn close_sender(&self) -> Result<()> {
        unistd::close(self.sender.as_raw_fd())
```

Furisto (approved):

src/process/channel.rs
yihuaf: unistd functions will return a different error type than anyhow result, so we have to use `?` for a conversion.

PR #241, opened 2021-08-27
Final fix to linux_ns_itype tests

### LINKED ISSUE #240: When using the host mount namespace, we should not pivot_root
Scary things will happen if we run pivot_root without entering into a mount namespace. According to `runc`, we want to use chroot instead.

yihuaf:
Do not pivot_root when under the host mount namespace. Chroot instead.

Fix #240

Also passes "linux_ns_itype/linux_ns_itype.t" test.


yihuaf:
Need #238 to go in first for the integration test to pass.

PR #242, opened 2021-08-28
Use chroot when not entering into mount namespace

### LINKED ISSUE #240: When using the host mount namespace, we should not pivot_root

yihuaf:
Do not pivot_root when under the host mount namespace. Chroot instead.

Fix #240

Also passes "linux_ns_itype/linux_ns_itype.t" test.

### DIFF
--- a/integration_test.sh
+++ b/integration_test.sh
@@ -39,7 +39,7 @@ test_cases=(
   "linux_devices/linux_devices.t"
   # "linux_masked_paths/linux_masked_paths.t"
   "linux_mount_label/linux_mount_label.t"
-  # "linux_ns_itype/linux_ns_itype.t"
+  "linux_ns_itype/linux_ns_itype.t"
   "linux_ns_nopath/linux_ns_nopath.t"
   "linux_ns_path/linux_ns_path.t"
   "linux_ns_path_type/linux_ns_path_type.t"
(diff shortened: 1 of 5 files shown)


utam0k (approved):
perfect!

utam0k:
@yihuaf This test has been running forever. Can you check it out?
```
Running linux_ns_itype/linux_ns_itype.t
```

yihuaf:
> @yihuaf This test has been running forever. Can you check it out?
> 
> ```
> Running linux_ns_itype/linux_ns_itype.t
> ```

Yes I will.

yihuaf:
@utam0k So the hang has been reported since 2019. (https://github.com/opencontainers/runtime-tools/issues/698). I'd rather get this change in without this test. On my local machine, I can't reproduce this. The test passes with flying color. On the Github action, the test passes, but as the issue indicated hang on the clean stage. What do you think?

utam0k:
> @utam0k So the hang has been reported since 2019. ([opencontainers/runtime-tools#698](https://github.com/opencontainers/runtime-tools/issues/698)). I'd rather get this change in without this test. On my local machine, I can't reproduce this. The test passes with flying color. On the Github action, the test passes, but as the issue indicated hang on the clean stage. What do you think?

Oops... Thanks for your research. Let's comment on what goes through locally and the problems and comment it out.
Lastly, just to confirm, if it doesn't work under certain conditions, we can consider skipping it with `check_enviroment()`, but is that going to be difficult? (I don't think you need to  spend that much time investigating this.)

When you will do, it would be helpful if you could also update the README.md in a similar fashion to following PR.

PR #243, opened 2021-08-28
Pass process user integration test

Furisto:
Making this test pass is easy enough (just add the supplementary groups using setgroups system call) but there are some subtleties around this for rootless containers due to CVE-2014-8989.

The gist of this priviledge escalation is: Before 3.19 it was possible for an unprivileged user to enter an user namespace, become root and then call setgroups in order to drop membership in supplementary groups. This allowed access to files which blocked access based on being a member of these groups.

This leaves us with three scenarios:

Unprivileged user starting a rootless container: The main process is running as an unprivileged user and therefore cannot write the mapping until "deny" has been written to /proc/{pid}/setgroups. Once written /proc/{pid}/setgroups cannot be reset and the setgroups system call will be disabled for all processes in this user namespace. This also means that we should detect if the user is unprivileged and additional gids have been specified and bail out early as this can never work.

Privileged user starting a rootless container: It is not necessary to write "deny" to /proc/setgroups in order to create the gid mapping and therefore we don't. This means that setgroups could be used to drop groups, but this is fine as the user is privileged and could do so anyway.
We should check if the specified supplemental groups fall into the range that are specified in the gid mapping and bail out early if they do not. 

Privileged user starting a normal container: Just add the supplementary groups.

I noticed some problems when testing with rootless containers, which are not directly related to this feature and fixed them:
- An attempt was made to move the init process into cgroups which should not be done without systemd support
- When a privileged user starts a rootless container we try to change back into the old directory [here](https://github.com/containers/youki/blob/ff582b7f7f6bf4346294b9191916b105da885a3d/src/rootfs.rs#L80), but this fails in this case because the old directory belonged to the privileged user and we are now in a user namespace with a user mapped to an unprivileged user on the host. I removed changing back to the old directory because we pivot_root right after anyway and therefore this doesn't seem needed. Someone correct me if I am wrong here.

### DIFF
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -368,7 +368,8 @@ pub fn container_init(
         }
     };
 
-    set_supplementary_gids(&proc.user, &args.rootless)?;
+    set_supplementary_gids(&proc.user, &args.rootless)
+        .context("failed to set supplementary gids")?;
 
     command
         .set_id(Uid::from_raw(proc.user.uid), Gid::from_raw(proc.user.gid))
@@ -462,6 +463,30 @@ pub fn container_init(
     unreachable!();
 }
 
+// Before 3.19 it was possible for an unprivileged user to enter an user namespace,
+// become root and then call setgroups in order to drop membership in supplementary
+// groups. This allowed access to files which blocked access based on being a member
+// of these groups (see CVE-2014-8989)
+//
+// This leaves us with three scenarios:
+//
+// Unprivileged user starting a rootless container: The main process is running as an
+// unprivileged user and therefore cannot write the mapping until "deny" has been written
+// to /proc/{pid}/setgroups. Once written /proc/{pid}/setgroups cannot be reset and the
+// setgroups system call will be disabled for all processes in this user namespace. This
+// also means that we should detect if the user is unprivileged and additional gids have
+// been specified and bail out early as this can never work. This is not handled here,
+// but during the validation for rootless containers.
+//
+// Privileged user starting a rootless container: It is not necessary to write "deny" to
+// /proc/setgroups in order to create the gid mapping and therefore we don't. This means
+// that setgroups could be used to drop groups, but this is fine as the user is privileged
+// and could do so anyway.
+// We already have checked during validation if the specified supplemental groups fall into
+// the range that are specified in the gid mapping and bail out early if they do not.
+//
+// Privileged user starting a normal container: Just add the supplementary groups.
+//
 fn set_supplementary_gids(user: &User, rootless: &Option<Rootless>) -> Result<()> {
     if let Some(additional_gids) = &user.additional_gids {
         if additional_gids.is_empty() {
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -480,7 +480,7 @@ fn set_supplementary_gids(user: &User, rootless: &Option<Rootless>) -> Result<()
             .collect();
 
         match rootless {
-            Some(r) if r.priviledged => {
+            Some(r) if r.privileged => {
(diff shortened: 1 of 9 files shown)


src/rootfs.rs
```diff
@@ -68,17 +68,15 @@ pub fn prepare_rootfs(spec: &Spec, rootfs: &Path, bind_devices: bool) -> Result<
         }
     }
 
-    let olddir = getcwd()?;
-    chdir(rootfs)?;
+    chdir(rootfs).context("failed to set current directory to rootfs")?;
```
yihuaf, line 71: Usually I don't like chdir inside one function since it has side effects, making it much harder for ourselves in the future. I prefer to either chdir back, leaving the cwd the same after the function returns, or not chdir at all. In this case, I think the right thing to do is to remove chdir logic all together here. Instead, we make the `setup_default_symlinks` work with rootfs.

yihuaf (commented):
> 
>This leaves us with three scenarios:
>
>Unprivileged user starting a rootless container: The main process is running as an unprivileged user and therefore cannot write the mapping until "deny" has been written to /proc/{pid}/setgroups. Once written /proc/{pid}/setgroups cannot be reset and the setgroups system call will be disabled for all processes in this user namespace. This also means that we should detect if the user is unprivileged and additional gids have been specified and bail out early as this can never work.

>Privileged user starting a rootless container: It is not necessary to write "deny" to /proc/setgroups in order to create the gid mapping and therefore we don't. This means that setgroups could be used to drop groups, but this is fine as the user is privileged and could do so anyway.
We should check if the specified supplemental groups fall into the range that are specified in the gid mapping and bail out early if they do not.
Privileged user starting a normal container: Just add the supplementary groups.

Can you put the discussion into comments about the use of  in the code. Those discussion are really good and I think they belong in the code, rather than PR.

Furisto:
@yihuaf PTAL

src/process/init.rs
```diff
@@ -456,6 +463,64 @@ pub fn container_init(
...
+
+        match rootless {
+            Some(r) if r.privileged => {
+                nix::unistd::setgroups(&gids).context("failed to set supplementary gids")?;
+            }
+            None => {
+                nix::unistd::setgroups(&gids).context("failed to set supplementary gids")?;
+            }
+            // this should have been detected during validation
+            _ => unreachable!(
```
yihuaf, line 515: Under what circumstance where is branch get triggered? Isn't match option can only be either Some or None?

src/process/init.rs
```diff
@@ -456,6 +463,64 @@ pub fn container_init(
...
+//
+// Privileged user starting a normal container: Just add the supplementary groups.
+//
+fn set_supplementary_gids(user: &User, rootless: &Option<Rootless>) -> Result<()> {
+    if let Some(additional_gids) = &user.additional_gids {
+        if additional_gids.is_empty() {
+            return Ok(());
+        }
+
+        let setgroups =
```
yihuaf, line 496: Do we have to check this step and error out early? Can we just go ahead call `setgroups` and the kernel will refuse to setgroups and error out?

yihuaf (approved):
LGTM, and in my opinion much easier to follow with the added comments. Just two minor questions.

src/process/init.rs
Furisto: It should return EPERM. This way we get a better error message.

src/process/init.rs
Furisto: The Some has a match guard, so this is only triggered when the user is not privileged. Ideally I could write
`None | Some(r) if r. privileged`
but Rust does not allow that (yet).

PR #244, opened 2021-08-29
distinguish channels more clearly between each process

utam0k:


src/process/channel.rs
```diff
@@ -10,53 +10,62 @@ use std::io::Read;
...
-
-    pub fn send_child_ready(&mut self, pid: Pid) -> Result<()> {
-        // Send over the ChildReady follow by the pid.
-        log::debug!("sending init pid ({:?})", pid);
-        self.write_message(Message::ChildReady)?;
-        self.sender.write_all(&(pid.as_raw()).to_be_bytes())?;
+impl SenderExt for Sender {
+    #[inline]
+    fn write_message(&mut self, msg: Message) -> Result<()> {
+        self.write_all(&(msg as u8).to_be_bytes())?;
```
yihuaf, line 20: Can we add error context here for error msg?

src/process/channel.rs
```diff
@@ -67,6 +76,58 @@ impl Channel {
...
     }
 
+    pub fn close(&self) -> Result<()> {
+        unistd::close(self.receiver.as_raw_fd())?;
+        Ok(())
+    }
+}
+
+pub fn intermediate_to_main() -> Result<(SenderIntermediateToMain, ReceiverFromIntermediate)> {
+    let (sender, receiver) = new_pipe()?;
```
yihuaf, line 86: Isn't new_pipe already has this part of the code?

src/process/channel.rs
```diff
@@ -83,51 +144,113 @@ impl Channel {
...
+        SenderInitToIntermediate { sender },
+        ReceiverFromInit { receiver },
+    ))
+}
+
+pub struct SenderInitToIntermediate {
+    sender: Sender,
+}
+
+impl SenderInitToIntermediate {
```
yihuaf, line 191: Since we differentiate channel here, the init process don't ready send pid to the intermediate process, so the `init_ready` here don't have to send Pid. In the init process, since it is already inside the pid namespace, it doesn't know what is the real pid outside the pid namespace. The intermediate process knows the pid anyway since it is one doing the fork.

yihuaf (commented):
Can we be more aggressive in putting down context for error message? Other than the comments, LGTM.

utam0k:
@yihuaf PTAL

yihuaf (approved):

Issue #246, opened 2021-08-30
Pass oom_score_adj related integration test


yihuaf:
The goal of this issue is to uncomment this line and pass the integration test.
https://github.com/containers/youki/blob/2d02d22caf2763dd249d943f95d90374618eb10f/integration_test.sh#L64

The oom_score_adj is implemented already.
```rust
    // if Out-of-memory score adjustment is set in specification.  set the score
    // value for the current process check
    // https://dev.to/rrampage/surviving-the-linux-oom-killer-2ki9 for some more
    // information
    if let Some(ref resource) = linux.resources {
        if let Some(oom_score_adj) = resource.oom_score_adj {
            let mut f = fs::File::create("/proc/self/oom_score_adj")?;
            f.write_all(oom_score_adj.to_string().as_bytes())?;
        }
    }
```

However, it is not play nice with `prctl dumpable`, so the operation is actually not having the effect. Likely we have to understand how dumpable and oom_score_adj are interacting.

Ref: https://github.com/opencontainers/runc/blob/ba7a87730dd05937cb1c7f33528b1d460e620434/libcontainer/nsenter/nsexec.c#L651

utam0k:
The goal of this issue is to uncomment this line and pass the integration test.
https://github.com/containers/youki/blob/2d02d22caf2763dd249d943f95d90374618eb10f/integration_test.sh#L64

Onyinye91-ctrl:
@utam0k I will like to work on this issue

utam0k:
@Onyinye91-ctrl sure!

PR #248, opened 2021-08-30
uncomment integration test

Onyinye91-ctrl:
I worked on issue #246 (Uncomment integration test on line 66)

### DIFF
--- a/integration_test.sh
+++ b/integration_test.sh
@@ -63,7 +63,7 @@ test_cases=(
   "process/process.t"
   "process_capabilities/process_capabilities.t"
   "process_capabilities_fail/process_capabilities_fail.t"
-  # "process_oom_score_adj/process_oom_score_adj.t"
+  "process_oom_score_adj/process_oom_score_adj.t"
   "process_rlimits/process_rlimits.t"
   "process_rlimits_fail/process_rlimits_fail.t"
   # "process_user/process_user.t"


utam0k (changes requested):
@Onyinye91-ctrl Thanks for your first contribution! Did `process_oom_score_adj/process_oom_scocre_ajd.t` pass on your machine? This is how it failed on my machine.
```
$ cat process_oom_score_adj/process_oom_score_adj.t.log | grep "not ok"
not ok 279 - has expected OOM score adjustment
```

utam0k:
@Onyinye91-ctrl Sorry for the lack of explanation, the purpose of that issue is to uncomment and the test will pass. Uncommenting is the final result.

yihuaf:
@utam0k Apparently Github Action passed the test. However, the test fails for me on my local box as well. Maybe this is not as straight forward as we initially thought.

utam0k:
@yihuaf Let's just see if it works with runc.

yihuaf:
> @yihuaf Let's just see if it works with runc.

`runc` passes on my box, so the issue is likely with Youki. I will have to dig a bit deeper.

yihuaf:
@utam0k Now I am really confused. I know why it is not passing on the local box. There are actually two `oom_score_adj` definition, one inside `linux.resource.oom_score_adj`. The other one inside `spec.process.oom_score_adj`. The integration test on the local box sets the `process.oom_score_adj`, but not the `linux.resource.oom_score_adj`. The code only look for `linux.resource.oom_score_adj` and sets it, so of course the test fails.

Now it begs the question, how on earth does it pass on Github Action.......

utam0k:
Sorry, this issue was resolved by following PR.
https://github.com/containers/youki/pull/251/files

PR #262, opened 2021-09-03
Add `new` method to instantiate Delete command

alfonsoros88:
We were trying to use `youki` as a library and notice that the `Delete` command was missing the `new` method. It can not be instantiated otherwise. This patch also improves the api consistency with other commands.

### DIFF
--- a/src/commands/delete.rs
+++ b/src/commands/delete.rs
@@ -20,6 +20,17 @@ pub struct Delete {
 }
 
 impl Delete {
+    /// instant Delete Command
+    ///
+    /// This method is provided for those using `youki` as a library to enable the use of the
+    /// Delete command to remove containers.
+    pub fn new(container_id: String, force: bool) -> Self {
+        Self {
+            container_id,
+            force,
+        }
+    }
+
     pub fn exec(&self, root_path: PathBuf, systemd_cgroup: bool) -> Result<()> {
         log::debug!("start deleting {}", self.container_id);
         // state of container is stored in a directory named as container id inside


utam0k:
@alfonsoros88 Thanks for your contribution!  There are `new()` because only Create and Start are special, and there are used for Run. youki's commands aren't currently planned to be provided as a library, but what do you plan to use them for?

alfonsoros88:
@utam0k sorry for the delay! We were trying to use youki's commands by including it as a library dependency to our project. We thought it would be easier to handle it that way instead of using the `youki` binary and having to handle the dependency with a `build.rs`. 

For creating containers we are not using the `Create` command directly, as it is just a wrapper of `ContainerBuilder`. Unfortunately, for cleaning up, the `Delete` command seems to be the one containing the logic for removing containers. Instead of copy/pasting the `exec` function, we thought that we could just use the `Delete` command directly and for that we needed the `new()` method.

utam0k:
@alfonsoros88 Can I ask you to add a comment such as providing for software that uses youki as a library in the comments?

utam0k:
@alfonsoros88 
I see I understand. However, I don't really want to think about providing anything under `src/` as a library. Is there any problem to generate Delete structure directly instead?
```rust
Delete {
  contaienr_id,
  force,
}
```

alfonsoros88:
@utam0k we are trying to use youki in https://github.com/esrlabs/northstar. This project uses its own "kind" of containers, we wanted to add support for OCI containers by using youki underneath.

It is not possible to instantiate `Delete` in that way because the fields are not public.

utam0k:
> @utam0k we are trying to use youki in https://github.com/esrlabs/northstar. This project uses its own "kind" of containers, we wanted to add support for OCI containers by using youki underneath.
> 
> It is not possible to instantiate `Delete` in that way because the fields are not public.

@alfonsoros88  I understand. Let's provide it. Could you add a comment to the effect that you are providing `new()` for those who are not using it as youki, but are using youki as a library?

alfonsoros88:
@utam0k sure thing! Do refer to a commit comment or as a doc comment for the `new()` method?

utam0k:
> @utam0k sure thing! Do refer to a commit comment or as a doc comment for the `new()` method?

Please do so in the doc comments.

utam0k (approved):
Thanks!

PR #265, opened 2021-09-05
Print logfile when test case crash

chenyukang:
Because we run bash with -e,  when a test case crash, the parent process will also exit, without print out the logfile.
This is confusing for run a single crash test case since the printed result is same with a successful one.
Like the "mounts/mounts.t"


utam0k (approved):
cool!

PR #267, opened 2021-09-05
fix unstable the channel tests.

utam0k:
I don't know the cause, but I want to merge this as soon as possible.

### DIFF
--- a/src/process/channel.rs
+++ b/src/process/channel.rs
@@ -230,48 +230,53 @@ fn new_pipe() -> Result<(Sender, Receiver)> {
 }
 
 #[cfg(test)]
+// Tests become unstable if not serial. The cause is not known.
 mod tests {
     use super::*;
     use anyhow::Context;
     use nix::sys::wait;
     use nix::unistd;
+    use serial_test::serial;
 
     #[test]
+    #[serial]
     fn test_channel_intermadiate_ready() -> Result<()> {
         let (sender, receiver) = &mut intermediate_to_main()?;
         match unsafe { unistd::fork()? } {
             unistd::ForkResult::Parent { child } => {
+                wait::waitpid(child, None)?;
                 let pid = receiver
                     .wait_for_intermediate_ready()
                     .with_context(|| "Failed to wait for intermadiate ready")?;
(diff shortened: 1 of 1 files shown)


src/process/channel.rs
```diff
@@ -230,48 +230,53 @@ fn new_pipe() -> Result<(Sender, Receiver)> {
...
                     .with_context(|| "Failed to wait for intermadiate ready")?;
+                receiver.close()?;
                 assert_eq!(pid, child);
-                wait::waitpid(child, None)?;
             }
             unistd::ForkResult::Child => {
                 let pid = unistd::getpid();
-                sender
-                    .intermediate_ready(pid)
-                    .with_context(|| "Failed to send intermadiate ready")?;
```
utam0k, line 254: I removed this because it would have overwritten the error message in the function that the test is running.

Furisto (approved):

yihuaf:
The test should not fail even if running in parallel, in theory. If it fails, I worry there is a race condition somewhere we should fix.

yihuaf:
Is this the error we saw? I can't find the error log in the referenced PR since they seem to be overridden in the PR.
```
failures:

---- hooks::test::test_run_hook stdout ----
Error: Broken pipe (os error 32)

Caused by:
    Broken pipe (os error 32)
thread 'hooks::test::test_run_hook' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/test/src/lib.rs:193:5

---- tty::tests::test_setup_console stdout ----
thread 'tty::tests::test_setup_console' panicked at 'called `Result::unwrap()` on an `Err` value: failed to open console-socket', src/tty.rs:157:40


failures:
    hooks::test::test_run_hook
    tty::tests::test_setup_console

test result: FAILED. 21 passed; 2 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass '--lib'
```

PR #269, opened 2021-09-05
Rework cgroup detection

Furisto:
The original implementation assumed that cgroup v1 und cgroup v2 could be used side by side, but resource controllers are either attached to the one or the other. A system supports one of three modes (if it is systemd based): 
- Legacy: pure v1
- Unified: pure v2
- Hybrid: v1, but with some v2 features activated. Resource control needs to be done with v1 as all controllers are attached to the v1 hierarchy. Going forward, this mode will not be available anymore. 

One can check which filesystems are mounted where in order to determine in which mode the system is (see code comments).

### DIFF
--- a/cgroups/src/common.rs
+++ b/cgroups/src/common.rs
@@ -1,5 +1,4 @@
 use std::{
-    env,
     fmt::{Debug, Display},
     fs::{self, File},
     io::{BufRead, BufReader, Write},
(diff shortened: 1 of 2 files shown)


cgroups/src/common.rs
```diff
@@ -91,92 +94,79 @@ pub fn read_cgroup_file<P: AsRef<Path>>(path: P) -> Result<String> {
...
-    if cgroup2_mount.is_some() {
-        cgroups.push(Cgroup::V2);
+                return Ok(CgroupSetup::Legacy);
+            }
+        }
+        false => bail!("non default cgroup root not supported"),
     }
 
-    Ok(cgroups)
+    bail!("failed to detect cgroup setup");
```
utam0k, line 139: I don't think I'll actually reach the code here.

utam0k:
I have a feeling that when using cgroup v2, the values need to be converted just like crun, is that a problem?
https://github.com/containers/crun/blob/main/crun.1.md#cgroup-v2

cgroups/src/common.rs
Furisto: It will if the filesystem type is not tmpfs.

Furisto:
Cpu is converted [here](https://github.com/containers/youki/blob/b8442af26f7017a38f516fd7db26a29d0cac9a91/cgroups/src/v2/cpu.rs#L97-L103), but memory doesn't seem to be converted. I will fix this in another PR...

utam0k (approved):
LGTM

PR #270, opened 2021-09-05
fix a failure because it is running before checkout.

utam0k:
fix: https://github.com/containers/youki/runs/3518110331


utam0k:
@Furisto If we don't make this fix, can we have a look at the main ci as it keeps failing?

Furisto (approved):

Issue #226, opened 2021-08-24
Introducing Code Coverage


utam0k:
I'd like to consider implementing code coverage, referring to the following repository.
The only thing we need to consider is that we can probably only do it nightly.
https://github.com/containers/oci-spec-rs

utam0k:
I'd like to measure the coverage of the cgroups directory as well, but is it difficult?
As far as the current page is concerned, it only measures below src. I would like to measure cgroups as well if possible.
https://app.codecov.io/gh/containers/youki

PR #273, opened 2021-09-06
CI Code Coverage Fix

YJDoc2:
ref #226 #272 
This attempts to Fix code coverage issue :
After seeing https://github.com/actions-rs/grcov/issues/105, which asks to update code coverage using inbuilt features of llvm, as well as PyO3 also seems to use it, (which is quite a big lib). So I tried to use similar workflow as https://github.com/PyO3/pyo3/blob/main/.github/workflows/ci.yml to see if it can fix the issue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,8 +20,9 @@ jobs:
             .: 'src/*'
             ./test_framework: test_framework/*
             ./youki_integration_test: youki_integration_test/*
+            ./cgroups: cgroups/*
   check:
-    needs: changes
+    needs: [changes]
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


YJDoc2:
So I think this does work for accounting code coverage in cgroups as well. On codecov docs, it states that reports are merged : https://docs.codecov.com/docs/merging-reports So I think this merges the code coverage for youki as well as cgroups both.

On another node, this didn't spawn format and clippy jobs as no code was changed, which means the check job is working as expected !!! :tada: :tada: :tada: 

@utam0k please take a look once, and verify.

YJDoc2:
I also think the grcov.yaml file is not needed anymore and can be deleted, but not sure.

YJDoc2:
Also, I just came around https://github.com/Swatinem/rust-cache, should we try changing our plain cache action with this one specifically for rust?

Furisto (approved):

utam0k:
Let's install it and measure the benchmarks!
> Also, I just came around https://github.com/Swatinem/rust-cache, should we try changing our plain cache action with this one specifically for rust?

utam0k (approved):

PR #274, opened 2021-09-06
Bump procfs

Furisto:
A new version of procfs has been released that contains my patch for reading the namespace information from proc. We therefore can remove the duplicated code from youki.


utam0k (approved):
nice!

PR #277, opened 2021-09-06
Stablize cargo test

yihuaf:
The root cause of the flakiness observed in the hook, clean up file descriptors, and channel tests are the result of running the tests in parallel in the same process by cargo test. Many of the operations have side effects on the underlying OS and the process, such as closing down fds. These tests also depends on the underlying OS state such as file descriptor states. The best we can do is to mark these tests as `serial` and properly clean up any left over resources such as opened fds before the test ends. If the underlying OS decided to generate errors for some of the IO calls we are doing, then there is not much we can do. Not sure if retry is a valid solution here and retry generally will make the code much more complex.

Note: `runc` actually doesn't have unit tests on most of these operations. The other solution is to fully depend on the `integration` tests to test out these code path. It may be the reasonable thing to do if these unit tests prove to be hard to get right.

Note2: Currently, I ran cargo test for 100 iterations and no failed tests. There may still be flakiness, but it should be extremely rare.

### DIFF
--- a/hack/stress_cargo_test.sh
+++ b/hack/stress_cargo_test.sh
@@ -1,0 +1,11 @@
+#!/bin/bash -ue
+
+# This is a simple script to stress test `cargo test` to rule out flaky tests.
+
+COUNT=${1:-20}
+
+for i in $(seq 1 ${COUNT})
+do 
+    echo "Run test ${i} iteration..."
+    cargo test -- --nocapture
+done
--- a/src/container/tenant_builder.rs
+++ b/src/container/tenant_builder.rs
@@ -8,7 +8,7 @@ use std::{
     collections::HashMap,
     convert::TryFrom,
     fs,
-    os::unix::prelude::{RawFd},
+    os::unix::prelude::RawFd,
     path::{Path, PathBuf},
     str::FromStr,
 };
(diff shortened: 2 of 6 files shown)


utam0k:
Thank you for your investigation. Can I ask you to write a clear description of the results of this investigation in the section of the test that we are using `serialize` other tests?
> The root cause of the flakiness observed in the hook, clean up file descriptors, and channel tests are the result of running the tests in parallel in the same process by cargo test. Many of the operations have side effects on the underlying OS and the process, such as closing down fds. These tests also depends on the underlying OS state such as file descriptor states. The best we can do is to mark these tests as serial and properly clean up any left over resources such as opened fds before the test ends. If the underlying OS decided to generate errors for some of the IO calls we are doing, then there is not much we can do. Not sure if retry is a valid solution here and retry generally will make the code much more complex.

yihuaf:
> Thank you for your investigation. Can I ask you to write a clear description of the results of this investigation in the section of the test that we are using `serial` other tests?
> 
> > The root cause of the flakiness observed in the hook, clean up file descriptors, and channel tests are the result of running the tests in parallel in the same process by cargo test. Many of the operations have side effects on the underlying OS and the process, such as closing down fds. These tests also depends on the underlying OS state such as file descriptor states. The best we can do is to mark these tests as serial and properly clean up any left over resources such as opened fds before the test ends. If the underlying OS decided to generate errors for some of the IO calls we are doing, then there is not much we can do. Not sure if retry is a valid solution here and retry generally will make the code much more complex.

Done. I left the `tty` test alone because it was the original user of serial tests I think, so it is not part of this change.

utam0k (approved):
thanks

PR #278, opened 2021-09-07
Change cache action from default to Swatinem/rust-cache@v1 

YJDoc2:
This reduces CI time : for tests and coverage from 4 mins to 1 min, for integration test from ~7 mins to 5 mins. There is not much improvement in integrations tests, as go builds are not cached, maybe that can improve time even more.
Also as coverage uses nightly, the cache will be flushed daily, so the first PR on each day will have slow coverage report, but after that the cache should persist till end of the day. see https://github.com/marketplace/actions/rust-cache
Hope this will improve CI/CD experience for youki!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,19 +116,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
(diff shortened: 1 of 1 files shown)


YJDoc2:
As this is running for the first time here, it will take longer, but after merging the PRs should have faster workflows, as the cache will work. check my notes in https://github.com/YJDoc2/youki/pull/6 
Also in case this doesn't work directly, we may need to add shared key field as stated in https://github.com/marketplace/actions/rust-cache , maybe that will work as expected.

utam0k (approved):
awesome

PR #280, opened 2021-09-07
cache runtime-tools.

utam0k:
It's a small but definite speedup.
https://github.com/utam0k/youki/pull/7#issuecomment-914310311


YJDoc2 (approved):
lgtm :+1:

PR #283, opened 2021-09-07
fix: Mismatch of PWD in tutorial

kenoss:
## What is fixed

At this line, readers are in `<repository-root>/tutorial`, while the youki binary is in `<repository-root>`.
I fixed the mismatch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ We can edit the `config.json` to add customized behaviors for container. Here, w
 
 Then we can explore the lifecycle of a container:
 ```sh
+$ cd ..                                                # go back to the repository root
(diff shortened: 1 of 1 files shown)


yihuaf:
Thank you for the catch. But can I ask you to change the tutorial to be done in the root, instead of tutorial? In another word, can we remove https://github.com/containers/youki/pull/283/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R110 and update the path to be from the project root?

yihuaf (approved):

kenoss:
Thanks for the review and the suggestion. An obstruction for the suggestion is `youki spec` can only generate `config.json` in the current directory and it needs additional `mv`. I feel the current version is simple enough, so I agree with your conclusion.

PR #285, opened 2021-09-08
Convert memory swap values

Furisto:
This [converts ](https://github.com/containers/crun/blob/main/crun.1.md#memory-controller) cgroup v1 memory swap values to cgroup v2.


utam0k (approved):
lgtm

PR #286, opened 2021-09-09
Extend info cmd with status of cgroup controllers

Furisto:
This adds which controllers are attached to the cgroup v2 hierarchy. For devices I just check if it is possible to attach epbf programs to a cgroup. I also moved the freezer controller to pseudo controllers. While contrary to devices and unified, this is a real subsystem it is also quite different from the other controllers. More importantly you cannot read it from cgroup.controllers, nor write it to subtree_control which was one of the original motivations for introducing this distinction.

### DIFF
--- a/cgroups/src/v2/controller_type.rs
+++ b/cgroups/src/v2/controller_type.rs
@@ -1,5 +1,6 @@
 use std::fmt::Display;
 
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 pub enum ControllerType {
     Cpu,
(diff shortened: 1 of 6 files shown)


Furisto:
We should hold this off until the upgrade to the new oci-spec release is completed.

utam0k:
> We should hold this off until the upgrade to the new oci-spec release is completed.

I agree with you.

Furisto:
@utam0k PTAL

utam0k (approved):
cool

PR #266, opened 2021-09-05
Upgrade oci-spec-rs to 0.4.0

guni1192:
fix: #225 

## Common
- [x] Fix import path: `use oci_spec::XXX` -> `use oci_spec::runtime::XXX`

## Upgrade oci-spec-rs in `cgroups`
- [x] Port FreezerState to `cgroups` crate
  - [x] `Controller::apply()` receive `ContainerOpt` instead of `LinuxResources`
- ~~Remove `A` from `LinuxDeviceType`~~

## Upgrade oci-spec-rs in `youki`
- [x] Fix capability type (Capability type change: Vec -> HashSet)
- [x] Implement functions equivalent to `LinuxDeviceType::to_sflag` in youki.

## ~Upgrade oci-spec-rs to `v0.5.1`~

- Apply builder pattern
  - https://github.com/containers/oci-spec-rs/pull/69
  - `src/container/tenant_builder.rs` rewrite

PR #287, opened 2021-09-09
 Introduce a workspace to enable execution of commands in bulk.

utam0k:


YJDoc2 (commented):
lgtm :+1:
I think as workspace makes the /target dir common for both (youki and cgroups), so we might be able to remove the separate cache action for cgroups as well, will have to check that.

In cgroups I think you were trying to extract common statement outside if-else, but it might create issues for #266 ? It is already having issues with merge conflicts with existing files, so I feel it'd be better to leave those changes for @guni1192 to add in their PR. What do you think?

Furisto:
You can leave it out. My open PR contains the same change, so it will not get lost.

Furisto:
Oh wait, I don't think this will work, because with the workspace clippy will flag the code now.

utam0k:
@Furisto @YJDoc2 
Thank you for your comments. Well, I don't want to bother @guni1192  too much since https://github.com/containers/youki/pull/266 is a big change, so I'll draft this PR and wait for it to be merged.

utam0k:
@Furisto @YJDoc2 PTAL

YJDoc2:
Additionally we can remove the Cargo.lock fine in cgroups, and combine the separate cache actions, as the compiled cgroups files are kept in the youki's target/ only.

https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html
>A workspace is a set of packages that share the same Cargo.lock and output directory.

Other than that lgtm :+1:

utam0k:
> Additionally we can remove the Cargo.lock file in cgroups, and combine the separate cache actions, as the compiled cgroups files are kept in the youki's target/ only.
> 
> https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html
> 
> > A workspace is a set of packages that share the same Cargo.lock and output directory.
> 
> Other than that lgtm

@YJDoc2 Oh thanks for the review. I've taken care of it.

YJDoc2 (approved):
lgtm :+1: if you want to keep cache actions separate for now.

utam0k:
> lgtm  if you want to keep cache actions separate for now.

We could certainly organize it. I'll consider it in another PR.

PR #292, opened 2021-09-10
Implemented seccomp and pass the integration test

### LINKED ISSUE #25: Add seccomp
Hi, thanks for the project, it's very intresting!
I would like to work on seccomp feature if possible.

yihuaf:
Fix #25 

So this turns out to be simpler than I thought, thanks to the existance of `libseccomp`. `runc` also uses `libseccomp` which does all the heavy lifting.

One concern I have is the `seccomp-sys` repo. We don't have a lot of choices for a rust binding for libseccomp and this one looks reasonable. However, it is LGPL, so we may want to roll our own bindings. I used this repo just for prototype. If we are NOT OK with LGPL code into our repo, then we should discuss on what is the right thing to do here.

Regardless, this is a functional prototypes now, so I would like you guys to take a look. The PR can be merged as is, given the issue mentioned above.

TODO:
- [ ] `oci-spec-rs` is missing a few fields for seccomp. For example, the default error code is hardcoded to EPERM at the moment.
- [ ] Our own libssecomp bindings??


flxo:
If you do decide to to your own bindings consider an option to include the build of libseccomp from your bindings build.rs  because crates relying of prebuilt dependencies are just painful when cross compiling.

yihuaf:
> If you do decide to to your own bindings consider an option to include the build of libseccomp from your bindings build.rs because crates relying of prebuilt dependencies are just painful when cross compiling.

I think eventually this would be a good thing to support, especially for arch other than X86, and we will definitely keep this in mind. I am interested to understand your use cases more, especially if you are thinking about statically linking on other platform. On the other hand, I am fairly convinced that offloading to `libseccomp` is the right thing to do going forward, so let's sort this PR out first.

yihuaf:
Finally.... @utam0k PTAL. Also let me know what you think about LGPL and the `seccomp-sys` crate. 

A big chunk is the `config.json` I use for testing. I pulled that straight from docker, to make sure everything works.

utam0k:
@yihuaf Sorry I didn't have time to review this PR today, so I'll review it tomorrow.

Furisto:
> 
> 
> @yihuaf Sorry I didn't have time to review this PR today, so I'll review it tomorrow.

Same here. Sorry for the delay.

yihuaf:
Take your time :) It is a more complex change.

src/seccomp/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+        return Ok(());
+    }
+
+    let seccomp = seccomp.unwrap();
+    if seccomp.flags.is_some() {
+        // runc did not support this, so let's skip it for now.
+        bail!("seccomp flags are not yet supported");
+    }
+
+    // log::debug!("XXX seccomp: {:?}", seccomp);
```
utam0k, line 211: Did you forget to turn it off?

src/process/init.rs
```diff
@@ -299,6 +299,13 @@ pub fn container_init(
         .set_id(Uid::from_raw(proc.user.uid), Gid::from_raw(proc.user.gid))
         .context("Failed to configure uid and gid")?;
 
+    // Without no new privileges, seccomp is a privileged operation. We have to
+    // do this before dropping capabilities. Otherwise, we should do it later,
+    // as close to exec as possible.
+    if linux.seccomp.is_some() && proc.no_new_privileges.is_none() {
+        seccomp::initialize_seccomp(linux.seccomp.as_ref()).context("Failed to execute seccomp")?;
+    }
```
utam0k, line 307: I think it's better not to use Option when passing to `initialize_seccomp()`. That way, you don't have to bother to check it in `initialize_seccomp()`.
```suggestion
    if proc.no_new_privileges.is_none() {
        if let Some(seccomp) = linux.seccomp.as_ref() {
            seccomp::initialize_seccomp(seccomp).context("Failed to execute seccomp")?;
        }
    }
```

src/seccomp/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    pub fn build(self) -> Result<scmp_arg_cmp> {
+        if self.op.is_some() && self.datum_a.is_some() {
+            Ok(scmp_arg_cmp {
+                arg: self.arg,
+                op: self.op.unwrap(),
+                datum_a: self.datum_a.unwrap(),
+                datum_b: self.datum_b.unwrap_or(0),
+            })
```
utam0k, line 57: How about this code? This can remove `unwrap()`.
```suggestion
        if let (Some(op), Some(datum_a)) = (self.op, self.datum_a) {
            Ok(scmp_arg_cmp {
                arg: self.arg,
                op,
                datum_a,
                datum_b: self.datum_b.unwrap_or(0),
            })
```

src/seccomp/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+        if res != 0 {
+            bail!("Failed to load seccomp profile: {}", res);
+        }
+
+        Ok(())
+    }
+}
+
+fn translate_syscall(syscall_name: String) -> Result<i32> {
+    let c_syscall_name = CString::new(syscall_name.as_str())
```
utam0k, line 142:
```suggestion
fn translate_syscall(syscall_name: &str) -> Result<i32> {
    let c_syscall_name = CString::new(syscall_name)
```

src/seccomp/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+        self
+    }
+
+    pub fn build(self) -> Result<scmp_arg_cmp> {
+        if self.op.is_some() && self.datum_a.is_some() {
+            Ok(scmp_arg_cmp {
+                arg: self.arg,
+                op: self.op.unwrap(),
+                datum_a: self.datum_a.unwrap(),
+                datum_b: self.datum_b.unwrap_or(0),
```
utam0k, line 56: Is there any documentation that shows that 0 is the correct default value?

utam0k:
I would love to know your thoughts on the merits of this and other options? Isn't it realistic to not depend on libseccomp for example? I'm new to this field, so sorry if I'm saying something crazy.
>  Our own libssecomp bindings??

src/seccomp/mod.rs
yihuaf: The `scmp_arg_cmp` struct represent a comparison. Some of the `op` require one and some require two value. For example, an equal comparison requires only one value, testing if the syscall argument is equal to this value. In the case where the only one value is needed, I believe the value two or `datum_b` is ignored. The implementation in libseccomp sets the value to 0 anyway, but I don't think it is specifically documented in the man page.

src/seccomp/mod.rs
yihuaf: The `scmp_arg_cmp` struct represent a comparison. Some of the `op` require one and some require two value. For example, an equal comparison requires only one value, testing if the syscall argument is equal to this value. In the case where the only one value is needed, I believe the value two or `datum_b` is ignored. The implementation in libseccomp sets the value to 0 anyway, but I don't think it is specifically documented in the man page.

yihuaf:
> I would love to know your thoughts on the merits of this and other options? Isn't it realistic to not depend on libseccomp for example? I'm new to this field, so sorry if I'm saying something crazy.
> 
> > Our own libssecomp bindings??

First of all, this PR contains all the necessary changes needed in Youki to use seccomp. Any further work would be to replace the library code underneath, depending on how far we want to go.

The first question is whether to use `seccomp-sys` repo or our own binding. The downside of `seccomp-sys` repo is the LGPL license and the fact that it may not be well maintained. The work to replace it doesn't look hard since it is a thin layer of FFI bindings. I believe rust even has a auto generator for generating bindings for C.

The second question is whether to use `libseccomp`. My short answer is we should use libseccomp unless at some point in the future it is not working out for us. `libseccomp` is well maintained and used by `runc`, so I don't worry about the project becomes unmaintained. To some extent, OCI spec is designed around the `libseccomp` API or designed with `libseccomp` in mind. We are linking against `libseccomp`, so its LGPL license is not an issue. So at this point, rewriting `libseccomp` logic does not buy us much in my opinion.

With that being said, it would be a great project to rewrite libseccomp in rust, but it should be out of the scoop of Youki.

src/seccomp/mod.rs
Furisto: Can you add this explanation as a comment to the code?

src/seccomp/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+                    // If we failed to resolve the syscall by name, likely the kernel
+                    // doeesn't support this syscall. So it is safe to skip...
+                    log::warn!(
+                        "Failed to resolve syscall, likely kernel doesn't support this. {:?}",
+                        name
+                    );
+                    continue;
+                }
+
+                let syscall_number = translate_syscall(name)?;
```
Furisto, line 258: Do you have to call translate_syscall here again? Couldn't you do just
```suggestion
                let syscall_number = ret.unwrap()
```
given that you have have already checked for an error above?

src/seccomp/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+    use std::io::Write;
+    use std::os::unix::prelude::AsRawFd;
+    use std::path;
+
+    #[test]
+    #[serial]
+    fn test_basic() -> Result<()> {
+        // Note: seccomp profile is really hard to write unit test for. First,
+        // we can't really test default error or kill action, since rust test
+        // actually replies on certain syscalls. Second, some of the syscall
```
Furisto, line 322:
```suggestion
        // actually relies on certain syscalls. Second, some of the syscall
```

Furisto:
> The first question is whether to use `seccomp-sys` repo or our own binding. The downside of `seccomp-sys` repo is the LGPL license and the fact that it may not be well maintained. The work to replace it doesn't look hard since it is a thin layer of FFI bindings. I believe rust even has a auto generator for generating bindings for C.

I am not familiar with LGPL in detail, so please correct me here, but could that mean that we would have to also [release](https://stackoverflow.com/questions/10130143/gpl-lgpl-and-static-linking) youki as LGPL if we statically link to it?

yihuaf:
> > The first question is whether to use `seccomp-sys` repo or our own binding. The downside of `seccomp-sys` repo is the LGPL license and the fact that it may not be well maintained. The work to replace it doesn't look hard since it is a thin layer of FFI bindings. I believe rust even has a auto generator for generating bindings for C.
> 
> I am not familiar with LGPL in detail, so please correct me here, but could that mean that we would have to also [release](https://stackoverflow.com/questions/10130143/gpl-lgpl-and-static-linking) youki as LGPL if we statically link to it?



> > The first question is whether to use `seccomp-sys` repo or our own binding. The downside of `seccomp-sys` repo is the LGPL license and the fact that it may not be well maintained. The work to replace it doesn't look hard since it is a thin layer of FFI bindings. I believe rust even has a auto generator for generating bindings for C.
> 
> I am not familiar with LGPL in detail, so please correct me here, but could that mean that we would have to also [release](https://stackoverflow.com/questions/10130143/gpl-lgpl-and-static-linking) youki as LGPL if we statically link to it?

In short, no. You are thinking of GPL license. 

If we dynamically link LGPL code, we are fine. If we statically link, we need to provide our code so people can re-link to a different version of the LGPL lib, under any license. Youki being open sourced meets the requirement. So legally we can use it without any trouble. In addition, we are not really distributing Youki as compiled binary, so we aren't limited by the license.

Still I thought I should bring it up just so we are aware :) I believe we already use the `systemd` lib in Youki which is also LGPL code.

Furisto (approved):

yihuaf:
Turns out, writing FFI binding in rust is way easier than I imagined... I included it in this PR and removed the need for `seccomp-sys` crate. @Furisto Can I get another approve from you?

Furisto (approved):

utam0k (approved):
great!!!!!

PR #294, opened 2021-09-11
fix a failure when dirs is empty at changes job.

utam0k:
fix https://github.com/containers/youki/issues/284

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,7 @@ jobs:
             ./cgroups: cgroups/*
   check:
     needs: [changes]
+    if: ${{ !contains(needs.changes.outputs.dirs, '[]') }}
     runs-on: ubuntu-latest
     strategy:
       matrix:


utam0k:
- Patterns with changes in Rust code https://github.com/containers/youki/actions/runs/1223845447

YJDoc2 (commented):
Hey, have you tested this when no code changes are there? Asking because in https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contains it states that for arrays,  it checks if the element is in the array. So as needs.changes.outputs.dirs itself is an array , wouldn't 
```
contains(needs.changes.outputs.dirs, '[]')
```
check if `[]` is an element in the dirs, and not dirs itself is an empty array? Maybe I am wrong, as  I haven't worked with github actions expressions either, so if you have run this with no code changes and this works, lgtm :+1:

utam0k:
@YJDoc2 This seems to work for now. The pattern with no code changes is ci of this PR ;)

YJDoc2 (approved):
Oh yes! I missed that :sweat_smile: All well then :+1:

YJDoc2:
Also in one of previous commits in this PR, you had made a separate job for dependencies, was there any issue with that? Or were the dependencies not sharable across jobs?

utam0k:
> Also in one of previous commits in this PR, you had made a separate job for dependencies, was there any issue with that? Or were the dependencies not sharable across jobs?

That way of writing was completely wrong and I needed to think of another way.

PR #297, opened 2021-09-12
sipliy split init.rs into several files.

utam0k:
relate https://github.com/containers/youki/issues/295

### DIFF
--- a/src/process/intermediate.rs
+++ b/src/process/intermediate.rs
@@ -1,16 +1,7 @@
+use crate::{namespaces::Namespaces, process::channel, process::fork, syscall::Syscall};
 use anyhow::{Context, Result};
 use nix::unistd::{Gid, Uid};
-use oci_spec::runtime::{LinuxNamespaceType};
-
-
-
-
-use crate::{
-    namespaces::Namespaces,
-    process::channel,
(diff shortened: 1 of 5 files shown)


utam0k:
@yihuaf @Furisto @YJDoc2 What do you think of this PR?

yihuaf:
Changes looks good. I think splitting in this case can go either way. On one hand, the file is not that long and having everything in one file actually helps the understanding. I will admit, `init.rs` is probably the wrong name if we keep these functions in one file. On the other hand, I like how you split up the code.

src/process/intermediate.rs
```diff
@@ -0,0 +1,101 @@
+use anyhow::{Context, Result};
+use nix::unistd::{Gid, Uid};
+use oci_spec::runtime::{LinuxNamespaceType};
+
+
+
```
yihuaf, line 6: Nit: that's a lot of empty lines :)

yihuaf (approved):
As mentioned, the change itself LGTM. I am OK with splitting or not.

Furisto (approved):

YJDoc2 (approved):
Looks good :+1:

Issue #299, opened 2021-09-15
[Tracking Issue] Upgrade `oci-spec-rs`


yihuaf:
There are a few major and breaking changes since `oci-spec-rs` repo is broken off. We should catch up as soon as possible to avoid bigger gap. Using this as tracking issue.

- [X] Upgrade to 0.4.0 #266 
- [x] Upgrade to 0.5.0 #303 
- [x] Upgrade to 0.5.1 #303 
- [ ] Upgrade to 0.5.2 #329

utam0k:
@guni1192 Has development already started?

yihuaf:
If @guni1192 hasn't started, I can help to finish this. This is probably a higher priority.

guni1192:
@utam0k @yihuaf 
The cgroup crate has been updated to oci-spec-rs 0.5.1.
Work is in progress on youki.
There may be a lot of rewriting of tenant_builder.rs.
I'm trying to submit a Draft PR on Saturday.

utam0k:
@guni1192  Is it possible to send out PR only for cgroups first? I believe separate PRs for `src` and `cgroups` would be simpler to review.

guni1192:
@utam0k OK, I'll give it a try.
However, it may fail to resolve some dependencies.

yihuaf:
Use #329 to track.

PR #303, opened 2021-09-18
Updated cgroups oci-spec-rs to 0.5.1 or later

guni1192:
ref: #299 

Apply builder pattern.

Getter
```diff
- let y = linux_xxx.y
+ let y = linux_xxx.y()
```

Builder
```diff
- let x = LinuxXXX { y: y };
+ let x = LinuxXXXBuilder::default().y(y).build().unwrap();
```


cgroups/Cargo.toml
```diff
@@ -14,7 +14,10 @@ nix = "0.22.0"
 procfs = "0.10.1"
 log = "0.4"
 anyhow = "1.0"
-oci-spec = { git="https://github.com/utam0k/oci-spec-rs/", tag = "v0.4.0-with-bugfix"}
+# TODO: Fetch from crate.io instead of git when next release oci-spec-rs
+# ref: https://github.com/containers/oci-spec-rs/pull/69
+# oci-spec = "0.5.1"
+oci-spec = { git = "https://github.com/containers/oci-spec-rs",  rev = "5018f8e5b0355a82c08962cefa5ab07a05b930c6" }
```
yihuaf, line 20: Is the current commit beyong 0.5.1 or before?

guni1192: @yihuaf No, it's a later commit than 0.5.1.
I'm waiting for the release of a version that incorporates the following patch.
https://github.com/containers/oci-spec-rs/pull/69

guni1192:
@utam0k 
It seems that if you use a different version of oci-spec-rs, you can't build cargo.
Therefore, can I update youki's oci-spec-rs in this PR?

```
error[E0308]: mismatched types
   --> src/container/builder_impl.rs:161:28
    |
161 |                 resources: linux.resources.clone().unwrap(),
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `oci_spec::runtime::linux::LinuxResources`, found struct `LinuxResources`
    |
    = note: perhaps two different versions of crate `oci_spec` are being used?
````

**Comment by utam0k** [comment]
> @utam0k
> It seems that if you use a different version of oci-spec-rs, you can't build cargo.
> Therefore, can I update youki's oci-spec-rs in this PR?
> 
> ```
> error[E0308]: mismatched types
>    --> src/container/builder_impl.rs:161:28
>     |
> 161 |                 resources: linux.resources.clone().unwrap(),
>     |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `oci_spec::runtime::linux::LinuxResources`, found struct `LinuxResources`
>     |
>     = note: perhaps two different versions of crate `oci_spec` are being used?
> ```

Oh... Can I ask you to implement it in this PR?

**Comment by guni1192** [comment]
@utam0k I see. I'll try it.

File: cgroups/src/v1/memory.rs [hunk 2]
```diff
@@ -101,11 +101,12 @@ impl Controller for Memory {
...
-            return Some(memory);
-        }
-
+    fn needs_to_handle<'a>(_controller_opt: &'a ControllerOpt) -> Option<&'a Self::Resource> {
+        // TODO: fix compile error
+        // error[E0515]: cannot return value referencing temporary value
+        // if let Some(memory) = &controller_opt.resources.memory() {
+        //     return Some(memory);
+        // }
         None
```
guni1192, line 110: Is there a better way to fix this?

Cargo.toml
```diff
@@ -39,7 +39,10 @@ mio = { version = "0.7.13", features = ["os-ext", "os-poll"] }
 chrono = { version="0.4", features = ["serde"] }
 once_cell = "1.6.0"
 futures = { version = "0.3", features = ["thread-pool"] }
-oci-spec = { git="https://github.com/utam0k/oci-spec-rs/", tag = "v0.4.0-with-bugfix"}
+# TODO: Fetch from crate.io instead of git when next release oci-spec-rs
```
yihuaf, line 42: Can you file an issue so we don't forget? Can you also file an issue in `oci-spec-rs` to ask for a minor release?

cgroups/src/v1/memory.rs
yihuaf: I think you can remove this function.

yihuaf: To answer your question, the reason compiler tells you is that you are checking a Option with reference inside, but you also return a new option with the same reference. That's not memory safe because now you have two Option containing the same reference. You can `clone` the value of `memory` and returns an entirely new option. However, I do find if we have to use clone in this case, we should question a bit what do we really want the code to do. In this case, it is just checking if the option is some and use the value. As you already did, we can safely to just not use this function and delete it.

cgroups/src/v1/pids.rs
```diff
@@ -20,17 +20,19 @@ impl Controller for Pids {
...
         }
 
         Ok(())
     }
 
-    fn needs_to_handle<'a>(controller_opt: &'a ControllerOpt) -> Option<&'a Self::Resource> {
-        if let Some(pids) = &controller_opt.resources.pids {
-            return Some(pids);
-        }
+    fn needs_to_handle<'a>(_controller_opt: &'a ControllerOpt) -> Option<&'a Self::Resource> {
```
yihuaf, line 30: Same thing.

src/container/tenant_builder.rs
```diff
@@ -156,28 +159,72 @@ impl<'a> TenantContainerBuilder<'a> {
...
             bail!("Could not retrieve container init pid");
         }
 
         let init_process = procfs::process::Process::new(container.pid().unwrap().as_raw())?;
-        self.set_namespaces(spec, init_process.namespaces()?)?;
+        let ns = self.set_namespaces(init_process.namespaces()?)?;
+        let linux = LinuxBuilder::default().namespaces(ns).build()?;
 
-        Ok(())
+        let mut spec_builder = SpecBuilder::default()
```
yihuaf, line 197: This is fine, but I think we should implement in `oci-spec-rs` From spec to spec_builder. Otherwise, there will just be a lot of boilerplate code.

src/container/tenant_builder.rs
```diff
@@ -253,70 +280,84 @@ impl<'a> TenantContainerBuilder<'a> {
...
-                    .as_mut()
-                    .context("no process in spec")?
-                    .capabilities = Some(LinuxCapabilities {
-                    ambient: caps.clone().into(),
-                    bounding: caps.clone().into(),
-                    effective: caps.clone().into(),
-                    inheritable: caps.clone().into(),
-                    permitted: caps.into(),
-                })
+                let mut cb = LinuxCapabilitiesBuilder::default();
```
yihuaf, line 289: can we change to `capability_builder`? `cb` gets confused to `call back`. Or just builder.

yihuaf (approved):
LGTM. 

Just a consideration in the future, we should implement From spec to spec builder, so we don't have boilerplate code to copy a spec. It is error prune since we have to copy every field with a default builder.

cgroups/src/v1/memory.rs
guni1192: I understand now. Thank you.
Since this function is defined in the Controller trait, it needs to be removed from the trait definition.
I know it will be another amount of modification, but should I remove it within this PR?

src/container/tenant_builder.rs
guni1192: I think so too.

Cargo.toml
guni1192: OK. The following was submitted.

https://github.com/containers/youki/issues/329
https://github.com/containers/oci-spec-rs/issues/72

cgroups/src/v1/memory.rs
utam0k: Wouldn't it be easier to do if this PR is merged?
https://github.com/containers/oci-spec-rs/pull/73

yihuaf: Then don't remove it. Maybe put a TODO to explain.

utam0k: @guni1192 The above PR has been merged.

utam0k: @guni1192 cc: @yihuaf The timing was bad and this PR has been merged. Could you use another PR to handle it?

PR #305, opened 2021-09-18
Forbid empty string values for container id in commands

YJDoc2:
Currently container_id is a required argument for all commands that require container_id, but there is no restriction that empty string value cannot be provided as container_id. For example :
```sh
youki --root (root_path) create --bundle (bundle path)
```
gives error, but
```sh
youki --root (root_path) create --bundle (bundle path) ""
```
does not, as (at least bash) passes "" as an empty string argument, and clap allows it, as it is a valid string value.
This PR adds forbid_empty_value to all container_id fields, which forces it to give an error message whenever empty string is provided.

### DIFF
--- a/src/commands/create.rs
+++ b/src/commands/create.rs
@@ -23,6 +23,7 @@ pub struct Create {
     #[clap(long, default_value = "0")]
     preserve_fds: i32,
     /// name of the container instance to be started
+    #[clap(forbid_empty_values = true, required = true)]
     pub container_id: String,
 }
 
--- a/src/commands/delete.rs
+++ b/src/commands/delete.rs
@@ -13,6 +13,7 @@ use nix::sys::signal as nix_signal;
 
 #[derive(Clap, Debug)]
 pub struct Delete {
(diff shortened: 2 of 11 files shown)


YJDoc2:
Please take a look @utam0k @Furisto

YJDoc2:
Also, here changes job is not run even though there are changes in src/* file. this is probably due to 
https://github.com/containers/youki/blob/main/.github/workflows/main.yml#L20
where the src/* is in quotes ('') due to which it does not match any path. 
@utam0k  Should I try updating the workflow file in this PR only, or should I open a new PR for that?

utam0k:
> Also, here changes job is not run even though there are changes in src/* file. this is probably due to
> https://github.com/containers/youki/blob/main/.github/workflows/main.yml#L20
> where the src/* is in quotes ('') due to which it does not match any path.
> @utam0k @Furisto Should I try updating the workflow file in this PR only, or should I open a new PR for that?

Can I ask you for another PR?

utam0k (approved):
a nice improvement!

PR #306, opened 2021-09-18
Fix Changes Job in CI

YJDoc2:
the quotes ('') in the changes job failed to glob match the changed files, and thus the check job was not run. This PR fixes that, as well as add seccomp directory in changes.


YJDoc2:
@utam0k Please take a look, I have changed the glob pattern

utam0k (approved):
lgtm

PR #308, opened 2021-09-19
handle name as a str instead of a String.

utam0k:


YJDoc2:
Hey, can I ask you to change required fields in the youki_integration_tests as well? It depends on test_framework for its tests, so it needs changes as well. Also for the failing CI, you'll need to add  a blank `[workspace]` section in cargo.toml of both crates.

YJDoc2:
Also, I'm currently working in the youki_integration_tests section, where I'm adding required utils to implement the tests. If required, I can make these changes in my end as well, so there won't be any conflicts.

utam0k:
@YJDoc2 `test_frame` is excluded from the youki workspace because I think it will require a binary that is independent of `youki`.

YJDoc2:
Yes, this works as well. I was suggesting adding an empty `[workspace]` section in test_framework crate, as that what the rust lint suggests :sweat_smile: 
>test_frame is excluded from the youki workspace because I think it will require a binary that is independent of youki.

Otherwise the changes look good, but if possible, can you also make the necessary changes in youki_integration_tests as well? These changes break it, dues to changed types.

utam0k:
I'm sorry, what should I do?
> Otherwise the changes look good, but if possible, can you also make the necessary changes in youki_integration_tests as well? These changes break it, dues to changed types.

YJDoc2:
Currently the youki_integration_test uses the old definition of test_framework, so it still uses String instead of &str ;  thus it does not compile anymore with these changes. Currently there are two tests : container_create and container_lifecycle. Both of these use the String return types etc. So we will need to change those to &str , add lifetime annotations, maybe make the test names global const &str, as those does not change.If you want, we can merge changes in test_framework in this PR, and I can fix the breakage of youki_integration_tests in a new PR.

utam0k:
> Currently the youki_integration_test uses the old definition of test_framework, so it still uses String instead of &str ; thus it does not compile anymore with these changes. Currently there are two tests : container_create and container_lifecycle. Both of these use the String return types etc. So we will need to change those to &str , add lifetime annotations, maybe make the test names global const &str, as those does not change.If you want, we can merge changes in test_framework in this PR, and I can fix the breakage of youki_integration_tests in a new PR.

Oops, I forgot it. I have fixed it.

YJDoc2 (approved):
Amazing! lgtm :+1:

PR #309, opened 2021-09-19
Support 'shared' and 'unbindable' rootfs propagations

### LINKED ISSUE #256: Pass rootfs propagation test
## Overview
Need to pass the linux_rootfs_propagation integration test

# Goal
Pass the [rootfs propagation](https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_rootfs_propagation/linux_rootfs_propagation.go) integration test in runtime-tools. 
If you comment in this part and the test passes, you are all set.
https://github.com/containers/youki/blob/ef4b7932e80a91c67503325a76d322f0fc8f670b/integration_test.sh#L50

tiqwab:
Resolve #256.

This pull request enables youki to handle `shared` and `unbindable` [rootfs propagation](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#rootfs-mount-propagation).

Most of the change in the pull request follows runc implementation, but one difference is that runc actually doesn't pass `linux_rootfs_propagation/linux_rootfs_propagation.t` integration test (it fails in setting `shared` or `unbindable` propagation type).

This pull request added `rootfs::adjust_root_mount_propagation` after pivot_root to set appropriate propagation type at root mount point of container if `shared` or `unbindable` rootfs propagation is specified. This approach comes from [an open pull request of runc](https://github.com/opencontainers/runc/pull/1815).

### DIFF
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -248,6 +248,8 @@ pub fn container_init(
                 .with_context(|| format!("Failed to chroot to {:?}", rootfs))?;
         }
 
+        rootfs::adjust_root_mount_propagation(spec)?;
+
         if let Some(kernel_params) = &linux.sysctl {
             sysctl(kernel_params)
                 .with_context(|| format!("Failed to sysctl: {:?}", kernel_params))?;
--- a/src/rootfs.rs
+++ b/src/rootfs.rs
@@ -2,7 +2,7 @@
 //! Most systems mount another filesystem over it
 
 use crate::utils::PathBufExt;
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use nix::errno::Errno;
 use nix::fcntl::{open, OFlag};
 use nix::mount::mount as nix_mount;
@@ -11,7 +11,9 @@ use nix::sys::stat::{mknod, umask};
 use nix::sys::stat::{Mode, SFlag};
 use nix::unistd::{chown, close};
 use nix::unistd::{Gid, Uid};
+use nix::NixPath;
 use oci_spec::runtime::{LinuxDevice, LinuxDeviceType, Mount, Spec};
+use procfs::process::{MountInfo, MountOptFields, Process};
 use std::fs::OpenOptions;
 use std::fs::{canonicalize, create_dir_all, remove_file};
 use std::os::unix::fs::symlink;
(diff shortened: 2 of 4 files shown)


utam0k:
@tiqwab Thanks for your PR! Could you add unit tests whenever possible?

tiqwab:
Of course. I feel for now it is difficult to add enough unit tests, but I'll try it.

utam0k:
> Of course. I feel for now it is difficult to add enough unit tests, but I'll try it.

I know it is difficult to add complete unit tests in this area. So it's okay as long as you can.

src/process/init.rs
```diff
@@ -248,6 +248,8 @@ pub fn container_init(
                 .with_context(|| format!("Failed to chroot to {:?}", rootfs))?;
         }
 
+        rootfs::adjust_root_mount_propagation(spec)?;
```
yihuaf, line 251: Can we add context to the error handling? `with_context` or `context`.

src/rootfs.rs
```diff
@@ -35,6 +38,8 @@ pub fn prepare_rootfs(spec: &Spec, rootfs: &Path, bind_devices: bool) -> Result<
     nix_mount(None::<&str>, "/", None::<&str>, flags, None::<&str>)
         .context("Failed to mount rootfs")?;
 
+    make_parent_mount_private(rootfs)?;
```
yihuaf, line 41: Same, add error context.

src/rootfs.rs
```diff
@@ -387,3 +392,55 @@ fn parse_mount(m: &Mount) -> (MsFlags, String) {
...
+            MsFlags::MS_PRIVATE,
+            None::<&str>,
+        )?;
+    }
+
+    Ok(())
+}
+
+/// Change propagation type of rootfs as specified in spec.
+pub fn adjust_root_mount_propagation(spec: &Spec) -> Result<()> {
```
yihuaf, line 431: Can we pass in linux rather than the whole spec. Looks like only the linux is used.

yihuaf (commented):
Overall looks good. Thanks for doing this.

Can you please add error handling and context when you can. I know some of the existing code are not doing this, but we are trying to have better error handling whenever we can :)

You may have to create the unit test using fork and entering into user and mount namespace. Let me know if you run into any troubles.

tiqwab:
I added some error contexts at the suggest places and a simple unit test for `rootfs::find_parent_mount`, which doesn't require real mount nor namespaces. I'll work on more complicated tests in a few days.

utam0k (approved):
LGTM

Issue #282, opened 2021-09-07
Set up utils and port one integration test


YJDoc2:
As we are trying to port the integration tests to rust, we have created test framework, as well as set up container lifecycle tests in #186 . Now we have to set up utils required to port rest of the tests and set up one of the integration test to show as a template. After this we can create a tracking issue and open tickets for rest of the tests as well.

PR #310, opened 2021-09-19
Add integration test utils necessary for implementing rest integration tests

YJDoc2:
ref #282
Part 1 for closing the issue.

This changes hard-coded `youki` to more generic `runtime`
This also Sets up utils required for integration tests, which will be used for implementing rest of integration tests. Currently these function are not used, so are marked `#[allow(dead_code)]` so clippy doesn't flag them, this will be removed later.
This also switched using local directory in youki folder for setting up workspace to using temporary folders, taken after cgroups/temp_dir. This will prevent possible clashes between tests, and possibly allow running tests in parallel in future.

This is a bit big PR :sweat_smile: , but many of changes are not in implementation, but to accommodate the temp_dir.


YJDoc2:
@utam0k Please take a look.

youki_integration_test/src/tests/lifecycle/container_lifecycle.rs
```diff
@@ -1,20 +1,33 @@
...
+use std::thread::sleep;
+use std::time::Duration;
 use test_framework::{TestResult, TestableGroup};
 
 use super::{create, delete, kill, start, state};
 
+// By experimenting, somewhere around 50 is enough for youki process
+// to get the kill signal and shut down
+// here we add a little buffer time as well
+const SLEEP_TIME: u64 = 75;
```
utam0k, line 11: How about using `std::time::Duration` insted ot `u64`.

YJDoc2: A duration is created later from this u64, and then given to sleep because I thought we could not invoke functions for defining consts, but I checked ans we can use `duration::from_millis` directly in constant definition. I will change that along with other required changes in additional commit :+1:

YJDoc2:
Changed the type of `SLEEP_TIME` from u64->Duration

utam0k:
@YJDoc2 How about to search for `runtime` in the `$PATH` variable as well?

YJDoc2:
I personally feel it is better to accept the path as a argument than env var, as with env var, we will need make sure it is set, and when testing multiple runtime, the variable will need to be updated ,thinking from this integration test can be their own standalone program used to test multiple runtimes. 

Personally I feel taking the path as an argument is more self contained and preferable than an env var.

utam0k:
> I personally feel it is better to accept the path as a argument than env var, as with env var, we will need make sure it is set, and when testing multiple runtime, the variable will need to be updated ,thinking from this integration test can be their own standalone program used to test multiple runtimes.
> 
> Personally I feel taking the path as an argument is more self contained and preferable than an env var.

The PATH variable refers to a commonly used environment variable. In other words, I thought it would be more convenient to run the program when just passing `runc` instead of the full path, but what do you think?
I think we need to pass `/usr/bin/runc` now.

YJDoc2:
Oh, extremely sorry, I misunderstood your comment :sweat: :sweat:
Yes currently we need to pass the full path right now, and indeed it would be much more convenient to search in `$PATH`
I will make the appropriate changes. Sorry for the misunderstanding :sweat:

utam0k:
> Oh, extremely sorry, I misunderstood your comment  
> Yes currently we need to pass the full path right now, and indeed it would be much more convenient to search in `$PATH`
> I will make the appropriate changes. Sorry for the misunderstanding

My explanation was also inadequate. Please don't worry about it.

YJDoc2:
@utam0k Now have added the path resolution using `which` crate which takes care of underlying platform to resolve the path.
Now
```sh
sudo youki_integration_test -r runc
```
also works as expected.

utam0k (approved):

PR #314, opened 2021-09-20
Make container commands more suitable for use as a library

Furisto:
The idea behind this PR is to make it easier to use youki as library, so it can be reused in different contexts.

```
let mut container = ContainerBuilder::new(self.container_id.clone(), syscall.as_ref())
            .build()?;
container.start()?;
container.pause()?;
container.resume()?;
container.kill()?;
container.delete()?;
```

### DIFF
--- a/src/commands/create.rs
+++ b/src/commands/create.rs
@@ -33,23 +33,6 @@ pub struct Create {
 // it is running, it is just another process, and has attributes such as pid, file descriptors, etc.
 // associated with it like any other process.
 impl Create {
-    /// instant Create Command
-    pub fn new(
-        container_id: String,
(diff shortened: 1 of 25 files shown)


utam0k:
@alfonsoros88 Hi! Maybe you are interested in this PR. so I'll give you the mention.

Furisto:
- I also have found that some commands used the systemd global option to determine if systemd should be used. Instead the decision has to be made by examining the container state as once the container has been created we have to use the same cgroup manager throughout the lifecycle of the container.
- Added context to many more results.

utam0k (approved):
perfect!

PR #318, opened 2021-09-21
Fix error message(`LinuixIdMapping` to `uid_mappings`)

shorii:


src/rootless.rs
```diff
@@ -100,7 +100,7 @@ fn validate(spec: &Spec) -> Result<()> {
     let uid_mappings = linux
         .uid_mappings
         .as_ref()
-        .context("rootless containers require LinuxIdMapping in spec")?;
+        .context("rootless containers require uid_mappings in spec")?;
```
utam0k, line 103: Thanks for your PR. How about aligning to the runtime spec? This means it should be `uidMappings`.
https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#user-namespace-mappings

shorii: Thank you for your advice. I fixed this.(`uid_mappings` -> `uidMappings`,  `gid_mappings` -> `gidMappings`)

utam0k: perfect!

utam0k (approved):
LGTM

PR #319, opened 2021-09-21
style: Fix indentation

kenoss:
Fixed trivial indentation.

I don't know why `cargo fmt` allows both before/after versions. Glancing issues of rustfmt, I suspect rustfmt allows [no indent for match arms](https://github.com/rust-lang/rustfmt/issues/4616).

While that being said, I think this should be fixed because:

- Originally, this match is properly indented. https://github.com/containers/youki/pull/98/files#diff-6b8c70294dc20a59afa843ccbe8e3ab2357842d13d737643d80bf325ad931a69R105
- AFAIK, using indent for match arms is the majority.


utam0k (approved):
LGTm

PR #324, opened 2021-09-23
rootful mode for vagrant

zidoshare:
I added a rootful development mode to vagrant, which may provide us with convenience for debugging in vagrant using a similar vscode editor remote ssh.

It is very simple to use, like the following:
```shell
VAGRANT_VAGRANTFILE=Vagrantfile.root vagrant up
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,13 +192,19 @@ $ ./integration_test.sh linux_*
 
 ### Setting up Vagrant
 
-You can try youki on platforms other than linux by using the Vagrantfile we have prepared.
+You can try youki on platforms other than linux by using the Vagrantfile we have prepared. We have prepared two environments for vagrant, namely rootless mode and rootful mode
 
 ```
 $ git clone git@github.com:containers/youki.git
 $ cd youki
+
+# If you want to develop in rootless mode, and this is the default mode
 $ vagrant up
 $ vagrant ssh
+
+# or if you want to develop in rootful mode
+$ VAGRANT_VAGRANTFILE=Vagrantfile.root vagrant up
+$ VAGRANT_VAGRANTFILE=Vagrantfile.root vagrant ssh
 
 # in virtual machine
 $ cd youki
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -192,12 +192,15 @@ $ ./integration_test.sh linux_*
 
 ### Setting up Vagrant
 
-You can try youki on platforms other than linux by using the Vagrantfile we have prepared.
+You can try youki on platforms other than linux by using the Vagrantfile we have prepared.We have prepared two environments for vagrant, namely rootless mode and rootful mode
```
utam0k, line 195:
```suggestion
You can try youki on platforms other than linux by using the Vagrantfile we have prepared. We have prepared two environments for vagrant, namely rootless mode and rootful mode
```

README.md
```diff
@@ -192,12 +192,15 @@ $ ./integration_test.sh linux_*
...
 
 ```
 $ git clone git@github.com:containers/youki.git
 $ cd youki
+# If you want to develop in rootless mode,and this is default mode
```
utam0k, line 200:
```suggestion
# If you want to develop in rootless mode, and this is the default mode
```

utam0k:
@zidoshare Thanks for your PR.  I feel that we only support root full mode; is there any good point in continuing to support the two vagrant environments?

zidoshare:
@utam0k 
We may pay more attention to the support of rootless mode than root full, because it is more secure, which is as attractive as no daemon.

On the other hand, when we add new features, we always debug successfully in root full mode before adding patches to rootless mode.

So I think there should be two modes for selective development, and they do not conflict with each other.

utam0k:
> @utam0k
> We may pay more attention to the support of rootless mode than root full, because it is more secure, which is as attractive as no daemon.
> 
> On the other hand, when we add new features, we always debug successfully in root full mode before adding patches to rootless mode.
> 
> So I think there should be two modes for selective development, and they do not conflict with each other.

I understand. It's not that costly to support at the moment, so let's support it. Please fix only what I commented.

zidoshare:
I know, but I found some errors, please wait a while, I am fixing it @utam0k

zidoshare:
Should it be rootfull instead of rootful? @utam0k

utam0k (approved):
Thasnks

PR #330, opened 2021-09-25
Implement seccomp notify

yihuaf:
This is the first part of the change where we support seccomp notify and returning a fd for seccomp.


src/seccomp/mod.rs
```diff
@@ -417,4 +491,50 @@ mod tests {
...
+                        );
+                    }
+                    _ => {
+                        bail!("Child process failed to exit correctly: {:?}", status);
+                    }
+                }
+            }
+            nix::unistd::ForkResult::Child => {
+                let _ = prctl::set_no_new_privileges(true);
+                let _ = prctl::set_no_new_privileges(true);
```
utam0k, line 527: Why do we need twice?

src/seccomp/mod.rs
```diff
@@ -139,6 +140,33 @@ impl FilterContext {
...
+            bail!("internal libseccomp fault; likely no seccomp filter is loaded");
+        }
+
+        if res == -(libc::EOPNOTSUPP as i32) {
+            bail!("seccomp notify filter not supported");
+        }
+
+        if res > 0 {
+            return Ok(Some(res));
+        }
```
utam0k, line 166: How about using `match res`?

utam0k:
Is it possible to break the `mod.rs` down into a few more files?

src/seccomp/mod.rs
yihuaf: I thought about it. However, note the return code is actually negative, so it is comparison to an expression instead of a value. Match expression doesn't really support this unless I am using match guard. At that point, Using if statement here is likely easier to read.

src/seccomp/mod.rs
yihuaf: We don't. LOL.

yihuaf:
> Is it possible to break the `mod.rs` down into a few more files?

We can, but it is still a coherent block at the moment. The only relevant pub function here is `initialize_seccomp`. All the other bits are in the service of this function. Maybe next time when this file gets larger.

utam0k:
> We can, but it is still a coherent block at the moment. The only relevant pub function here is `initialize_seccomp`. All the other bits are in the service of this function. Maybe next time when this file gets larger.

Ok, I agree with you.

src/seccomp/mod.rs
utam0k: I undetstand. How about using `nix::errorno`?
https://docs.rs/nix/0.22.1/nix/errno/enum.Errno.html#method.from_i32

yihuaf: That's a good idea.

utam0k (approved):
LGTM

PR #333, opened 2021-09-25
Ensure cgroup error behavior is consistent with runc

### LINKED ISSUE #328: Invalid TLB configuration still creates cgroup, and youki process
@utam0k @Furisto @tsturzl please take a look

I was implementing HugeTLB test as part of #282, and for the invalid TLB version, as per https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_cgroups_hugetlb/linux_cgroups_hugetlb.go#L62-L89 I found an issue :
When testing with runc, the runc process starts, founds out that the TLB config is invalid, and exists with message :
```
time="2021-09-24T18:47:07+05:30" level=warning msg="signal: killed"
time="2021-09-24T18:47:07+05:30" level=error msg="container_linux.go:349: 
starting container process caused "process_linux.go:449: container init caused "process_linux.go:415:  setting cgroup config for procHooks process caused 
"failed to write "314572800" to "/sys/fs/cgroup/hugetlb/dbac89a0-effa-7e71-8df4-c98facbca452/hugetlb.3MB.limit_in_bytes": 
open /sys/fs/cgroup/hugetlb/dbac89a0-effa-7e71-8df4-c98facbca452/hugetlb.3MB.limit_in_bytes: permission denied"

container_linux.go:349: starting container process caused "process_linux.go:449: 
container init caused \"process_linux.go:415: setting cgroup config for procHooks process caused 
"failed to write "314572800" to "/sys/fs/cgroup/hugetlb/dbac89a0-effa-7e71-8df4-c98facbca452/hugetlb.3MB.limit_in_bytes": 
open /sys/fs/cgroup/hugetlb/dbac89a0-effa-7e71-8df4-c98facbca452/hugetlb.3MB.limit_in_bytes: permission denied"
```

The runc does not create a cgroup, and trying to get the state using state command also give error that container does not exist.

On the other hand the youki process starts, does not create the container, but still creates the cgroup, and when checking the state, it shows that the container is in the 'stopped' state, when if fact is was not created/ should not have been created.

Another difference I noticed was that runc creates cgroup with container id as name in the root of respective cgroup, for eg :
for id 123abc, runc creates cgroup at path /sys/fs/cgroup/hugetlb/123abc 
whereas youki creates a cgroup at path /sys/fs/cgroup/hugetlb/youki/123abc 

I'm not sure if the exact path needs to be same across runtime implementations, but if the tools need to manually remove cgroups at end of testing, this will cause issues, as for each runtime, the path might be different.
Can someone verify if the path is same or not with some other runtime ?

Furisto:
This should fix #328. 

- If no cgroup path is specified, we will default to /sys/fs/cgroup/[containerid]. 
- If any error is encountered during container creation, the cgroup will be removed and the container state directory will be deleted.

@YJDoc2 Can you confirm that this fixes your problem?

### DIFF
--- a/cgroups/src/common.rs
+++ b/cgroups/src/common.rs
@@ -3,6 +3,7 @@ use std::{
     fs::{self, File},
     io::{BufRead, BufReader, Write},
     path::{Path, PathBuf},
+    time::Duration,
 };
 
 use anyhow::{bail, Context, Result};
@@ -354,4 +355,22 @@ pub(crate) fn default_devices() -> Vec<LinuxDevice> {
             gid: None,
         },
     ]
+}
+
+pub(crate) fn delete_with_retry<P: AsRef<Path>>(path: P) -> Result<()> {
+    let mut attempts = 0;
+    let mut delay = Duration::from_millis(10);
+    let path = path.as_ref();
+
+    while attempts < 5 {
+        if fs::remove_dir(path).is_ok() {
+            return Ok(());
+        }
+
+        std::thread::sleep(delay);
+        attempts += attempts;
+        delay *= attempts;
+    }
+
+    bail!("could not delete {:?}", path)
 }
--- a/cgroups/src/common.rs
+++ b/cgroups/src/common.rs
@@ -357,12 +357,22 @@ pub(crate) fn default_devices() -> Vec<LinuxDevice> {
     ]
 }
 
-pub(crate) fn delete_with_retry<P: AsRef<Path>>(path: P) -> Result<()> {
+/// Attempts to delete the path the requested number of times.
+pub(crate) fn delete_with_retry<P: AsRef<Path>, L: Into<Option<Duration>>>(
+    path: P,
+    retries: u32,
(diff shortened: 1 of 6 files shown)


cgroups/src/common.rs
```diff
@@ -355,3 +356,21 @@ pub(crate) fn default_devices() -> Vec<LinuxDevice> {
         },
     ]
 }
+
+pub(crate) fn delete_with_retry<P: AsRef<Path>>(path: P) -> Result<()> {
```
utam0k, line 360: I think it is more versatile to have the number of retries and the delay as arguments. What do you think?

YJDoc2:
Sorry for the delay, I checked, and this is now consistent with runc. This should fix and close #328 . Thanks for the fix, now I will go ahead with rest of the integration test :+1: :smile:

Furisto:
@utam0k PTAL

utam0k (approved):
LGTM

PR #339, opened 2021-09-27
Add HugeTLB tests

### LINKED ISSUE #282: Set up utils and port one integration test

YJDoc2:
This PR aims to close #282, by setting up initially required utils, and adding huge tlb test as a template for adding rest of the tests.

This also parallelize test running, which cuts down time form ~6 seconds to run all of current tests to ~3 seconds, I feel this will be quire useful as the number of tests increase.

After this is accepted, we can close #282 and open tracking issue to implement rest of integration tests.


utam0k:
Why not combine `youki_integration_test/` and `test_framework/` into one directory? Of course, this response can be done with a another PR.

youki_integration_test/src/utils/support.rs
```diff
@@ -21,3 +28,41 @@ pub fn get_project_path() -> PathBuf {
...
+    let tar_gz = File::open(tar_path)?;
+    let tar = GzDecoder::new(tar_gz);
+    let mut archive = Archive::new(tar);
+    archive.unpack(&temp_dir)?;
+    Ok(temp_dir)
+}
+
+/// Sets the config.json file as per given spec
+pub fn set_config<P: AsRef<Path>>(project_path: P, config: &Spec) -> Result<()> {
+    let path = project_path.as_ref().join("bundle").join("config.json");
```
utam0k, line 65: Why we can use `oci-spec-rs` builder to dynamically create our own config.json instead of using the `config.json`? In other words, could we generate a `config.json` that could be easily configured programmatically for each test?

YJDoc2:
Originally when developing this, I wanted to keep the framework separate from the tests themselves, so it can be developed separately, and can be used for other testing as required, (even as a separate crate if required). Of course, currently it is only used by integration tests, so can be merged.
> Why not combine youki_integration_test/ and test_framework/ into one directory?

Also I had another concern about duplication of code : If you see some part of code in cgroups/common is repeated here, which is sued to get page size data from system. I felt that having a dependency on any part of cgroups or youki in tests would not be a good thing, and errors in there would not be noticed. But that leads to duplication of code here. I would like to now your opinion on this.

youki_integration_test/src/utils/support.rs
YJDoc2: If I understand your concern correctly, I am doing exactly that here : in tlb_test.rs:13 there is a function which generates the spec using oci-spec-rs builder.
The function set_config is used to save that generated config in the directory on disk, so runtime can use it when creating the container.
If you mean that we should add an interface over oci-spec-rs which will generate a single spec, and modify it as required, I don't think that the editing the config that way would be good, given that the builder already does something similar.
In case I have misunderstood something, can you please once clarify it? :sweat_smile:

utam0k: Oops, I missed that! It's perfect.

utam0k:
> Originally when developing this, I wanted to keep the framework separate from the tests themselves, so it can be developed separately, and can be used for other testing as required, (even as a separate crate if required). Of course, currently it is only used by integration tests, so can be merged.

My image was to create one new directory and put the current two directories in it. It may be easier to develop using Cargo members with one of them as `src`.

utam0k (approved):
LGTM

PR #345, opened 2021-09-29
Implemented util function to test in child process

yihuaf:
We have many cases where we want to write unit test in a child process, because the functions we want to test will change the underlying OS state. Since cargo test runs in a single process, we have to fork unto child process to run functions. Before, using only fork, we can only pass `int` back to parent process. Using `ipc-channel` crate, we can pass richer test results back to the parent process so the error message is much clearer.

I only made change to seccomp tests as an example. Let me know what you think.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ tabwriter = "1"
 fastrand = "1.4.1"
 crossbeam-channel = "0.5"
 seccomp = { version = "0.1.0", path = "./seccomp" }
+ipc-channel = "0.15.0"
 
 [dev-dependencies]
 # TODO: Fetch from crate.io instead of git when next release oci-spec-rs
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -50,6 +50,7 @@ tabwriter = "1"
 fastrand = "1.4.1"
 crossbeam-channel = "0.5"
 seccomp = { version = "0.1.0", path = "./seccomp" }
+ipc-channel = "0.15.0"
```
utam0k, line 53: I'd like to make `ipc-channel` depend on only `dev-dependencies`, is it possible?

utam0k:
I've been trying to work on this issue too to see if I can't do something about it. I agree with you about putting this processing together.

Cargo.toml
yihuaf: Yes it can be done. Take a look at the latest commit. I may want to use ipc-channel in other part of Youki's code, but let's keep it as dev-dependencies for now, and discuss this when we get there.

utam0k (approved):
Thanks!

PR #340, opened 2021-09-27
279 increate the code coverage of src rootfs

tommady:
#279 
the rootfs part

PR #347, opened 2021-09-30
[WIP] Add test for rootfs propagation with mock mount

tiqwab:
This is a follow-up pull request to add unit tests for functions implemented in #309.

This pull request calls mount system call through `Syscall` trait to mock it in unit tests. Although we can use fork and then execute real mount system call in child process, I feel checking mount arguments is simple and enough in this case.

@utam0k @yihuaf 
What do you think about the idea of using mock for these tests?

### DIFF
--- a/src/process/init.rs
+++ b/src/process/init.rs
@@ -1,11 +1,11 @@
 use super::args::ContainerArgs;
 use crate::apparmor;
+use crate::syscall::Syscall;
 use crate::{
     capabilities, hooks, namespaces::Namespaces, process::channel, rootfs, rootless::Rootless,
     seccomp, tty, utils,
 };
 use anyhow::{bail, Context, Result};
-use nix::mount::mount as nix_mount;
 use nix::mount::MsFlags;
 use nix::sched::CloneFlags;
 use nix::{
@@ -91,13 +91,13 @@ fn sysctl(kernel_params: &HashMap<String, String>) -> Result<()> {
 // The first time we bind mount, other flags are ignored,
 // so we need to mount it once and then remount it with the necessary flags specified.
 // https://man7.org/linux/man-pages/man2/mount.2.html
-fn readonly_path(path: &str) -> Result<()> {
-    match nix_mount::<str, str, str, str>(
+fn readonly_path(path: &Path, command: &dyn Syscall) -> Result<()> {
+    match command.mount(
         Some(path),
         path,
-        None::<&str>,
+        None,
         MsFlags::MS_BIND | MsFlags::MS_REC,
-        None::<&str>,
+        None,
     ) {
         // ignore error if path is not exist.
         Err(nix::errno::Errno::ENOENT) => {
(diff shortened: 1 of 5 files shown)


yihuaf:
Haven't looked the entire PR yet. I think to test the readonly_path and mask_path individually, we should fork and use namespaces. mock only shows the command is being called, but testing the actual assumption we make.

tiqwab:
I actually wrote a draft for the test of `rootfs::make_parent_mount_private` function with the actual mount system call in https://github.com/tiqwab/youki/commit/4e49290ac190cb649eb74dd9e2e1cd9ed8d59e80. Is it similar to what you think?

I just focused on writing unit tests for `rootfs::make_parent_mount_private` and `rootfs::adjust_root_mount_propagation` and in that case I thought checking mount args is simpler and enough. However I understand the benefit of using the actual mount call and we might had better to use it considering other test cases in rootfs.

tiqwab:
and I realized that there is a similar work in https://github.com/containers/youki/pull/340, so we might had better to close this and leave it for @tommady.

tommady:
> and I realized that there is a similar work in #340, so we might had better to close this and leave it for @tommady.

ya thank you and please allow me to copy some of your masterwork there 🙇🏻

yihuaf:
> I actually wrote a draft for the test of `rootfs::make_parent_mount_private` function with the actual mount system call in [tiqwab@4e49290](https://github.com/tiqwab/youki/commit/4e49290ac190cb649eb74dd9e2e1cd9ed8d59e80). Is it similar to what you think?
> 
> I just focused on writing unit tests for `rootfs::make_parent_mount_private` and `rootfs::adjust_root_mount_propagation` and in that case I thought checking mount args is simpler and enough. However I understand the benefit of using the actual mount call and we might had better to use it considering other test cases in rootfs.

Yes, this is what I had in mind as well. A few minor points:
1. Let's use fork instead of clone, or we should implement `clone` properly. Rust memory management has some weird issue with regarding to stack vs. heap allocation for stack. We used to use clone but discovered issue in rust. So fork is much safer since you trust the kernel code to copy all the memory correctly. If clone prove to be better suited than fork, I would bring back the old implementation for a clone wrapper. See  #167 for reference.

2. We can now use `ipc-channel` to pass richer results from child process back to parent process. See #345 for reference.

yihuaf:
> and I realized that there is a similar work in #340, so we might had better to close this and leave it for @tommady.

@tommady Are you able to incorporate this change? Since the title of #340 is vague, can I ask what is the scope of the PR? Let us know what you plan to keep from this PR for #340. In general, please keep the PR as small as reasonably possible. This will be much easier for review and coordinate.

tommady:
> @tommady Are you able to incorporate this change? Since the title of #340 is vague, can I ask what is the scope of the PR? Let us know what you plan to keep from this PR for #340. In general, please keep the PR as small as reasonably possible. This will be much easier for review and coordinate.

hi @yihuaf the goal of PR #340 is trying to increase the testing coverage of rootfs.rs file only.

so the plain is that
1. won't modify the existing working flow
2. only try to mock the system calls
3. try to increase the coverage as much as possible

is that acceptable to you? thanks

PR #350, opened 2021-10-01
Add rootless option for spec

chenyukang:
```shell
./target/x86_64-unknown-linux-gnu/debug/youki spec --rootless
```

Will generate a config.json for rootless.

### DIFF
--- a/src/commands/spec_json.rs
+++ b/src/commands/spec_json.rs
@@ -1,15 +1,14 @@
 use anyhow::Result;
 use clap::Clap;
-use std::path::PathBuf;
-
 use nix;
 use oci_spec::runtime::{
-    LinuxBuilder, LinuxIdMappingBuilder, LinuxNamespaceBuilder, LinuxNamespaceType, MountBuilder,
-    Spec, SpecBuilder,
+    Linux, LinuxBuilder, LinuxIdMappingBuilder, LinuxNamespace, LinuxNamespaceBuilder,
+    LinuxNamespaceType, MountBuilder, Spec, SpecBuilder,
 };
 use path_clean;
 use serde_json::to_writer_pretty;
 use std::fs::File;
+use std::path::PathBuf;
 /// Command generates a config.json
 #[derive(Clap, Debug)]
 pub struct SpecJson {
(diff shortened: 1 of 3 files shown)


chenyukang:
Reference: [runc](https://github.com/opencontainers/runc/blob/d1c9b43e94b49d77cede5af42ef802961584b5bc/libcontainer/specconv/example.go#L171)

src/commands/spec_json.rs
```diff
@@ -1,18 +1,106 @@
...
+            .host_id(gid)
+            .container_id(0_u32)
+            .size(1_u32)
+            .build()?]);
+
+    // Fix the mounts
+    let mut mounts = vec![];
+    for mount in spec.mounts().as_ref().unwrap().iter() {
+        let dest = mount.destination().clone();
+        if path_clean::clean(dest.as_path().to_str().unwrap()) == "/sys" {
```
yihuaf, line 63: Add comments to explain why you need to use `path_clean` here?

yihuaf (commented):
Otherwise LGTM

src/commands/spec_json.rs
```diff
@@ -1,18 +1,106 @@
...
+pub fn set_for_rootless(spec: &Spec) -> Result<Spec> {
+    let uid = nix::unistd::geteuid().as_raw();
+    let gid = nix::unistd::getegid().as_raw();
+
+    // Remove network from the default spec
+    let mut namespaces = vec![];
+    for ns in spec
+        .linux()
+        .as_ref()
+        .unwrap()
```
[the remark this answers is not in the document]
chenyukang: > Can we not use `unwrap`? I know it may take a few more line to return results, but let's avoid unwrap in general.

how about use `unwrap_or` ?

src/commands/spec_json.rs
```diff
@@ -1,20 +1,123 @@
...
+#[cfg(test)]
+// Tests become unstable if not serial. The cause is not known.
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_spec_json() -> Result<()> {
+        let mut spec = Default::default();
+        spec = set_for_rootless(&spec)?;
+        to_writer_pretty(&File::create("config.json")?, &spec)?;
```
yihuaf, line 120: This may be the cause for unstable. Are you creating the same `config.json` with the other test? You may want to use temp file.

chenyukang: > This may be the cause for unstable. Are you creating the same `config.json` with the other test? You may want to use temp file.

Yes, will fix it.

src/commands/spec_json.rs
```diff
@@ -1,20 +1,125 @@
...
+#[cfg(test)]
+// Tests become unstable if not serial. The cause is not known.
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_spec_json() -> Result<()> {
+        let mut spec = Default::default();
+        spec = set_for_rootless(&spec)?;
+        let tmpdir = tempfile::tempdir().unwrap();
```
utam0k, line 120: How about using this?
https://github.com/containers/youki/blob/9bb360a48db6094fad89144b522e31ed01384e72/src/utils.rs#L148

utam0k (approved):
Awesome!

PR #352, opened 2021-10-02
remove dead code in src/utils.rs

hle0:
There is some dead code I found in src/utils.py when starting to write unit tests for it. This PR removes it.

Thank you!


utam0k (approved):
Thanks!

PR #353, opened 2021-10-02
Add pid to newuidmap/newgidmap as argument

shorii:
`newuidmap`/`newgidmap` expects pid as first argument.

https://www.man7.org/linux/man-pages/man1/newuidmap.1.html
https://www.man7.org/linux/man-pages/man1/newgidmap.1.html


utam0k (approved):
LGTM

Issue #346, opened 2021-09-29
Implement secure join


yihuaf:
We need a `secure_join` implementation that's similar to `https://github.com/cyphar/filepath-securejoin`. In short, this function guarantee `secure_join(rootfs, path)` the `path` is within `rootfs`. When `path` is `../../` or other malicious symbolic links, it can cause youki to make changes outside of rootfs, creating security issue. This can be a follow up to #342 

An example of usage in runc: 

https://github.com/opencontainers/runc/blob/51beb5c436b159ae2d483b219c37ecfde13b006a/libcontainer/utils/utils.go#L119

Ian-Yy:
Hi, I'm quite new to youki. Can I try to work on this?

yihuaf:
@Ian-Yy Sure. Do you know where to start? Please don't hesitate to ask us on how we can help you.

Ian-Yy:
@yihuaf I will try going through the code first. Thanks.

Ian-Yy:
Hi @yihuaf , correct me if I am wrong. This task require me to implement a new function ```secure_join``` in the trait ```PathBufExt```? Is that right?

yihuaf:
> Hi @yihuaf , correct me if I am wrong. This task require me to implement a new function `secure_join` in the trait `PathBufExt`? Is that right?

Yes that's a very good idea. Or you can just implement a function `secure_join(rootfs: Path, path: Path)` as a single function.

Ian-Yy:
Sorry but there's one thing I'm quite unsure of now. The implementation of ```secure_join``` in ```https://github.com/cyphar/filepath-securejoin``` uses a lot of ```filepath.Clean``` and we do not have it in Rust. I saw there's a crate [https://docs.rs/path-clean/0.1.0/path_clean/index.html] that does exactly the same as Go ```filepath.Clean``` but is quite old. Should I use this crate or code it?

yihuaf:
Let's use it to at least prototype.

Ian-Yy:
Hi @yihuaf , I've done with a prototype with some tests for it. After working on it for some time, I realized I don't need to use the crate ```path_clean``` as mentioned above. Please let me know if there's anything wrong with the PR.

PR #354, opened 2021-10-02
Implement secure_join for path

Ian-Yy:
This PR serve as a prototype for issue #346

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -8,6 +8,7 @@ use nix::unistd;
 use std::collections::HashMap;
 use std::ffi::CString;
 use std::fs::{self, DirBuilder, File};
+use std::io::ErrorKind;
 use std::ops::Deref;
 use std::os::linux::fs::MetadataExt;
 use std::os::unix::fs::DirBuilderExt;
@@ -153,7 +154,6 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
 }
 
 pub fn secure_join(rootfs: &Path, unsafe_path: &Path) -> Result<PathBuf> {
-    
     let mut clean_path = PathBuf::from(rootfs);
     let mut path = PathBuf::from(unsafe_path);
     let mut p = PathBuf::new();
@@ -164,14 +164,14 @@ pub fn secure_join(rootfs: &Path, unsafe_path: &Path) -> Result<PathBuf> {
     loop {
         if i > 255 {
             // Dereference too many symlinks, may be infinite loop
-            break;
+            bail!("dereference too many symlinks, may be infinite loop");
         }
 
         let part_path;
         match part.next() {
             Some(part) => {
                 part_path = PathBuf::from(part);
-            },
+            }
             None => {
                 break;
             }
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -152,6 +152,66 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
...
+    let mut clean_path = PathBuf::from(rootfs);
+    let mut path = PathBuf::from(unsafe_path);
+    let mut p = PathBuf::new();
+
+    let mut part = path.iter();
+    let mut i = 0;
+
+    loop {
+        if i > 255 {
+            // Dereference too many symlinks, may be infinite loop
```
yihuaf, line 166: Can we error out here?

yihuaf, line 166: Can we error out in this use? You can use `bail!("dereference too many symlinks, may be infinite loop");`

src/utils.rs
```diff
@@ -152,6 +152,66 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
...
+                break;
+            }
+        }
+
+        if !part_path.is_absolute() {
+            if part_path.starts_with("..") {
+                p.pop();
+            } else {
+                // check if symlink then dereference
+                let metadata = match PathBuf::from(&clean_path).join(&p).join(&part_path).symlink_metadata() {
```
yihuaf, line 185: We can error out here as well?

src/utils.rs
```diff
@@ -152,6 +152,66 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
...
+                    Ok(metadata) => Some(metadata),
+                    Err(_) => None,
+                };
+
+                match metadata {
```
yihuaf, line 190: This can be `if let Some(metadata) = metadata {...}`.

yihuaf (commented):
Looks good overall.

yihuaf:
You may have to run `cargo fmt --all -- --check` and `cargo clippy --all`.

src/utils.rs
Ian-Yy: Do you mean error out if the metadata is not obtained? The ```symlink_metadata``` will throw error if the file does not exists and in my ```test_secure_join```, the paths I'm using don't exist. So, those tests will fail if I error out here. One way is to create those paths like in ```test_secure_join_symlink```. What do you think?

Ofenhed:
This is overly complex. Did you check out [fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html)?

utam0k:
> This is overly complex. Did you check out [fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html)?

I was also wondering if it could be implemented more simply with `canonicalize`. If you have tried to implement using it and found it difficult, please let me know.

yihuaf:
It is not the same thing unfortunately. First, canonicalize requires the path to first exist, which is not true when we have to evaluate path inside the container root. For example, device path may not be created or mounted, but we have to resolve the path first. Second, the requirement for evaluate symlink to be relative to the rootfs, not relative to the system host root. For example, this is equivalent of:
```
chroot rootfs
readlink --canonicalize-missing --no-newline unsafe_path
```

See https://github.com/cyphar/filepath-securejoin for details and this is the same function used in the `runc` implementation.

utam0k:
> It is not the same thing unfortunately. First, `fs::canonicalize` requires the path to first exist, which is not true when we have to evaluate path inside the container root. For example, device path may not be created or mounted, but we have to resolve the path first. Second, the requirement for evaluate symlink to be relative to the rootfs, not relative to the system host root. For example, this is functionally equivalent of:
> 
> ```
> chroot rootfs
> readlink --canonicalize-missing --no-newline unsafe_path
> ```
> 
> See https://github.com/cyphar/filepath-securejoin for details and this is the same function used in the `runc` implementation.

I understood. Thanks.

src/utils.rs
yihuaf: I see. Then let's add a comment here to explain. Also can be just ignore the specific error and error out when it's the other error? I assume we only want to return `None` for the case symblink doesn't exist.

Ian-Yy:
Hi, I've updated the code according to @yihuaf reviews and checked both ```cargo fmt --all -- --check``` and ```cargo clippy --all``` too this time. Please let me know if there's anything to add.

src/utils.rs
yihuaf: You can remove this comment now.

src/utils.rs
```diff
@@ -152,6 +153,77 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
     Ok(())
 }
 
+pub fn secure_join(rootfs: &Path, unsafe_path: &Path) -> Result<PathBuf> {
+    let mut clean_path = PathBuf::from(rootfs);
+    let mut path = PathBuf::from(unsafe_path);
+    let mut p = PathBuf::new();
```
yihuaf, line 159: Can we use another name instead of `p`. It's not clear what this does.

src/utils.rs
```diff
@@ -152,6 +153,77 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
...
+                    .symlink_metadata()
+                {
+                    Ok(metadata) => Some(metadata),
+                    Err(error) => match error.kind() {
+                        // if file does not exists, treat it as normal path
+                        ErrorKind::NotFound => None,
+                        other_error => {
+                            bail!(
+                                "unable to obtain symlink metadata for file {:?}: {:?}",
+                                PathBuf::from(&clean_path).join(&p).join(&part_path),
```
yihuaf, line 197: We can save this as a variable so we don't have to calculate twice.

src/utils.rs
```diff
@@ -152,6 +153,77 @@ pub fn ensure_procfs(path: &Path) -> Result<()> {
...
+                                other_error
+                            );
+                        }
+                    },
+                };
+
+                if let Some(metadata) = metadata {
+                    if metadata.file_type().is_symlink() {
+                        let link_path =
+                            fs::read_link(PathBuf::from(&clean_path).join(&p).join(&part_path))
```
yihuaf, line 207: You can use `?` here instead of `unwrap`? We want to return error here instead of `unwrap`.

yihuaf (commented):
`cargo clippy --all-targets --all-features -- -D warnings` to make CI happy

Ian-Yy:
> `cargo clippy --all-targets --all-features -- -D warnings` to make CI happy

Sorry. I thought running `cargo clippy --all` is enough. Can we perhaps put this and `cargo fmt` somewhere in the documentation for newcomers? 

Meanwhile, I will be making the changes.

yihuaf (approved):

PR #362, opened 2021-10-04
part of PR 340 adding test_bind_dev,test_mknod_dev and test_create_devices

tommady:
as requested by separating #340 into small PRs

this one is the fourth part which is adding three test cases:

- test_bind_dev
- test_mknod_dev
- test_create_devices

please help review 🙇🏻

### DIFF
--- a/src/rootfs.rs
+++ b/src/rootfs.rs
@@ -306,13 +306,11 @@ impl RootFS {
                     bail!("{} is not a valid device path", dev.path().display());
                 }
 
-                crate::utils::create_dir_all_with_mode(
+                crate::utils::create_dir_all(
                     rootfs
                         .join(dev.path().as_in_container()?)
                         .parent()
-                        .unwrap_or_else(|| Path::new("/")),
-                    dev.uid().unwrap_or(0),
-                    Mode::from_bits_truncate(0o755),
+                        .unwrap_or_else(|| Path::new("")),
                 )?;
 
                 if bind {
--- a/src/rootfs.rs
+++ b/src/rootfs.rs
@@ -299,29 +299,29 @@ impl RootFS {
         I: IntoIterator<Item = &'a LinuxDevice>,
     {
         let old_mode = umask(Mode::from_bits_truncate(0o000));
-        if bind {
-            let _ = devices
-                .into_iter()
-                .map(|dev| {
-                    if !dev.path().starts_with("/dev") {
-                        panic!("{} is not a valid device path", dev.path().display());
-                    }
+        devices
+            .into_iter()
+            .map(|dev| {
+                if !dev.path().starts_with("/dev") {
+                    bail!("{} is not a valid device path", dev.path().display());
+                }
 
+                crate::utils::create_dir_all_with_mode(
+                    rootfs
+                        .join(dev.path().as_in_container()?)
+                        .parent()
+                        .unwrap_or_else(|| Path::new("/")),
+                    dev.uid().unwrap_or(0),
+                    Mode::from_bits_truncate(0o755),
+                )?;
+
+                if bind {
                     self.bind_dev(rootfs, dev)
-                })
-                .collect::<Result<Vec<_>>>()?;
-        } else {
-            devices
-                .into_iter()
-                .map(|dev| {
-                    if !dev.path().starts_with("/dev") {
-                        panic!("{} is not a valid device path", dev.path().display());
-                    }
-
+                } else {
                     self.mknod_dev(rootfs, dev)
-                })
-                .collect::<Result<Vec<_>>>()?;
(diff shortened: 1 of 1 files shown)


tommady:
sorry, the test failures are some kind of I don't understand...
```bash
cargo test --all --all-features --no-fail-fast
/build-script-build` (exit status: 101)
  --- stdout
    MKDIR    staticobjs
    CC       bpf.o
    CC       btf.o

  --- stderr
  Package libelf was not found in the pkg-config search path.
  Perhaps you should add the directory containing `libelf.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'libelf' found
  btf.c:18:10: fatal error: gelf.h: No such file or directory
     18 | #include <gelf.h>
        |          ^~~~~~~~
  compilation terminated.
  make: *** [Makefile:113: /home/tommady/rust/youki_tommady/target/x86_64-unknown-linux-gnu/debug/build/libbpf-sys-238adb98268543af/out/obj/staticobjs/btf.o] Error 1
```
```bash
info: when --target option is used, coverage for proc-macro and build script will not be displayed because cargo does not pass RUSTFLAGS to them
   Compiling seccomp v0.1.0 (/home/runner/work/youki/youki/seccomp)
   Compiling cgroups v0.1.0 (/home/runner/work/youki/youki/cgroups)
   Compiling youki v0.0.1 (/home/runner/work/youki/youki)
warning: field is never read: `user_namespace`
  --> src/rootless.rs:20:5
   |
20 |     pub(crate) user_namespace: Option<LinuxNamespace>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `youki` (lib) generated 1 warning
warning: `youki` (lib test) generated 1 warning (1 duplicate)
    Finished test [unoptimized + debuginfo] target(s) in 22.04s
     Running unittests (target/llvm-cov-target/x86_64-unknown-linux-gnu/debug/deps/youki-a853ea8c9159406b)
```
need to check more deeper, or is that a known issue? that I can wait for the fix 🙇🏻‍♂️

Furisto:
Looks like a temporary issue with the package server. I restarted the workflow.

tommady:
hi the current status is that will have a flaky failure 

```bash
test stats::tests::test_parse_space_separated_as_nested_keyed_data ... FAILED

failures:

---- stats::tests::test_parse_space_separated_as_nested_keyed_data stdout ----
thread 'stats::tests::test_parse_space_separated_as_nested_keyed_data' panicked at 'called `Result::unwrap()` on an
`Err` value: failed to open "/tmp/test_parse_newline_separated_as_nested_keyed_data/space_separated"

Caused by:
    No such file or directory (os error 2)', cgroups/src/stats.rs:589:77
```
trying to make some fix

src/rootfs.rs
```diff
@@ -940,4 +938,139 @@ mod tests {
             .get_symlink_args();
         assert_eq!(want, got)
     }
+
+    #[test]
+    #[serial]
```
utam0k, line 943: Why does this test need `#[serial]`?

src/rootfs.rs
```diff
@@ -940,4 +938,139 @@ mod tests {
...
+        };
+        let got = &rootfs
+            .syscall
+            .as_any()
+            .downcast_ref::<TestHelperSyscall>()
+            .unwrap()
+            .get_mount_args()[0];
+        assert_eq!(want, *got);
+
+        remove_file("/tmp/null").unwrap();
```
utam0k, line 971: I should have pointed this out in an earlier test, but how about using this?
https://github.com/containers/youki/blob/c4264a1096f417f2f5348c56dba375bea97af22b/src/utils.rs#L221

src/rootfs.rs
tommady: hi @utam0k , sorry I do not quite understand about the serial, 
about it, for me about this derive is telling the cargo to do the testing step by step not in parallel.

so please guide if I am wrong 🙇🏻

src/rootfs.rs
tommady: sure will address this

src/rootfs.rs
utam0k: Sure! Your understanding is correct. The `serial` is used for tests that are unstable in parallel depending on the environment in which they are run, such as using the `pipe`. It seems to me that this test does not need to be serial, perhaps with some effort.
Also, here is a script that will run the unit tests over and over again.
https://github.com/containers/youki/blob/main/hack/stress_cargo_test.sh

utam0k (approved):
Perfect

PR #363, opened 2021-10-05
Use generic for signal argument in container_kill

alfonsoros88:
`nix::sys::signal::Signal` is part of the API for the `kill` related calls. This causes a problem for lib users when they have a different nix version.

This patch introduces a public `youki::signal::Signal` type to be used instead.


yihuaf:
Can you describe the issue you ran into in more details?

Furisto (approved):

Furisto:
Yeah, originally youki was not designed to be used as a library, so people will run into these kind of kinks. We will iron them out over time.

PR #364, opened 2021-10-05
Update Youki with latest oci-spec-rs

yihuaf:
Now that setters are in, remove the boilerplate code.


yihuaf:
There is a bug in the `oci-spec-rs`. Pending PR: https://github.com/containers/oci-spec-rs/pull/78

PR #365, opened 2021-10-06
[WIP] part of PR 340 adding test_mount_to_container

tommady:
as requested by separating #340 into small PRs

this part is adding:

1. test_mount_to_container

please help review 🙇🏻

### DIFF
--- a/src/rootfs.rs
+++ b/src/rootfs.rs
@@ -378,26 +378,22 @@ impl RootFS {
         label: Option<&str>,
     ) -> Result<()> {
         let typ = m.typ().as_deref();
-        let d = if let Some(l) = label {
+        let mut d = data.to_string();
+
+        if let Some(l) = label {
             if typ != Some("proc") && typ != Some("sysfs") {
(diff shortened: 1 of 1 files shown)


utam0k:
It's time to create a rootfs directory and separate the files.

tommady:
the only logic I modified at [here](https://github.com/containers/youki/pull/365/files#diff-314a54fbe7cedbdbac5064dea18be7800c584124bb51550b58fee172cd624ee8R443)
is because referenced to runC at [here](https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L450)

they were using device type is bind as condition,
not like youki did flags contains bind,
and it makes more sense because the youki condition was

`if flags contain MS_BIND and there is no intersects with MS_BIND`

this sentence looks very weird.

utam0k:
@tommady Instead of creatting an exclusive file for test, it is more natural in Rust to separate some functions into appropriate files based on some criteria, and put the tests associated with the functions into the same file.

PR #373, opened 2021-10-08
Support systemd named hierarchy and emulate cgroup namespaces for v1 control cgroups

Furisto:
This adds two features:
- If cgroup namespaces are not available we can emulate them using bind mounts.
- Named hierarchies are cgroup hierarchies with no attached controllers. Systemd uses these to track services. More info [here](https://man7.org/linux/man-pages/man7/cgroups.7.html).


utam0k:
@Furisto I am sorry, but I have conflated this PR.

Issue #279, opened 2021-09-07
Increate the code coverage of `src/` to 75%


utam0k:
## Why?
Since youki is a container runtime, it is software that should be security-sensitive. Therefore, it should be covered as much as possible by unit tests to avoid bugs as much as possible.
However, because it is a container runtime, it is difficult to write unit tests because it is difficult to handle mocks. In other words, this is a challenge. It is also a chance to design well.

## How to participate in this challenge
Please indicate that you would like to challenge this issue. I will look at the code coverage and recommend where to challenge. If you are new to youki development, it would be helpful if you could describe how familiar you are with Rust and container runtimes.

tommady:
Hi @utam0k ,
may I take this challenge (or part of it)?

I am very inexperienced on both 1. the rust and 2. the container runtime.

about the rust, I only have some small contributions to
* [spodifyd](https://github.com/Spotifyd/spotifyd)
* [ta-rs](https://github.com/greyblake/ta-rs)
* [gbdt-rs](https://github.com/mesalock-linux/gbdt-rs)

and a small project I owned
[fugle-rs](https://github.com/tommady/fugle-rs)

about the container runtime, 
I only know the [issue](https://github.com/containers/youki/pull/289) I just solved.

so hope you will accept my self-recommended and give me this chance to learn more? 🙇🏻 
thank you.

utam0k:
@tommady Of cource! Can I ask you to try adding tests to it?
https://codecov.io/gh/containers/youki/src/main/src/capabilities.rs

tommady:
hi @utam0k please help to review #296 
thank you 🙇🏻 

> If you want, I can prepare more difficult parts.

of course, I'd glade to

utam0k:
@tommady Can I ask you to update the test coverage for this file?
https://app.codecov.io/gh/containers/youki/blob/main/src/hooks.rs

If you want something more difficult, I can ask for the most difficult, init.rs or intermediate.rs. Please take a look at these files after you have done the previous ones first and let me know if you want to try them. Implementing unit tests for these files will be a tremendous challenge and an exciting one. However, it will probably take time and patience to take on these challenges, so I would like to hear your opinion. Of course, I'm here to help.

yihuaf:
For hook.rs, there is a test that tests the timeout path, but it's skipped at the moment. The min timeout is 1 second, so the unit test would take at least 1 second to run. I wasn't sure if we should include that. Maybe remove the #[skip] and try again for the coverage first?

[...]

PR #376, opened 2021-10-09
279 increate the code coverage of src container

tommady:
#279
the container/container part


tommady:
Thanks for teaching me that!
I didn't know that,
Will addressed.

Thanks again!

PR #378, opened 2021-10-11
Support cgroup v2 mounts

Furisto:
This supports specifying cgroup mounts on systems with unified hierarchy. In case using cgroup namespaces do not work (could not get it to work with rootless containers), the host cgroup hierarchy is bind mounted into the container instead. Also fixed some path handling issues I noticed.


utam0k:
Awesome

PR #384, opened 2021-10-14
implement seccomp notify

### LINKED ISSUE #307: seccomp notify

yihuaf:
1. Implemented seccomp notify listener logic based on the oci spec.
2. Because seccomp notify listener needs to know the container init process id, I have to make intermediate process to return init process pid right after forking the init process. After that, the intermediate process will return and the init process will sync with the main process directly.

Note: I just realize while implementing this PR that seccomp notify is a fairly new thing that is less than a year old. seccomp v2.5.2 was released Aug 2021 and still includes changes to the seccomp notify. Likely there will be cases where we need a finer grained control over seccomp api level, which is not implemented in the PR. Most of the time, seccomp listener and notify features are not invoked, so this PR code path won't get triggered. I will iron out these kinks with follow up PRs.

Fix #307


utam0k:
@yihuaf Cool!
Can I ask you to update `docs/.draw.io.svg` and add an integration test about seccomp notify?

yihuaf:
I updated the drawio.svg.

> add an integration test about seccomp notify?

It is probably easier to do a follow up PR. Likely I have to port over "seccompagent" for the testing:
ref: https://github.com/opencontainers/runc/tree/master/contrib/cmd/seccompagent

utam0k:
@yihuaf Thanks for your update. I think it would be worthwhile to add the interactions between each process about seccomp/seccomp notify to the sequence diagram. What do you think?

Is it difficult to add this into youki's original integration test that we are working on? I have no problem at all with another PR.
> It is probably easier to do a follow up PR. Likely I have to port over "seccompagent" for the testing:
> ref: https://github.com/opencontainers/runc/tree/master/contrib/cmd/seccompagent

PR #389, opened 2021-10-16
Add ns_itype test

YJDoc2:
Setup ns_itype test, but I'm not sure if this is set correctly or not :
The original test seem to do things in complecated ways, so I'm worried that I might have missed something. 
https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_ns_itype/linux_ns_itype.go#L58 here, it loops through the namespaces, which are defined https://github.com/opencontainers/runtime-tools/blob/master/validation/util/linux_namespace.go here ; and then removes them from the default spec, https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_ns_itype/linux_ns_itype.go#L70 here. But this essentially removes all namespaces from the spec. so in my implementation, I have used the builder with empty vec to remove all namespaces. Then using procfs, I have directly compared the namespace inodes of host and container.
Is it correct? @utam0k @yihuaf can you please take a look?

### DIFF
--- a/youki_integration_test/src/tests/linux_ns_itype/ns_itype_test.rs
+++ b/youki_integration_test/src/tests/linux_ns_itype/ns_itype_test.rs
@@ -33,27 +33,33 @@ fn get_test<'a>(test_name: &'static str) -> Test<'a> {
     Test::new(
         test_name,
         Box::new(move || {
-            let host_proc = Process::myself().unwrap();
+            let host_proc = Process::myself().expect("error in getting /proc/self");
             let host_namespaces = match host_proc.namespaces() {
                 Ok(n) => n,
                 Err(e) => {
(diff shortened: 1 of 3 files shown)


YJDoc2:
Currently I have left a few extra comments and commented code, If this is correct, I'll fix that in another commit, I have opened this PR because I wasn't sure if the current implementation is correct or not.

yihuaf:
Super. Looking forward to the final PR :)

Furisto:
Just fyi, the correct way to compare namespaces is with inode _and_ device id, but the eq implementation of procfs namespace does that already for you, so you are fine.

YJDoc2:
Ohh, I originally compared only by inode before changing it to direct comparison of HashMaps :sweat_smile:  Thanks for the info!

utam0k:
How about making sure that unwrap comes with the appropriate error wording?

Issue #361, opened 2021-10-04
[Tracking Issue] Implement Integraion Tests


YJDoc2:
As per #56 , we are porting the OCI validation tests into rust. This is the tracking issue for the same.
The folder /youki_integration_test contains the currently implemented tests which are lifecyle, creation and huge_tlb

Here is the documentation for developing the integration tests:
https://youki-dev.github.io/youki/developer/e2e/rust_oci_test.html

**Please only work at one test at a time, and wait for its PR to get merged before opening another one**. It helps us to have an easier way for reviewing the PRs 🙏 

## Default
List of all test, as per https://github.com/opencontainers/runtime-tools/tree/master/validation :
- [ ] config_updates_without_affect
  - Original test code in Go: https://github.com/opencontainers/runtime-tools/blob/master/validation/config_updates_without_affect/config_updates_without_affect.go
- [ ] create(@sivchari )
  - Original test code in Go: https://github.com/opencontainers/runtime-tools/blob/master/validation/create/create.go
- [ ] default
  - Original test code in Go: https://github.com/opencontainers/runtime-tools/blob/master/validation/default/default.go
- [x] delete (@sou1118 ) https://github.com/youki-dev/youki/pull/3082
- [ ] delete_only_create_resources
- [ ] delete_resources
- [x] hooks
- [ ] hooks_stdin
- [x] hostname(@chermehdi)
- [x] kill (@YamasouA  https://github.com/youki-dev/youki/pull/2996)
- [ ] kill_no_effect
- [ ] killsig (@YamasouA )
- [x] linux_cgroups_blkio(#537)
- [x] linux_cgroups_cpus(#462)
- [ ] linux_cgroups_devices ( @sat0ken )
- [x] linux_cgroups_hugetlb
- [x] linux_cgroups_memory(#473)
- [x] linux_cgroups_network(#516)
- [x] linux_cgroups_pids(#391)
- [ ] linux_cgroups_relative_blkio(@furon-kuina)
- [ ] linux_cgroups_relative_cpus // *Please take over from https://github.com/youki-dev/youki/pull/2898*
- [ ] linux_cgroups_relative_devices (@sat0ken )
- [ ] linux_cgroups_relative_hugetlb(@kmpzr)
- [ ] linux_cgroups_relative_memory(@omprakaash)
- [x] linux_cgroups_relative_network (@moz-sec  https://github.com/youki-dev/youki/pull/2986 )
[...]

PR #391, opened 2021-10-16
Cgroup v1 pid integration tests

Furisto:
- Implements the cgroup v1 pid tests
- Added more context to error messages and added a new function for checking if the container was created successfully
- I realize this is subjective but I found Passed, Skipped and Failed more understandable for TestResult so I changed the names. I can change it back if you prefer the older names.
- I think we should have a convention for the used cgroups path. I set it as [cgroup_root]/runtime-test/[test_name]
- The spec does not state what should happen if the limit is zero or negative, but we and runc just set unlimited in this case. We should write down where the spec is ambiguous, so that we do not forget it. This can then be used to improve the spec. For now I just added //SPEC in front of such comments in the code. Other suggestions welcome.

#361

### DIFF
--- a/youki_integration_test/README.md
+++ b/youki_integration_test/README.md
@@ -47,6 +47,8 @@ This framework also has some test utils, meant to help doing common operations i
 - delete_container : runs the runtime command with delete argument, with given id and with given bundle directory
 - get_state : runs the runtime command with state argument, with given id and with given bundle directory
 - test_outside_container : this is meant to mimic [validateOutsideContainer](https://github.com/opencontainers/runtime-tools/blob/59cdde06764be8d761db120664020f0415f36045/validation/util/test.go#L263) function of original tests.
+- check_container_created: this checks if the container was created succesfully.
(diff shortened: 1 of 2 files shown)


YJDoc2:
Hey, I will take a good detailed look once again tomorrow, but some things that I noticed :
- context messages and helper function is an excellent addition! Thank you for adding that !
- Our TestResult enum was meant to follow after Rust's Result, with adding a Skip variant, so I kept names of others same as Result. I do understand that Passed, Skipped and Failed make more sense, but I also feel a bit that if we are using Result in the name, we should keep the variant names similar :sweat_smile: I'll think more on this...
- Using a common convention for cgroup path is good, but the created cgroup is not removed once the test are over (runtime-test directory remains after tests are over). We can manually remove this with remove_dir, after all container processes are exited, but then we should add the clean up function. I don't feel leaving an extra cgroup directory (even if it doesn't give issues) on tester's system is a good thing.

Furisto:
@YJDoc2 Done

Issue #396, opened 2021-10-18
[Discussion] Share code between youki_integration_test and youki


yihuaf:
It is inevitable that the youki_integration_test and youki will share some code. For example, the definition of container state is already duplicated, as well as tempdir from util. Should we find a way to share code between youki and youki_integration_test?

One proposal is to make most part of Youki into a library crate and Youki binary as another crate. Then we can make youki_integration_test another binary crate and share the youki library crate.

@utam0k @YJDoc2 @Furisto would love to hear your opinions on this.

yihuaf:
This is almost exactly what I have in mind as well. A few minor things. How about we call `containers` crate `libcontainer` to reflect that it is a library crate? Should we separate integration tests and the test framework into first level crate as well?   
```
crates
  - ... (what you mentioned above)
  - test/integration_test (bin crate)
  - test/framework (library crate)
hack
docs
...

```

Furisto:
I would like to do it. I have a free day tomorrow, so even if it turns out to be more complicated I should be finished tomorrow latest.

yihuaf:
Go right ahead then :)

utam0k:
Thanks for the great suggestion. Let's try it. `lib*` is a traditional name in the container runtime community, but I haven't seen it used elsewhere as a library much. Let's try to find a name that works best for us.
Please add me as a PR reviewer.

PR #403, opened 2021-10-21
Make youki a library crate

Furisto:
As discussed in #396, this creates a new crate libcontainer, the youki binary itself will only be a thin wrapper around this library. All crates have been moved under the crate folder, so that we now have the following structure
```
crates
- integration_test
- libcgroups
- libcontainer
- libseccomp
- test-framework
docs
hacks
...
```
Despite my comments in #396, I have now added a lib prefix as paths like container::container::Container, not only look weird, but also confused the compiler. 
No that we have one cargo.lock file, there was  a problem with multiple versions of clap incompatible to each other, so I had to settle on clap 3.0.0.0-beta.2, beta.3 and beta.4 did not work for the integration test, and with beta.5 the compiler complained that the macro resolution got so complicated that the compilation had to be aborted. That also means that I had to remove "forbid_empty_values" from the command args as it is not available in this version. Besides this, there should not be any functional changes.


yihuaf:
Also may be wait for @utam0k to comment/approve this as well.

PR #445, opened 2021-10-30
refactoring the syscall test

tommady:
#279 
part of it.

- refactoring testing syscall
- taking off syscall Linux coverage rate because it cannot be covered

### DIFF
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -515,7 +515,6 @@ mod tests {
         syscall::create_syscall,
         test::{ArgName, MountArgs, TestHelperSyscall},
     };
-    use anyhow::{anyhow, Context};
     use nix::{fcntl, sys, unistd};
     use oci_spec::runtime::{LinuxNamespaceBuilder, SpecBuilder, UserBuilder};
     use serial_test::serial;
@@ -726,7 +725,7 @@ mod tests {
             .as_any()
             .downcast_ref::<TestHelperSyscall>()
             .unwrap();
-        mocks.set_ret_err(ArgName::Mount, || Err(anyhow!(nix::errno::Errno::ENOENT)));
+        mocks.set_ret_err(ArgName::Mount, || bail!(nix::errno::Errno::ENOENT));
 
         assert!(masked_path(Path::new("/proc/self"), &None, syscall.as_ref()).is_ok());
         let got = mocks.get_mount_args();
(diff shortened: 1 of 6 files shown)


crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -722,37 +722,30 @@ mod tests {
...
+        let mocks = syscall
             .as_any()
             .downcast_ref::<TestHelperSyscall>()
-            .unwrap()
-            .set_mount_ret_err(
-                Some(|| Err(anyhow!(nix::errno::Errno::ENOENT).context(""))),
-                1,
-            );
+            .unwrap();
+        mocks.set_ret_err(ArgName::Mount, || Err(anyhow!(nix::errno::Errno::ENOENT)));
```
utam0k, line 729: How about this?
```suggestion
        mocks.set_ret_err(ArgName::Mount, || bail!(nix::errno::Errno::ENOENT));
```

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -28,13 +28,15 @@ use crate::capabilities;
 pub struct LinuxSyscall;
 
 impl LinuxSyscall {
+    #[cfg_attr(coverage, no_coverage)]
```
utam0k, line 31: Can't we just apply this to the whole file?

utam0k (commented):
It's almost great! I've made a few comments.

crates/libcontainer/src/syscall/linux.rs
tommady: @utam0k I tried in the last commit, seems no use.
#[cfg_attr(coverage, no_coverage)]
this mark seems to need wrapping all functions to avoid being counted into coverage.

please guide me if I am doing anything wrong.
I'll reverse the last commit if there is really no way to achieve the goal. 🙇🏻

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -141,10 +142,12 @@ impl Syscall for LinuxSyscall {
                         _ => caps::drop(None, CapSet::Bounding, *c)?,
                     }
                 }
-                Ok(())
             }
-            _ => caps::set(None, cset, value),
+            _ => {
+                caps::set(None, cset, value)?;
+            }
```
utam0k, line 148: Why do we need this change?

tommady: because I changed the trait contract from
```
set_capability(&self, cset: CapSet, value: &CapsHashSet) -> Result<(), CapsError>
```
to 
```
set_capability(&self, cset: CapSet, value: &CapsHashSet) -> Result<()>
```
to fulfill anyhow as the Result not using the std Result.

utam0k (approved):
looks good

PR #451, opened 2021-11-03
Support resource control via systemd

Furisto:
This adds support for resource control using dbus and systemd. Currently supported are:
- Starting and removing transient units
- Resource control of cpu, cpusets and pids

Not yet supported is resource control with rootless containers. I have this working locally, but it still needs some more work and testing. This means that the following tasks are still open and will be followed up in another PR:
- Support for rootless containers
- Resource restrictions defined in the unified section
- Memory resource restrictions
- Integration tests

Thank you to @nimrodshn for his previous work on this, which is also included in this PR.


crates/libcgroups/src/systemd/dbus/client.rs
```diff
@@ -0,0 +1,126 @@
...
+        pid: u32,
+        parent: &str,
+        unit_name: &str,
+    ) -> Result<()> {
+        // To view and introspect the methods under the 'org.freedesktop.systemd1' destination
+        // and object path under it use the following command:
+        // `gdbus introspect --system --dest org.freedesktop.systemd1 --object-path /org/freedesktop/systemd1`
+        let proxy = self.create_proxy();
+
+        // To align with runc, yuoki will always add the following properties to its container units:
```
utam0k, line 43:
```suggestion
        // To align with runc, youki will always add the following properties to its container units:
```

utam0k:
Please give me some more time to understand this PR.

crates/libcgroups/src/systemd/cpu.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn is_realtime_requested(cpu: &LinuxCpu) -> bool {
+        if cpu.realtime_period().is_some() {
+            return true;
+        }
+
+        if cpu.realtime_runtime().is_some() {
+            return true;
+        }
+
+        false
```
utam0k, line 70: How about this?
```suggestion
        cpu.realtime_period().is_some() || cpu.realtime_runtime().is_some()
```

crates/libcgroups/src/systemd/cpuset.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl CpuSet {
+    fn apply(
+        cpu: &LinuxCpu,
+        systemd_version: u32,
+        properties: &mut HashMap<String, Box<dyn RefArg>>,
+    ) -> Result<()> {
+        if systemd_version < 244 {
+            bail!(
+                "Systemd version ({}) is too old to support cpuset restrictions",
```
utam0k, line 38:
```suggestion
                "systemd version ({}) is too old to support cpuset restrictions",
```

utam0k:
Why don't you collect the property keys in `mod.rs` or something?
https://github.com/systemd/systemd/blob/20f7ada699f41882ed455bb6de4ab96a7c012427/src/core/dbus-cgroup.c#L453-L508

Furisto:
@utam0k PTAL

utam0k:
We'll continue here. @nimrodshn Thanks a lot.
https://github.com/containers/youki/pull/451

Furisto:
@utam0k I think you closed the wrong PR. Did you want to close #120?

utam0k:
> @utam0k I think you closed the wrong PR. Did you want to close #120?

oops! sorry!

a1ien:
We can also remove systemd dependency from libcontainer crate since it not use this dependency at all.

Furisto:
@a1ien Good catch! I forgot about that one.

utam0k (approved):
Awesome

PR #452, opened 2021-11-04
adding test_sync_seccomp for process/container_main_process

tommady:
#279
the process/container_main_process 
test_sync_seccomp

### DIFF
--- a/crates/libcontainer/src/process/container_main_process.rs
+++ b/crates/libcontainer/src/process/container_main_process.rs
@@ -300,38 +300,43 @@ mod tests {
 
         let (mut main_sender, mut main_receiver) = channel::main_channel()?;
         let (mut init_sender, mut init_receiver) = channel::init_channel()?;
-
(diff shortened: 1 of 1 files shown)


crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -276,4 +278,65 @@ mod tests {
...
+                &mut main_receiver,
+            )
+            .unwrap();
+        });
+
+        let fd = scmp_file.into_raw_fd();
+        let notify_th = thread::spawn(move || {
+            assert!(main_sender.seccomp_notify_request(fd).is_ok());
+            assert!(init_receiver.wait_for_seccomp_request_done().is_ok());
+        });
```
utam0k, line 329: Why do we need to use thread?

tommady: yes you are right, we don't need that thread. 🙇🏻

utam0k (approved):

PR #465, opened 2021-11-10
Add debug flag

### LINKED ISSUE #454: Feature request: add `--debug` flag to enable debug logging
It is useful to use youki as a model to observe some of the behavior of other OCI runtimes
Also this may help to find bugs in youki in the future

~~offtopic: I'm looking for the default folder location when runsc doesn't specify --root~~

unknowndevQwQ:
fix #454

### DIFF
--- a/crates/youki/src/logger.rs
+++ b/crates/youki/src/logger.rs
@@ -8,6 +8,7 @@ use std::io::Write;
 use std::path::PathBuf;
 use std::process::exit;
 use std::str::FromStr;
+use std::string::String;
 
 /// If in debug mode, default level is debug to get maximum logging
 #[cfg(debug_assertions)]
@@ -35,14 +36,11 @@ pub fn init(
     };
     let log_level_from_env = match env::var("YOUKI_LOG_LEVEL") {
         Ok(val) => val,
-        Err(err) => {
-            log::error!("unknown log level env: {}", err);
-            exit(1)
-        }
+        Err(_) => String::from("i don't know"),
     };
     let log_level = if log_debug_flag {
         "debug"
-    } else if log_level_from_env != "" {
+    } else if log_level_from_env != "i don't know" {
         &log_level_from_env
     } else {
         DEFAULT_LOG_LEVEL
@@ -50,9 +48,9 @@ pub fn init(
     let log_level = match LevelFilter::from_str(log_level) {
         Ok(val) => val,
         Err(err) => {
-            log::error!("error: {}", err);
+            bail!("error: '{}', val from env: '{}'", err, log_level);
             exit(1)
-        }
+        },
     };
     let target = if let Some(log_file) = log_file {
         let file = OpenOptions::new()
--- a/crates/youki/src/logger.rs
+++ b/crates/youki/src/logger.rs
@@ -2,13 +2,11 @@
 
 use anyhow::{bail, Context, Result};
 use log::LevelFilter;
-use std::env;
+use std::borrow::Cow;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::PathBuf;
-use std::process::exit;
 use std::str::FromStr;
-use std::string::String;
 
 /// If in debug mode, default level is debug to get maximum logging
 #[cfg(debug_assertions)]
@@ -29,28 +27,17 @@ pub fn init(
     log_file: Option<PathBuf>,
     log_format: Option<String>,
 ) -> Result<()> {
+    let filter: Cow<str> = if log_debug_flag {
+        "debug".into()
+    } else if let Ok(level) = std::env::var("YOUKI_LOG_LEVEL") {
+        level.into()
+    } else {
+        DEFAULT_LOG_LEVEL.into()
+    };
     let formatter = match log_format.as_deref() {
         None | Some(LOG_FORMAT_TEXT) => text_write,
         Some(LOG_FORMAT_JSON) => json_write,
         Some(unknown) => bail!("unknown log format: {}", unknown),
-    };
-    let log_level_from_env = match env::var("YOUKI_LOG_LEVEL") {
-        Ok(val) => val,
-        Err(_) => String::from("i don't know"),
-    };
-    let log_level = if log_debug_flag {
-        "debug"
-    } else if log_level_from_env != "i don't know" {
-        &log_level_from_env
-    } else {
-        DEFAULT_LOG_LEVEL
-    };
-    let log_level = match LevelFilter::from_str(log_level) {
-        Ok(val) => val,
-        Err(err) => {
-            bail!("error: '{}', val from env: '{}'", err, log_level);
-            exit(1)
-        },
     };
     let target = if let Some(log_file) = log_file {
         let file = OpenOptions::new()
@@ -64,7 +51,7 @@ pub fn init(
         env_logger::Target::Stderr
     };
     env_logger::Builder::new()
-        .filter_level(log_level)
+        .filter_level(LevelFilter::from_str(filter.as_ref()).context("failed to parse log level")?)
         .format(formatter)
         .target(target)
         .init();
--- a/crates/youki/src/logger.rs
+++ b/crates/youki/src/logger.rs
@@ -1,9 +1,13 @@
 //! Default Youki Logger
 
 use anyhow::{bail, Context, Result};
+use log::LevelFilter;
+use std::env;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process::exit;
+use std::str::FromStr;
 
 /// If in debug mode, default level is debug to get maximum logging
 #[cfg(debug_assertions)]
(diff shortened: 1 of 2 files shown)


crates/youki/src/logger.rs
```diff
@@ -19,12 +19,13 @@ const LOG_FORMAT_JSON: &str = "json";
...
 /// Multiple parts might call this at once, but the actual initialization
 /// is done only once due to use of OnceCell
-pub fn init(log_format: Option<String>, log_file: Option<PathBuf>) -> Result<()> {
+pub fn init(log_debug_flag: bool, log_file: Option<PathBuf>, log_format: Option<String>) -> Result<()> {
     let formatter = match log_format.as_deref() {
         None | Some(LOG_FORMAT_TEXT) => text_write,
         Some(LOG_FORMAT_JSON) => json_write,
         Some(unknown) => bail!("unknown log format: {}", unknown),
     };
+    let loglevel = if log_debug_flag { "debug" } else { "warn" };
```
utam0k, line 28: Do you use `loglevel` anywhere? I think you need to use this flag, perhaps around line 41?

unknowndevQwQ: > Do you use loglevel anywhere? I think you need to use this flag, perhaps around line 41?  

I don't know what I should do after this, can you give me some tips?

Furisto: You need to pass the log level to the logger builder. It has a filter_level method that should do the trick. You should set the log level to debug if the debug flag is specified, otherwise you need to check the youki log environment variable. If both are not set, you should fall back to the default log level.

unknowndevQwQ: > You need to pass the log level to the logger builder. It has a filter_level method that should do the trick. You should set the log level to debug if the debug flag is specified, otherwise you need to check the youki log environment variable. If both are not set, you should fall back to the default log level.

image
like this? Or should I make some other changes?

crates/youki/src/main.rs
```diff
@@ -38,13 +38,15 @@ use nix::unistd::getuid;
 #[derive(Parser, Debug)]
 #[clap(version = crate_version!(), author = "youki team")]
 struct Opts {
-    /// root directory to store container state
-    #[clap(short, long)]
-    root: Option<PathBuf>,
+    #[clap(long)]
+    debug: bool,
```
yihuaf, line 42: should this be an option as well? The `--debug` flag can be true, false, or absent.

yihuaf (changes requested):
First of all, there is a `DEFAULT_LOG_LEVEL` variable you should be using for default. Second, in addition to `--debug` flag, you will have to handle `YOUKI_LOG_LEVEL` from environment variable. We use `env_logger`, and you should read the documentation here: https://docs.rs/env_logger/0.9.0/env_logger

crates/youki/src/main.rs
unknowndevQwQ: I noticed that there is no default value for --systemd-cgroup either, should I add one?

crates/youki/src/logger.rs
utam0k:
```suggestion
    let loglevel = if log_debug_flag {
        "debug"
    } else {
        DEFAULT_LOG_LEVEL
    };
```

and set a variable `loglevel` instead of `DEFAULT_LOG_LEVEL` at L41.
How about this?

crates/youki/src/main.rs
unknowndevQwQ: > should this be an option as well? The `--debug` flag can be true, false, or absent. Does the variable default to `false`?

Variable is false when flag is not used

yihuaf: If the variable defaults to false, then we are good.

crates/youki/src/logger.rs
unknowndevQwQ: hmm

crates/youki/src/logger.rs
```diff
@@ -19,12 +20,20 @@ const LOG_FORMAT_JSON: &str = "json";
...
         Some(LOG_FORMAT_JSON) => json_write,
         Some(unknown) => bail!("unknown log format: {}", unknown),
     };
+    let loglevel = if log_debug_flag {
+        "debug"
+    } else match env::var("YOUKI_LOG_LEVEL") {
+        Ok(val) => val,
+
+        // Need to add warning log
+        Err(e) => DEFAULT_LOG_LEVEL,
```
unknowndevQwQ, line 35: I seem to have a lot of things not done

unknowndevQwQ:
@utam0k Seems to be ready to go

utam0k:
@unknowndevQwQ Sorry, this pr seems to be conflicting.

unknowndevQwQ:
> @unknowndevQwQ Sorry, this pr seems to be conflicting.

fixed

crates/youki/src/main.rs
```diff
@@ -38,6 +38,11 @@ use nix::unistd::getuid;
 #[derive(Parser, Debug)]
 #[clap(version = crate_version!(), author = "youki team")]
 struct Opts {
+    // I don't know how to get the log level when the --debug flag is not set (I want to show some default values on the help page when the options are not set)
+    // Example: '--debug     change log level to debug. (default: "warn")'
+    /// change log level to debug.
```
utam0k, line 43: These comments in this section are `./youki -h` will be output.
When I thought about it, I felt that the first line was unnecessary.

utam0k: These codes may be helpful for you.
https://github.com/containers/youki/blob/ba08bb323165f4d0b98a0321fb990725bb32de69/crates/youki/src/commands/create.rs#L10-L27

crates/youki/src/logger.rs
```diff
@@ -36,12 +46,12 @@ pub fn init(log_file: Option<PathBuf>, log_format: Option<String>) -> Result<()>
...
-    )
-    .format(formatter)
-    .target(target)
-    .init();
+    env_logger::Builder::new()
+        .parse_env(env_logger::Env::new().filter("YOUKI_LOG_LEVEL"))
+        .filter_level(FromStr::from_str(log_level).unwrap())
+        .format(formatter)
+        .target(target)
+        .init();
```
utam0k, line 54: This doesn't give a clear idea of which level has the highest priority.
So it is better to decide which log level you want in lines 33-37 and just set the value in the builder and do not overwrite it.

crates/youki/src/main.rs
unknowndevQwQ: hmm

unknowndevQwQ: > These codes may be helpful for you.
> 
> https://github.com/containers/youki/blob/ba08bb323165f4d0b98a0321fb990725bb32de69/crates/youki/src/commands/create.rs#L10-L27

The default value in some options is dynamic, for example, `--root`

crates/youki/src/logger.rs
unknowndevQwQ: This thing is starting to get tricky for me

utam0k: What areas became tricky?

crates/youki/src/main.rs
utam0k: @unknowndevQwQ 
My point is that the comments in this section are not for developers, but for users. I would like you to take that into account when making your comments.

crates/youki/src/logger.rs
unknowndevQwQ: Get the value from the environment variable and convert it to the appropriate type

utam0k: > Get the value from the environment variable and convert it to the appropriate type

It didn't seem like it would be a very complex code, but what are your concerns?

utam0k: Doesn't this code overwrite the default level even if `YOUKI_LOG_LEVEL` is set?

unknowndevQwQ: > > Get the value from the environment variable and convert it to the appropriate type
> 
> It didn't seem like it would be a very complex code, but what are your concerns?

Get the value from the environment variable and convert it to &str, output a friendly error if it fails

unknowndevQwQ:
```rust
let log_level = if log_debug_flag {
        "debug"
    } else {
        match env::var("YOUKI_LOG_LEVEL") {
            Ok(val) => 
            Err(_e) => DEFAULT_LOG_LEVEL
        }
    };
```

crates/youki/src/main.rs
unknowndevQwQ: > @unknowndevQwQ My point is that the comments in this section are not for developers, but for users. I would like you to take that into account when making your comments.

Should I delete L41-L42?

crates/youki/src/logger.rs
```diff
@@ -19,12 +24,34 @@ const LOG_FORMAT_JSON: &str = "json";
...
+    } else {
+        DEFAULT_LOG_LEVEL
+    };
+    let log_level = match LevelFilter::from_str(log_level) {
+        Ok(val) => val,
+        Err(err) => {
+            bail!("error: '{}', val from env: '{}'", err, log_level);
+            exit(1)
+        },
+    };
```
Furisto, line 54: How about this?

```suggestion
    let filter: Cow<str> = if debug_flag {
        "debug".into()
    } else if let Ok(level) = std::env::var("YOUKI_LOG_LEVEL") {
        level.into()
    } else {
        DEFAULT_LOG_LEVEL.into()
    };

    env_logger::Builder::new()
        .filter_level(LevelFilter::from_str(filter.as_ref()).context("failed to parse log level")?)
        .format(formatter)
        .target(target)
        .init();
```

unknowndevQwQ: thx

crates/youki/src/main.rs
unknowndevQwQ: @utam0k

crates/youki/src/logger.rs
unknowndevQwQ: `YOUKI_LOG_LEVEL=o sudo -E ./youki/target/x86_64-unknown-linux-gnu/release/youki state test`

```
log init failed: failed to parse log level

Caused by:
    attempted to convert a string that doesn't match an existing log level
{
  "ociVersion": "v1.0.2",
  "id": "test",
  "status": "stopped",
  "pid": 114514,
  "bundle": "test",
  "annotations": {},
  "created": "2021-11-11T11:11:11.111111111Z",
  "creator": 0,
  "useSystemd": false
}
```

Is it normal to continue doing other things after a loglevel failure?

Furisto:
Yes, a faulty log level should not prevent execution of the command.

Furisto (approved):

Furisto:
lgtm

unknowndevQwQ:
> lgtm

Can we compress the redundant commits?

utam0k:
> > lgtm
> 
> Can we compress the redundant commits?

I wish We could have, but I don't have to be so obsessed.

unknowndevQwQ:
> > > lgtm
> > 
> > 
> > Can we compress the redundant commits?
> 
> I wish We could have, but I don't have to be so obsessed.

Then let nature take its course.

PR #472, opened 2021-11-13
Fix test_make_parent_mount_private

### LINKED ISSUE #471: test_make_parent_mount_private is finding corrent parent mount of when `/tmp` is mounted by tmpfs
Some systems mount a `tmpfs` filesystem at `/tmp` and some systems just treat it as a regular directory on the root filesystem. When `/tmp` is mounted, the temporary directory used by the test has the parent mount `/tmp` instead of the expected `/`. This means that currently this test is failing on any system that mounts `/tmp`. This is a bug with the test, not the function it's testing, because it's expected that the parent is actually `/tmp` in this case, and the rest of the assertion is correct.

Output:

```
thread 'rootfs::mount::tests::test_make_parent_mount_private' panicked at 'assertion failed: `(left == right)`
  left: `MountArgs { source: None, target: "/", fstype: None, flags: MS_PRIVATE, data: None }`,
 right: `MountArgs { source: None, target: "/tmp", fstype: None, flags: MS_PRIVATE, data: None }`', crates/libcontainer/src/rootfs/mount.rs:546:9
```

I think the appropriate fix would either be to expect `/tmp` or `/`, otherwise we could simply move this test outside of `/tmp` entirely, but then we may have to write to the home directory or somewhere weird on the filesystem which may or may not also be mounted depending on the specific system. Is it even important to test the `target`? If so I think the first solution might be best because we know the 2 cases we can expect, versus if we move to another directory the results can be even more inconsistent between systems.

tsturzl:
This fixes #471

The checks have just been broken out to their individual pieces rather than testing the entire comparison of the `MountArgs` so we can test for `/` or `/tmp` set as target, the reasoning for this is outlined in the issue.

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -528,22 +528,24 @@ mod tests {
         let m = Mount::new();
         assert!(m.make_parent_mount_private(tmp_dir.path()).is_ok());
 
-        let want = MountArgs {
-            source: None,
-            target: PathBuf::from("/"),
-            fstype: None,
-            flags: MsFlags::MS_PRIVATE,
-            data: None,
-        };
-        let got = m
+        let set = m
             .syscall
             .as_any()
             .downcast_ref::<TestHelperSyscall>()
             .unwrap()
             .get_mount_args();
 
-        assert_eq!(got.len(), 1);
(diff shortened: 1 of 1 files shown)


tommady (approved):
LGTM 🙇🏻

PR #473, opened 2021-11-13
cgroups v1 memory integration test

tsturzl:
Essentially exactly like runtime-tools. This should however address this unresolved issue with runtime-tools: https://github.com/opencontainers/runtime-tools/issues/721

On systems with swap disabled in the kernel this test will likely be skipped entirely. Thing might not be the ideal behavior in that case, but I didn't want to change behavior from runtime-tools too much.

### DIFF
--- a/crates/integration_test/src/tests/cgroups/memory.rs
+++ b/crates/integration_test/src/tests/cgroups/memory.rs
@@ -1,0 +1,81 @@
+use std::path::Path;
+
+use anyhow::{Context, Result};
+use oci_spec::runtime::{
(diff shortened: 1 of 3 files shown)


crates/integration_test/src/tests/cgroups/memory.rs
```diff
@@ -0,0 +1,81 @@
...
+            return test_result;
+        }
+    }
+
+    TestResult::Passed
+}
+
+fn can_run() -> bool {
+    Path::new(CGROUP_MEMORY_LIMIT).exists() && Path::new(CGROUP_MEMORY_SWAPPINESS).exists()
+}
```
utam0k, line 68: 👍

utam0k (approved):

PR #474, opened 2021-11-13
style: adjusting the position of parameters and flag declarations

unknowndevQwQ:
Split from #465
Because this pr does not have much to do with #465

### DIFF
--- a/crates/youki/src/logger.rs
+++ b/crates/youki/src/logger.rs
@@ -19,7 +19,7 @@ const LOG_FORMAT_JSON: &str = "json";
 /// Initialize the logger, must be called before accessing the logger
 /// Multiple parts might call this at once, but the actual initialization
 /// is done only once due to use of OnceCell
-pub fn init(log_format: Option<String>, log_file: Option<PathBuf>) -> Result<()> {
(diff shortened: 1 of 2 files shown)


utam0k:
@unknowndevQwQ Why we need this change?

unknowndevQwQ:
> @unknowndevQwQ Why we need this change?

It will become a little more organized

utam0k:
> > @unknowndevQwQ Why we need this change?
> 
> It will become a little more organized
> 
> When trying to add the --debug flag, I didn't even realize where the --root flag was declared

I don't understand what changes of this PR improve. Would you please explain it to me?

unknowndevQwQ:
> > > @unknowndevQwQ Why we need this change?
> > 
> > 
> > It will become a little more organized
> > When trying to add the --debug flag, I didn't even realize where the --root flag was declared
> 
> I don't understand what changes of this PR improve. Would you please explain it to me?

The ordering of the parameter names in a-z order facilitates a quicker and more accurate search for the relevant content of a parameter when there are quite a few of them.

utam0k (approved):

Issue #464, opened 2021-11-10
establish a benchmarking method


utam0k:
## Why
For youki, performance is one of the most important concerns. Until now, we have relied on our individual senses for this, but we should prepare the official and easy way to measure it more accurately.

## Goal
Provide documentation and scripts to measure performance.
Ideally, it should be incorporated into the CI and measured by PR every time.

utam0k:
a example https://github.com/containers/youki/pull/447#issuecomment-958991982

utam0k:
I am looking for people who are willing to take on this challenge.

tsturzl:
I had actually been fooling around with benchmarking Youki. I had previously used the Criterion crate to benchmark Rust projects for work and have really enjoyed it. I'm also aware of flamegraph-rs which can give performance profiles. I'm not sure how well these things might work since Youki works in a different way from many other applications, such as forking processes, but currently I have a mostly working Criterion setup on [this branch](https://github.com/tsturzl/youki/tree/benchmarks).

A couple of problems I face are having to redo some of the setup logic, some general weirdness around relative paths, it's really a pain to run anything but rootless mode since I almost never have a system wide install of the rust toolchain, and I currently cannot have the benchmarks cleanup their containers since the current delete routine exits the process which means the test process exits whenever I try to delete a container.

This is just my adventure in trying to leverage Rust tooling to get some benchmarks, but perhaps there are other tools that might make more sense which can just execute youki as a separate process. Though I think benchmarks are just the most surface level thing we can do, especially in such a coarse manner. It would be really nice to get profiling so we can understand how things are performing and where the issues might be.

I'd be interested in diving in further here.

tommady:
@tsturzl I agree with your point, so what I wanna raise here is a part of the benchmark progress.

Since I have some idea with the libcontainer lib wanna do some modifications, but I am afraid to effect the performance, so I want to add a PR to do the relative benchmark by the GitHub action with a label triggering way to get the benchmark criteria of each PR'd build compares to main branch's build.

And since the GitHub action running independently and remotely won't screwed up the local env I think this one can be added up for the first step.

The comparison will be quite simple, in my imagination it just using the tool @utam0k used then running few commands like echo a string, ls a path by the youki with the same conditions, that's all.

If this is not an acceptable way please correct and guide me 🙇‍♂️

utam0k:
> @tsturzl I agree with your point, so what I wanna raise here is a part of the benchmark progress.
> 
[...]

PR #478, opened 2021-11-15
adding benchmark execution time github action

tommady:
part of #464 
this PR is the first part of that issue.

how to use it:
1. label a PR with benchmark-exec-time
2. then every push in this PR will run a benchmark action

image

### DIFF
--- a/.github/workflows/benchmark_execution_time.yml
+++ b/.github/workflows/benchmark_execution_time.yml
@@ -1,0 +1,138 @@
+name: Benchmark execution time comparison with the main branch
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  building-pr-branch:
+    if: contains( github.event.pull_request.labels.*.name, 'benchmark-exec-time' )
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout to PR branch
+      uses: actions/checkout@v2
+
+    - name: Setup Linux env
+      run: |
+        sudo apt -y update
+        sudo apt install libsystemd-dev librust-libdbus-sys-dev libseccomp-dev
+
+    - name: Setting rust toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        override: true
+        profile: minimal
+
+    - name: Building PR branch
+      run: ./build.sh --release
+
+    - name: Uploading PR build to artifact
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark_execution_time.yml
```diff
@@ -0,0 +1,139 @@
...
+
+    - name: Adding info into result
+      run: |
+        sed -i '1i > commit (${{ github.sha }})\n' bench_exec_result.md
+
+    - name: Writting report to PR comment
+      uses: marocchino/sticky-pull-request-comment@v2 
+      with:
+        append: true
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
utam0k, line 138: Maybe this is not enough permissions in the organization, can you change it to print in actions?

tommady: OK,
I'll figure out a way to print the resulting beauty.

tommady:
Hi, so I change to print in the terminal, the result will like below in the [action](https://github.com/containers/youki/runs/4221231295?check_suite_focus=true)

image

utam0k (approved):
This is a great first step.

.github/workflows/benchmark_execution_time.yml
utam0k: @saschagrunert The youki team is trying to measure the results of a very simple benchmark (just for reference) using github actions. I'd like to comment on the results to the PR, but I don't seem to have permission. Could you ask the organizer to check it?

saschagrunert: @rhatdan can we make @utam0k admin of this repository to add secrets for using GitHub actions?

utam0k: Thanks for your help.
I am already the owner of this repository.
It looks like there is no write on `GITHUB_TOKEN` permissions. Isn't `GITHUB_TOKEN` set to read-only on the organizer side? I could not check the permissions of the organizer with my permissions. In the repository, `GITHUB_TOKEN` is set to write, but it may be possible to override this permission in the organizer.
https://github.com/containers/youki/pull/478/files#diff-28fe3108ee857a8865d400d072ffeab3d06d0c14073313bd0b39cf4fa41a57fdR134-R142

PR #482, opened 2021-11-20
make complex loglevel decision easy to understand.

utam0k:


utam0k:
@tommady This doesn't relate to this PR but the benchmark ci is not responding with `benchmark-exec-time`. Do you know what it is?

tommady:
> @tommady This doesn't relate to this PR but the benchmark ci is not responding with `benchmark-exec-time`. Do you know what it is?

Sorry because, that label needs to be added from the beginning of a PR created otherwise the action will passing no label arg into it which is the one of the benchmark exec condition.

I just rerun it seems not work because the action only runs the ran one again...

So the solution might be 
1. Please commit again 
2. We can add a manual trigger condition 

Sorry to cause you inconvenience.

tommady:
!github easy-benchmark

tommady:
https://github.com/containers/youki/actions/runs/1485921927

cooool @utam0k thank you to improve this 🙇🏻

crates/youki/src/logger.rs
```diff
@@ -93,3 +100,31 @@ where
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::env;
+
+    #[test]
+    fn test_detect_log_level_is_debug() {
+        env::set_var(LOG_LEVEL_ENV_NAME, "error");
```
yihuaf, line 111: We may have to serialized this. cargo test runs in a single process and env variables are shared across the same process and test cases.

yihuaf:
Otherwise, LGTM

crates/youki/src/logger.rs
utam0k: Sounds good. I will try to fix it.

utam0k:
@yihuaf @Furisto PTAL

crates/youki/src/logger.rs
```diff
@@ -51,14 +47,28 @@ pub fn init(
...
 
     Ok(())
 }
 
+fn detect_log_level(is_debug: bool) -> Result<LevelFilter> {
+    let filter: Cow<str> = if is_debug {
+        dbg!(is_debug);
+        "debug".into()
+    } else if let Ok(level) = std::env::var(LOG_LEVEL_ENV_NAME) {
+        println!("from env: {:?}", level);
```
yihuaf, line 63: Is this print for debug?

crates/youki/src/logger.rs
```diff
@@ -93,3 +103,55 @@ where
...
+    #[test]
+    fn test_detect_log_level_is_debug() {
+        let _guard = LogLevelGuard::new("error").unwrap();
+        assert_eq!(detect_log_level(true).unwrap(), LevelFilter::Debug)
+    }
+
+    #[test]
+    #[serial]
+    fn test_detect_log_level_default() {
+        env::remove_var(LOG_LEVEL_ENV_NAME);
```
yihuaf, line 143: Do you also want to take advantage of the LogGuard here?
1. remember the current value
2. remove the current value
3. afterwards, restore the previous value

Make `new` take in Option<log_level>.

yihuaf:
Not sure why integration test is failing.

crates/youki/src/logger.rs
utam0k: 😆

crates/youki/src/logger.rs
utam0k: It is important that the environment variable is not set in this test, but it looks like we can set the original environment variable by using the guard after this test. Let's do it.

yihuaf (approved):

PR #488, opened 2021-11-23
Create a subdirectory under XDG_RUNTIME_DIR

### LINKED ISSUE #487: determine_root_path() isn't quite right for XDG_RUNTIME_DIR and /run/user/{} cases
For rootless operation, `determine_root_path()` will select from one of four options for the root state directory:

1. `$XDG_RUNTIME_DIR`
2. `/run/user/<uid>`
3. `$HOME/.youki/run`
4. `/tmp/youki/<uid>`

Options (1) and (2) will often amount to the same thing, since `/run/user/<uid>` is the default value for `$XDG_RUNTIME_DIR` on most systems.  But, in both these cases we don't include "youki", or a similar tag in the path at all, so state directories for each container will be placed directly underneath the runtime directory.

That doesn't match normal conventions, where each program puts its own subdir under the runtime dir.  More importantly, it means that `youki list` which works by iterating through the subdirectories of the root dir will list unrelated things from other tools as though they were containers.

dgibson:
When using a root state directory from $XDG_RUNTIME_DIR, or from it's
default value of /run/user/<uid>, we don't add a tag specific to youki
to the path.  That means the directories for individual containers
will be placed directly in the general use runtime dir.

That's against normal conventions, and could mean that "youki list"
will see files or directories from other software as if they were
youki managed containers.  Therefore, add "youki" to the base runtime
path from XDG.

fixes #487

Signed-off-by: David Gibson <david@gibson.dropbear.id.au>

### DIFF
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -5,7 +5,7 @@ mod commands;
 mod logger;
 
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use anyhow::bail;
 use anyhow::Context;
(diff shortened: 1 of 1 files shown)


crates/youki/src/main.rs
```diff
@@ -145,15 +145,18 @@ fn determine_root_path(root_path: Option<PathBuf>) -> Result<PathBuf> {
     }
 
     // see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+    let uid = getuid().as_raw();
     if let Ok(path) = std::env::var("XDG_RUNTIME_DIR") {
-        return Ok(PathBuf::from(path));
+        let path = Path::new(&path).join("youki");
+        if create_dir_all_with_mode(&path, uid, Mode::S_IRWXU).is_ok() {
+            return Ok(path);
+        }
```
utam0k, line 153: If it fails, it must return with an error or the process will continue, which is what you are aiming for, right?

dgibson: Well, not immediately.  If this create fails we continue onto the remaining options (`$HOME/.youki/run` then `/tmp/youki/<uid>`).  AFAICT that's what happened before, and I'm not intending to change that.

Furisto: That is correct. If we cannot get the preferred location (which is XDG_RUNTIME_DIR), we try a bunch of other locations that may be suitable before we give up.

Furisto:
Looks good to me.

Furisto (approved):

Furisto:
@dgibson Thanks for your contribution!

PR #497, opened 2021-11-24
Use /tmp/youki-<uid> rather than /tmp/youki/<uid> in determine_root_path

### LINKED ISSUE #496: Final fallback for determine_root_path() will fail with multiple users
`determine_root_path` goes through various options to find a state storage location, the last of which is `/tmp/youki/<uid>`.

Unfortunately, if two different users are using youki and this fallback option is used, then the second user will fail on this option as well.

That's because not just the final directory, but newly created directories leading up to it will be created with the permissions specified.  So, when the first user uses youki `/tmp/youki` will be created owned by them and with write permission only by them.  That will prevent the next user from creating `/tmp/youki/NNN`.

dgibson:
determine_root_path goes through various options to find a state storage
location, the last of which is /tmp/youki/<uid>.

If a user (say, UID 1000) uses youki, and this final option is selected,
/tmp/youki will be created as well as /tmp/youki/1000.  Both will be
created owned by UID 1000 and with write permissions only for that user.

Them, if another user (say, UID 1001) attempts to use youki and the same
final option is selected, it will fail, because it cannot create
/tmp/youki/1001 under the /tmp/youki owned by UID 1000.

There's really no way to safely create a multi-user shared subdirectory in
/tmp, so instead we should create our per-user directory directly under
/tmp.  We do this by calling it /tmp/youki-<uid> instead.

fixes #496

Signed-off-by: David Gibson <david@gibson.dropbear.id.au>

### DIFF
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -168,7 +168,7 @@ fn determine_root_path(root_path: Option<PathBuf>) -> Result<PathBuf> {
         }
     }
 
-    let tmp_dir = PathBuf::from(format!("/tmp/youki/{}", uid));
+    let tmp_dir = PathBuf::from(format!("/tmp/youki-{}", uid));
     if create_dir_all_with_mode(&tmp_dir, uid, Mode::S_IRWXU).is_ok() {
         return Ok(tmp_dir);
     }


utam0k (approved):
@dgibson 
Looks good.
Thanks for the very clear explanation. Maybe it's time to put some testing into the process around here.

dgibson:
> @dgibson Looks good. Thanks for the very clear explanation. Maybe it's time to put some testing into the process around here.

Yes, I was thinking it didn't look like the rootless paths were exercised much in testing, which isn't ideal.  AFAICT that will be a moderately significant amount of work, though, since the runtime-tools testsuite doesn't make container specs which are suitable for rootless.

PR #506, opened 2021-11-29
[WIP] increase apparmor test coverage

tommady:
#279 
the AppArmor part

### DIFF
--- a/crates/libcontainer/src/apparmor.rs
+++ b/crates/libcontainer/src/apparmor.rs
@@ -1,8 +1,5 @@
 use anyhow::{Context, Result};
-use std::{
-    fs::{self},
-    path::Path,
-};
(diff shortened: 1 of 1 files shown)


tommady:
Hi, all,
I have a silly question, about the AppArmor.

I noticed that the paths of this mod were using
- /proc/self/attr/apparmor/exec
- /proc/self/attr/exec

their permissions are all
`-rw-rw-rw- 1 root root 0`

from my understanding, it means self, group and  `others`  can read and write as well right?

so I simply write a test to call the apply_profile function which will lead us to write all the profiles into those files,
but still getting `permission denied os error (13)`

I am very curious about why, please could someone teach me? 
thanks a lot.

utam0k:
@Furisto Hi! Do you know what is going on?

Issue #508, opened 2021-11-30
Command line parsing front end would be useful in its own crate


dgibson:
Some colleagues and I have a use case for a rather specialized OCI runtime tool.  The idea is that it wouldn't implement the container backend itself, but rather would just be a frontend which then calls into another implementation to do the actual container operations.  We have a couple of uses in mind for this:

- Instrumenting / tracing an existing runtime for debugging purposes
- Translating the OCI command line interface into semantically similar, but not identical interfaces (e.g. the shimv2 gRPC protocol used by Kata containers v2)

We were planning to make this in Rust, and youki already has a very clean command line parsing implementation that is not deeply tied into the back end implementation.  Would it be ok to split the front end CLI parsing into a separate crate, which we could re-use for our purposes?

/cc @c3d

Furisto:
Seems reasonable. We have done similar things with other parts of our code in the past, as one of our goals is to make it easier to write container implementations in Rust, so there is no conflict there. The required changes seem to be relatively low impact.

dgibson:
> Seems reasonable. We have done similar things with other parts of our code in the past, as one of our goals is to make it easier to write container implementations in Rust, so there is no conflict there. The required changes seem to be relatively low impact.

Terrific!  I've made a suggested implementation of this change in PR #509

utam0k:
@dgibson 
Furisto is right and I agree with him. Are you talking about kata-agent?
https://github.com/kata-containers/kata-containers/pull/2785
I had heard a bit of this story before. Let's move forward cooperatively. Thank you very much for your contribution to youki :100:

dgibson:
> @dgibson Furisto is right and I agree with him. Are you talking about kata-agent? [kata-containers/kata-containers#2785](https://github.com/kata-containers/kata-containers/pull/2785) I had heard a bit of this story before. Let's move forward cooperatively. Thank you very much for your contribution to youki 100

Actually, no.  I'm aware of that PR, and re-using the command line parsing there as well could certainly be useful, but it's not something I'm looking at myself.  Rather I'm looking at adding an OCI command line to the full Kata stack, with runtime, VM and agent.  Currently Kata (2.x) is controlled using the "shimv2" gRPC protocol which has similar operations to the OCI runtime spec, but with a few differences.  That means that Kata 2.x can't be used with podman, for example, only engines which speak the shimv2 protocol (containerd and CRI-O both have such support).

[...]

PR #509, opened 2021-11-30
Split CLI parsing front end into a separate crate

dgibson:
As discussed in #508, we have a specialized use case where we'd like to re-use just the CLI parsing parts of youki in another project.

Doing such a split in youki is reasonably straightforward, as this draft PR shows.

### DIFF
--- a/crates/liboci-cli/README.md
+++ b/crates/liboci-cli/README.md
@@ -14,13 +14,13 @@ Interface][https://github.com/opencontainers/runtime-tools/blob/master/docs/comm
 | kill       | ✅         | ✅                | ✅   | ✅   | ✅    |
 | delete     | ✅         | ✅                | ✅   | ✅   | ✅    |
 | checkpoint |            |                   | ✅   | ✅   |       |
-| events     |            |                   | ✅   |      | ✅    |
-| exec       |            |                   | ✅   | ✅   | ✅    |
-| list       |            |                   | ✅   | ✅   | ✅    |
-| pause      |            |                   | ✅   | ✅   | ✅    |
-| ps         |            |                   | ✅   | ✅   | ✅    |
+| events     | ✅         |                   | ✅   |      | ✅    |
+| exec       | ✅         |                   | ✅   | ✅   | ✅    |
+| list       | ✅         |                   | ✅   | ✅   | ✅    |
+| pause      | ✅         |                   | ✅   | ✅   | ✅    |
+| ps         | ✅         |                   | ✅   | ✅   | ✅    |
 | restore    |            |                   | ✅   | ✅   |       |
-| resume     |            |                   | ✅   | ✅   | ✅    |
-| run        |            |                   | ✅   | ✅   | ✅    |
-| spec       |            |                   | ✅   | ✅   | ✅    |
+| resume     | ✅         |                   | ✅   | ✅   | ✅    |
+| run        | ✅         |                   | ✅   | ✅   | ✅    |
+| spec       | ✅         |                   | ✅   | ✅   | ✅    |
 | update     |            |                   | ✅   | ✅   |       |
--- a/crates/liboci-cli/README.md
+++ b/crates/liboci-cli/README.md
@@ -1,0 +1,26 @@
+# liboci-cli
+
+This is a crate to parse command line arguments for OCI container
+runtimes as specified in the [OCI Runtime Command Line
+Interface][https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md).
+
+## Implemented subcommands
+
+| Command    | liboci-cli | CLI Specification | runc | crun | youki |
+| :--------: | :--------: | :---------------: | :--: | :--: | :---: |
+| create     | ✅         | ✅                | ✅   | ✅   | ✅    |
+| start      | ✅         | ✅                | ✅   | ✅   | ✅    |
+| state      | ✅         | ✅                | ✅   | ✅   | ✅    |
+| kill       | ✅         | ✅                | ✅   | ✅   | ✅    |
+| delete     | ✅         | ✅                | ✅   | ✅   | ✅    |
+| checkpoint |            |                   | ✅   | ✅   |       |
+| events     |            |                   | ✅   |      | ✅    |
+| exec       |            |                   | ✅   | ✅   | ✅    |
+| list       |            |                   | ✅   | ✅   | ✅    |
+| pause      |            |                   | ✅   | ✅   | ✅    |
+| ps         |            |                   | ✅   | ✅   | ✅    |
+| restore    |            |                   | ✅   | ✅   |       |
+| resume     |            |                   | ✅   | ✅   | ✅    |
+| run        |            |                   | ✅   | ✅   | ✅    |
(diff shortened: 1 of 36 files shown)


crates/youki/src/main.rs
```diff
@@ -117,24 +76,31 @@ fn main() -> Result<()> {
...
+            CommonCmd::Exec(exec) => commands::exec::exec(exec, root_path),
+            CommonCmd::List(list) => commands::list::list(list, root_path),
+            CommonCmd::Pause(pause) => commands::pause::pause(pause, root_path),
+            CommonCmd::Ps(ps) => commands::ps::ps(ps, root_path),
+            CommonCmd::Resume(resume) => commands::resume::resume(resume, root_path),
+            CommonCmd::Run(run) => commands::run::run(run, root_path, systemd_cgroup),
+            CommonCmd::Spec(spec) => commands::spec_json::spec(spec),
+        },
+
+        SubCommand::Info(info) => commands::info::info(info),
```
utam0k, line 103: I think it would make more sense to create a `SubCommand::Ext` and have the info command belong to it, in the same treatment as `Standard` and `Common`, but what do you think?

crates/oci-runtime-clap/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Implemented subcommands
+
+| Command    | oci-runtime-clap | CLI Specification | runc | crun | youki |
+| :--------: | :--------------: | :---------------: | :--: | :--: | :---: |
+| create     | ✅               | ✅                | ✅   | ✅   | ✅    |
+| start      | ✅               | ✅                | ✅   | ✅   | ✅    |
+| state      | ✅               | ✅                | ✅   | ✅   | ✅    |
+| kill       | ✅               | ✅                | ✅   | ✅   | ✅    |
+| delete     | ✅               | ✅                | ✅   | ✅   | ✅    |
+| :--------: | :--------------: | :---------------: | :--: | :--: | :---: |
```
utam0k, line 16: If you want to separate Common and Standard, why don't you write that down and separate them in a table?

utam0k:
I understand the general policy. It looks good, but I made a few comments.

crates/oci-runtime-clap/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "oci-runtime-clap"
```
Furisto, line 2: So far we have used the lib prefix for our crates. I also would drop 'clap' in the name. I suggest libruntime-cli. @utam0k Thoughts?

crates/oci-runtime-clap/README.md
```diff
@@ -0,0 +1,27 @@
...
+| events     | ✅               |                   | ✅   |      | ✅    |
+| exec       | ✅               |                   | ✅   | ✅   | ✅    |
+| list       | ✅               |                   | ✅   | ✅   | ✅    |
+| pause      | ✅               |                   | ✅   | ✅   | ✅    |
+| ps         | ✅               |                   | ✅   | ✅   | ✅    |
+| restore    |                  |                   | ✅   | ✅   |       |
+| resume     | ✅               |                   | ✅   | ✅   | ✅    |
+| run        | ✅               |                   | ✅   | ✅   | ✅    |
+| spec       | ✅               |                   | ✅   | ✅   | ✅    |
+| update     |                  |                   | ✅   | ✅   |       |
```
Furisto, line 27: 'info' is missing 😉

crates/oci-runtime-clap/Cargo.toml
utam0k: I agree.  I think so too. There may be cases like this one, where someone wants to create a cli but only using the youki backend, but let's think about that when the time comes.

crates/oci-runtime-clap/README.md
utam0k: Since `info` is an extended command, perhaps it's not written on purpose? Even if that's the case, I'd like to write somewhere that there is an info command.

crates/youki/src/main.rs
dgibson: I don't want `Info` to be in the command line crate, because it is (AFAICT) a youki specific extension, whereas the CLI crate is supposed to be only for handling things common to all (or at least most) runtimes.  I could create an `ExtCommand` enum with the `youki` crate if you like, although I don't see a lot of point to it (we can always add it, if youki gets more extension commands in future).

crates/oci-runtime-clap/Cargo.toml
dgibson: I like the idea of including "oci" in the title: as a component of the youki tree alone, it's redundant since the whole project is about OCI.  But I'm hoping this can become a crate published on its own, in which case putting OCI in the name makes its function clearer.  I'm not particularly attached to the current name though; names are hard.

How about `liboci-cli`?

crates/oci-runtime-clap/README.md
dgibson: Sorry, I'm not sure what you mean.

crates/oci-runtime-clap/README.md
dgibson: `info` is not implement in *this* crate, so it doesn't belong in this list.  `info` is already described in several places in the `youki` crate - most notably in the output of `youki --help`.

crates/oci-runtime-clap/README.md
utam0k: This doesn't seem to separate the tables. Is this the convention?
https://github.com/containers/youki/blob/deec552874e3d9b2ce2d2b3cb124853550cab347/crates/liboci-cli/README.md

crates/oci-runtime-clap/Cargo.toml
utam0k: Both are great names! I think it should include `oci`.  So, let's go with `liboci-cli`

crates/oci-runtime-clap/README.md
Furisto: Ah got it

Furisto (approved):

Furisto:
lgtm

crates/oci-runtime-clap/Cargo.toml
dgibson: Done!

crates/oci-runtime-clap/README.md
dgibson: Oh, right, I see what you mean.  Clearly the `:---:` line doesn't mean what I thought it did in github markdown.  Should be fixed in the latest revision.

crates/liboci-cli/README.md
```diff
@@ -0,0 +1,26 @@
+# liboci-cli
+
+This is a crate to parse command line arguments for OCI container
+runtimes as specified in the [OCI Runtime Command Line
+Interface][https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md).
```
utam0k, line 5: Is this a mistake?
```suggestion
runtimes as specified in the (OCI Runtime Command Line
Interface)[https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md).
```

utam0k (approved):
LGTM with nits

PR #511, opened 2021-12-01
Fix log files and remove env_logger

### LINKED ISSUE #475: Flag `--log` is broken?
I found that the current youki can't write logs to a file.

```
$ sudo ./youki --log /tmp/youki.log create -b ~/workspace/tmp/mycontainer mycontainer

# file is created, but empty
$ ls -l /tmp/youki.log
-rw-r--r-- 1 root root 0 Nov 14 12:45 /tmp/youki.log
```

youki uses env_logger, but `env_logger::Target::Pipe` is broken according to https://github.com/env-logger-rs/env_logger/issues/208.

yihuaf:
Fix #475 

As mentioned in #475, the log file is broken due to env_logger breaks. Originally we were hoping using env_logger can save us from maintaining our own logger, but it seems that env_logger is not a good fit. While there is an issue filed with env_logger to fix file logging target, it is unlikely to be fixed in a reasonable time frame.

For logger, we require:
1. Log to file
2. Ability to set formatter for JSON
3. Not complex to configure and lightweight since our use case is simple and we would like to maintain performance.

For now, let's fall back to use our own logger implementation. It is simple enough and fits our use case. I am surprised that there is no good out of box logger for Rust similar to `logrus` for Golang. Most of the logging libraries in Rust either lacks features or is not well maintained. https://github.com/tokio-rs/tracing seems to be a promising substitute for `log` in rust, but it requires more work.

### DIFF
--- a/crates/youki/src/logger.rs
+++ b/crates/youki/src/logger.rs
@@ -34,9 +34,7 @@ pub fn init(
     log_file: Option<PathBuf>,
     log_format: Option<String>,
 ) -> Result<()> {
-    let level = detect_log_level(log_debug_flag)
-        .context("failed to parse log level")?
-        .to_level();
+    let level = detect_log_level(log_debug_flag).context("failed to parse log level")?;
     let format = detect_log_format(log_format).context("failed to detect log format")?;
     let _ = LOG_FILE.get_or_init(|| -> Option<File> {
         log_file.map(|path| {
@@ -49,8 +47,10 @@ pub fn init(
         })
     });
 
-    let logger = YoukiLogger::new(level, format);
-    log::set_boxed_logger(Box::new(logger))?;
+    let logger = YoukiLogger::new(level.to_level(), format);
+    log::set_boxed_logger(Box::new(logger))
+        .map(|()| log::set_max_level(level))
+        .expect("set logger failed");
 
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


crates/youki/src/logger.rs
```diff
@@ -66,47 +74,94 @@ fn detect_log_level(is_debug: bool) -> Result<LevelFilter> {
...
-    write!(f, "{{")?;
-    write!(f, "\"level\":\"{}\",", record.level())?;
-    write!(f, "\"time\":\"{}\"", chrono::Local::now().to_rfc3339(),)?;
-    write!(f, ",\"msg\":")?;
-    // Use serde_json here so we don't have to worry about escaping special characters in the string.
-    serde_json::to_writer(f.by_ref(), &record.args().to_string())?;
-    writeln!(f, "}}")?;
+struct YoukiLogger {
+    /// Indicates level up to which logs are to be printed
+    level: Option<log::Level>,
```
utam0k, line 79: What is the case of `None`?

Furisto: None means logging is disabled. None is returned by to_level if a logging level of 'Off' is specified.

Furisto (approved):

Furisto:
Looks good!

crates/youki/src/logger.rs
utam0k: I forgot about it! You're right.

PR #513, opened 2021-12-02
Improvements to cgroup support

Furisto:
This contains multiple smaller improvements that did not really warrant their own PR.
- When a transient unit is created by systemd we now ensure that all controllers that have been delegated are inherited by the subtree
- Pause/Resume and stats are now enabled for systemd managed cgroups through the v2 cgroup manager
- The libcgroups crate now provides feature flags for v1, v2 and systemd managed cgroups
- Fixed some commands that failed due to trying to read the runtime config.json instead of the youki config.json

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -179,15 +179,13 @@ pub fn create_cgroup_manager<P: Into<PathBuf>>(
     let cgroup_path = cgroup_path.into();
 
     match cgroup_setup {
-        CgroupSetup::Legacy | CgroupSetup::Hybrid => {
-            return create_v1_cgroup_manager(cgroup_path);
-        },
+        CgroupSetup::Legacy | CgroupSetup::Hybrid => create_v1_cgroup_manager(cgroup_path),
         CgroupSetup::Unified => {
             if systemd_cgroup {
-               return create_systemd_cgroup_manager(cgroup_path, container_name);
+                return create_systemd_cgroup_manager(cgroup_path, container_name);
             }
 
-            return create_v2_cgroup_manager(cgroup_path);
+            create_v2_cgroup_manager(cgroup_path)
         }
     }
 }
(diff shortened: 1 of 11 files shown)


crates/libcgroups/src/common.rs
```diff
@@ -173,40 +176,78 @@ pub fn create_cgroup_manager<P: Into<PathBuf>>(
...
 
+#[cfg(feature = "v1")]
+fn create_v1_cgroup_manager(cgroup_path: PathBuf) -> Result<Box<dyn CgroupManager>> {
+    log::info!("cgroup manager V1 will be used");
+    Ok(Box::new(v1::manager::Manager::new(cgroup_path)?))
+}
+
+#[cfg(not(feature = "v1"))]
+fn create_v1_cgroup_manager(_cgroup_path: PathBuf) -> Result<Box<dyn CgroupManager>> {
+    bail!("cgroup v1 feature is not activated");
```
tsturzl, line 201: Maybe for these features we should add some context around the fact that the build excluded this feature? I know we didn't do this in the past, but maybe it's worth doing moving forward.

tsturzl: I really appreciate this being broken out and put behind feature flags. I think this is much cleaner.

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -363,11 +377,11 @@ impl CgroupManager for Manager {
     }
 
     fn freeze(&self, state: FreezerState) -> Result<()> {
-        todo!();
+        self.fs_manager.freeze(state)
```
tsturzl, line 380: Have you tested this? I'm assuming without a previous implementation there are no test cases for this.

tsturzl (changes requested):
Looks good. Minor nit picks, and if you disagree feel free to resolve.

crates/libcgroups/src/systemd/manager.rs
Furisto: I am reusing the cgroup v2 manager here (the systemd dbus interface does not have support for the freezer), so there are tests for it.

tsturzl (approved):
This looks great

PR #514, opened 2021-12-03
add the integration test of containerd with youki.

utam0k:
But the integration tests of containerd can't pass.
https://github.com/containers/youki/issues/331


utam0k:
I'll make it a little smaller

PR #515, opened 2021-12-04
Add shell completion

creepinson:
## Changes

I added shell completion for zsh, bash or any other shell that clap_generate supports.

This is how I used it in my setup with zsh:

```bash
youki completion -g zsh > ~/.zfunc/youki.zsh
```

### DIFF
--- a/crates/youki/src/commands/completion.rs
+++ b/crates/youki/src/commands/completion.rs
@@ -1,0 +1,15 @@
+use clap::{App, Parser};
+use clap_generate::{generate, Generator, Shell};
+use std::io;
+
+#[derive(Debug, Parser)]
+pub struct CompletionParser {
+    #[clap(long = "generator", short = 'g', arg_enum)]
+    pub generator: Shell,
+}
+
+pub fn print_completions<G: Generator>(gen: G, app: &mut App) -> Result<(), anyhow::Error> {
+    generate(gen, app, app.get_name().to_string(), &mut io::stdout());
+
+    Ok(())
+}
--- a/crates/youki/src/commands/completion.rs
+++ b/crates/youki/src/commands/completion.rs
@@ -1,15 +1,22 @@
+use anyhow::Result;
 use clap::{App, Parser};
-use clap_generate::{generate, Generator, Shell};
+use clap_generate::{generate, Shell};
 use std::io;
 
 #[derive(Debug, Parser)]
-pub struct CompletionParser {
-    #[clap(long = "generator", short = 'g', arg_enum)]
-    pub generator: Shell,
+/// Generate scripts for shell completion
+pub struct Completion {
+    #[clap(long = "shell", short = 's', arg_enum)]
+    pub shell: Shell,
 }
 
-pub fn print_completions<G: Generator>(gen: G, app: &mut App) -> Result<(), anyhow::Error> {
-    generate(gen, app, app.get_name().to_string(), &mut io::stdout());
+pub fn completion(args: Completion, app: &mut App) -> Result<()> {
+    generate(
+        args.shell,
+        app,
+        app.get_name().to_string(),
+        &mut io::stdout(),
+    );
 
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


utam0k:
@creepinson 
Hi! Thanks for your suggestion. But you may  have forgotten running format.

creepinson:
You're right, I had used vscode's rust-analyzer formatter assuming that would work fine.

crates/youki/src/commands/completion.rs
```diff
@@ -0,0 +1,15 @@
+use clap::{App, Parser};
+use clap_generate::{generate, Generator, Shell};
+use std::io;
+
+#[derive(Debug, Parser)]
```
Furisto, line 5: youki -h currently does not have a description for this command

```suggestion
Generate scripts for shell completion
#[derive(Debug, Parser)]
```

crates/youki/src/commands/completion.rs
```diff
@@ -0,0 +1,15 @@
...
+pub struct CompletionParser {
+    #[clap(long = "generator", short = 'g', arg_enum)]
+    pub generator: Shell,
+}
+
+pub fn print_completions<G: Generator>(gen: G, app: &mut App) -> Result<(), anyhow::Error> {
```
utam0k, line 11: How about using `anyhow::Result`?

crates/youki/src/main.rs
```diff
@@ -100,6 +102,9 @@ fn main() -> Result<()> {
         },
 
         SubCommand::Info(info) => commands::info::info(info),
+        SubCommand::Completion(parser) => {
+            commands::completion::print_completions(parser.generator, &mut app)
+        }
```
Furisto, line 107:
```suggestion
        SubCommand::Completion(completion) => {
            commands::completion::completion(completion, &mut app)
        }
```

crates/youki/src/commands/completion.rs
```diff
@@ -0,0 +1,15 @@
+use clap::{App, Parser};
+use clap_generate::{generate, Generator, Shell};
+use std::io;
+
+#[derive(Debug, Parser)]
+pub struct CompletionParser {
+    #[clap(long = "generator", short = 'g', arg_enum)]
+    pub generator: Shell,
+}
```
Furisto, line 9:
```suggestion
pub struct Completion {
    #[clap(long = "shell", short = 's', arg_enum)]
    pub shell: Shell,
}
```

crates/youki/src/main.rs
```diff
@@ -46,6 +46,7 @@ enum SubCommand {
 
     // Youki specific extensions
     Info(info::Info),
+    Completion(commands::completion::CompletionParser),
```
Furisto, line 49:
```suggestion
    Completion(completion::Completion),
```

crates/youki/src/commands/completion.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub fn print_completions<G: Generator>(gen: G, app: &mut App) -> Result<(), anyhow::Error> {
+    generate(gen, app, app.get_name().to_string(), &mut io::stdout());
+
+    Ok(())
+}
```
Furisto, line 15:
```suggestion
pub fn print_completions(args: Completion, app: &mut App) -> Result<()> {
    generate(args.shell, app, app.get_name().to_string(), &mut io::stdout());

    Ok(())
}
```

Furisto (changes requested):

Furisto:
Minor stuff, the style should be consistent with the other commands.

creepinson:
Looks like I messed up the order of the comment and the `#[derive(Debug, Parser)]` line next to the struct, although `cargo fmt --all --check` is successful. Otherwise I believe I made the requested changes.

utam0k (approved):
Thanks!

Furisto (approved):

Furisto:
Great!

PR #516, opened 2021-12-05
Integration test: cgroup v1 network tests, fix to memory tests

tsturzl:
I have implemented a similar test to the oci-spec runtime tools, and I've verified that it is passing consistently. I also noticed that the cgroups path for memory was not set correctly and included the fix in this PR. I also verified the change the memory cgroups test is passing consistently. I also made sure both are able to fail, as I suspect the memory cgroup test wasn't even running before because the wrong path was set for the `can_run` check.

### DIFF
--- a/crates/integration_test/src/tests/cgroups/memory.rs
+++ b/crates/integration_test/src/tests/cgroups/memory.rs
@@ -8,8 +8,8 @@ use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
 
 use crate::utils::{test_outside_container, test_utils::check_container_created};
 
-const CGROUP_MEMORY_LIMIT: &str = "memory.limit_in_bytes";
-const CGROUP_MEMORY_SWAPPINESS: &str = "memory.swappiness";
+const CGROUP_MEMORY_LIMIT: &str = "/sys/fs/cgroup/memory/memory.limit_in_bytes";
+const CGROUP_MEMORY_SWAPPINESS: &str = "/sys/fs/cgroup/memory/memory.swappiness";
 
 fn create_spec(cgroup_name: &str, limit: i64, swappiness: u64) -> Result<Spec> {
     let spec = SpecBuilder::default()
--- a/crates/integration_test/src/tests/cgroups/network.rs
+++ b/crates/integration_test/src/tests/cgroups/network.rs
@@ -67,7 +67,7 @@ fn get_network_interfaces() -> Option<(String, String)> {
     let lo_if_name = interfaces.get(0).map(|iface| &iface.name)?;
     let eth_if_name = interfaces.get(1).map(|iface| &iface.name)?;
 
-    return Some((lo_if_name.to_string(), eth_if_name.to_string()));
+    Some((lo_if_name.to_string(), eth_if_name.to_string()))
 }
 
 fn test_network_cgroups() -> TestResult {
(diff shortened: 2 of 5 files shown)


tsturzl:
It's very ironic that the code coverage drops when adding tests. I wonder if we can exclude the integration tests crate from the code coverage reports.

tsturzl:
I made another PR to ignore coverage reports for the integration tests:  #517

crates/integration_test/src/tests/cgroups/memory.rs
```diff
@@ -8,8 +8,8 @@ use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
 
 use crate::utils::{test_outside_container, test_utils::check_container_created};
 
-const CGROUP_MEMORY_LIMIT: &str = "memory.limit_in_bytes";
-const CGROUP_MEMORY_SWAPPINESS: &str = "memory.swappiness";
+const CGROUP_MEMORY_LIMIT: &str = "/sys/fs/cgroup/memory/memory.limit_in_bytes";
+const CGROUP_MEMORY_SWAPPINESS: &str = "/sys/fs/cgroup/memory/memory.swappiness";
```
utam0k, line 12: 👍

crates/integration_test/src/tests/cgroups/network.rs
```diff
@@ -0,0 +1,146 @@
...
+    let cgroup_name = "test_network_cgroups";
+
+    let interfaces = datalink::interfaces();
+
+    // Gets the loopback interface, or defaults to "lo"
+    let lo_if_name = interfaces.get(0).map_or_else(|| "lo", |iface| &iface.name);
+    // Gets the first ethernet interface, or defaults to "eth0"
+    let eth_if_name = interfaces
+        .get(1)
+        .map_or_else(|| "eth0", |iface| &iface.name);
```
utam0k, line 74: It is better to check about `interfaces` in `can_run()`?

tsturzl: This is what the runtime tools tests are currently doing, they check the interfaces and default them to `lo` and `eth0`. As far as I can tell the tests may still pass if the interfaces aren't found.

tsturzl: See here: https://github.com/opencontainers/runtime-tools/blob/1684d131456a6bc99b8e96aa4a99783f21e58d79/validation/linux_cgroups_network/linux_cgroups_network.go#L21

I wish they explained the reasoning for setting a default more in the comment. It seems the rationale here, based on [this commit message](https://github.com/opencontainers/runtime-tools/commit/cf9decfcf37364d67f0d93ec80a0d0d847cef480) is to allow the test to run on systems with different interface names. For example as far as I can tell on every system the first interface is loopback (lo), and then on many systems the second interface may actually be something other than `eth0`. In fact on my system I do not have `eth0`, and the second interface is `wlp0s20f3`, probably because I don't have an ethernet interface on my laptop.

When I run the test with "eth0" explicitly set as the interface on my laptop the test does in fact fail. So I think you're right, we should probably add this to `can_run`.

tsturzl: Take a look at the latest revision.

utam0k (approved):
Looks good

PR #528, opened 2021-12-11
Implement integration tests for cgroup v2 cpu

Furisto:
Extends our integration with tests for cpu resource restrictions on cgroup v2 systems. It also enables logging for the tests in case you need to debug them.

### DIFF
--- a/crates/integration_test/src/logger.rs
+++ b/crates/integration_test/src/logger.rs
@@ -10,7 +10,6 @@ use std::str::FromStr;
 pub static LOG_FILE: OnceCell<Option<File>> = OnceCell::new();
 const LOG_LEVEL_ENV_NAME: &str = "YOUKI_INTEGRATION_LOG_LEVEL";
 
-
 /// Initialize the logger, must be called before accessing the logger
 /// Multiple parts might call this at once, but the actual initialization
 /// is done only once due to use of OnceCell
@@ -73,7 +72,7 @@ impl Log for IntegrationLogger {
     /// Function to carry out logging
     fn log(&self, record: &Record) {
         if self.enabled(record.metadata()) {
-            let log_msg =text_format(record);
+            let log_msg = text_format(record);
             // if log file is set, write to it, else write to stderr
             if let Some(mut log_file) = LOG_FILE.get().unwrap().as_ref() {
                 let _ = writeln!(log_file, "{}", log_msg);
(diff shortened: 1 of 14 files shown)


Furisto:
I suspect that the test error has nothing to do with my changes and the tests might have been broken for a while. The last time the tests were run was [here](https://github.com/containers/youki/runs/4423582546?check_suite_focus=true), but the youki binary is actually not found even though no error is reported. This seems to have been fixed with [this](https://github.com/containers/youki/pull/524) change, but this did not trigger the integration tests, so it was not detected there either. Investigating this further...

utam0k:
maybe it is solved by this PR 
https://github.com/containers/youki/pull/514

Furisto:
@utam0k PTAL

crates/integration_test/src/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::path::PathBuf;
+use std::str::FromStr;
+
+pub static LOG_FILE: OnceCell<Option<File>> = OnceCell::new();
+const LOG_LEVEL_ENV_NAME: &str = "YOUKI_INTEGRATION_LOG_LEVEL";
+
+/// Initialize the logger, must be called before accessing the logger
+/// Multiple parts might call this at once, but the actual initialization
+/// is done only once due to use of OnceCell
+pub fn init(log_file: Option<PathBuf>, debug: bool) -> Result<()> {
```
utam0k, line 16: I think it is better that the output of the integration test should be handled by only stderr/stdout. If you want to write to the file, you can use redirect, tee, etc

YJDoc2:
Hey @Furisto any idea why the integration test error is happening? The fix is to use the release version of youki instead of debug, but the debug / release should not differ in functionalities right? The CI log indicates that there was an error in reading the container state, which should either work or be broken in both debug and release builds... :thinking: 
If the exact reason is not yet found, I'll once try to git bisect tomorrow to try and get the faulty commit, and see if I can find anything from it.
Thanks :)

Furisto:
@YJDoc2 There are three parts
- I added a log statement some time ago that would at the start of main log the arguments with which youki was called
- The logger implementation logs to stderr
- The integration test calls the state command and fails if output was written to stderr

The reason that it works in release and not in debug is that the log level in release and debug is different. Using the release flags is the simplest fix and we should test in release anyway, but I am planning to take another look at the issue once this PR is through.

Furisto:
@utam0k Done

utam0k (approved):
LGTM

Issue #529, opened 2021-12-12
dockerd/moby - failed to parse youki version: unknown output format


jhult:
`sudo systemctl status docker.service` produces this `WARNING`:

`Dec 10 15:14:29 hostname dockerd[20706]: time="2021-12-10T15:14:29.259763302-05:00" level=warning msg="failed to parse /usr/bin/youki version: unknown output format: youki 0.0.1\n"`

Here is how the parsing is being done: https://github.com/moby/moby/blob/65cc84abc522a564699bb171ca54ea1857256d10/daemon/info_unix.go#L280

Related issues/commits:
- https://github.com/moby/moby/issues/41628
- https://github.com/moby/moby/commit/2137b8ccf27c16f230a50f9019ed0dbeee0840c1
- https://github.com/kata-containers/runtime/pull/2386

PR #530, opened 2021-12-12
Update version for runc compatibility for Moby

jhult:
Update version for compatibility with Moby. Fix for #529.

image

### DIFF
--- a/crates/youki/Cargo.toml
+++ b/crates/youki/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.1"
 authors = ["youki team"]
 edition = "2021"
 description = "A container runtime written in Rust"
+build = "build.rs"
 
 [dependencies.clap]
 version = "3.0.0-beta.5"
@@ -12,7 +13,7 @@ features = ["std", "suggestions", "derive", "cargo"]
 
 
 [dependencies]
-anyhow = "1.0"
+anyhow = "1.0.51"
 chrono = { version="0.4", features = ["serde"] }
 libcgroups = { path = "../libcgroups" }
 libcontainer = { path = "../libcontainer" }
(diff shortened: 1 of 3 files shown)


utam0k:
Sorry, we are probably the cause of the CI failures 😭  Please wait a moment.

crates/youki/Cargo.toml
```diff
@@ -31,3 +32,6 @@ clap_generate = { version = "3.0.0-beta.5" }
 [dev-dependencies]
 serial_test = "0.5.1"
 
+[build-dependencies]
+anyhow = "1.0"
```
utam0k, line 36:
```suggestion
anyhow = "1.0.51"
```

crates/youki/build.rs
```diff
@@ -0,0 +1,8 @@
+use anyhow::Result;
+use vergen::{Config, ShaKind, vergen};
+
+fn main() -> Result<()> {
```
utam0k, line 4: Why do we need to use `anyhow::Result` in `build.rs`?

crates/youki/Cargo.toml
jhult: Bumped this in `[build-dependencies]` and also bumped `anyhow` in the `[dependencies]` section.

crates/youki/build.rs
jhult: The [`vergen` function](https://docs.rs/vergen/latest/vergen/fn.vergen.html) returns a [`Result`](https://docs.rs/anyhow/1.0.51/anyhow/type.Result.html).

I based the code on [this](https://docs.rs/vergen/latest/vergen/#buildrs) example.

If there is another (or better) way to accomplish this, feel free to suggest it :)

utam0k: Oh sorry, my point was ignorant.

jhult:
Rebased on latest `main` and fixed formatting.

utam0k (approved):
Thanks!

Issue #147, opened 2021-07-20
impelementation of update command


utam0k:
This issue creates an implementation that will serve as a template for now, and the other parameters that will be updated will be managed in a separate issue as a `good first issue`.
ref: https://github.com/opencontainers/runc/blob/master/man/runc-update.8.md

utam0k:
@fbrv It's a little more challenging, but are you interested?

fbrv:
Sure, I can give a try

utam0k:
@fbrv I'm looking forward to your PR!

knight42:
@utam0k Hi! I would like to work on this!

utam0k:
@knight42 Sure!

hayama17:
@saku3 @knight42 
Hi! Can I work on the CPU flags (--cpu-period, --cpu-quota, --cpu-share, --cpu-burst, --cpu-idle) used by the update tests in #3576 ? They are currently skipped for youki.

saku3:
@hayama17 
Yes, please go ahead! Thank you for taking this on.

hayama17:
@saku3 

>  the other parameters that will be updated will be managed in a separate issue as a good first issue.

I don't have permission to create sub-issues, so here is a proposal for splitting this issue, based on the flags in [runc-update.8.md](https://github.com/opencontainers/runc/blob/main/man/runc-update.8.md). Grouped by resource since some flags share the same cgroup file (e.g. `--cpu-period`/`--cpu-quota` → `cpu.max`):

| sub-issue | flags |
| --- | --- |
| cpu | `--cpu-period`, `--cpu-quota`, `--cpu-share`, `--cpu-burst`, `--cpu-idle` |
| cpuset | `--cpuset-cpus`, `--cpuset-mems` |
| memory | `--memory`, `--memory-reservation`, `--memory-swap` |
| blkio | `--blkio-weight` |
| Intel RDT | `--l3-cache-schema`, `--mem-bw-schema` |
| cpu-rt (cgroup v1 only, low priority) | `--cpu-rt-period`, `--cpu-rt-runtime` |


If this looks good, could you create the sub-issues? I'd like to start with the cpu one.

hayama17:
@utam0k @saku3

I don't have permission to create sub-issues, so unless there are objections to the grouping above, I'll open them as regular issues in a few days and link them back to this one — feel free to convert them to sub-issues afterwards if that's preferred.

saku3:
@hayama17 

[...]

PR #536, opened 2021-12-13
implement the update subcommand(partially)

knight42:
Ref #147 

This PR did several things:
* Update the version of `oci-spec` to the latest
* Extract a common function `create_cgroup_manager` which could be used in `ps` and `update` command
* Add the `update` subcommand and make it be abled to changed the pids limit in the container

### DIFF
--- a/crates/liboci-cli/src/update.rs
+++ b/crates/liboci-cli/src/update.rs
@@ -9,5 +9,4 @@ pub struct Update {
     /// Set the maximum number of processes allowed in the container
     #[clap(long)]
     pub pids_limit: Option<i64>,
-    // TODO(knight42): support more options
 }
--- a/crates/liboci-cli/src/update.rs
+++ b/crates/liboci-cli/src/update.rs
@@ -1,0 +1,13 @@
+use clap::Parser;
+
+/// Update running container resource constraints
+#[derive(Parser, Debug)]
+pub struct Update {
+    #[clap(forbid_empty_values = true, required = true)]
+    pub container_id: String,
+
+    /// Set the maximum number of processes allowed in the container
+    #[clap(long)]
+    pub pids_limit: Option<i64>,
+    // TODO(knight42): support more options
+}
--- a/crates/youki/src/commands/mod.rs
+++ b/crates/youki/src/commands/mod.rs
@@ -3,7 +3,6 @@ use std::{fs, path::Path};
 
 use libcgroups::common::CgroupManager;
 use libcontainer::container::Container;
-use libcontainer::utils;
 
 pub mod completion;
 pub mod create;
@@ -40,25 +39,11 @@ fn create_cgroup_manager<P: AsRef<Path>>(
     root_path: P,
     container_id: &str,
 ) -> Result<Box<dyn CgroupManager>> {
-    let container = load_container(root_path, &container_id)?;
-    let config_absolute_path = container.root.join("config.json");
-    log::debug!("load spec from {:?}", config_absolute_path);
-    let spec = oci_spec::runtime::Spec::load(config_absolute_path)?;
-    log::debug!("spec: {:?}", spec);
-    let cgroups_path = utils::get_cgroup_path(
-        spec.linux()
-            .as_ref()
-            .context("no linux in spec")?
-            .cgroups_path(),
-        container.id(),
-    );
+    let container = load_container(root_path, container_id)?;
+    let cgroups_path = container.spec()?.cgroup_path;
     let systemd_cgroup = container
         .systemd()
         .context("could not determine cgroup manager")?;
 
-    Ok(libcgroups::common::create_cgroup_manager(
-        cgroups_path,
-        systemd_cgroup,
-        container.id(),
-    )?)
+    libcgroups::common::create_cgroup_manager(cgroups_path, systemd_cgroup, container.id())
 }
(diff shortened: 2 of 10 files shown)


knight42:
/cc @utam0k 

One thing I am curious is that now we read the cgroup path from the `config.json` under container root path https://github.com/containers/youki/blob/cbba0bc641ae2cb3aa3f0a060676532296c5dd97/crates/youki/src/commands/ps.rs#L14, but `config.json` is not copied to the root dir when creating container, so the `config.json` is missing when I run the `ps` command, is this intended?

crates/youki/src/commands/mod.rs
```diff
@@ -32,3 +35,30 @@ fn load_container<P: AsRef<Path>>(root_path: P, container_id: &str) -> Result<Co
...
+    log::debug!("load spec from {:?}", config_absolute_path);
+    let spec = oci_spec::runtime::Spec::load(config_absolute_path)?;
+    log::debug!("spec: {:?}", spec);
+    let cgroups_path = utils::get_cgroup_path(
+        spec.linux()
+            .as_ref()
+            .context("no linux in spec")?
+            .cgroups_path(),
+        container.id(),
+    );
```
Furisto, line 54: You can use container.spec() for this. We do not save the whole runtime spec during container creation anymore, only the parts that will be needed later on. This is the reason why trying to load config.json fails.

Furisto (changes requested):

Furisto:
The CI failure is due to some clippy warnings that need to be fixed

crates/liboci-cli/src/update.rs
```diff
@@ -0,0 +1,13 @@
...
+/// Update running container resource constraints
+#[derive(Parser, Debug)]
+pub struct Update {
+    #[clap(forbid_empty_values = true, required = true)]
+    pub container_id: String,
+
+    /// Set the maximum number of processes allowed in the container
+    #[clap(long)]
+    pub pids_limit: Option<i64>,
+    // TODO(knight42): support more options
```
Furisto, line 12: Please remove the TODO. If there are still open points they should be put in the issue.

Furisto:
Also the clippy warnings still need to be fixed and the pnet package seems to be specified twice in the cargo.lock file now.

utam0k:
> 
@knight42 
Oh I forget to take care of `ps` command :sob: 
When this PR https://github.com/containers/youki/pull/447 merged, we stopped copying `config.rs` for improving performance.
Now, we use `YoukiConfig` instead of `config.rs`
https://github.com/containers/youki/blob/781f9fa49b1fc6aa2534c28fb2c97a4b5a5bc47b/crates/libcontainer/src/container/container_start.rs#L43
If you like, I will ask you to fix it. What do you think? Of course, if you are busy, I'll fix it on my end.

crates/youki/src/commands/update.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub fn update(args: Update, root_path: PathBuf) -> Result<()> {
+    let cmanager = create_cgroup_manager(root_path, &args.container_id)?;
+
+    let mut linux_res = LinuxResources::default();
+    if args.pids_limit.is_some() {
+        let mut pids = LinuxPids::default();
+        pids.set_limit(args.pids_limit.unwrap());
+        linux_res.set_pids(Some(pids));
+    }
```
utam0k, line 17: How about this? This will eliminate the need `unwrap()` :wink: 
```suggestion
    if let Some(new_pids_limit) = args.pids_limit {
        let mut pids = LinuxPids::default();
        pids.set_limit(new_pids_limit);
        linux_res.set_pids(Some(pids));
    }
```

crates/liboci-cli/src/update.rs
```diff
@@ -0,0 +1,12 @@
...
+
+/// Update running container resource constraints
+#[derive(Parser, Debug)]
+pub struct Update {
+    #[clap(forbid_empty_values = true, required = true)]
+    pub container_id: String,
+
+    /// Set the maximum number of processes allowed in the container
+    #[clap(long)]
+    pub pids_limit: Option<i64>,
```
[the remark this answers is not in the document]
Furisto: The runtime spec specifies a signed integer for the pid limit. We (and runc and also crun if I remember correctly) treat a negative value as unrestricted (i.e. 'max').

knight42:
> @knight42 Oh I forget to take care of `ps` command 😭 When this PR #447 merged, we stopped copying `config.rs` for improving performance. Now, we use `YoukiConfig` instead of `config.rs`
> 

@utam0k Hi! I believe this issue has been fixed in this PR based on @Furisto 's suggestion [here](https://github.com/containers/youki/pull/536#discussion_r767817099).

@Furisto Would you like to take another look?

crates/youki/src/commands/update.rs
knight42: Nice catch! Fixed.

Furisto:
@knight42 Looks good! One last thing: Can you please squash the last 4 commits into one and name it "Address review comments" or something like that? Otherwise LGTM!

knight42:
@Furisto Hi! I have already squashed the commits.

utam0k (approved):
Looks good. Thanks!

Furisto:
@knight42 Can you resolve the merge conflicts?

knight42:
@Furisto Fixed now.

Furisto (approved):

utam0k:
@knight42 Awesome!

PR #537, opened 2021-12-13
Add Cgroup V1 block IO integration test

YJDoc2:
This adds Cgroups V1 block io integration test

### DIFF
--- a/crates/integration_test/src/tests/cgroups/blkio.rs
+++ b/crates/integration_test/src/tests/cgroups/blkio.rs
@@ -15,6 +15,17 @@ use crate::utils::{
     test_utils::{check_container_created, CGROUP_ROOT},
 };
 
+// -----> README
+// for this test to work for all parameters, the kernel needs to be compiled with CFQ IO schedular
+// for some reason, Ubuntu and other distributions might come with a kernel compiled with mq-deadline
+// schedular, which does not expose/support options such as blkio.weight, blkio.weight_device
+// for these we can skip all the tests, or skip the corresponding tests
+// the current implementation skips corresponding tests, so one can test atleast bps and iops
+// device on such systems
+// check https://superuser.com/questions/1449688/a-couple-of-blkio-cgroup-files-are-not-present-in-linux-kernel-5
+// and https://github.com/opencontainers/runc/issues/140
+// for more info on this
+
 #[derive(Debug, Default)]
 struct WeightDevice {
     major: i64,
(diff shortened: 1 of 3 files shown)


YJDoc2:
The CI is likely to fail for the same reason as https://github.com/containers/youki/pull/528#issuecomment-991806762
The test also give same error for youki on my local machine, but runc passes all the tests.
Once that issue is fixed, I'll run the tests on youki as well and update here.

Furisto:
@YJDoc2 I fixed this problem [here](https://github.com/containers/youki/pull/528/commits/d1bdf10859cb0077ba828ee0318e4212b29e78b2)

YJDoc2:
Oh yes, I tested youki with that change and it passes the tests as well. I'm waiting for that PR to be merged so it will pass in the CI itself, and the I can ask for reviews. Thanks :+1:

crates/integration_test/src/tests/cgroups/blkio.rs
```diff
@@ -0,0 +1,578 @@
...
+            fs::read_to_string(format!("{}/blkio.throttle.write_bps_device", path))
+                .context("error in reading block io write bps device")?;
+        let mut throttle_devices = Vec::new();
+        for line in throttle_write_bps_string.lines() {
+            let (device_data, rate) = line
+                .split_once(" ")
+                .context("invalid weight device format")?;
+            let (major_str, minor_str) = device_data
+                .split_once(":")
+                .context("invalid major-minor number format")?;
```
Furisto, line 215: This snippet seems to be used a few times. Could be worth creating a method for this.

crates/integration_test/src/tests/cgroups/blkio.rs
```diff
@@ -0,0 +1,578 @@
...
+        .join("blkio/runtime-test")
+        .join(cgroup_name);
+    let block_io = get_blkio_data(cgroup_path.to_str().unwrap())?;
+
+    let resources = spec.linux().as_ref().unwrap().resources().as_ref().unwrap();
+    let spec_block_io = resources.block_io().as_ref().unwrap();
+    // weight ------
+    if supports_weight() {
+        if spec_block_io.weight().is_none() {
+            return Err(anyhow::anyhow!("spec block io weight is none"));
```
Furisto, line 292:
```suggestion
            bail!("spec block io weight is none");
```
A few more instances of this.

YJDoc2:
Hey, sorry for the mistakes :sweat_smile: :sweat_smile:
I had realized that the device reading code can be extracted into function, but forgot to do that before making commit and asking review :sweat_smile: 

I have fixed that, as well as improved error reporting as you have suggested. One of the issue that it caused is that we need to clone the path, as fs::read_to_string takes the path by value. 

I'm not sure why read_to_string needs to take it by value, as I checked rust src, and it eventually calls as_ref on it, which needs only by ref, not by value... :thinking: 

Anyways, please take a look now, thanks :)

Furisto:
> I have fixed that, as well as improved error reporting as you have suggested. One of the issue that it caused is that we need to clone the path, as fs::read_to_string takes the path by value.
> 
Can you point to a specific example? PathBuf implements AsRef<Path> so you should not need to clone it.

YJDoc2:
> Can you point to a specific example? PathBuf implements AsRef so you should not need to clone it.

Yes you're right, I have been using it incorrectly :sweat: 

I have removed the clone calls, so PTAL.

Furisto:
Looks good! @utam0k Will this conflict with #514? If it does it might make more sense to merge yours first, as yours looks trickier to adapt to the changes than the other way around.

utam0k:
> Looks good! @utam0k Will this conflict with #514? If it does it might make more sense to merge yours first, as yours looks trickier to adapt to the changes than the other way around.

Thank you for taking care of my pr. I still have a few things to do on that PR of mine, so you can merge this one first!

Furisto (approved):

PR #569, opened 2021-12-27
Interpret a cpu quota of zero as default value

Furisto:
Runc interpretes a cpu quota of zero as an instruction to set the default value for cpu quota. As runtimes like CRI-O [depend](https://github.com/containers/youki/issues/78#issuecomment-1001039717) on this behavior we should do the same.


utam0k (approved):
Looks good

PR #573, opened 2021-12-28
Add option to list test groups

Furisto:
When I only want to run specific tests, I have to look up the test group name in the code, which is inconvenient. This adds a command to the integration test executable to list all test groups.

### DIFF
--- a/crates/integration_test/src/main.rs
+++ b/crates/integration_test/src/main.rs
@@ -7,7 +7,7 @@ use crate::tests::pidfile::get_pidfile_test;
 use crate::tests::seccomp_notify::get_seccomp_notify_test;
 use crate::tests::tlb::get_tlb_test;
 use crate::utils::support::set_runtime_path;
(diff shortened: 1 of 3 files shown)


crates/integration_test/src/main.rs
```diff
@@ -95,11 +96,42 @@ fn main() -> Result<()> {
...
+        // runtime path is relative or resolved correctly
+        Ok(path) => set_runtime_path(&path),
+        // runtime path is name of program which probably exists in $PATH
+        Err(_) => match which::which(&opts.runtime) {
+            Ok(path) => set_runtime_path(&path),
+            Err(e) => {
+                eprintln!("Error in finding runtime : {}\nexiting.", e);
+                std::process::exit(66);
+            }
+        },
```
utam0k, line 118: 💯

utam0k (approved):
LTGM

PR #576, opened 2021-12-29
Add gitpod as development option

Furisto:
@utam0k Do you have permissions to install the [app](https://github.com/apps/gitpod-io) for this repository? Then we could also enable prebuilds.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/576"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,9 @@
+image:
+  file: .gitpod.Dockerfile
+tasks:
+  - name: Build
+    init: cargo build 
+github:
+  prebuilds:
+    pullRequestsFromForks: true
+    addBadge: true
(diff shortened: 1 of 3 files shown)


utam0k:
@Furisto I've tried. What do you think?

Furisto:
@utam0k Ok, seems to be working.

utam0k:
> @utam0k Ok, seems to be working.

@Furisto 
Can I do the tutorial which be written
in README?
I have tried this a few month ago and failed😢

.gitpod.yml
```diff
@@ -0,0 +1,9 @@
+image:
+  file: .gitpod.Dockerfile
+tasks:
+  - name: Build
+    init: cargo build 
+github:
+  prebuilds:
+    pullRequestsFromForks: true
+    addBadge: true
```
utam0k, line 9: How about adding the `rust-analyzer` extension?

utam0k:
Maybe we need to use `qemu` to run the tutorial, but gitpod is enough helpful to review PR and so on.

utam0k (approved):
This is enough useful to review PR or something. Looks good.

PR #579, opened 2021-12-30
Use correct hugetlb interface file name

Furisto:


utam0k (approved):
Perfect!
https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#hugetlb

PR #580, opened 2021-12-30
style(spellcheck): Run a spellchecker and fix typos

em-:
Fix a broken link and a few misspellings.

### DIFF
--- a/crates/libcontainer/src/capabilities.rs
+++ b/crates/libcontainer/src/capabilities.rs
@@ -544,7 +544,7 @@ mod tests {
         struct Testcase {
             name: String,
             input: LinuxCapabilities,
-            // be awared of that calling sequence in the drop_privileges function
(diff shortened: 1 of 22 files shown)


crates/libcontainer/src/capabilities.rs
```diff
@@ -544,7 +544,7 @@ mod tests {
         struct Testcase {
             name: String,
             input: LinuxCapabilities,
-            // be awared of that calling sequence in the drop_privileges function
+            // be awarded of that calling sequence in the drop_privileges function
```
Furisto, line 547:
```suggestion
            // be aware that the calling sequence in the drop_privileges function
```

em-: Err, thank you!

utam0k (approved):
Thanks for your first contribution! Looks nice.

PR #582, opened 2021-12-31
Readonly paths

YJDoc2:
Add readonly paths integration tests


YJDoc2:
Hey @Furisto , @yihuaf can you take a look? This still has some things to be cleaned up, but I made this PR, as I had some questions regarding my implementation, as well as the original tests. I'll add them as comments, please check 
Thanks

crates/integration_test/src/tests/readonly_paths/readonly_paths.rs
```diff
@@ -0,0 +1,275 @@
...
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
+
+        let test_file = bundle_path.join(&ro_file);
+        match fs::File::create(&test_file) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
```
YJDoc2, line 90: In the readonly paths test,



The link to the go test is : https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_readonly_paths/linux_readonly_paths.go#L14

What I can't understand is under which conditions/errors in runtime itself is the test supposed to fail : the test sets some paths as read-only in the spec, then sends a fn to validate inside the runtime, and returns error if any from the validation

Now the fn that is sends to inside validate creates the first-dir, sub dirs and files that are supposed to be read-only, and writes some bytes to the files ; and returns err, if any of this returns err. I initially thought that these operations should fail, as the paths are read-only, but as the function is called not from inside, but from outside the container, these operations are valid. Even the (original Go) test itself expect these to be successful. Then I checked the validate-inside fn, but that simply call this fn, then creates and starts the runtime, waits for it to stop and then returns if any error occurred in running of the container. 

So I thought that the test should fail if the read only first are not created, which will fail the creation of container, so I used an empty fn to give to inside-validate ; but that too is passing for runc and youki.

Thus I cannot understand under which circumstances is the test supposed to fail, cause otherwise right now it seems a redundant test.

Basically, as test can only show presence of bugs, not the absence of them, I don't understand the purpose of this test, as I don't understand the condition under which it is supposed to fail.

crates/integration_test/src/tests/readonly_paths/readonly_paths.rs
```diff
@@ -0,0 +1,275 @@
...
+        let r = unsafe { libc::mknod(_path, mode, 0) };
+        if r != 0 {
+            bail!(
+                "error in creating a device node at {:?} with mode {:?}, got return code {}",
+                test_file,
+                mode,
+                r
+            );
+        }
+
```
YJDoc2, line 217: The use of unsafe irks me here, as there should be a better way to implement this. I tried setting the mode in open options, but it always gives OsErr 22, invalid argument, as specified in the comment here. Is there a better way to do this, or is there something that I am doing wrong because of which the open option method is failing?

crates/integration_test/src/tests/readonly_paths/readonly_paths.rs
```diff
@@ -0,0 +1,275 @@
...
+    );
+    let ro_device_nodes_chr = Test::new(
+        "readonly_device_node_chr",
+        Box::new(|| test_node(libc::S_IFCHR | 0o666)),
+    );
+
+    let ro_device_nodes_fifo = Test::new(
+        "readonly_device_nodes_fifo",
+        Box::new(|| test_node(libc::S_IFIFO | 0o666)),
+    );
```
YJDoc2, line 263: Another issue with the device node is that if they are run separately, as they are here then all three tests pass. If I try to put them through a loop, as in the commented test function, then only the first in the loop succeeds, and the second fails. I tried this with different permutations, and always the first iteration of loops succeeds, the second one fails, irrespective device type which is first and second. I suspected that it might be issue that I'm trying to bind same device major & minor (0 in the test_device) to block / char / fifo device node simultaneously, but 
- as tests are run one after other in loop, this should not be an issue
- I tried binding same device (major,minor) = (0,0) locally manually simultaneously to blk/charr/fifo devices using mknod, and that actually succeeds.

I cannot understand what issue might be there, as each device node will be created in a separate container, which separate invocations of check_inside_runtime in the loop version

crates/integration_test/src/tests/readonly_paths/readonly_paths.rs
Furisto: The part you linked does not actually test the writability of the readonly-paths. The runtime test tools start a validation executable within a container in order to test that, see https://github.com/opencontainers/runtime-tools/blob/db715ee5eb891b81a031e680fe92392bff11e31b/cmd/runtimetest/main.go#L939-L962

crates/integration_test/src/tests/readonly_paths/readonly_paths.rs
yihuaf: Not sure if this is the issue, but can you chain the opts calls? See example: https://doc.rust-lang.org/std/fs/struct.OpenOptions.html

`if let io::Result::Err(e) = fs::OpenOptions::new().mode(mode).create(true).open(&test_file) {...}`

YJDoc2: Hey, I did try chaining the opts, and that compiled, but still gave the same issue at runtime as the separate value setting.

yihuaf: How about: `In order for the file to be created, OpenOptions::write or OpenOptions::append access must be used.` from https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create

yihuaf: Also, can we use nix::mknod instead of libc::mknod?

YJDoc2: Hey, using `write(true)` fixed the issue :+1:  thanks for the suggestion!
I originally thought `create(true)` by itself would set `write(true)` internally, but we need to do it manually.
This solves the unsafe issue, as well as the separate tests issue :+1:
I also have removed the libc direct dependency and used nix all the way :)

crates/integration_test/src/tests/readonly_paths/readonly_paths.rs
YJDoc2: solved by using OpenOptions with `create(true).write(true)`

YJDoc2:
This PR is adds an important Part of the integration tests : The test_inside_container function, which allows testing the restrictions from inside the container ; along with its counterpart the runtimetest . The PR contains the readonly_paths test, as well as a new crate for runtimetest, which is the binary which runs inside the container process and runs tests inside.

Apologies for such a big PR, but it needed to add both of the things together, as to make sure runtimetest works, we need a test which will run test_inside_container and to run a test_inside_container, we need runtiemtest, thus both of those ended up being in this same PR.

As the runtimetest tool and its configuration can be tricky, as well as I had several issues while figuring out how all works and fits together for the original Go tests, I have added detailed explanations in the Readme as well as have added docs for it as well.

Also I would like to apologies to @utam0k because this PR will likely make even more conflicts for #514 . If there is any way I can help with resolving the conflicts, or reducing the issue, please let me know I would like to help. If we need to wait for it to be merged before this, that is fine as well, we can review this and keep it in queue to be merged after that PR.

Please take a look, thanks :)

utam0k:
There is no problem. As this pr is too large to merge, I'll try to make some smaller PRs.

> Also I would like to apologies to @utam0k because this PR will likely make even more conflicts for #514 . If there is any way I can help with resolving the conflicts, or reducing the issue, please let me know I would like to help. If we need to wait for it to be merged before this, that is fine as well, we can review this and keep it in queue to be merged after that PR.

utam0k (approved):
Awesome!
However, this PR will inevitably have a large impact on other people's work. Therefore, I'll ignore the details and merge this once. Let me improve some of them on myend.

Issue #595, opened 2022-01-06
Panic when running rootless container example


Will-Shanks:
After following the rootless docker install steps in the [ArchWiki](https://wiki.archlinux.org/title/Docker#Docker_rootless) and this docker [guide](https://docs.docker.com/engine/security/rootless/) I attempted to follow the rootless container example in the readme, which caused a panic.

Steps to repro: (follow rootless guide in readme)
```bash 
$ systemctl --user start docker
$ export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
$ ./build.sh
$ mkdir -p tutorial/rootfs
$ cd tutorial
$ docker export $(docker create busybox) | tar -C rootfs -xf -
$ ../youki spec --rootless
[DEBUG crates/youki/src/main.rs:92] 2022-01-05T16:08:14.758729077-08:00 started by user 1000 with ArgsOs { inner: ["../youki", "spec", "--rootless"] }

$ RUST_BACKTRACE=1 ../youki run rootless-container
[DEBUG crates/youki/src/main.rs:92] 2022-01-05T16:13:10.880630125-08:00 started by user 1000 with ArgsOs { inner: ["../youki", "run", "rootless-container"] }
[DEBUG crates/libcontainer/src/container/init_builder.rs:94] 2022-01-05T16:13:10.883668566-08:00 container directory will be "/run/user/1000/youki/rootless-container"
[DEBUG crates/libcontainer/src/container/container.rs:191] 2022-01-05T16:13:10.883730716-08:00 Save container status: Container { state: State { oci_version: "v1.0.2", id: "rootless-container", status: Creating, pid: None, bundle: ".", annotations: Some({}), created: None, creator: None, use_systemd: None }, root: "/run/user/1000/youki/rootless-container" } in "/run/user/1000/youki/rootless-container"
[DEBUG crates/libcontainer/src/rootless.rs:39] 2022-01-05T16:13:10.883918610-08:00 rootless container should be created
[INFO crates/libcgroups/src/common.rs:232] 2022-01-05T16:13:10.884004438-08:00 systemd cgroup manager with system bus false will be used
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 1', crates/libcgroups/src/systemd/manager.rs:88:22
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
[...]

PR #597, opened 2022-01-06
Improve cgroup path handling for rootless containers

Furisto:
- Ensure that a valid cgroup path is created for rootless containers if the cgroup path is not specified
- Handle that a cgroup path without a parent is provided

See #595

### DIFF
--- a/crates/libcgroups/src/systemd/manager.rs
+++ b/crates/libcgroups/src/systemd/manager.rs
@@ -7,7 +7,7 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use dbus::arg::RefArg;
-use nix::unistd::Pid;
+use nix::{unistd::Pid, NixPath};
 use std::path::{Path, PathBuf};
 
 use super::{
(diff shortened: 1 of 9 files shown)


Will-Shanks:
This PR fixes #595

Furisto:
Thanks for the confirmation @Will-Shanks.

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -65,35 +65,33 @@ impl TryFrom<&Path> for CgroupsPath {
...
+                prefix: parts[0].to_owned(),
+                name: parts[1].to_owned(),
+            },
+            3 => CgroupsPath {
+                parent: parts[0].to_owned(),
+                prefix: parts[1].to_owned(),
+                name: parts[2].to_owned(),
+            },
+            _ => bail!("cgroup path {:?} is invalid", cgroups_path),
+        };
```
utam0k, line 92: Although I looked for to avoid specifying an index of the array, I didn't find out 😭

utam0k (approved):

utam0k:
Looks good!

PR #614, opened 2022-01-16
make the runtime-tools directory one level deeper.

utam0k:
First step of https://github.com/containers/youki/issues/601


YJDoc2:
Hey, I tried this with different combinations, and when running on my system locally it works fine. After searching a bit, I feel this issue is due to some problem in current nightly. I'm updating my nightly will update here after checking with current nightly.

YJDoc2:
Hey, I could reproduce the same error that is in the CI with today's (16 Jan 2022) nightly on my local system. The CI steps worked fine with the previous nightly I had installed, so I don't think this is any error in the PR changes. 

That said , can you take a look at https://github.com/YJDoc2/youki/runs/4831134294?check_suite_focus=true ,  in this the `integration_test` workflow seems to be failing due to some git checkout issue, which is not occurring in this PR.

Thanks :)

.github/workflows/main.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         run: |
           cargo llvm-cov clean --workspace
           cargo llvm-cov --no-report
-          cargo llvm-cov --no-run --lcov --ignore-filename-regex "libseccomp/src|integration_test/src|test_framework/src|systemd_api.rs" --output-path ./coverage.lcov
+          cargo llvm-cov --no-run --lcov --ignore-filename-regex "libseccomp/src|crates/integration_test/src|test_framework/src|systemd_api.rs" --output-path ./coverage.lcov
```
YJDoc2, line 96: If this change was done only to fix the CI coverage issue, this can be reverted, as that seems nightly issue, not the PR changes, and previous (without this change) works fine without this change.

utam0k: This change doesn't seems to fix it 😂

YJDoc2: As I said in https://github.com/containers/youki/pull/614#issuecomment-1013831105 I think the issue is not in your changes issue is in the current nightly rust version, with which CI is running this.

If you haven't updated your rust using `rustup update` in at least last 5-6 days, try running cargo-llvm-cov manually using the commands we use in CI. I did that for a nightly version of roughly 1 month before. The coverage report was generated normally. Then I updated the rust , and ran it again, and with the updated nightly version I got the exact same error we are getting in this PR.

As it seems now, only way is to wait next few days, in which the updated nightly might fix this issue. I'm thinking of opening an issue regarding this, but not sure if it should be filed on cargo-llvm-cov or on rust itself.

utam0k: @YJDoc2 Thanks for your information. I have put them back.

utam0k:
> Hey, I could reproduce the same error that is in the CI with today's (16 Jan 2022) nightly on my local system. The CI steps worked fine with the previous nightly I had installed, so I don't think this is any error in the PR changes.
> 
> That said , can you take a look at https://github.com/YJDoc2/youki/runs/4831134294?check_suite_focus=true , in this the `integration_test` workflow seems to be failing due to some git checkout issue, which is not occurring in this PR.
> 
> Thanks :)

Maybe you forgot to change the submodule, i.e. `git mv`?

YJDoc2 (approved):
Hey, this looks good to me :+1:
Are you still making any changes here, or is it fine to merge?

utam0k:
> Hey, this looks good to me 👍 Are you still making any changes here, or is it fine to merge?

No, I am. I'll merge this.

PR #615, opened 2022-01-16
support rust 1.58.0

utam0k:


.github/workflows/integration_tests_validation.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [1.56.1, 1.57.0]
+        rust: [1.56.1, 1.58.0]
```
Furisto, line 28: Should this be 
```suggestion
        rust: [1.57.0, 1.58.0]
```

utam0k: Sorry for not explaining, here I want to run CI with min and max to support.

Furisto (approved):

PR #622, opened 2022-01-19
ready for integration test for the exec command.

utam0k:
youki's exec command is now broken 😭 
cc: @Furisto 

I've checked that this pr works fine using `runc`.

### DIFF
--- a/crates/integration_test/src/tests/lifecycle/container_lifecycle.rs
+++ b/crates/integration_test/src/tests/lifecycle/container_lifecycle.rs
@@ -36,8 +36,8 @@ impl ContainerLifecycle {
     }
 
     #[allow(dead_code)]
-    pub fn exec(&self, cmd: &str) -> TestResult {
-        exec::exec(&self.project_path, &self.container_id, cmd)
+    pub fn exec(&self, cmd: Vec<&str>, expected_output: Option<&str>) -> TestResult {
+        exec::exec(&self.project_path, &self.container_id, cmd, expected_output)
     }
 
     pub fn start(&self) -> TestResult {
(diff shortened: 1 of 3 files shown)


crates/integration_test/src/tests/lifecycle/container_lifecycle.rs
```diff
@@ -64,6 +69,7 @@ impl<'a> TestableGroup<'a> for ContainerLifecycle {
         vec![
             ("create", self.create()),
             ("start", self.start()),
+            // ("exec", self.exec("true")),
```
YJDoc2, line 72: Can we change the true command to something that gives output like `echo "Hello world"` ? That way we will be able to see if the exec command actually executes the given command inside the container. Otherwise even a no-op exec will pass the test.

crates/integration_test/src/tests/lifecycle/exec.rs
```diff
@@ -0,0 +1,20 @@
...
+        .stderr(Stdio::piped())
+        .arg("--root")
+        .arg(project_path.join("runtime"))
+        .arg("exec")
+        .arg(id)
+        .arg(exec_cmd)
+        .spawn()
+        .expect("failed to execute exec command")
+        .wait_with_output();
+    get_result_from_output(res)
```
YJDoc2, line 19: In case we change the command to `echo "Hello world"` , you will have to manually get the stdout and stderr here, and check for no-error and that stdout has the expected output.

YJDoc2 (commented):
Hey this is a very good idea to add an exec test! I don't think the original oci-tests had exec separately tests, which is why we might have forgotten to add this originally, but definitely a needed addition :+1:

I have added a couple of comments, please take a look at them :)

crates/integration_test/src/tests/lifecycle/exec.rs
utam0k: 👍

crates/integration_test/src/tests/lifecycle/container_lifecycle.rs
utam0k: 👍

utam0k:
@YJDoc2 PTAL

YJDoc2 (approved):
LGTM :+1:

YJDoc2:
Hey, This is great, we can merge this now :+1:
The CI seems to be failing due to DCO app, but I'm not sure when and why did we enable this for the repo. :thinking:

PR #626, opened 2022-01-21
Add minimum rust version requirement to libcgroups and libcontainers

YJDoc2:
CVE-2022-21658 was announced today which affects the rust std::fs::remove_dir_all function, where due to a race condition, a non-privileged process can request a privileged process to delete a dir which they do not have permission to delete. More information on this can be found at https://blog.rust-lang.org/2022/01/20/cve-2022-21658.html

Youki uses this function in libcgroups and libcontainer, where it is used to either remove a cgroup or remove a container bundle dir. Thus this PR adds `rust-version` field to both of these, and sets it to `1.58.1` which has a patch for this CVE. Now to compile youki one must have rust compiler of version at least `1.58.1` or newer. If tried to compile with previous version, it will give a compile time error.

This also changes the rust versions in the CI rust matrix to be compatible with this.

### DIFF
--- a/crates/libcontainer/Cargo.toml
+++ b/crates/libcontainer/Cargo.toml
@@ -1,8 +1,9 @@
 [package]
 name = "libcontainer"
-version = "0.0.1"
+version = "0.0.2"
 authors = ["youki team"]
 edition = "2021"
+rust-version = "1.58.1"
 description = "Library for container creation"
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


YJDoc2:
@utam0k PTAL

crates/libcontainer/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "libcontainer"
-version = "0.0.1"
+rust-version = "1.58.1"
+version = "0.0.2"
```
utam0k, line 4: Could you have all crates set to 0.0.2?

YJDoc2: Sure, I only set these two as I had added the rust-version field to these two, but I can add those, as well as fix some inconsistent styling and whitespace issues

YJDoc2: Hey I have updated all crate's version numbers to 0.0.2, please take a look now.

utam0k (approved):
Perfect! Thanks a lot.

Issue #629, opened 2022-01-21
Re-open /dev/null inside the container


adrianreber:
To use checkpoint/restore without `--console-socket` for testing it is necessary in runc/crun to start the container with `&> /dev/null < /dev/null`. This however results in using `/dev/null` from the host and not from the container. In this case stdin, stdout and stderr needs to be connected to `/dev/null` in the container. I did this for crun: https://github.com/containers/crun/commit/bbb1fa9f380b0606165222a2414cb7d4d45dd97f

Please also add this to youki.

PR #630, opened 2022-01-21
Use /dev/null inside of the container

adrianreber:
If stdin, stdout or stderr are redirected to /dev/null they need to be
pointed to /dev/null in the container. This is necessary for
checkpointing as CRIU cannot checkpoint processes using devices outside
of the container.

Fixes: #629


Furisto (commented):
This looks good, you just need to run cargo fmt --all to resolve the CI failure.

adrianreber:
Fixed formating

Furisto (approved):

Issue #628, opened 2022-01-21
Add --detach -d option


adrianreber:
To implement checkpoint/restore containers need to be started with `--detach`. This is currently missing.

The important part for checkpoint/restore is that `--detach` does `setsid()`. Hardcoding `setsid()` already helped me to be able to checkpoint a container.
```diff
diff --git a/crates/libcontainer/src/process/container_intermediate_process.rs b/crates/libcontainer/src/process/container_intermediate_process.rs
index b2a5e0d..a9e1f00 100644
--- a/crates/libcontainer/src/process/container_intermediate_process.rs
+++ b/crates/libcontainer/src/process/container_intermediate_process.rs
@@ -95,6 +95,8 @@ pub fn container_intermediate_process(
         intermediate_sender
             .close()
             .context("failed to close sender in the intermediate process")?;
+        nix::unistd::setsid()?;
+
         container_init_process(args, main_sender, init_receiver)
     })?;
     // Once we fork the container init process, the job for intermediate process
```
I am not sure that is the right location, but that works for me. Having it done correctly in `--detach` is probably much better than what I have done.

Furisto:
We already call setsid(), but only if we allocate a pseudo terminal. Just looking through the runc code i see that they always seem to call setsid regardless of whether --detach is specified or not. Crun does the [same](https://github.com/containers/crun/pull/267).

yihuaf:
I will take care of the detached and foreground mode. Although the core of this issue seems to be unrelated to `detached` anymore lol.

PR #632, opened 2022-01-22
Always call setsid

Furisto:
See #628


adrianreber:
Ah, if that is better then that works also for me.

utam0k (approved):
LGTM 💯 
https://github.com/containers/youki/issues/628#issuecomment-1019312585
I'm sorry for being late for this review.

PR #641, opened 2022-01-24
Initial checkpoint support

adrianreber:
This is the initial draft for checkpoint support. No restore support yet.

Tests are still missing and it relies on the not yet officially release Rust CRIU bindings. For now it is just a draft.

It depends on #632

I called the checkpoint command ´checkpointt' with two `t`s so that Podman does not think youki already fully supports checkpoint/restore. Once everything works I will rename it.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/641"></a>

### DIFF
--- a/.github/workflows/integration_tests_validation.yaml
+++ b/.github/workflows/integration_tests_validation.yaml
@@ -36,9 +36,14 @@ jobs:
         uses: Swatinem/rust-cache@v1
       - run: sudo apt-get -y update
       - run: sudo apt-get install -y pkg-config libsystemd-dev libdbus-glib-1-dev libelf-dev libseccomp-dev
+      - name: Install runc 1.1.0
+        run: |
+          wget -q https://github.com/opencontainers/runc/releases/download/v1.1.0/runc.amd64
+          sudo mv runc.amd64 /usr/bin/runc
+          sudo chmod 755 /usr/bin/runc
       - name: Build
         run: ./build.sh
       - name: Validate tests on runc
-        run: cd ./crates/integration_test && sudo ./tests.sh run runc
+        run: runc --version && cd ./crates/integration_test && sudo ./tests.sh run runc
       - name: Validate tests on youki
         run: cd ./crates/integration_test && sudo ./tests.sh run ../../youki
--- a/crates/integration_test/src/tests/lifecycle/checkpoint.rs
+++ b/crates/integration_test/src/tests/lifecycle/checkpoint.rs
@@ -1,0 +1,168 @@
+use super::get_result_from_output;
+use crate::utils::get_runtime_path;
+use crate::utils::test_utils::State;
+use crate::utils::{create_temp_dir, generate_uuid};
+use anyhow::anyhow;
+use std::path::Path;
+use std::process::{Command, Stdio};
+use test_framework::TestResult;
+
+// Simple function to figure out the PID of the first container process
+fn get_container_pid(project_path: &Path, id: &str) -> Result<i32, TestResult> {
+    let res_state = match Command::new(get_runtime_path())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .arg("--root")
+        .arg(project_path.join("runtime"))
+        .arg("state")
+        .arg(id)
+        .spawn()
+        .expect("failed to execute state command")
+        .wait_with_output()
+    {
+        Ok(o) => o,
+        Err(e) => {
+            return Err(TestResult::Failed(anyhow!(
+                "error getting container state {}",
+                e
+            )))
+        }
+    };
+    let stdout = match String::from_utf8(res_state.stdout) {
+        Ok(s) => s,
+        Err(e) => {
+            return Err(TestResult::Failed(anyhow!(
+                "failed to parse container stdout {}",
+                e
+            )))
+        }
+    };
+    let state: State = match serde_json::from_str(&stdout) {
+        Ok(v) => v,
+        Err(e) => {
+            return Err(TestResult::Failed(anyhow!(
+                "error in parsing state of container: stdout : {}, parse error : {}",
+                stdout,
+                e
+            )))
+        }
+    };
+
+    Ok(match state.pid {
+        Some(p) => p,
+        _ => -1,
+    })
+}
+
+// CRIU requires a minimal network setup in the network namespace
+fn setup_network_namespace(project_path: &Path, id: &str) -> Result<(), TestResult> {
+    let pid = get_container_pid(project_path, id)?;
+
+    if let Err(e) = Command::new("nsenter")
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .arg("-t")
+        .arg(format!("{}", pid))
+        .arg("-a")
+        .args(vec!["/bin/ip", "link", "set", "up", "dev", "lo"])
+        .spawn()
+        .expect("failed to exec ip")
+        .wait_with_output()
+    {
+        return Err(TestResult::Failed(anyhow!(
+            "error setting up network namespace {}",
+            e
+        )));
+    }
+
+    Ok(())
+}
+
+fn checkpoint(
+    project_path: &Path,
+    id: &str,
+    args: Vec<&str>,
+    work_path: Option<&str>,
+) -> TestResult {
+    if let Err(e) = setup_network_namespace(project_path, id) {
+        return e;
+    }
+
+    let temp_dir = match create_temp_dir(&generate_uuid()) {
+        Ok(td) => td,
+        Err(e) => {
+            return TestResult::Failed(anyhow::anyhow!(
+                "failed creating temporary directory {:?}",
+                e
+            ))
+        }
+    };
+    let checkpoint_dir = temp_dir.as_ref().join("checkpoint");
+    if let Err(e) = std::fs::create_dir(&checkpoint_dir) {
+        return TestResult::Failed(anyhow::anyhow!(
+            "failed creating checkpoint directory ({:?}): {}",
+            &checkpoint_dir,
+            e
+        ));
+    }
+
+    let additional_args = match work_path {
+        Some(wp) => vec!["--work-path", wp],
+        _ => Vec::new(),
+    };
+
+    let runtime_path = get_runtime_path();
+
+    let checkpoint = Command::new(runtime_path)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .arg("--root")
+        .arg(project_path.join("runtime"))
+        .arg(match runtime_path {
+            _ if runtime_path.ends_with("youki") => "checkpointt",
+            _ => "checkpoint",
+        })
+        .arg("--image-path")
+        .arg(&checkpoint_dir)
+        .args(additional_args)
+        .args(args)
+        .arg(id)
+        .spawn()
+        .expect("failed to execute checkpoint command")
+        .wait_with_output();
+
+    let result = get_result_from_output(checkpoint);
+    if let TestResult::Failed(_) = result {
+        return result;
+    }
+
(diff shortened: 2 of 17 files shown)


adrianreber:
Hmm, the test succeeds locally with runc. Need to figure out why it does not work in CI.

adrianreber:
Tests are still failing as I only tested on a cgroup v2 system. v1 needs some additional code.

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,124 @@
...
+                        }
+                        _ => (),
+                    }
+                }
+                _ => (),
+            }
+        }
+
+        let directory = std::fs::File::open(&opts.image_path)?;
+        criu.set_images_dir_fd(directory.as_raw_fd());
```
rst0git, line 73: Do we need to create the directory specified with `image_path`?

adrianreber:
Ready for review. All tests are successful.

utam0k:
Thanks for your PR. But I don't have the time to review this PR on my weekday. So I'll check on next holiday, please wait just a moment 🙏

adrianreber:
> Thanks for your PR. But I don't have the time to review this PR on my weekday. So I'll check on next holiday, please wait just a moment pray

No problem.

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,124 @@
...
+                            }
+                        }
+                        _ => (),
+                    }
+                }
+                _ => (),
+            }
+        }
+
+        let directory = std::fs::File::open(&opts.image_path)?;
```
Furisto, line 72:
```suggestion
        let directory = File::open(&opts.image_path)
            .with_context(|| format!("failed to open {:?}", opts.image_path))?;
```

crates/libcontainer/src/container/container_checkpoint.rs
Furisto: +1

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,124 @@
...
+                    Some(wp) => wp,
+                    _ => &opts.image_path,
+                }
+                .clone()
+                .into_os_string()
+                .into_string()
+                .unwrap(),
+                CRIU_CHECKPOINT_LOG_FILE
+            );
+        }
```
Furisto, line 115:
```suggestion
 if let Err(e) = criu.dump() {
            bail!(
                "checkpointing container {} failed with {:?}. Please check CRIU logfile {:}/{}",
                self.id(),
                e,
                opts.work_path
                    .as_ref()
                    .unwrap_or_else(|| &opts.image_path)
                    .display(),
                CRIU_CHECKPOINT_LOG_FILE
            );
        }
```

Furisto:
Did not manage to go through everything today, so feel free to wait until I am finished.

crates/libcontainer/src/container/container_checkpoint.rs
adrianreber: We do not need to create it, but we could if it does not exist. I think in Podman we create it to make sure it gets created with 0600. We could do it here. Or not. I have not strong opinion. If it does not exist the user will see an error. By creating the directory we could avoid that error. On the other hand, if it does not exist then the user has to make a conscious decision where the checkpoint directory is. I have a couple of checkpoint directories all around my development system because the directory is automatically created. Both options have advantages and disadvantages.

We could leave it for now and if users are complaining then it can still be added later. Or if one of the container engines needs it, it can also be added later.

adrianreber:
Changed as suggested by @Furisto

crates/integration_test/src/tests/lifecycle/checkpoint.rs
```diff
@@ -0,0 +1,168 @@
...
+) -> TestResult {
+    if let Err(e) = setup_network_namespace(project_path, id) {
+        return e;
+    }
+
+    let temp_dir = match create_temp_dir(&generate_uuid()) {
+        Ok(td) => td,
+        Err(e) => {
+            return TestResult::Failed(anyhow::anyhow!(
+                "Failed creating temporary directory {:?}",
```
utam0k, line 95: Can I ask you to make all error messages start lowercase letters?
```suggestion
                "failed creating temporary directory {:?}",
```

crates/youki/src/commands/checkpoint.rs
```diff
@@ -0,0 +1,24 @@
...
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
+
+use liboci_cli::Checkpoint;
+
+pub fn checkpoint(args: Checkpoint, root_path: PathBuf) -> Result<()> {
+    log::debug!("start checkpointing container {}", args.container_id);
+    let mut container = load_container(root_path, &args.container_id)?;
+    let opts = libcontainer::container::CheckpointOptions {
```
utam0k, line 12: Why don't you implement `impl CheckpointOption from Checkpoint`?

utam0k:
Great idea. I think we can represent the WIP by adding it to the feature table in the README.md, WDYT?
> I called the checkpoint command ´checkpointt' with two ts so that Podman does not think youki already fully supports checkpoint/restore. Once everything works I will rename it.

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,121 @@
...
+        let mounts = spec.mounts().clone();
+        for m in mounts.unwrap() {
+            match m.typ().as_deref() {
+                Some("bind") => {
+                    let dest = m
+                        .destination()
+                        .clone()
+                        .into_os_string()
+                        .into_string()
+                        .unwrap();
```
Furisto, line 48: We generally try to avoid using unwrap. 
```suggestion
                        .expect("failed to convert mount destination")
```

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,121 @@
...
+                    criu.set_external_mount(dest.clone(), dest);
+                }
+                Some("cgroup") => {
+                    match libcgroups::common::get_cgroup_setup()
+                        .context("failed to determine cgroup setup")?
+                    {
+                        // For v1 it is necessary to list all cgroup mounts as external mounts
+                        Legacy | Hybrid => {
+                            for mp in libcgroups::v1::util::list_subsystem_mount_points().unwrap() {
```
Furisto, line 57:
```suggestion
                            for mp in libcgroups::v1::util::list_subsystem_mount_points().context("failed to get subsystem mount points")? {
```

crates/liboci-cli/src/checkpoint.rs
```diff
@@ -0,0 +1,30 @@
+use clap::Parser;
+use std::path::PathBuf;
+
+/// Checkpoint a running container
+#[derive(Parser, Debug)]
+pub struct Checkpoint {
+    #[clap(forbid_empty_values = true, required = true)]
+    pub container_id: String,
+    /// allow external unix sockets
```
Furisto, line 9: These comments also show up in the help for youki, so to be consistent with the other commands they should start with a capital letter.

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,121 @@
...
+                                let cgroup_mount =
+                                    mp.clone().into_os_string().into_string().unwrap();
```
Furisto, line 59:
```suggestion
                                    mp.clone().into_os_string().into_string().expect("failed to convert mount point")
```

.github/workflows/integration_tests_validation.yaml
```diff
@@ -36,9 +36,14 @@ jobs:
         uses: Swatinem/rust-cache@v1
       - run: sudo apt-get -y update
       - run: sudo apt-get install -y pkg-config libsystemd-dev libdbus-glib-1-dev libelf-dev libseccomp-dev
+      - name: Install runc 1.1.0
```
Furisto, line 39: Why do we need to install this version of runc?

crates/youki/src/commands/checkpoint.rs
adrianreber: I don't understand this question. Maybe my rust knowledge is not good enough. Can you be more specific about what needs to change here?

adrianreber:
> Great idea. I think we can represent the WIP by adding it to the feature table in the README.md, WDYT?
> 
> > I called the checkpoint command ´checkpointt' with two ts so that Podman does not think youki already fully supports checkpoint/restore. Once everything works I will rename it.

In this PR or in another PR?

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -0,0 +1,121 @@
...
+        // It seems to be necessary to be defined outside of 'if' to
+        // keep the FD open until CRIU uses it.
+        let work_dir: std::fs::File;
+        if let Some(wp) = &opts.work_path {
+            work_dir = std::fs::File::open(wp)?;
+            criu.set_work_dir_fd(work_dir.as_raw_fd());
+        }
+
+        criu.set_log_file(CRIU_CHECKPOINT_LOG_FILE.to_string());
+        criu.set_log_level(4);
```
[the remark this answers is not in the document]
adrianreber: All runtimes using CRIU (runc and crun) currently hard-code the loglevel to CRIU's maximum. That is why I did the same. The main reason is, that there is a complete log file for the user if something fails. If it fails with a lesser log level it is sometimes difficult to understand why CRIU failed.

.github/workflows/integration_tests_validation.yaml
adrianreber: runc 1.1.0 changed how it handles `--work-dir` (or `--work-path`). Before 1.1.0 runc would create a directory called `criu-work` if the user does not specify a working directory. CRIU's default behaviour is to put the log and statistics files in the image path if work dir is not set. With runc 1.1.0 runc behaves like CRIU. That is also what I did for this PR and therefore the tests do not pass with older versions of runc.

utam0k:
> > Great idea. I think we can represent the WIP by adding it to the feature table in the README.md, WDYT?
> > > I called the checkpoint command ´checkpointt' with two ts so that Podman does not think youki already fully supports checkpoint/restore. Once everything works I will rename it.
> 
> In this PR or in another PR?

@adrianreber Can I ask for this PR?

crates/youki/src/commands/checkpoint.rs
utam0k: @adrianreber I'm sorry for my inadequate explanation. Why don't you implement this trait? I feel it would be more Rust-like if the conversion from `Checkpoint` to `CheckpointOptions` could be done using `from()`. WDYT?
https://doc.rust-lang.org/std/convert/trait.From.html

crates/libcontainer/src/container/container_checkpoint.rs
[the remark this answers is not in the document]
utam0k: I understand. makes sense.

crates/youki/src/commands/checkpoint.rs
adrianreber: Ah, okay. I tried it, but I am not sure where. If I try to implement it in the youki create I get:
```
impl doesn't use only types from inside the current crate
```
If I add it to the liboci_cli crate I need to access information from libcontainer and also the other way around. Not sure if that is desired or even possible.

Any recommendations where I need to add the corresponding implementation?

adrianreber:
Did most of the suggested changes (besides the one with `impl From<`) and also added a line to README.md.

utam0k (approved):
@adrianreber LGTM 💯 
Thanks!!!

PR #673, opened 2022-01-28
Remove duplication from commands execution in Integration tests

Szymongib:
Hey 👋 
I am looking into trying to add/extend integration tests and notice some duplication around commands execution code.

I did not find any reason why not to reuse some of the code from `test_utils.rs`, but in case I missed something and duplication was intentional let me know 🙂

### DIFF
--- a/crates/integration_test/src/tests/lifecycle/state.rs
+++ b/crates/integration_test/src/tests/lifecycle/state.rs
@@ -1,33 +1,24 @@
-use crate::utils::test_utils::get_state;
-use std::io;
+use crate::utils::get_state;
+use anyhow::{anyhow, Result};
 use std::path::Path;
 use test_framework::TestResult;
 
 pub fn state(project_path: &Path, id: &str) -> TestResult {
-    let res = get_state(id, project_path)
-        .expect("failed to execute state command")
-        .wait_with_output();
-    match res {
-        io::Result::Ok(output) => {
-            let stderr = String::from_utf8(output.stderr).unwrap();
-            let stdout = String::from_utf8(output.stdout).unwrap();
+    match get_state(id, project_path) {
+        Result::Ok((stdout, stderr)) => {
             if stderr.contains("Error") || stderr.contains("error") {
-                TestResult::Failed(anyhow::anyhow!(
-                    "Error :\nstdout : {}\nstderr : {}",
-                    stdout,
-                    stderr
-                ))
+                TestResult::Failed(anyhow!("Error :\nstdout : {}\nstderr : {}", stdout, stderr))
             } else {
                 // confirm that the status is stopped, as this is executed after the kill command
                 if !(stdout.contains(&format!(r#""id": "{}""#, id))
                     && stdout.contains(r#""status": "stopped""#))
                 {
-                    TestResult::Failed(anyhow::anyhow!("Expected state stopped, got : {}", stdout))
+                    TestResult::Failed(anyhow!("Expected state stopped, got : {}", stdout))
                 } else {
                     TestResult::Passed
                 }
             }
         }
-        io::Result::Err(e) => TestResult::Failed(anyhow::Error::new(e)),
+        Result::Err(e) => TestResult::Failed(e.context("failed to get container state")),
     }
 }
--- a/crates/integration_test/src/tests/lifecycle/state.rs
+++ b/crates/integration_test/src/tests/lifecycle/state.rs
@@ -1,18 +1,10 @@
-use crate::utils::get_runtime_path;
 use std::io;
 use std::path::Path;
-use std::process::{Command, Stdio};
 use test_framework::TestResult;
+use crate::utils::test_utils::get_state;
 
 pub fn state(project_path: &Path, id: &str) -> TestResult {
-    let res = Command::new(get_runtime_path())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .arg("--root")
-        .arg(project_path.join("runtime"))
-        .arg("state")
-        .arg(id)
-        .spawn()
+    let res = get_state(id, project_path)
         .expect("failed to execute state command")
         .wait_with_output();
     match res {
--- a/crates/integration_test/src/tests/lifecycle/state.rs
+++ b/crates/integration_test/src/tests/lifecycle/state.rs
@@ -1,7 +1,7 @@
+use crate::utils::test_utils::get_state;
 use std::io;
 use std::path::Path;
 use test_framework::TestResult;
-use crate::utils::test_utils::get_state;
 
 pub fn state(project_path: &Path, id: &str) -> TestResult {
     let res = get_state(id, project_path)
--- a/crates/integration_test/src/utils/support.rs
+++ b/crates/integration_test/src/utils/support.rs
@@ -78,7 +78,7 @@ pub fn prepare_bundle(id: &Uuid) -> Result<TempDir> {
 
     let mut spec = Spec::default();
     let mut process = Process::default();
-    process.set_args(Some(vec!["sleep".into(), "5".into()]));
+    process.set_args(Some(vec!["sleep".into(), "10".into()]));
     spec.set_process(Some(process));
     set_config(&temp_dir, &spec).unwrap();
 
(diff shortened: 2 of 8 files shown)


YJDoc2:
Hey, thanks a lot for your contribution !

So the reason behind the code duplication is that those tests were created before creation of `test_utils` and was supposed to test the individual commands (create, exec etc.), whereas the commands of `test_util`were created later to extract common commands and assume that these will succeed given correct input. 

That said, now looking at it, this change might be a good idea, as this will also test those functions themselves.

I haven't yet taken a detailed look at this, but I'll check it as soon as possible :)

Also, I request you not to make further addition/ changes in the integration tests (except this PR) for some times after this, because I'm currently working on changing the dir structure of tests, which might cause conflicts.

Thanks :)

Szymongib:
Thanks for your response @YJDoc2 

> Also, I request you not to make further addition/ changes in the integration tests (except this PR) for some times after this, because I'm currently working on changing the dir structure of tests, which might cause conflicts.

Sure! Let me know in case there is something I can help with 🙂

crates/integration_test/src/utils/test_utils.rs
```diff
@@ -189,7 +198,10 @@ pub fn test_inside_container(
     // assume that the create command was successful. If it wasn't we can catch that error
     // in the start_container, as we can not start a non-created container anyways
     std::thread::sleep(std::time::Duration::from_millis(1000));
-    match start_container(&id, &bundle).unwrap().wait_with_output() {
+    match start_container(&id.to_string(), &bundle)
```
YJDoc2, line 201: Hey can we extract the id_str as a variable, as it seems to be used in quite a few places later as well

crates/integration_test/src/utils/test_utils.rs
```diff
@@ -207,23 +219,29 @@ pub fn test_inside_container(
         ));
     }
 
-    let (out, err) = get_state(&id, &bundle).unwrap();
+    let (out, err) = get_state_output(&id.to_string(), &bundle).unwrap();
```
YJDoc2, line 222: id_str : here

crates/integration_test/src/utils/support.rs
```diff
@@ -78,7 +78,7 @@ pub fn prepare_bundle(id: &Uuid) -> Result<TempDir> {
 
     let mut spec = Spec::default();
     let mut process = Process::default();
-    process.set_args(Some(vec!["sleep".into(), "5".into()]));
+    process.set_args(Some(vec!["sleep".into(), "10".into()]));
```
YJDoc2, line 81: Was this changed for any specific reason? There is no issue with changing this, as the original value 5 was also arbitrary, but if there is a specific reason, can you either put a comment here or extract the &'static str as a const and name it appropriately?

crates/integration_test/src/tests/lifecycle/state.rs
```diff
@@ -1,18 +1,10 @@
...
 pub fn state(project_path: &Path, id: &str) -> TestResult {
-    let res = Command::new(get_runtime_path())
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .arg("--root")
-        .arg(project_path.join("runtime"))
-        .arg("state")
-        .arg(id)
-        .spawn()
+    let res = get_state(id, project_path)
```
YJDoc2, line 7: Given that you have added the function get_state_output, can we use it directly here? As when the state command is successful, we are converting the stdout and stderr to string, both here and in the get_state_output fn. So can we just call that and the work on the result returned by it?

crates/integration_test/src/utils/test_utils.rs
```diff
@@ -64,62 +64,65 @@ pub fn create_container<P: AsRef<Path>>(id: &Uuid, dir: P) -> Result<Child> {
...
-        .spawn()?
-        .wait_with_output()?;
+        .spawn()
+        .context("could not get container state")?;
+    Ok(res)
+}
+
+pub fn get_state_output<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
+    sleep(SLEEP_TIME);
+    let output = get_state(id, dir)?.wait_with_output()?;
```
YJDoc2, line 101: please add context to the wait_with_output error here as well, so we can know which step exactly gave error, spawning or waiting.

crates/integration_test/src/utils/test_utils.rs
```diff
@@ -64,62 +64,65 @@ pub fn create_container<P: AsRef<Path>>(id: &Uuid, dir: P) -> Result<Child> {
...
         .arg("state")
         .arg(id.to_string())
-        .spawn()?
-        .wait_with_output()?;
+        .spawn()
+        .context("could not get container state")?;
+    Ok(res)
+}
+
+pub fn get_state_output<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
```
YJDoc2, line 99: Hey, I don't have any better ideas either, but can we change names of these functions? `get_state` name indicates that it will return state of the container, rather than the child process running `runtime state` command. I don't have any better suggestions as of now either, but ideally, the function name should indicate if its gonna run the command or get you the o/p

One suggestion is that, as far as I can see, `get_state` is only used in the state test, and if we can replace it with `get_state output` as per my comment on it, then we can make the current `get_state` private function of the module and not expose it, maybe name it `run_state_command` or something, and rename current `get_state_output` to `get_state`.

YJDoc2, line 99: I don't have any better suggestions right now either, but can we change the names of these functions? `get_state` name indicates that it will return the state of the container rather than the child process running `runtime state` command.

One suggestion is as far as I can see, current `get_state` is only used in the state test, and as per my comment on it, if we can replace it with current `get_state_output` then we can make the current `get_state` a private function of this module, and rename to something like `run_state_command` and rename `get_state_output` to `get_state`

YJDoc2 (changes requested):
Hey, overall a great change, really nice that the duplication is removed. I have few nitpicks, please take a look at them :)

crates/integration_test/src/utils/support.rs
Szymongib: Ah sorry, I was going to highlight it here. The reason for bumping it to 10 was that lifecycle tests were failing on my machine (before any changes). The process was exiting before the test managed to kill it, so kill was failing on the stopped container.

I suppose this does not happen on CI so I can revert this change, however, it might be useful for other people running on slower machines.

crates/integration_test/src/utils/test_utils.rs
Szymongib: I have ended up removing the `get_state` function completely and just renamed `get_state_output`->`get_state` as after refactoring `lifecycle/state.rs` it was no longer needed.

crates/integration_test/src/utils/test_utils.rs
Szymongib: I have also changed the `create_container` function to take `&str` instead of `&Uuid` so that all those functions are consistent.

YJDoc2:
Hey, @Szymongib I'm not sure why but the CI of validating tests on youki has been running for about an hour, and tests on runc were passed in 39sec :thinking: :thinking:

Szymongib:
Hey @YJDoc2, yes noticed it too. I ran them a few times locally but they pass without any issues 🤔

YJDoc2:
I'm currently also trying to pull this PR and run locally, in case they run fine also on my end, maybe I'll try to stop and re-run the CI.

YJDoc2:
Hmm.. they seem to run fine on my system as well.. I'll try to stop and re-run the CI, although I'm not sure what is causing the hang-up in the current run :thinking:

crates/integration_test/src/utils/support.rs
YJDoc2: I guess that's fine, as I said there was no specific reason for keeping it 5 either ...

YJDoc2 (approved):
LGTM :+1:

YJDoc2:
Hey @Szymongib Thanks a lot for your contribution! 



>>    Also, I request you not to make further addition/ changes in the integration tests (except this PR) for some times after this, because I'm currently working on changing the dir structure of tests, which might cause conflicts.

>Sure! Let me know in case there is something I can help with :slightly_smiling_face:

If you would like to work more on the tests, check out the #361 , there we have listed all the tests we need to port from original OCI runtime tools to here in rust. If you would like to work on any one, comment and claim it, but as I said please wait till I'm done with the refactoring, otherwise there will be a lot of merge conflicts.

Thanks :)

PR #674, opened 2022-01-29
Ensure namespaces are entered in correct order

Furisto:
During exec namespaces were entered in arbitrary order. This leads to a high probability of the exec call failing because once a mount namespace is entered which contains /proc mounted by a process from a different pid namespace, we are unable to locate the rest of the namespaces we want to enter.


utam0k (approved):
Looks good!

PR #683, opened 2022-02-06
Use buffered io for reading state file

Furisto:
Deserializing the state file triggers dozens of small reads. With this only one read is performed.


utam0k:
!github easy-benchmark

utam0k (approved):
LGTM

PR #702, opened 2022-02-11
Add unit tests to libcgroups/src/v2/devices/emulator.rs

cr0ax:
I want to contribute more tests in this area.

This seemed like a good place to start.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1135,9 +1135,9 @@ checksum = "308cc39be01b73d0d18f82a0e7b2a3df85245f84af96fdddc5d202d27e47b86a"
 
 [[package]]
 name = "memmap2"
-version = "0.5.2"
+version = "0.5.3"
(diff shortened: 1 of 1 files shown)


crates/libcgroups/src/v2/devices/emulator.rs
```diff
@@ -54,4 +54,67 @@ impl Emulator {
...
+        // assert
+        assert_eq!(emulator.rules.len(), 0);
+        assert_eq!(emulator.default_allow, false);
+    }
+
+    #[test]
+    fn test_add_some_rule() {
+        // arrange
+        let mut emulator = Emulator::with_default_allow(false);
+        const PERMISSION: &str = "PERMISSION";
```
cr0ax, line 107: Is there a better way? I didn't look too far into what this param is supposed to be, at this point I'm just trying to successfully exercise code paths.

PR #703, opened 2022-02-11
Add tests to libcgroups/src/v2/devices/emulator.rs

cr0ax:
I want to contribute more tests in this area.

This seemed like a good place to start.

### DIFF
--- a/crates/libcgroups/src/v2/devices/emulator.rs
+++ b/crates/libcgroups/src/v2/devices/emulator.rs
@@ -54,4 +54,69 @@ impl Emulator {
     }
 }
 
-// FIXME: add some tests
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use oci_spec::runtime::LinuxDeviceCgroupBuilder;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


crates/libcgroups/src/v2/devices/emulator.rs
```diff
@@ -54,4 +54,69 @@ impl Emulator {
...
+        // assert
+        assert_eq!(emulator.rules.len(), 0);
+        assert!(!emulator.default_allow);
+    }
+
+    #[test]
+    fn test_add_some_rule() {
+        // arrange
+        let mut emulator = Emulator::with_default_allow(false);
+        const PERMISSION: &str = "PERMISSION";
```
cr0ax, line 107: Is there a better way to do this?

utam0k: How do you want to do this? Since str is static, I don't see the need to use const.
```suggestion
        let permission: &str = "PERMISSION";
```

cr0ax:
Closed in favor of #704

PR #704, opened 2022-02-11
Add tests to libcgroups/src/v2/devices/emulator.rs

cr0ax:
Repeat of https://github.com/containers/youki/pull/703

Branched off an older commit on a hunch about what was going wrong with the codecov bot before


utam0k (approved):
Looks good 💯

PR #706, opened 2022-02-13
Add tests to crates/libcgroups/src/v2/devices/controller.rs

cr0ax:
In order to add these tests, I added `mockall` to the dev dependencies.

There are some things that could be cleaner, eg using mockall's `double`
or `cfg_if` but I did not want to bloat the release dependencies with extra crates.

This required some changes to the v2/devices/bpf module, introducing an
intermediate module `prog` to wrap the various `prog_*` functions inside.
Those functions were then renamed to remove the `prog_` prefix.
The addition of this intermediate module allowed us to use the `automock`
macro to automatically generate mocked functions. This is a useful
testing pattern that can be used in the future on the bpf module itself.

Signed-off-by: Christian Burke <cr0ax64@gmail.com>

### DIFF
--- a/crates/libcgroups/src/v2/devices/bpf.rs
+++ b/crates/libcgroups/src/v2/devices/bpf.rs
@@ -1,117 +1,124 @@
-use anyhow::{bail, Result};
-use std::os::unix::io::RawFd;
-use std::ptr;
-
-// FIXME: add tests
-
-pub fn prog_load(license: &str, insns: &[u8]) -> Result<RawFd> {
-    let insns_cnt = insns.len() / std::mem::size_of::<libbpf_sys::bpf_insn>();
-    let insns = insns as *const _ as *const libbpf_sys::bpf_insn;
-
-    let prog_fd = unsafe {
-        libbpf_sys::bpf_load_program(
-            libbpf_sys::BPF_PROG_TYPE_CGROUP_DEVICE,
-            insns,
-            insns_cnt as u64,
-            license as *const _ as *const i8,
-            0,
-            ptr::null_mut::<i8>(),
-            0,
-        )
-    };
-
-    if prog_fd < 0 {
-        return Err(errno::errno().into());
-    }
-    Ok(prog_fd)
-}
-
+#[derive(Clone)]
 pub struct ProgramInfo {
     pub id: u32,
     pub fd: i32,
 }
 
-pub fn prog_query(cgroup_fd: RawFd) -> Result<Vec<ProgramInfo>> {
-    let mut prog_ids: Vec<u32> = vec![0_u32; 64];
-    let mut attach_flags = 0_u32;
-    for _ in 0..10 {
-        let mut prog_cnt = prog_ids.len() as u32;
-        let ret = unsafe {
-            libbpf_sys::bpf_prog_query(
-                cgroup_fd,
-                libbpf_sys::BPF_CGROUP_DEVICE,
+#[cfg_attr(test, automock)]
+pub mod prog {
+    use super::ProgramInfo;
+    use anyhow::{bail, Result};
+    use std::os::unix::io::RawFd;
+    use std::ptr;
+
+    pub fn load(license: &str, insns: &[u8]) -> Result<RawFd> {
+        let insns_cnt = insns.len() / std::mem::size_of::<libbpf_sys::bpf_insn>();
+        let insns = insns as *const _ as *const libbpf_sys::bpf_insn;
+
+        let prog_fd = unsafe {
+            libbpf_sys::bpf_load_program(
+                libbpf_sys::BPF_PROG_TYPE_CGROUP_DEVICE,
+                insns,
+                insns_cnt as u64,
+                license as *const _ as *const i8,
                 0,
-                &mut attach_flags,
-                &prog_ids[0] as *const u32 as *mut u32,
-                &mut prog_cnt,
+                ptr::null_mut::<i8>(),
+                0,
             )
         };
-        if ret != 0 {
-            let err = errno::errno();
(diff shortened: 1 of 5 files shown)


crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -1,117 +1,124 @@
...
-                cgroup_fd,
-                libbpf_sys::BPF_CGROUP_DEVICE,
+#[cfg_attr(test, automock)]
+pub mod prog {
+    use super::ProgramInfo;
+    use anyhow::{bail, Result};
+    use std::os::unix::io::RawFd;
+    use std::ptr;
+
+    pub fn load(license: &str, insns: &[u8]) -> Result<RawFd> {
```
cr0ax, line 14: changed names to remove stuttering eg `bpf::prog::prog_load` -> `bpf::prog::load`

crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -1,117 +1,124 @@
...
-                // allocate more space and try again
-                prog_ids.resize(prog_cnt as usize, 0);
-                continue;
+        if prog_fd < 0 {
+            return Err(errno::errno().into());
+        }
+        Ok(prog_fd)
+    }
+
+    /// Given a fd for a cgroup, collect the programs associated with it
```
cr0ax, line 36: I wrote some comments as I was going around, let me know if they are wrong or otherwise inappropriate and I will remove them.

crates/libcgroups/src/v2/devices/controller.rs
```diff
@@ -11,6 +11,12 @@ use oci_spec::runtime::LinuxDeviceCgroup;
 use crate::common::{default_allow_devices, default_devices, ControllerOpt};
 use crate::v2::controller::Controller;
 
+#[cfg(test)]
+use bpf::mock_prog::{attach, bump_memlock_rlimit, detach2, load, query};
+
+#[cfg(not(test))]
+use bpf::prog::{attach, bump_memlock_rlimit, detach2, load, query};
```
cr0ax, line 18: this can be cleaned up by mockall's `double`, but would require it to be in the release dependencies, which I wanted to avoid

crates/libcgroups/src/v2/devices/controller.rs
```diff
@@ -57,8 +63,8 @@ impl Devices {
 
         // Increase `ulimit -l` limit to avoid BPF_PROG_LOAD error (#2167).
         // This limit is not inherited into the container.
-        bpf::bump_memlock_rlimit()?;
```
cr0ax, line 60: I actually liked the `bpf::` namespacing around here and think I could reintroduce it by adding `as bpf` above.

cr0ax (commented):
Reccomended to look at it using these params: https://github.com/containers/youki/pull/706/files?diff=unified&w=1 because of the whitespace changes.

crates/libcgroups/src/v2/devices/controller.rs
cr0ax: I have reintroduced this

cr0ax:
@utam0k what do you think about this strategy?
I want to continue using it in this area, in particular for the syscall wrappers inside the `bpf::prog` module.
I'm thinking of adding it onto this PR

cr0ax:
Expanded in https://github.com/containers/youki/pull/711

utam0k:
@cr0ax Hi, thanks for your PR. Could you confirm that compiling the release flag doesn't change the size of the binary?

cr0ax:
@utam0k I compiled once each for

* `b162cf58f559d948792e7736fe0ad367d90c0cd1` and
* `devices_tests`

using `./build.sh --release`

I got the same result each time:

```
du -s youki 
5436	youki
```

cr0ax:
Added 1 last thing here, made these tests `serial`, if they use the same mocks it seems those mock contexts are shared.

utam0k:
> Added 1 last thing here, made these tests `serial`, if they use the same mocks it seems those mock contexts are shared.

This is a little hard for us. The system call mocks the following, can we consider this one?
https://github.com/containers/youki/blob/main/crates/libcontainer/src/syscall/test.rs

cr0ax:
@utam0k after some back-and-forth with you and some testing of the `mockall` library, I have changed the `serial` annotations to be parameterized with `bpf`, to limit the scope of the `serial` tests.

This is based on some experiments where we found that static methods share a context, whether in a module or a trait.

The recommended approach from `mockall` is to create a mutex and share context. Adding the `serial(bpf)` attribute is effectively the same strategy, and simpler.

crates/libcgroups/src/v2/devices/controller.rs
```diff
@@ -73,20 +79,99 @@ impl Devices {
...
+        query.expect().once().returning(|_| Ok(vec![]));
+        attach.expect().once().returning(|_, _| Ok(()));
+        detach2.expect().never();
+
+        // act
+        Devices::apply_devices(&tmp, &Some(vec![a_type])).expect("Could not apply devices");
+    }
+
+    #[test]
+    #[serial(bpf)] // mock contexts are shared
```
cr0ax, line 143: [From the docs for serial](https://docs.rs/serial_test/latest/serial_test/attr.serial.html)

> Multiple tests with the serial attribute are guaranteed to be executed in serial. Ordering of the tests is not guaranteed however. If you want different subsets of tests to be serialised with each other, but not depend on other subsets, you can add an argument to serial, and all calls with identical arguments will be called in serial.

cr0ax (commented):
Added a note about use of serial

utam0k (approved):
@cr0ax LGTM 💯

PR #711, opened 2022-02-14
Adds more mocked tests

cr0ax:
A follow-up to https://github.com/containers/youki/pull/706

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/711"></a>

### DIFF
--- a/crates/libcgroups/src/v2/devices/bpf.rs
+++ b/crates/libcgroups/src/v2/devices/bpf.rs
@@ -142,15 +142,15 @@ pub mod prog {
 
 #[cfg(test)]
 mod tests {
-    use serial_test::serial;
-
     use super::prog;
     use crate::v2::devices::mocks::{mock_libbpf_sys, mock_libc};
     use errno::{set_errno, Errno};
     use libc::{ENOSPC, ENOSYS};
 
+    use serial_test::serial;
+
     #[test]
-    #[serial] // mock contexts are shared
+    #[serial(libbpf_sys)] // mock contexts are shared
     fn test_bpf_load() {
         // eBPF uses 64-bit instructions
         let instruction_zero: &[u8] = &[0x0, 0x0, 0x0, 0x0];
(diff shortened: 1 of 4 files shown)


cr0ax:
I think this is the better PR, I will try to annotate it in the near future

crates/libcgroups/src/v2/devices/mocks.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    pub fn bpf_prog_attach(
+        _prog_fd: ::std::os::raw::c_int,
+        _attachable_fd: ::std::os::raw::c_int,
+        _type_: libbpf_sys::bpf_attach_type,
+        _flags: ::std::os::raw::c_uint,
+    ) -> ::std::os::raw::c_int {
+        unimplemented!();
+    }
+}
```
cr0ax, line 60: I needed a stubbed module to apply the mock to.

Do we already do something like this elsewhere?

crates/libcgroups/src/v2/devices/mod.rs
```diff
@@ -3,4 +3,9 @@ pub mod controller;
 pub mod emulator;
 pub mod program;
 
+#[cfg(test)]
+#[allow(clippy::too_many_arguments)]
+#[cfg_attr(coverage, feature(no_coverage))]
+pub mod mocks;
```
cr0ax, line 9: only used in tests, also libc functions trigger the `too_many_arguments` clippy warning

crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -11,13 +11,33 @@ pub mod prog {
...
+    };
+    #[cfg(not(test))]
+    use libbpf_sys::{
+        bpf_load_program, bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_query,
+    };
 
+    pub fn load(license: &str, insns: &[u8]) -> Result<RawFd> {
+        let insns_cnt = insns.len() / std::mem::size_of::<bpf_insn>();
+        let insns = insns as *const _ as *const bpf_insn;
+        #[allow(unused_unsafe)]
```
cr0ax, line 37: I don't like this one, but the mocked stuff isn't actually unsafe.

We could probably make it look `unsafe` though...

crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -11,13 +11,33 @@ pub mod prog {
...
+
+    // TODO: consider use of #[mockall_double]
+    #[cfg(test)]
+    use crate::v2::devices::mocks::mock_libbpf_sys::{
+        bpf_load_program, bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_query,
+    };
+    #[cfg(not(test))]
+    use libbpf_sys::{
+        bpf_load_program, bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_query,
+    };
```
cr0ax, line 32: For the above, we are importing mocked versions or not based on if we are in a test environment

cr0ax (commented):
Added some annotations for things I want to call out for discussion.

utam0k (approved):
LGTM

PR #722, opened 2022-02-20
Use cgroup.kill file if available

Furisto:
Kernel 5.14 has introduced a cgroup.kill file to reliably kill processes in a cgroup.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/722"></a>


utam0k (approved):
LGTM.
I've checked non-root users. There isn't some problem with a non-root user.

Issue #721, opened 2022-02-19
Requesting Temporary ignoring of default.go oci-runtime-validation test


YJDoc2:
After two weeks of testing and a lot of digging of git commit history, I think I have finally found what has been causing CI failure for #694 's OCI validation test (there are some other failing tests, but this is not related to them).

The surprising reason behind that test failing is a knock-on effect of the way -
1. Our CI is set up (I have partial blame in this :sweat: )
2. We recently adding these capabilities
3. runtime-tools not validating there tests on runc (?)

If you see in [commet](https://github.com/containers/youki/pull/694#issuecomment-1031352614) I have mentioned that the CI is faioing on __youki__ for these three capabilities, although if you check the integration test CI of any other PR there is no issue. The reason behind this is :
1. We added caching mechanism to the CI for integration tests on 7th Sep [Here](https://github.com/containers/youki/commit/964c4fca0a6ef3287634e0397fc7d38164caa650). This caches the build artifacts of the runtime-tests builds, which is the compiled tests, and has a very good reason of doing so, as without it the CI takes 7-8 mins and with caching it takes about 3-4 mins. The downside is that the tests are not built each time, and we always use the version that was cached on 7th Sep.
2. runtime-tools itself updated one of their dependencies go-capabilities on 26th of October [here](https://github.com/opencontainers/runtime-tools/commit/8e1a3b587864ef065f7eff09e54933720d950117) 
3. The go-capabilities itself was updated on Aug 5 [here](https://github.com/syndtr/gocapability/commit/9c89fbc73e2eec76b2e366ffa3dfa05c4d0cec6d) to include the three new capabilities, and thus the runtime tools itself as well inherited these updates.
4. runc released version 1.1 [here](https://github.com/opencontainers/runc/releases) on 18th Jan , which included the support for these capabilities. I think we also have only recently implemented these as well.
5. As our CI was using the cached version of tests, they did not check for these and thus tests were passed. But when I moved the git module in the refactor, the cache was invalidated, and youki was tested with new version and failed the tests.
6. The probable reason is that the CI runner's kernel might not be updated or compiled with correct flags or has permissions to set these capabilities, so runc cannot use them OR the oci-tests has bug in it, (as far as I checked the do not validate the tests on runc or some other runtime)

It is too late here, so I'm not sure what approach might be best, my suggestion is to either comment the default.go test and open an issue if not yet opened on oci-runtime-tools

[...]

PR #727, opened 2022-02-21
Remove caching of OCI tests in CI

YJDoc2:
This removes caching of OCI tests in CI. This is regarding to #721 , and should work as a temporary solution to it. This will increase the CI time, but will always use the latest commit version. There might be some work-around such that we can cache and invalidate cache when new commits are added to OCi repo, but this does not address it, it can be done in a separate PR.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/727"></a>

### DIFF
--- a/crates/libcontainer/src/syscall/linux.rs
+++ b/crates/libcontainer/src/syscall/linux.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use std::{any::Any, mem, path::Path, ptr};
 
 use anyhow::{anyhow, bail, Result};
-use caps::{CapSet, Capability, CapsHashSet};
+use caps::{CapSet, CapsHashSet};
 use libc::{c_char, uid_t};
 use nix::{
     errno::Errno,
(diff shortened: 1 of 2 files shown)


YJDoc2:
@utam0k This should potentially solve #721 for now. This might result in slightly bad CI times (integration tests needing 5-7 mins) but that is preferable than having broken/outdated integration tests
Please take a look :)

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -132,15 +132,7 @@ impl Syscall for LinuxSyscall {
...
-                    match c {
-                        Capability::CAP_PERFMON
-                        | Capability::CAP_CHECKPOINT_RESTORE
-                        | Capability::CAP_BPF => {
-                            log::warn!("{:?} is not supported.", c);
-                            continue;
-                        }
-                        _ => caps::drop(None, CapSet::Bounding, *c)?,
-                    }
+                    caps::drop(None, CapSet::Bounding, *c)?
```
utam0k, line 135: This is fine for now, but I'd like to consider checking the linux version

utam0k (approved):

utam0k:
@YJDoc2 Thanks for your hard work!!! 💯

PR #740, opened 2022-02-26
Handle relative paths

### LINKED ISSUE #720: Failed to create container when using relative root path
When using youki with relative root path creating container fails. From looking at a code it is probably due to not converting path to absolute path and using `chdir`.

Steps to reproduce:
```
mkdir -p ./some/dir
./youki --root ./some/dir create -b [BUNDLE] some-id
```
Error: `Error: No such file or directory`

I can look into it and try to fix it.

Szymongib:
This PR improves the handling of relative paths for `root_path` and `pid_file`. I decided to create the specified directory early in `determine_root_path` if it does not exist, as this seems to be the only way to `canonicalize` path prior to creating container dir (path needs to exist otherwise `canonicalize` will fail). This results in the directory being created for other commands like `list` or `delete` also which might not be ideal, but `runc` seems to work this way either so decided to stick with it.

Also not sure about canonicalizing those paths in `ContainerBuilder` as it disrupts the interface slightly by introducing `Result<Self>` as a return type to some methods, however, I did not find a better place for that so if anyone has some suggestions please let me know.

Additionally, I have added context to a few errors that made it easier to initially troubleshoot.

Fixes #720

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/740"></a>

### DIFF
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -1,7 +1,9 @@
 use crate::syscall::Syscall;
+use anyhow::{Context, Result};
 use std::path::PathBuf;
 
 use super::{init_builder::InitContainerBuilder, tenant_builder::TenantContainerBuilder};
+
 pub struct ContainerBuilder<'a> {
     /// Id of the container
     pub(super) container_id: String,
@@ -28,8 +30,8 @@ pub struct ContainerBuilder<'a> {
 /// use libcontainer::syscall::syscall::create_syscall;
 ///
 /// ContainerBuilder::new("74f1a4cb3801".to_owned(), create_syscall().as_ref())
-/// .with_root_path("/run/containers/youki")
-/// .with_pid_file(Some("/var/run/docker.pid"))
+/// .with_root_path("/run/containers/youki").expect("invalid root path")
+/// .with_pid_file(Some("/var/run/docker.pid")).expect("invalid pid file")
 /// .with_console_socket(Some("/var/run/docker/sock.tty"))
 /// .as_init("/var/run/docker/bundle")
 /// .build();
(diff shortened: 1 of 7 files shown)


crates/libcontainer/src/container/builder.rs
```diff
@@ -101,11 +103,14 @@ impl<'a> ContainerBuilder<'a> {
...
+    /// .with_root_path("/run/containers/youki").expect("invalid root path");
     /// ```
-    pub fn with_root_path<P: Into<PathBuf>>(mut self, path: P) -> Self {
-        self.root_path = path.into();
-        self
+    pub fn with_root_path<P: Into<PathBuf>>(mut self, path: P) -> Result<Self> {
+        self.root_path = path
+            .into()
+            .canonicalize()
+            .context("failed to canonicalize root path")?;
```
Furisto, line 112:
```suggestion
            .with_context(||format!("failed to canonicalize root path {:?}", path))?;
```

crates/libcontainer/src/container/builder.rs
```diff
@@ -117,11 +122,19 @@ impl<'a> ContainerBuilder<'a> {
...
     /// ```
-    pub fn with_pid_file<P: Into<PathBuf>>(mut self, path: Option<P>) -> Self {
-        self.pid_file = path.map(|p| p.into());
-        self
+    pub fn with_pid_file<P: Into<PathBuf>>(mut self, path: Option<P>) -> Result<Self> {
+        self.pid_file = if let Some(p) = path {
+            Some(
+                p.into()
+                    .canonicalize()
+                    .context("failed canonicalize pid file path")?,
```
Furisto, line 132:
```suggestion
                    .with_context(|| format!("failed canonicalize pid file path"))?,
```

crates/libcontainer/src/container/init_builder.rs
```diff
@@ -39,10 +39,14 @@ impl<'a> InitContainerBuilder<'a> {
...
-
-        let mut container = self.create_container_state(&container_dir)?;
+        let spec = self.load_spec().context("failed to load spec")?;
+        let container_dir = self
+            .create_container_dir()
+            .context("failed to create container dir")?;
+
+        let mut container = self
+            .create_container_state(&container_dir)
+            .context("failed to create container state")?;
```
Furisto, line 49: 👍

crates/libcontainer/src/container/builder.rs
Szymongib: Needed to adjust it slightly as `Into<PathBuf>` is not `Debug`.

crates/libcontainer/src/container/builder.rs
Szymongib: Did you mean to add the `path` here too? Adjusted it similarly to the other one.

Furisto (approved):

Furisto:
@Szymongib 💯

PR #773, opened 2022-03-11
Update the install instructions for Fedora

RMPR:
For some reason, when installing the build tools, the static glibc was
not installed.

Fixes https://github.com/containers/youki/issues/772

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/773"></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ $ sudo dnf install   \
       systemd-devel  \
       dbus-devel     \
       elfutils-libelf-devel \
-      libseccomp-devel
+      libseccomp-devel \
+      glibc-static
 ```
 
 ## Build


utam0k:
@RMPR First of all, thanks for your PR. I wonder if this is a problem only for youki. Can you build other rust projects?

RMPR:
Hi, I can build other Rust projects without any problem. [Here](https://asciinema.org/a/476166) for example, I can build the rustlings without this package installed.

utam0k:
What is your build target with? Why don't you try with `musl`?

RMPR:
I use whatever default build target is configured for Fedora, really don't know much about it. But when I try to use musl on the rustling repository I get this:
```shell
❯ cargo build --target x86_64-unknown-linux-musl
   Compiling cfg-if v1.0.0
   Compiling lazy_static v1.4.0
   Compiling regex-syntax v0.6.23
   Compiling scopeguard v1.1.0
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-linux-musl`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `scopeguard` due to previous error
warning: build failed, waiting for other jobs to finish...
error[E0463]: can't find crate for `std`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-linux-musl`

error: build failed
```

utam0k:
@RMPR Can I ask you to try this command and again?
```console
$ rustup target add x86_64-unknown-linux-musl
```

RMPR:
There is no rustup in my distros repository. [It seems](https://stackoverflow.com/questions/59385341/crosscompiling-rust-from-fedora-linux-host-to-windows-target-does-not-find-depen) the process of crosscompiling is a bit more involved, I can't do that right now, but as soon as I complete that I come back to you.

RMPR:
@Furisto you are right, using rustup it does work, but @utam0k I still think this is valuable because not everyone is comfortable not using packages from the official repositories. You can easily reproduce the error I highlighted on a fresh Fedora install:
```
$ sudo dnf install   \
      pkg-config     \
      systemd-devel  \
      dbus-devel     \
      elfutils-libelf-devel \
      libseccomp-devel \ 
      rust \
      cargo
```
Then if you `make build` it will fail. 

If you install `glibc-static` and you re-run `make build` the package builds without any issue.

utam0k:
@RMPR Great perspective. Thanks. But I don't go for how to install Rust at README. The README clearly states that Rust is required. Also, we have no control over which version of Rust is installed by the package manager.
https://github.com/containers/youki#requires

RMPR:
> But I don't go for how to install Rust at README

I don't either. This is solely for `glibc-static` I found it while trying to fiddle a bit with the project. Maybe you have other reasons I am not yet aware of why you don't want to use it? I look at it this way, in the Ubuntu section you prompt to install build-essentials, I see glibc-static as a package of the same type, helping with the build.

PR #774, opened 2022-03-11
Introduce the root Makefile.

utam0k:


scripts/build.sh
```diff
@@ -1,14 +1,17 @@
...
-mv ./youki_bin ${ROOT}/scripts/youki
+mv ./youki_bin ${OUTPUT}/youki
 
-cd ${ROOT}/integration_tests/rust-integration-tests
+cd ${ROOT}/tests/rust-integration-tests
 make FLAG=--release all
-mv ./runtimetest_bin ${ROOT}/scripts/runtimetest
-mv ./integration_test_bin ${ROOT}/scripts/integration_test
+mv ./runtimetest_bin ${OUTPUT}/runtimetest
+mv ./integration_test_bin ${OUTPUT}/integration_test
```
YJDoc2, line 16: :+1:

YJDoc2, line 16: :+1:

YJDoc2:
Hey, everything looks great! Thanks for improving this :)
One nitpick : I feel that the dir for integration tests should be named `integration_tests` or something similar, because just `tests` feels like unit or such tests, but if you feel that is better, it's ok :+1:

YJDoc2:
Hey, everything looks great! Thanks for improving this :)
One nitpick : I feel that the dir for integration tests should be named `integration_tests` or something similar, because just `tests` feels like unit or such tests, but if you feel that is better, it's ok :+1:

utam0k:
@YJDoc2 Thanks for your help.
I was wondering too what the name integration_tests should be. The problem was that it would match the name of the binary. Well, I don't think it would be that misleading.

YJDoc2:
Maybe then we can keep the root folder name as `integration_tests` but change the `rust-integration-tests/integration_tests` to `rust-integration-tests/tests` ? For me personally the root folder name `integration_tests` makes sense, as it currently contains oci and rust tests, and eventually should also contains dockerd and others as well.

utam0k:
> Maybe then we can keep the root folder name as `integration_tests` but change the `rust-integration-tests/integration_tests` to `rust-integration-tests/tests` ? For me personally the root folder name `integration_tests` makes sense, as it currently contains oci and rust tests, and eventually should also contains dockerd and others as well.

@YJDoc2 
In that case, the name of the binary would match the name of the directory. Do you have an idea to prevent the duplicate name?
By the way, what about e2e and the like?

utam0k:
@YJDoc2 PTAL

scripts/build.sh
```diff
@@ -1,14 +1,46 @@
...
+        o) output=${OPTARG}
+            ;;
+        r) VERSION=release
+            ;;
+        h) usage_exit
+            ;;
+        \?) usage_exit
+            ;;
+    esac
+done
```
YJDoc2, line 26: :100: :+1:

YJDoc2:
Well, I think currently this PR is a good solution. In case we come up with a better one, we can update it later.

> In that case, the name of the binary would match the name of the directory. Do you have an idea to prevent the duplicate name?

I'm sorry, I didn't get you :sweat_smile: 
> By the way, what about e2e and the like?

This looks good, just resolve the conflicts, then it can be merged :+1:

utam0k:
`e2e` means an end-to-end test.
> I'm sorry, I didn't get you 😅

YJDoc2 (approved):
LGTM :+1:
Thanks!

PR #775, opened 2022-03-13
Fix a comment explaining that `seccom_rule_add` requires multiple args to be broken into multiple rules.

yihuaf:
Credit to comments from here. https://github.com/containers/youki/commit/052ba25769500e2a2a1a0a3f9be41a80b7d0090d#r68527694


utam0k (approved):
@yihuaf Thanks!

Issue #771, opened 2022-03-09
Resolve deprecation warnings for Clap


YJDoc2:
Currently some of the clap options we use give deprecation warnings, with suggested changes. Perform those changes.

YJDoc2:
Resolved with #798

utam0k:
@YJDoc2 Thanks!!!

PR #798, opened 2022-03-27
Resolve deprecation warnings from clap

YJDoc2:
This resolves some of the clap deprecation warnings. This is part of resolving #771 .

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/798"></a>

### DIFF
--- a/.github/workflows/integration_tests_validation.yaml
+++ b/.github/workflows/integration_tests_validation.yaml
@@ -42,7 +42,9 @@ jobs:
           wget -q https://github.com/opencontainers/runc/releases/download/v1.1.0/runc.amd64
           sudo mv runc.amd64 /usr/bin/runc
           sudo chmod 755 /usr/bin/runc
+      - name: Build youki and tests
+        run: ./scripts/build.sh -o "$(git rev-parse --show-toplevel)"/scripts -r
       - name: Validate tests on runc
-        run: cd scripts && ./rust_integration_tests.sh runc
+        run: make validate-rust-tests
       - name: Validate tests on youki
         run: make integration-test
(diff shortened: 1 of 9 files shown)


YJDoc2:
This does not resolve all the deprecation warnings, there are two which still remain :
>warning: use of deprecated variant `clap::AppSettings::AllowHyphenValues`: Replaced with `Command::allow_hyphen_values` and `Arg::is_allow_hyphen_values_set`
  --> liboci-cli/src/lib.rs:57:41
   |
57 |     #[clap(setting = clap::AppSettings::AllowHyphenValues)]
   |                                         ^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

>warning: use of deprecated variant `clap::ArgSettings::Last`: Replaced with `Arg::last` and `Arg::is_last_set`
  --> liboci-cli/src/ps.rs:12:41
   |
12 |     #[clap(setting = clap::ArgSettings::Last)]
   |                                         ^^^^

Issue with resolving these, is that in the clap docs, both of them only have documentation for builder API, not the derive API which we use for youki. Thus in worst case, two subcommands which use these two might need to be refactored and ported to use the builder API instead of derive, and might also require some additional checks.

Furisto (commented):
Lets raise an issue with clap for the missing derive support and silence the warnings in the meanwhile.

YJDoc2:
I spent some time seeing if there was any workaround, but didn't find any. I have opened a discussion at clap : https://github.com/clap-rs/clap/discussions/3584

Also we don't have to silence the warnings immediately, as the CI will check them only if files in crates/* are changed. If this gets resolved before such a change occurs, then we won't need to change the CI, else we can change it temporarily in the PR where this issue occurs.

utam0k (approved):
💯

PR #820, opened 2022-04-04
move crates/youki to root

utam0k:
Signed-off-by: utam0k <k0ma@utam0k.jp>

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/820"></a>


Furisto:
What is the reasoning behind this change?

utam0k:
The youki directory is not expected to be used as a crate. Only this one should be treated as a bin, unlike other crates, so I brought it to the root. Also, the absence of `Cargo.toml` in the root directory may seem a bit strange for a Rust project. It is hard to know where to start looking at the code for first contributor.

YJDoc2:
Hey, I feel that the current crates/ setup is better. That way it separates the crates from tests and creates clear separation. Moving youki in root would make it a bit more confusing 

I don't feel this will be the case
>Also, the absence of Cargo.toml in the root directory may seem a bit strange for a Rust project. 

We are updating the documentation for this reason only.
>It is hard to know where to start looking at the code for first contributor.

Furisto:
The reason we choose the current way was to follow the example of [rust analyzer](https://matklad.github.io/2021/08/22/large-rust-workspaces.html), so the current structure should not be unexpected to someone with experience with rust projects. Also I feel that the reasons for this structure laid out in the linked article are still valid.

utam0k:
It sure seems I'm the only one who feels a little uncomfortable with this. Thanks so much for your comments, I really appreciate it. It helps.

PR #851, opened 2022-04-16
Ensure pid and root path are canonicalized

Furisto:
If the specified root folder and the pid file do not exist yet, canonicalize will fail. This handles these cases.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/851"></a>

### DIFF
--- a/crates/libcontainer/src/utils.rs
+++ b/crates/libcontainer/src/utils.rs
@@ -14,11 +14,13 @@ use std::ops::Deref;
 use std::os::linux::fs::MetadataExt;
 use std::os::unix::fs::DirBuilderExt;
 use std::os::unix::prelude::{AsRawFd, OsStrExt};
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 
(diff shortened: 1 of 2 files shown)


crates/libcontainer/src/utils.rs
```diff
@@ -42,6 +44,52 @@ impl PathBufExt for Path {
...
+                    .join(self);
+                return Ok(p.normalize());
+            }
+
+            Ok(self.normalize())
+        }
+    }
+
+    /// Normalizes a path. In contrast to canonicalize the path does not need to exist.
+    // adapted from https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
```
utam0k, line 66:
```suggestion
    /// adapted from https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
```

utam0k (approved):
I like these codes.

PR #897, opened 2022-05-06
Update CI coverage to use native option instead of action

YJDoc2:
This updates the CI to use native coverage generation option introduced in rust 1.60 instead of using CI action.
Test PR for https://github.com/containers/youki/issues/316#issuecomment-1119155913

DO NOT MERGE

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/897"></a>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,28 +77,23 @@ jobs:
       - name: Toolchain setup
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-01-14
+          toolchain: 1.60.0
           override: true
           profile: minimal
           components: llvm-tools-preview
       - name: Cache youki
         uses: Swatinem/rust-cache@v1
-      - name: install cargo-llvm-cov
-        env:
-          CARGO_LLVM_COV_VERSION: 0.1.5
-        run: |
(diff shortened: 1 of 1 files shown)


YJDoc2:
The decrease (-71%) is due to the coverage now considering the integration tests as well, which it should ignore. I'll try to make it ignore those files, which should fix it.

.github/workflows/main.yml
```diff
@@ -77,28 +77,23 @@ jobs:
...
+        run: cargo install grcov
       - name: Run Test Coverage for youki
         run: |
-          cd ./crates
-          cargo llvm-cov clean --workspace
-          cargo llvm-cov --no-report
-          cargo llvm-cov --no-run --lcov --ignore-filename-regex "libcgroups/src/systemd/dbus/systemd_api.rs" --output-path ./coverage.lcov
+          cargo clean
+          LLVM_PROFILE_FILE="coverage.profraw" RUSTFLAGS="-Cinstrument-coverage" cargo build 
+          grcov . --binary-path ./target/debug -s . -t lcov --branch --ignore-not-existing -o ./coverage.lcov
```
taiki-e, line 96: grcov itself does not have the ability to run tests, so the new commands actually just build the code. As a result, coverage is reported as 0 percent.

YJDoc2: Oops! I was following something, and forgot that we need to manually run tests :sweat_smile: :sweat_smile: 
Thanks a lot!

YJDoc2:
We might close this in favor of #898 . Keeping open for the time-being.

PR #898, opened 2022-05-06
Update cargo-llvm-cov and use rust 1.60.0 for coverage

taiki-e:
This should be enough to address https://github.com/containers/youki/issues/316#issuecomment-1119155913.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/898"></a>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,18 +77,16 @@ jobs:
       - name: Toolchain setup
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-01-14
+          toolchain: nightly-2022-05-06
           override: true
           profile: minimal
           components: llvm-tools-preview
       - name: Cache youki
         uses: Swatinem/rust-cache@v1
       - name: install cargo-llvm-cov
-        env:
-          CARGO_LLVM_COV_VERSION: 0.1.5
-        run: |
-          wget https://github.com/taiki-e/cargo-llvm-cov/releases/download/v${CARGO_LLVM_COV_VERSION}/cargo-llvm-cov-x86_64-unknown-linux-gnu.tar.gz -qO- | tar -xzvf -
-          mv cargo-llvm-cov ~/.cargo/bin
+        uses: taiki-e/install-action@v1
+        with:
+          tool: cargo-llvm-cov@0.3.2
       - name: Update System Libraries
         run: sudo apt-get -y update
       - name: Install System Libraries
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -77,18 +77,16 @@ jobs:
       - name: Toolchain setup
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-01-14
+          toolchain: nightly-2022-05-06
```
taiki-e, line 80: I originally tried to use 1.60.0 here, but ibcontainer enables the unstable feature when coverage is enabled, so I reverted to nightly.

https://github.com/containers/youki/blob/1591680993aa9c63cbc223e111af7c130aa319fe/crates/libcontainer/src/lib.rs#L1

taiki-e: After more looking into it, I found that the actual usage of no_coverage is only here, and it is no-op (because applied to `use` statement).

https://github.com/containers/youki/blob/e0843af2032f9e4d8aeefd87f8022feb5394e943/crates/libcontainer/src/syscall/linux.rs#L2-L3

So I believe we can remove the use of the unstable feature.

taiki-e: Removed the use of the unstable feature and changed to use rust 1.60.0: https://github.com/containers/youki/pull/898/commits/a7e800fb819b3d1f12b54fa0fe939cb8b9174d8e

YJDoc2:
Hey @taiki-e Thanks for your contribution!
This seems fine, but due to the changes in the file structure, and the coverage CI action not getting updated accordingly from us, this does not upload the coverage report to the codecov.

Can I ask you to remove this line in .github/workflows/main.yml?
```yaml
...
- name: Run Test Coverage for youki
        run: |
          cd ./crates # Please remove this line, as now we need to run this from project root instead of crates/
...
```

Thanks :)

YJDoc2:
Unrelated Note :

Also, test_guid_mapping was running for more than 15 mins, so it was cancelled in Rust 1.59, although ran fine in Rust 1.60. Might be just flaking issue, but needs a look at it.

taiki-e:
@YJDoc2 Thanks for the review! I applied your suggestion.

YJDoc2:
Hey, sincere apologies for delay :sweat: 

This looks good to me from CI perspective, @utam0k @Furisto can you once verify that removing the `cfg_attr` is fine? All tests are passing, so at least it does not create any known bugs.

Again, apologies for the delay :sweat:

utam0k (approved):
Looks good. Thank you @taiki-e

saschagrunert (approved):

PR #909, opened 2022-05-09
Remove the build dependency from some tests.

utam0k:
Remove the build dependency from some tests because it was inconvenient that we could always test only the latest builds. Instead, check in advance if the runtime is presents

### DIFF
--- a/scripts/oci_integration_tests.sh
+++ b/scripts/oci_integration_tests.sh
@@ -86,6 +86,13 @@ check_environment() {
   fi
 }
 
+if [[ ! -e $RUNTIME ]]; then
+  if ! which $RUNTIME ; then
+    echo "$RUNTIME not found"
+    exit 1
+  fi
+fi
+
 for case in "${test_cases[@]}"; do
   if [[ ! -e "${OCI_TEST_DIR}/validation/$case" ]]; then
     GO111MODULE=auto GOPATH=${ROOT}/tests/oci-runtime-tests make runtimetest validation-executables
--- a/.github/workflows/benchmark_execution_time.yml
+++ b/.github/workflows/benchmark_execution_time.yml
@@ -27,7 +27,7 @@ jobs:
           profile: minimal
 
       - name: Building PR branch
-        run: ./build.sh --release
+        run: make release-build
 
       - name: Uploading PR build to artifact
         uses: actions/upload-artifact@v2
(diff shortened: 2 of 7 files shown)


scripts/oci_integration_tests.sh
```diff
@@ -86,6 +86,13 @@ check_environment() {
   fi
 }
 
+if [[ ! -e $RUNTIME ]]; then
+  if ! which $RUNTIME ; then
+    echo "$RUNTIME not found"
+    exit 1
+  fi
+fi
+
```
YJDoc2, line 95: Shouldn't we set `$RUNTIME=$1` here like it is done in the `rust_integration_tests.sh` ?

YJDoc2:
Hey, this looks good , but I'm still a bit confused for why we are extracting the build step from `Makefile` into the CI? I'm not saying it is a bad choice to do so, but just wondering why we are doing it.

scripts/oci_integration_tests.sh
utam0k: Is this?
https://github.com/utam0k/youki/blob/0758953fd754452cf385c3bbb52fa813fed28420/scripts/oci_integration_tests.sh#L5

utam0k:
@YJDoc2 Nice question! The build dependencies during integration-test would have forced either release or debug builds in advance. This was too restrictive, so we removed this dependency. This made a build step necessary for CI.

In the future, I would like to refactor this CI, so that I can do the build in one place, upload the results, and reuse them elsewhere. I would also like to modify it so that unit tests are done only in one place.

scripts/oci_integration_tests.sh
YJDoc2: oh yes, I missed that :sweat_smile:

YJDoc2:
:+1:
>Nice question! The build dependencies during integration-test would have forced either release or debug builds in advance. This was too restrictive, so we removed this dependency. This made a build step necessary for CI.

Yes, currently the CI seems too monolithic, with a lot of steps repeated over and over :sweat: I will try to find a way to simplify. With the amount of checks we run, (tests x 2 rust versions, integration tests, lints) it might be useful to take a look the test pipeline which Rust language repo uses, although that will need paid resources.
>In the future, I would like to refactor this CI, so that I can do the build in one place, upload the results, and reuse them elsewhere. I would also like to modify it so that unit tests are done only in one place.

YJDoc2 (approved):
LGTM :+1:

PR #945, opened 2022-05-26
Fix CI compilation issues

Furisto:
- Update test matrix to 1.60 so that dependencies can be resolved
- Ensure that test binary is build and available in path
- Fix code coverage by updating to rust 1.61


utam0k (approved):
Thanks!

PR #959, opened 2022-06-04
test: hooks integration test

tony84727:
Add `hooks` integration test

See #361

Signed-off-by: Tony Duan <tony84727@gmail.com>

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/959"></a>

### DIFF
--- a/tests/rust-integration-tests/integration_test/src/main.rs
+++ b/tests/rust-integration-tests/integration_test/src/main.rs
@@ -1,6 +1,7 @@
 mod tests;
 mod utils;
 
+use crate::tests::hooks::get_hooks_tests;
 use crate::tests::lifecycle::{ContainerCreate, ContainerLifecycle};
 use crate::tests::linux_ns_itype::get_ns_itype_tests;
 use crate::tests::pidfile::get_pidfile_test;
@@ -78,6 +79,7 @@ fn main() -> Result<()> {
     let huge_tlb = get_tlb_test();
     let pidfile = get_pidfile_test();
     let ns_itype = get_ns_itype_tests();
+    let hooks = get_hooks_tests();
     let cgroup_v1_pids = cgroups::pids::get_test_group();
     let cgroup_v1_cpu = cgroups::cpu::v1::get_test_group();
     let cgroup_v2_cpu = cgroups::cpu::v2::get_test_group();
@@ -92,6 +94,7 @@ fn main() -> Result<()> {
     tm.add_test_group(&huge_tlb);
     tm.add_test_group(&pidfile);
     tm.add_test_group(&ns_itype);
+    tm.add_test_group(&hooks);
     tm.add_test_group(&cgroup_v1_pids);
     tm.add_test_group(&cgroup_v1_cpu);
     tm.add_test_group(&cgroup_v2_cpu);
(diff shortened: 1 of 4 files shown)


tests/rust-integration-tests/integration_test/src/tests/hooks/hooks_test.rs
```diff
@@ -0,0 +1,87 @@
...
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["true".to_string()])
+                .build()
+                .unwrap(),
+        )
+        .hooks(
+            HooksBuilder::default()
+                .prestart(vec![
+                    write_log_hook("pre-start1 called"),
```
utam0k, line 34: Why do we need to call twice per-hook?

utam0k (commented):
@tony84727 Thanks for your PR. Great! I left a comment, please check.

tests/rust-integration-tests/integration_test/src/tests/hooks/hooks_test.rs
tony84727: It's ported from here:
https://github.com/opencontainers/runtime-tools/blob/0105384f68e16803891d0a17d9067b1def6a2778/validation/hooks/hooks.go#L32-L67
I was having the same question. I guess calling twice is to ensure the implementation actually executes hooks in the given order. (hook 1 then hook 2 in sequence)

tony84727:
> @tony84727 This indicates that the runc did not pass the test, not youki. We run integration tests on runc to verify the integration tests themselves.
> 
> ```
> # Start group hooks
> 1 / 1 : hooks : not ok
> 	error : hooks must be called in the listed order
> ```

👌 I've made `runc` pass `hooks` test on my machine. Somehow, for hooks to write logs, we need to specify absolute path instead of simply `output`. I'm not sure why.

tests/rust-integration-tests/integration_test/src/tests/hooks/hooks_test.rs
```diff
@@ -1,20 +1,27 @@
...
     test_utils::start_container,
 };
 
+fn absolute_path(path: &str) -> PathBuf {
+    let pwd = std::env::current_dir().unwrap();
+    pwd.join(path)
+}
+
 fn write_log_hook(content: &str) -> Hook {
+    let output = absolute_path("./output");
```
utam0k, line 17: How about using `std::fs::canonicalize` instead of `absolute_path`?
https://doc.rust-lang.org/std/fs/fn.canonicalize.html#

tests/rust-integration-tests/integration_test/src/tests/hooks/hooks_test.rs
```diff
@@ -0,0 +1,94 @@
...
+    test_utils::start_container,
+};
+
+fn absolute_path(path: &str) -> PathBuf {
+    let pwd = std::env::current_dir().unwrap();
+    pwd.join(path)
+}
+
+fn write_log_hook(content: &str) -> Hook {
+    let output = absolute_path("./output");
```
utam0k, line 17: How about using std::fs::canonicalize instead of absolute_path?
https://doc.rust-lang.org/std/fs/fn.canonicalize.html#

tony84727: done.  
`canonicalize` seems to require the file to exist to resovle the aboslute path, so create the file prior running the tests.
Also, renamed the file to get rid of `_test` suffix.  
BTW, should I squash commits, maybe after reviews?

utam0k: Can I ask you to squash now?

tony84727: 👌 , squashed

utam0k: Thanks!

utam0k (approved):

PR #964, opened 2022-06-06
chore: update .gitignore

### LINKED ISSUE #958: .gitignore ignores `integration_test` files/directories
Hello, 

AFAIK, there was a directory refactor for integration tests.

When editing the integration tests (under `tests/rust-integration-tests/integration_test`), git complains that the source code is ignored.
because:
https://github.com/containers/youki/blob/3644c008f1604f698f9b40a21f06f024a107cad7/.gitignore#L15

Is that intentional? Or should we update the `.gitignore`?

Edit:
Just realized that the rule was to ignore the `integration_test` executable.
Maybe we should prefix the rule with a slash, so git only ignores the name within the directory, where the `.gitignore` file lives in, instead of ignoring any path with `integration_test` inside it. (i.e. tests/rust-integration-tests/**integration_test**)

tony84727:
prefixes some rules with slash (/) to make it only ignores
files/directories at the root directory

Fixes #958 

Signed-off-by: Tony Duan <tony84727@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-
-
 /tutorial
 .idea/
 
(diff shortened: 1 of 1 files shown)


utam0k (approved):
Thank you!!!

PR #993, opened 2022-06-19
Remove duplicated assignment

cyyzero:
In `exec` command, `process` field will be assigned twice when creating `TenantContainerBuilder`. I think it's unnecessary.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/993"></a>


utam0k (approved):
Thanks!!!

PR #1038, opened 2022-07-08
Update ID of rust-analyzer extension in gitpod.yml

eminence:
This extension is now known as `rust-lang.rust-analyzer`

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1038"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -9,4 +9,4 @@ github:
     addBadge: true
 vscode:
   extensions:
-    - matklad.rust-analyzer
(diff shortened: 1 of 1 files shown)


eminence:
Oh wait, I'm pretty sure this change is not correct... it seems that gitpod uses the open-vsx registry, which doesn't have the latest rust-analyzer extension.

eminence:
Going to close this until https://github.com/rust-lang/rust-analyzer/issues/11080 is resolved

PR #1128, opened 2022-08-18
Fix bug that attempts is always 0 in delete_with_retry

cyyzero:
In `delete_with_retry`, `attempts ` is initialized as 0 and add itself every time `remove_dir` fails. It makes `attempts` always 0 and `retries` has no effect here.

Signed-off-by: Chen Yiyang <cyyzero@qq.com>

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1128"></a>

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -420,7 +420,7 @@ pub(crate) fn delete_with_retry<P: AsRef<Path>, L: Into<Option<Duration>>>(
         }
 
         std::thread::sleep(delay);
-        attempts += attempts;
+        attempts += 1;
         delay *= attempts;
         if delay > limit {
             delay = limit;


cyyzero:
BTW, I found this bug when running commands below.

```sh
$ sudo youki run test
$ sudo youki pause test
$ sudo youki delete --force test
```

youki can't force delete a paused container , since container processes can't receive signal and quit until it is `THAWED`. Do I need to fix this issue?  Maybe resume the processes before removing cgroup can fix this.

crates/libcgroups/src/common.rs
```diff
@@ -420,7 +420,7 @@ pub(crate) fn delete_with_retry<P: AsRef<Path>, L: Into<Option<Duration>>>(
         }
 
         std::thread::sleep(delay);
-        attempts += attempts;
+        attempts += 1;
```
YJDoc2, line 423: Hey, maybe setting the initial value 1 would be better than adding 1 to attempts, as the delay in retries is generally exponential, and `attempts += attempts` accounts that exponentiality.
Adding 1 each time would keep a (almost) constant delay between attempts, which is not wrong, but I haven't really seen in context of delays (I have seen them to be exponential)

cyyzero: I think parameter `retries` represents the number of retries, so adding one each time is more appropriate.

YJDoc2 (approved):
LGTM :+1:
Thanks for your contribution!

Furisto (approved):

Furisto:
Thanks @cyyzero

Issue #1129, opened 2022-08-19
Can't kill a paused container in cgroup v1 environment.


cyyzero:
Test case is below:

```sh
$ sudo youki run test
$ sudo youki pause test
$ sudo youki list
ID    PID   STATUS   BUNDLE  CREATED                    CREATOR
test  7310  Paused   .       2022-08-19T12:51:28+08:00  root

$sudo youki kill test SIGKILL
$ sudo youki list
ID    PID   STATUS   BUNDLE  CREATED                    CREATOR
test  7310  Running  .       2022-08-19T12:51:28+08:00  root

$ cat /sys/fs/cgroup/freezer/test/freezer.state 
FROZEN
$ cat /sys/fs/cgroup/freezer/test/tasks        
7310
```

Container status is wrong because killing paused container failed. The container process won't receive SIGKILL and quit until it is thawed. It also makes `youki delete --force test` failing, since cgroup directory can't be removed until there's no tasks in it. 

Here's the solution. https://github.com/opencontainers/runc/pull/3217 Maybe I could help fix the issue.

YJDoc2:
Hey @cyyzero , sorry there hasn't been a reply. I'll try taking a look at the issue you have linked and getting back to you.

cc @Furisto

YJDoc2:
Hey @cyyzero , I took a look at the issue you linked, and we should also implement the same, i.e. if the cgroup on system is v1, then first thaw the container and then send a fatal kill signal , _iff_ --force option is given.

Would you be interested in working on this? My guess is following things will need to be done :
- whenever kill command is given, check if force is given or not
     - if force is not given, and cgroups is v1 and container is frozen, display error / warning saying could not be killed as it was forzen
     - if force option is given, and cgroups is v1 and container is frozen, thaw it and send the kill signal
     - if cgroups is v2 ,  nothing special needs to be done
- add an integration test to check this behavior, so we can make sure this keeps working (however, the test won't be the end-all be-all , as it will need to be run on cgroups v1 system to ensure that it the behavior is correct, cgroups v2 systems are already working as intended)

Let me know if you are interested in working on this, and in any case, Thanks for reporting and following up!

cyyzero:
Hey @YJDoc2., thanks for your reply!  I'm willing to fix this issue. I will firstly deal with the bug of `kill --force` under cgroup v1.  Maybe some time later I will ask you about some details of writing integration test (I'm not familiar with github CI).

YJDoc2:
[...]

PR #1204, opened 2022-09-21
Thaw a paused container in cgroup v1 when it is forcely deleted.

cyyzero:
If force is not given, and cgroups is v1 and container is frozen, display error saying could not be killed as it was forzen. If force option is given, and cgroups is v1 and container is frozen, thaw it and send the kill signal. If cgroups is v2 , nothing special needs to be done.

Fix: #1129

Signed-off-by: Chen Yiyang <cyyzero@qq.com>

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1204"></a>

### DIFF
--- a/crates/libcontainer/src/container/container_kill.rs
+++ b/crates/libcontainer/src/container/container_kill.rs
@@ -1,7 +1,7 @@
 use super::{Container, ContainerStatus};
 use crate::signal::Signal;
 use anyhow::{bail, Context, Result};
-use libcgroups::common::create_cgroup_manager;
+use libcgroups::common::{create_cgroup_manager, get_cgroup_setup};
 use nix::sys::signal::{self};
 
 impl Container {
@@ -24,29 +24,10 @@ impl Container {
     /// # }
     /// ```
     pub fn kill<S: Into<Signal>>(&mut self, signal: S, all: bool) -> Result<()> {
-        let signal = signal.into().into_raw();
-
-        let pids = if all {
-            let cgroups_path = self.spec()?.cgroup_path;
-            let use_systemd = self
-                .systemd()
-                .context("container state does not contain cgroup manager")?;
-            let cmanger = create_cgroup_manager(&cgroups_path, use_systemd, self.id())?;
-            cmanger.get_all_pids()?
-        } else {
-            vec![self
-                .pid()
-                .context("failed to get the pid of the container")?]
-        };
-
         self.refresh_status()
             .context("failed to refresh container status")?;
         if self.can_kill() {
-            pids.into_iter().try_for_each(|pid| {
-                log::debug!("kill signal {} to {}", signal, pid);
-                signal::kill(pid, signal)
-            })?;
-
+            self.do_kill(signal, all, false)?;
             self.set_status(ContainerStatus::Stopped).save()?;
             std::process::exit(0)
         } else {
@@ -56,5 +37,57 @@ impl Container {
                 self.status()
             )
         }
+    }
+
+    pub(crate) fn do_kill<S: Into<Signal>>(
+        &mut self,
+        signal: S,
+        all: bool,
+        force: bool,
+    ) -> Result<()> {
+        let signal = signal.into().into_raw();
+        let mut manger: Option<Box<dyn libcgroups::common::CgroupManager>> = None;
+        let pids = if all {
+            let cgroups_path = self.spec()?.cgroup_path;
+            let use_systemd = self
+                .systemd()
+                .context("container state does not contain cgroup manager")?;
+            let cmanger = create_cgroup_manager(&cgroups_path, use_systemd, self.id())?;
+            let all_pids = cmanger.get_all_pids()?;
+            manger = Some(cmanger);
+            all_pids
+        } else {
+            vec![self
+                .pid()
+                .context("failed to get the pid of the container")?]
+        };
+        if self.status() == ContainerStatus::Paused {
+            // In cgroup V1, paused processes cannot respond to signals,
+            // so if we want to kill a process forcely, we need to thaw it.
+            match get_cgroup_setup()? {
+                libcgroups::common::CgroupSetup::Legacy
+                | libcgroups::common::CgroupSetup::Hybrid => {
+                    if !force {
+                        bail!("in cgroup v1, frozon container can't be killed until it is thawed")
+                    }
+                    let cmanger = if let Some(cmanger) = manger {
+                        cmanger
+                    } else {
+                        let cgroups_path = self.spec()?.cgroup_path;
+                        let use_systemd = self
+                            .systemd()
+                            .context("container state does not contain cgroup manager")?;
+                        create_cgroup_manager(&cgroups_path, use_systemd, self.id())?
+                    };
+                    cmanger.freeze(libcgroups::common::FreezerState::Thawed)?;
+                }
+                libcgroups::common::CgroupSetup::Unified => {}
+            }
+        }
+        pids.into_iter().try_for_each(|pid| {
+            log::debug!("kill signal {} to {}", signal, pid);
+            signal::kill(pid, signal)
+        })?;
+        Ok(())
     }
 }
--- a/crates/libcontainer/src/container/container_kill.rs
+++ b/crates/libcontainer/src/container/container_kill.rs
@@ -27,67 +27,84 @@ impl Container {
         self.refresh_status()
             .context("failed to refresh container status")?;
         if self.can_kill() {
-            self.do_kill(signal, all, false)?;
-            self.set_status(ContainerStatus::Stopped).save()?;
-            std::process::exit(0)
+            self.do_kill(signal, all)?;
         } else {
-            bail!(
-                "{} could not be killed because it was {:?}",
-                self.id(),
-                self.status()
-            )
+            // just like runc, allow kill --all even if the container is stopped
+            if all && self.status() == ContainerStatus::Stopped {
+                self.do_kill(signal, all)?;
+            } else {
+                bail!(
+                    "{} could not be killed because it was {:?}",
+                    self.id(),
+                    self.status()
+                )
+            }
+        }
+        self.set_status(ContainerStatus::Stopped).save()?;
+        Ok(())
+    }
+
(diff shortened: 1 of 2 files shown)


YJDoc2:
Hey @cyyzero , I was going through implementation of runc of this, and one thing I noticed is that for single or multiple (https://github.com/opencontainers/runc/blob/main/libcontainer/container_linux.go#L370) , they first signal the processes, in the frozen state ( in fact they seem to freeze the cgroup before sending kill signal in case of all https://github.com/opencontainers/runc/blob/1c3b8dbaf440d16653d834b612258bbb28268730/libcontainer/init_linux.go#L530) then send kill signal, and then thaw the cgroup

In the PR you seem to thaw the cgroup before sending the kill signal. One of the issues with this is that between the time you get all processes from the cgroup and the time you send signal, there might be additional processes spawned as cgroup is not frozen. The runc's way seems to avoid it.

Also once you are done with the initial implementation that is in the current state, can I ask you to add `--all` option in the cmd interface? It is a pretty small change, and the value of it will be given as `all` argument to the  kill. That will help us to pass one of the containerd test.

Thank you!

cyyzero:
I might know why integration test TestContainerKillInitPidHost fails. I output command lines of each execution of youki:

```text
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json create --bundle /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost --pid-file /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/init.pid TestContainerKillInitPidHost
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json start TestContainerKillInitPidHost
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json kill TestContainerKillInitPidHost 9
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json kill --all TestContainerKillInitPidHost 9
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json kill --all TestContainerKillInitPidHost 9
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json delete TestContainerKillInitPidHost
runc --root /run/containerd/runc/testing --log /run/containerd-test/io.containerd.runtime.v2.task/testing/TestContainerKillInitPidHost/log.json --log-format json delete --force TestContainerKillInitPidHost
```

TestContainerKillInitPidHost runs the container without creating pid namespace, so the child process will not exit automatically after killing the init process, it is necessary to use `kill -all`.

https://github.com/containerd/containerd/blob/31f9d13f0cc3a7d60e009cb8e76ccb188d4a76d7/integration/client/container_linux_test.go#L1120

I haven't read the containerd or shim-v2 source code , so I'm not quite clear why the kill command is called multiple times. It seems the first `kill` has no options.  In the previous implementation, the container status becomes stopped after `kill`, and subsequent `kill --all`s can't pass the `can_kill` check, so the child process could not be killed correctly.

In fact, in the runc source code, it doesn't do status check for `--all` option. 
https://github.com/opencontainers/runc/blob/1102f3fc9d5208961f19d70618c3be251466ee3f/libcontainer/container_linux.go#L361-L371

crates/libcontainer/src/container/container_kill.rs
```diff
@@ -24,37 +24,87 @@ impl Container {
...
             .context("failed to refresh container status")?;
         if self.can_kill() {
-            pids.into_iter().try_for_each(|pid| {
-                log::debug!("kill signal {} to {}", signal, pid);
-                signal::kill(pid, signal)
-            })?;
+            self.do_kill(signal, all)?;
+        } else {
+            // just like runc, allow kill --all even if the container is stopped
+            if all && self.status() == ContainerStatus::Stopped {
```
YJDoc2, line 33: Shouldn't here we check for stopped and running both?  Otherwise it will give error when using --all with running container.
Also, given that we are having an if-else in the `do_kill` function, would it be better to move the checking logic there, or remove that function completely and move its logic here?

crates/libcontainer/src/container/container_kill.rs
```diff
@@ -24,37 +24,87 @@ impl Container {
...
+            } else {
+                bail!(
+                    "{} could not be killed because it was {:?}",
+                    self.id(),
+                    self.status()
+                )
+            }
+        }
+        self.set_status(ContainerStatus::Stopped).save()?;
+        std::process::exit(0);
```
YJDoc2, line 44: I'm not sure why we had exit here in the original implementation... libcontainers is a library and should not call exit unless extreme conditions. Ideally we should let users of the library decide what to do after calling this function. It would be a better idea to return `Ok(())` from here. @Furisto if you get time, can you check and let know if there was any particular reasoning behind exiting from here?

crates/libcontainer/src/container/container_kill.rs
```diff
@@ -24,37 +24,87 @@ impl Container {
...
+        if self.status() == ContainerStatus::Paused && signal == signal::Signal::SIGKILL {
+            match get_cgroup_setup()? {
+                libcgroups::common::CgroupSetup::Legacy
+                | libcgroups::common::CgroupSetup::Hybrid => {
+                    let cgroups_path = self.spec()?.cgroup_path;
+                    let use_systemd = self
+                        .systemd()
+                        .context("container state does not contain cgroup manager")?;
+                    let cmanger = create_cgroup_manager(&cgroups_path, use_systemd, self.id())?;
+                    cmanger.freeze(libcgroups::common::FreezerState::Thawed)?;
```
YJDoc2, line 72: Should we leave the cgroup thawed after this? I'm not sure myself. On one hand runc does seem to cgroup thawed for SIGKILL, but then the behavior is inconstant for other signals, which will not thaw the cgroup. If we do choose to re-freeze it, would that make sense, to re-freeze given that the container process is sigkill-ed ; and if so done, how will we handle the race condition between sending the signal, thawing the cgroup and re-freezing it? @utam0k what are your thoughts?

YJDoc2 (commented):
Hey, great work! I have left some comments, please take a look.

crates/libcontainer/src/container/container_kill.rs
utam0k: To be honest, I'm not sure, but I think there is no problem if we follow runc.
https://github.com/utam0k/runc/blob/2e8b7a1283838d7a8757b5c8a0425b525c28d269/libcontainer/container_linux.go#L381-L383

YJDoc2: Yeah, I guess that should be fine, as most of the things that use runc (containerd, podman, docker etc) seem to accept this or ignore this anyways.

utam0k: The real world isn't always ideal, maybe you should report the issue to runc.

crates/libcontainer/src/container/container_kill.rs
cyyzero: https://github.com/containers/youki/blob/23170c6a3225d951fe556ac652ccd0c5df3a9c68/crates/libcontainer/src/container/state.rs#L40-L46
https://github.com/containers/youki/blob/d4088a9a61f869d951739eb7b5bc5ccc23691292/crates/libcontainer/src/container/container_kill.rs#L29

If container is running, it will pass the `can_kill` check and be killed without error. Maybe youki should be the same as runc, remove the check for `Stopped` status here, and do the kill in all status if with the --all option.

I think `do_fork` should not add checking logic. In this way, it can be used more flexibly by other modules, such as  `delete` command.

crates/libcontainer/src/container/container_kill.rs
cyyzero: I agree. `return Ok(())` is better than exiting here.

crates/libcontainer/src/container/container_kill.rs
YJDoc2: Hey, I agree with you, this makes sense :+1: Let us keep it the way it is, I'll try to open an issue on runc to see what can be done.

crates/libcontainer/src/container/container_kill.rs
YJDoc2: Can you make that change, where we return the result instead of exiting? Thanks

YJDoc2:
@cyyzero Apologies for the delay in my response. I have updated the comments, and if you can change the process exit to return result and add the `--all` option to the cli part (liboci-cli  and the youki actually invoking the command), this PR is fine. Can you make that change and un-draft this PR, so we can approve and merge it? Thanks a lot!

@utam0k , as this does change the behavior of kill slightly, and adds an `--all` flag, should we bump the version of youki, liboci-cli and libcontainers?

Again, apologies to both for the delay on my side :sweat_smile:

utam0k:
> @utam0k , as this does change the behavior of kill slightly, and adds an --all flag, should we bump the version of youki, liboci-cli and libcontainers?

I'm planning to release v0.0.4 soon if the contained test is passed with youki after merging this PR, so there is no problem. Thanks for your concern ;)

crates/libcontainer/src/container/container_kill.rs
cyyzero: @YJDoc2 Sure😄

cyyzero:
https://github.com/containers/youki/blob/6d05dd2f60198aaecb7c93d08c08d7db4fbc6600/crates/liboci-cli/src/kill.rs#L5-L11
@YJDoc2 Thanks for your review! It seems that the `--all` option has been added to the liboci-cli before. And I have changed some `exit(0)` to return `Ok(())`.

YJDoc2 (approved):
lgtm :+1:

YJDoc2:
Hey @cyyzero Thanks for your contribution and efforts ! :)

Issue #1212, opened 2022-09-24
Git submodules using `git` urls can break `cargo install`


Hoverbear:
Currently if a user attempts to install `youki` via Cargo, they may meet an error if they do not have git keys configured fully:

```bash
❯ cargo install --git https://github.com/containers/youki
    Updating git repository `https://github.com/containers/youki`
    Updating git submodule `git@github.com:opencontainers/runtime-tools.git`
error: failed to update submodule `integration_tests/oci-runtime-tests`

Caused by:
  failed to fetch submodule `integration_tests/oci-runtime-tests` from git@github.com:opencontainers/runtime-tools.git

Caused by:
  failed to authenticate when downloading repository

  * attempted ssh-agent authentication, but no usernames succeeded: `git`

  if the git CLI succeeds then `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli
```

It may be desirable to use a different Git submodule remote URL protocol.

While attempting to build `youki` in a sandbox without configured git keys I notably saw this error which I think might be related?: 

```rust
youki>    Compiling futures-executor v0.3.24
youki>    Compiling futures v0.3.24
youki> error: failed to run custom build command for `youki v0.0.3 (/build/source/crates/youki)`
youki> Caused by:
youki>   process didn't exit successfully: `/build/source/target/release/build/youki-e2aad9c6a21f0e82/build-script-build` (exit status: 1)
youki>   --- stderr
youki>   Error: could not find repository from '/build/source/crates/youki'; class=Repository (6); code=NotFound (-3)
youki> warning: build failed, waiting for other jobs to finish...
error: builder for '/nix/store/4769q3a9n0d3ay1qmbha0q5brvg9jv6q-youki-0.0.3.drv' failed with exit code 101;
       last 10 log lines:
       >    Compiling youki v0.0.3 (/build/source/crates/youki)
       >    Compiling futures-executor v0.3.24
       >    Compiling futures v0.3.24
       > error: failed to run custom build command for `youki v0.0.3 (/build/source/crates/youki)`
       >
       > Caused by:
       >   process didn't exit successfully: `/build/source/target/release/build/youki-e2aad9c6a21f0e82/build-script-build` (exit status: 1)
       >   --- stderr
       >   Error: could not find repository from '/build/source/crates/youki'; class=Repository (6); code=NotFound (-3)
       > warning: build failed, waiting for other jobs to finish...
[...]

PR #1251, opened 2022-10-07
Add git commit sha placeholder if .git not found

YJDoc2:
Related to #1212 
Not sure if we actively support building through nix, and as the issue discussion states, it is not fully working either even with this change.
This will emit a placeholder sha of `unknown` and add continue-on-error option for vergen. Currently the only place we are using sha is in info and version cli commands. If .git is found, vergen overrides the `unknown` by actual sha. I have verified this, but if possible, someone else also verify once. 

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1251"></a>


utam0k (approved):
Thanks!

PR #1018, opened 2022-06-29
Get the result of exec command

utam0k:
Fix: https://github.com/containers/youki/issues/953

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1018"></a>

Issue #531, opened 2021-12-12
Support for containerd


utam0k:
# Goal

Pass the integration test of contained

# Status
Done!

### How to run containerd's integration test with youki

https://containers.github.io/youki/developer/containerd_integration_test_using_youki.html

guni1192:
@utamk0
What is the status of this issue?
I am doing containerd integration tests with runc replaced by youki.
Do you need the procedure and results of the test in my local environment?

utam0k:
@guni1192 I would love to have it. Can you please do it? Or can you give me a PR with a shell of the test means?

guni1192:
OK.　First, I will share the results of the test conducted in my environment.
I published it on my Gist because the lines in the log are very long.

https://gist.github.com/guni1192/af931310dca8dfc5b2a0f5b79acbad32

guni1192:
@utam0k 
I would like to go through the steps to automate integration testing using PR.
Can I edit the Vagrantfile in the repository for this purpose?
Specifically, I want to set up containerd and golang.
If you don't mind, I would like to send a PR this week.

guni1192:
I am concerned about a number of failed test cases in the containerd integration test, as follows.
I think containerd may have used runc-specific options( `runc kill --all <CONTAINER_ID> <SIGNAL>` ).

```
=== RUN   TestDaemonRestart
    container_test.go:2042: failed to delete task: unknown error after kill: runc did not terminate successfully: exit status 2: error: Found argument '--all' which wasn't expected, or isn't valid in this context
        
        	If you tried to supply `--all` as a value rather than a flag, use `-- --all`
        
        USAGE:
            runc kill <CONTAINER_ID> <SIGNAL>
        
        For more information try --help
        : unknown
--- PASS: TestDaemonRestart (0.31s)
=== RUN   TestContainerKillInitKillsChildWhenNotHostPid
    container_test.go:2199: unknown error after kill: runc did not terminate successfully: exit status 2: error: Found argument '--all' which wasn't expected, or isn't valid in this context
        
        	If you tried to supply `--all` as a value rather than a flag, use `-- --all`
        
        USAGE:
            runc kill <CONTAINER_ID> <SIGNAL>
        
        For more information try --help
        : unknown
--- FAIL: TestContainerKillInitKillsChildWhenNotHostPid (0.35s)
```

```
 $ runc kill --help                                                                                                                                                                                     
NAME:
   runc kill - kill sends the specified signal (default: SIGTERM) to the container's init process

[...]

PR #1252, opened 2022-10-07
Make exec behaviour consistent with runc's exec

YJDoc2:
related : #531 

NOTE : even though this does solve the issue, I'm not particularly happy with the way this is implemented. Comments and suggestions for improvements are most welcome.

This helps in passing following containerd's integration tests which were previously failing :
- TestContainerExec
- TestContaienrExecNoBinaryExists
- TestContainerExecLargeTTY (not sure if it was passing before and then broke, or was not passing from start, this PR will pass it)
- TestContainerLargeExecArgs
- TestContainerAttachProcess
- TestContainerMetrics
:tada: 

## Background
runc's exec has 4 distinct outcomes in terms of return value:
- when run without `--detach` and with a valid commannd, it waits and exits with the command's exit code
- when run without `--detach` and with an invalid command, it exits with and error, 255 exit code
- when run with `--detach` and with a valid command, it exits immediately after launching the process (tenant) and exits with a 0
- when run with `--detach` and with an invalid command, it exits with and error and exit code of 255

The 4rth case is subtle and made the main difference of why youki was not originally passing containerd's tests. 

On a sidde note, originally youki's exec did not respect the `--detach` flag and that was observed by the failing of TestContaienrExec after #1018 . This PR also fixes that.

Now, the reason behind the 4rth case exiting with error and not 0 like the 3rd, even though both are given detach flags, is runc executes the exec command similar to using process::spawn in Rust. Thus the exec occurs in two steps :
- First the exec process is spawned, and if there is any error in spawning the process, such as invalid command, invalid operation or OOM errors, it is checked at this stage, and immediately returned.
- If there was no error in spawning the command, then after doing some other work, runc checks if detach flag is given, if given immediately exits with 0, else waits for the exec'ed process to exit , and exits with its exit code.

Unlike it, youki uses fork-exec, which makes it difficult to do similar two-step exec as runc, because unlike process::spawn we do not really have a "handle" on the fork-exec'ed  process, at best we have parent-child waitpid relationship.

## This PR

This PR solves the issue by making sure the exec invocation done by youki binary gets the status of exec, AND add conditionality for waiting. 

The issue with the implementation is that the only way I could make it work was to open a pipe with O_CLOEXEC flag, and in the intermediate's fork for init process, _if_ the start_init_process returns, treat it as error (as the exec call should change process image and start the actual exec command, thus never returning). If the error occurs write the error message to the pipe. If exec succeeds, then the pipe is closed by OS. The tenant builder, after notifying start blocks by read-ing on this pipe, and if the read returns 0, indicating all write ends are closed, then the exec call has succeeded, otherwise, the buffer given to error contains the error message written by the init_process.
For this I had to drill down the RawFd from the tenant builder to the init process, and as the final `.build` uses the same implementation for init process and exec process, I had to introduce an enum to indicate the type AND pass the Fd. The problem is I don't feel this is much intuitive, and we have to make sure that the extra duplicated write-end fds are closed, for eg the intermediate process, after  forking from main process contains the duplicated write-end fd. If not closed, the read end blocks until the intermediate process has also exited. This took me a day to debug, and can be tricky to maintain.

Also, currently the exec-failed case always exits with 255, even though we might know the exact error code, as I couldn't find a way to downcast the anyhow error to specific type. The buffer size  is also hard-coded to 1024, so in case the error message is longer, it will be cut -off.

Personally I would have preferred a solution like notify-socket or main/intermediate sender-receiver  channels, but not sure if that supports O_CLOEXEC, and would work the same. Another issue preventing using main/intermediate channels is that they are created in the main_process which exists inside the `.build` call, however, the exec call is done only after notify_start is done, which is after the `.build` . This means that any other solution that is used will need some kind of parameter either being hard-coded or drilled down in a similar way RawFd is currently done.

Finally, one more thing to note is not only the youki cli invocation should not block/hand for exec if the detach flag is given, for some reason the containerd tests also monitor the intermediate process(?) which before this waited unconditionally on the init process. This PR thus also drills down the detach status, and both the cli and intermediate wait conditionally. I'm not sure why the test blocks on the exit of intermediate process , need to check if this is because we are leaking the pid of it somehow by incorrectly giving it instead of the init process. I have manually verified that the pid written to pid file is of the actual init process in case of create-start.

---

As mentioned in the start, not sure if this is the best approach, might be some better way to do this, comments and suggestion welcome. Thanks :)

cc: @utam0k @Furisto @yihuaf 

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1252"></a>

### DIFF
--- a/crates/libcontainer/src/container/tenant_builder.rs
+++ b/crates/libcontainer/src/container/tenant_builder.rs
@@ -19,6 +19,7 @@ use std::{
     str::FromStr,
 };
 
+use crate::process::args::ContainerType;
 use crate::{capabilities::CapabilityExt, container::builder_impl::ContainerBuilderImpl};
 use crate::{notify_socket::NotifySocket, rootless::Rootless, tty, utils};
 
@@ -38,6 +39,7 @@ pub struct TenantContainerBuilder<'a> {
     no_new_privs: Option<bool>,
     capabilities: Vec<String>,
     process: Option<PathBuf>,
+    detached: bool,
 }
 
 impl<'a> TenantContainerBuilder<'a> {
@@ -53,6 +55,7 @@ impl<'a> TenantContainerBuilder<'a> {
             no_new_privs: None,
             capabilities: Vec::new(),
             process: None,
+            detached: false,
         }
     }
 
@@ -89,6 +92,11 @@ impl<'a> TenantContainerBuilder<'a> {
         self
     }
 
+    pub fn detached(mut self, detached: bool) -> Self {
+        self.detached = detached;
+        self
+    }
+
     /// Joins an existing container
     pub fn build(self) -> Result<Pid> {
         let container_dir = self
@@ -120,7 +128,10 @@ impl<'a> TenantContainerBuilder<'a> {
         let (read_end, write_end) = pipe2(OFlag::O_CLOEXEC)?;
 
         let mut builder_impl = ContainerBuilderImpl {
-            init: false,
+            container_type: ContainerType::TenantContainer {
+                detached: self.detached,
+                exec_notify_fd: write_end,
+            },
             syscall: self.base.syscall,
             container_id: self.base.container_id,
             pid_file: self.base.pid_file,
@@ -132,7 +143,6 @@ impl<'a> TenantContainerBuilder<'a> {
             notify_path: notify_path.clone(),
             container: None,
             preserve_fds: self.base.preserve_fds,
-            exec_fd: Some(write_end),
         };
 
         let pid = builder_impl.create()?;
--- a/crates/libcontainer/src/container/tenant_builder.rs
+++ b/crates/libcontainer/src/container/tenant_builder.rs
@@ -1,6 +1,7 @@
 use anyhow::{bail, Context, Result};
 use caps::Capability;
-use nix::unistd::{self, Pid};
+use nix::unistd::{self, Pid,pipe2,read,close};
+use nix::fcntl::OFlag;
 use oci_spec::runtime::{
     Capabilities as SpecCapabilities, Capability as SpecCapability, LinuxBuilder,
     LinuxCapabilities, LinuxCapabilitiesBuilder, LinuxNamespace, LinuxNamespaceBuilder,
@@ -116,6 +117,8 @@ impl<'a> TenantContainerBuilder<'a> {
         let use_systemd = self.should_use_systemd(&container);
         let rootless = Rootless::new(&spec)?;
 
+        let (read_end,write_end) = pipe2(OFlag::O_CLOEXEC)?;
+
         let mut builder_impl = ContainerBuilderImpl {
             init: false,
             syscall: self.base.syscall,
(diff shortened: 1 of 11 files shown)


crates/libcontainer/src/process/args.rs
```diff
@@ -6,9 +6,18 @@ use std::path::PathBuf;
...
 use crate::{container::Container, notify_socket::NotifyListener, syscall::Syscall};
 
+#[derive(Debug, Copy, Clone)]
+pub enum ContainerType {
+    InitContainer,
+    TenantContainer {
+        detached: bool,
+        exec_notify_fd: RawFd,
+    },
+}
```
utam0k, line 16: I love this idea :heart:

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -136,7 +150,13 @@ impl<'a> TenantContainerBuilder<'a> {
         let mut notify_socket = NotifySocket::new(notify_path);
         notify_socket.notify_container_start()?;
 
-        Ok(pid)
+        close(write_end)?;
+
+        let mut buf = [0; 1024];
```
utam0k, line 155: This implementation has a limit on the number of characters in error messages from init process, right? I would like to remove this. I have considered some ideas
- Repeat read
- structured message
    - Pass the length of the message first, separated by `:`, etc.
- Make two communications
    1. Sent the length of a message
    2. Sent the body of a message.

utam0k:
@YJDoc2 Thanks for your hard work :pray: :bow:

utam0k:
Hi, @Furisto. May I ask you to review it? I would like to review it with you both.

crates/libcontainer/src/container/tenant_builder.rs
YJDoc2: Yes, I have also considered repeated reading until read returns 0, which will allow us to get complete message. Once rest of the code and approach is finalized and approved, I'll make that change.

I don't think we could use the third way, as far as know the data written to pipes is buffered, so the length and body will come out as a single data stream.

YJDoc2:
> @YJDoc2 Thanks for your hard work pray bow

Thank you for your kind words :blush:

crates/libcontainer/src/container/tenant_builder.rs
utam0k: I'm a little concerned when veeeeeeery long error message comes in

Furisto:
Hey, really sorry, I have not found the time yet to look at this. I will take a stab at this tomorrow.

Furisto:
I only see this dealing with the exec/tenant case, is my understanding correct? You can also detach from an init container (see https://github.com/opencontainers/runc/blob/main/docs/terminals.md for the supported modes). I am concerned that we introduce code that makes it harder to support this case in the future. Can you think of a way to support both?

Furisto:
Maybe I am overthinking it, I am trusting your judgement. Moving detach to the builder is the right call as we will be needing it later. Lets do this and then we are good to go I think.

crates/libcontainer/src/container/tenant_builder.rs
utam0k: @YJDoc2 If you need a lot of time to address it, please create an issue. and let's merge this PR ;)

YJDoc2: Hey, I was a bit busy for last couple of days, working on this now. Will probably get this done today or tomorrow. :)

YJDoc2:
@utam0k Apologies for the  delay, I have extracted the `detach` in the base builder as discussed with Furisto, as well as have implemented looping read for exec error message, so even longer error messages can be received. Please take a look. Thanks!

utam0k (approved):
@YJDoc2 :100: Thanks for your great work!

PR #1265, opened 2022-10-13
Update libbpf to 1.0.3+v1.0.1

YJDoc2:
This makes the relevant changes for the major release of libbpf-sys. Apart from a function name and signature difference, there is no visible breakage in the behavior.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1265"></a>

### DIFF
--- a/crates/libcgroups/src/v2/devices/bpf.rs
+++ b/crates/libcgroups/src/v2/devices/bpf.rs
@@ -24,26 +24,31 @@ pub mod prog {
     // TODO: consider use of #[mockall_double]
     #[cfg(test)]
     use crate::v2::devices::mocks::mock_libbpf_sys::{
-        bpf_load_program, bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_query,
+        bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_load, bpf_prog_query,
     };
     #[cfg(not(test))]
     use libbpf_sys::{
-        bpf_load_program, bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_query,
+        bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_load, bpf_prog_query,
     };
 
     pub fn load(license: &str, insns: &[u8]) -> Result<RawFd> {
         let insns_cnt = insns.len() / std::mem::size_of::<bpf_insn>();
         let insns = insns as *const _ as *const bpf_insn;
+        let opts = libbpf_sys::bpf_prog_load_opts {
+            kern_version: 0,
+            log_buf: ptr::null_mut::<i8>(),
+            log_size: 0,
+            ..Default::default()
+        };
         #[allow(unused_unsafe)]
         let prog_fd = unsafe {
-            bpf_load_program(
+            bpf_prog_load(
                 BPF_PROG_TYPE_CGROUP_DEVICE,
+                ptr::null::<i8>(),
+                license as *const _ as *const i8,
                 insns,
                 insns_cnt as u64,
-                license as *const _ as *const i8,
-                0,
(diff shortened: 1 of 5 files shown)


crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -24,26 +24,32 @@ pub mod prog {
...
     pub fn load(license: &str, insns: &[u8]) -> Result<RawFd> {
         let insns_cnt = insns.len() / std::mem::size_of::<bpf_insn>();
         let insns = insns as *const _ as *const bpf_insn;
+        let opts = {
+            let mut opts = bpf_prog_load_opts::default();
+            opts.kern_version = 0;
+            opts.log_buf = ptr::null_mut::<i8>();
+            opts.log_size = 0;
+            opts
+        };
```
YJDoc2, line 43: I'm pretty sure we can use just the default without modifying anything, as all fields are c_uints, whose defaults will be zero nonetheless. Have added this here for initial, can remove later

crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -24,26 +24,32 @@ pub mod prog {
...
                 BPF_PROG_TYPE_CGROUP_DEVICE,
+                ptr::null::<i8>(),
+                license as *const _ as *const i8,
                 insns,
                 insns_cnt as u64,
-                license as *const _ as *const i8,
-                0,
-                ptr::null_mut::<i8>(),
-                0,
+                &opts
```
YJDoc2, line 52: Logic behind this change :
- The second param is program name as string *. Now originally we used bpf_load_prog, where we can see in its [source](https://github.com/libbpf/libbpf/commit/9476dce6fe905a6bf1d4c483f7b2b8575c4ffb2d#diff-4b440c545696b4c1c729c0ebc776a5c478ebf893f01d4891c2cd7fd1694431baL430) it calls [bpf_load_program_xattr2](bpf_load_program_xattr2), in which the bpf_prog_load (named different before major release) was [called here](https://github.com/libbpf/libbpf/commit/9476dce6fe905a6bf1d4c483f7b2b8575c4ffb2d#diff-4b440c545696b4c1c729c0ebc776a5c478ebf893f01d4891c2cd7fd1694431baL426). The bpf_program_load would internally set the opts, and set the name as [null here](https://github.com/libbpf/libbpf/commit/9476dce6fe905a6bf1d4c483f7b2b8575c4ffb2d#diff-4b440c545696b4c1c729c0ebc776a5c478ebf893f01d4891c2cd7fd1694431baL440). Thus we should be able to safely do that without breaking original behavior (theoretically).
- After that the license param was moved up in the list, and the last three are set in the opts above.

YJDoc2: ! Not Outdated, marked so because of the forced fmt push

crates/libcgroups/src/v2/devices/bpf.rs
```diff
@@ -24,26 +24,33 @@ pub mod prog {
...
     use libbpf_sys::{
-        bpf_load_program, bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_query,
+        bpf_prog_attach, bpf_prog_detach2, bpf_prog_get_fd_by_id, bpf_prog_load,
+        bpf_prog_load_opts, bpf_prog_query,
     };
 
     pub fn load(license: &str, insns: &[u8]) -> Result<RawFd> {
         let insns_cnt = insns.len() / std::mem::size_of::<bpf_insn>();
         let insns = insns as *const _ as *const bpf_insn;
+        let opts = libbpf_sys::bpf_prog_load_opts {
```
YJDoc2, line 39: I'm pretty sure we can use just the default without modifying anything, as all fields are c_uints, whose defaults will be zero nonetheless. Have added this here for initial, can remove later

YJDoc2:
@utam0k Fixed!

utam0k (approved):
Thanks

PR #1297, opened 2022-10-24
Improve the flow of the containerd test with youki

utam0k:


Vagrantfile.containerd2youki
```diff
@@ -3,49 +3,54 @@
...
-      rm -rf /usr/local/go && tar -C /usr/local -xzf /tmp/go1.18.1.linux-amd64.tar.gz
-      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-      echo "export PATH=$PATH:$HOME/.cargo/bin:/usr/local/go/bin" >> ~/.bashrc
-      echo "export GOPATH=$HOME/go" >> ~/.bashrc
-      export PATH=$PATH:$HOME/.cargo/bin:/usr/local/go/bin
-      export GOPATH=$HOME/go
+    config.vm.provision "bootstrap", type: "shell" do |s|
+      s.inline = <<-SHELL
+        set -e -u -o pipefail
+        apt-get update && apt-get install -y \
```
raghulmz, line 16: The apt-packages / go / contianerd / cni /critools will be downloaded and installed for every single run of `make containerd-test`.

IMO it would make the test runs faster if a custom vagrant box was built (and possibly uploaded to the vagrant cloud) by a separate make target say `build vagrant`. The box has to be rebuilt only when there a change in the packages / base linux image is required.

This will make code -> test -> code process faster as only the rust code will be copied & rebuilt.

utam0k: Thanks for your review :bow: 

No, these install commands only run first.
> The apt-packages / go / contianerd / cni /critools will be downloaded and installed for every single run of make containerd-test.

utam0k:
@YJDoc2 @Furisto May I ask you a favor to review this one?

YJDoc2:
Hey @utam0k , I will review this by today or tomorrow :+1:

YJDoc2:
Hey @utam0k , Apologies for the delay :sweat: 
I checked, and there are some issues / improvements that can be done
- it seems that provision script is running each time we are doing the make. eg, after first run of `make containerd-test` , when we run it second time, it still hangs the same amount of time. When removed the `| true` part, it shows the log `default: Running provisioner: bootstrap (shell)...` also when running second time. we should really not run it twice. changing it as follows should make it work as expected :
```
config.vm.provision "shell" do |s|
      s.inline = <<-SHELL
...
```
- can we have logs to be explicit in the test ? currently the logs are stored in a file, can we have them also printed to stdout, because otherwise it looks like the command has hanged without doing anything. For this removing the `| true` from first will also help.

Also a thing to note is I'm not sure how the syncing of youki dir with vm work, will need to check that it still works as expected after these changes.

Again, apologies for the delay, I unexpectedly got caught up in some work, and couldn't check this earlier.

utam0k:
> Hey @utam0k , Apologies for the delay sweat I checked, and there are some issues / improvements that can be done
> 
> * it seems that provision script is running each time we are doing the make. eg, after first run of `make containerd-test` , when we run it second time, it still hangs the same amount of time. When removed the `| true` part, it shows the log `default: Running provisioner: bootstrap (shell)...` also when running second time. we should really not run it twice. changing it as follows should make it work as expected :
> 
> ```
> config.vm.provision "shell" do |s|
>       s.inline = <<-SHELL
> ...
> ```
> 
> * can we have logs to be explicit in the test ? currently the logs are stored in a file, can we have them also printed to stdout, because otherwise it looks like the command has hanged without doing anything. For this removing the `| true` from first will also help.
> 
> Also a thing to note is I'm not sure how the syncing of youki dir with vm work, will need to check that it still works as expected after these changes.
> 
> Again, apologies for the delay, I unexpectedly got caught up in some work, and couldn't check this earlier.

Thanks for your check. You don't need to say apologize. I am just grateful that you are always willing to help.
I haven't found the time yet to respond, so I'll address it later.

YJDoc2 (approved):
lgtm :+1:

PR #1302, opened 2022-10-25
Log result of the command before returning from main

YJDoc2:
This will log the error if there was any when executing the command. This will help in trying to figure out what was the issue from the logs, as right now, we do not get any info about it from the logs. This was particularly irritating when trying to fix the containerd test, as it would just report `runtime exited with 1`,  and even the logs would not show anything.


crates/youki/src/main.rs
```diff
@@ -132,7 +132,12 @@ fn main() -> Result<()> {
         SubCommand::Completion(completion) => {
             commands::completion::completion(completion, &mut app)
         }
+    };
+
+    if let Err(ref e) = output {
+        log::error!("error in executing command : {:?}", e);
```
utam0k, line 138:
```suggestion
        log::error!("error in executing command: {:?}", e);
```

crates/youki/src/main.rs
```diff
@@ -97,7 +97,7 @@ fn main() -> Result<()> {
     let root_path = determine_root_path(opts.global.root)?;
     let systemd_cgroup = opts.global.systemd_cgroup;
 
-    match opts.subcmd {
+    let output = match opts.subcmd {
```
utam0k, line 100: How about changing the name of the variable?
```suggestion
    let cmd_result  = match opts.subcmd {
```

YJDoc2:
@utam0k PTAL, have made both of the  suggested changes :+1:

utam0k (approved):
Thanks!

PR #1319, opened 2022-11-06
Add TestContainerKill required error message in kill command

### LINKED ISSUE #1314: Pass TestContainerKill containerd test
The reason TestContainerKill is failing is because the way it checks for error here https://github.com/containerd/containerd/blob/main/pkg/process/utils.go#L135 is by string comparison of error message (mostly) , and as youki's error message is different from what is checked, it assumes that an "incorrect" error was given and fails the test. Simply changing the error message to match one of the comparison passes the test. 

However, I'd prefer not to change the error message that is given, as I feel youki's error message is contains more information helpful for debugging. Maybe we can add that string after the current error message, or I'm looking how we can pass the ESRCH comparison, as I feel that would be a better way. Opened discussion on https://github.com/containerd/containerd/discussions/7605 .

Simply exiting with ESRCH (3) does not seem to pass the test.

YJDoc2:
This will pass the TestContainerKill of containerd

Fixes #1314 

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1319"></a>

### DIFF
--- a/crates/youki/src/commands/kill.rs
+++ b/crates/youki/src/commands/kill.rs
@@ -4,11 +4,20 @@ use std::{convert::TryInto, path::PathBuf};
 use anyhow::Result;
 
 use crate::commands::load_container;
-use libcontainer::signal::Signal;
+use libcontainer::{container::ContainerStatus, signal::Signal};
 use liboci_cli::Kill;
 
 pub fn kill(args: Kill, root_path: PathBuf) -> Result<()> {
     let mut container = load_container(root_path, &args.container_id)?;
     let signal: Signal = args.signal.as_str().try_into()?;
-    container.kill(signal, args.all)
+    match container.kill(signal, args.all) {
+        Ok(_) => Ok(()),
+        Err(e) => {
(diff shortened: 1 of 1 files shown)


crates/youki/src/commands/kill.rs
```diff
@@ -4,11 +4,21 @@ use std::{convert::TryInto, path::PathBuf};
...
 use crate::commands::load_container;
-use libcontainer::signal::Signal;
+use libcontainer::{container::ContainerStatus, signal::Signal};
 use liboci_cli::Kill;
 
 pub fn kill(args: Kill, root_path: PathBuf) -> Result<()> {
     let mut container = load_container(root_path, &args.container_id)?;
     let signal: Signal = args.signal.as_str().try_into()?;
-    container.kill(signal, args.all)
+    // container.kill(signal, args.all)
```
utam0k, line 13: I guess you forgot to remove this line.

YJDoc2: oops! Will fix that :+1:

utam0k (approved):
Thanks :100:

PR #1339, opened 2022-11-10
Ignore error when killing, if error is 'process does not exist'

YJDoc2:
Currently when killing a container, if the container process does not exist (already deleted) or when killing all processes in the container, a process exits between we getting its pid and sending the kill signal, kill command will produce error.

However, in this specific case, it is ok that killing failed, as the process to kill does not exist. So we ignore the error.

Note, I'd like the check of error code to be more idiomatic, but not sure how 

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1339"></a>

### DIFF
--- a/crates/libcontainer/src/container/container_kill.rs
+++ b/crates/libcontainer/src/container/container_kill.rs
@@ -57,7 +57,15 @@ impl Container {
         let pid = self.pid().unwrap();
 
         log::debug!("kill signal {} to {}", signal, pid);
-        signal::kill(pid, signal)?;
+        let res = signal::kill(pid, signal);
+
+        match res {
+            Err(nix::errno::Errno::ESRCH) => {
+                /* the process does not exist, which is what we want */
+            }
+            _ => res?,
+        }
+
         // For cgroup V1, a frozon process cannot respond to signals,
         // so we need to thaw it. Only thaw the cgroup for SIGKILL.
         if self.status() == ContainerStatus::Paused && signal == signal::Signal::SIGKILL {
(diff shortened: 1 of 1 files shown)


crates/libcontainer/src/container/container_kill.rs
```diff
@@ -57,7 +57,15 @@ impl Container {
...
         log::debug!("kill signal {} to {}", signal, pid);
-        signal::kill(pid, signal)?;
+        let res = signal::kill(pid, signal);
+
+        match res {
+            Err(nix::errno::Errno::ESRCH) => {
+                /* the process does not exist, which is what we want */
+            }
+            _ => res?,
+        }
```
YJDoc2, line 67: Is there a more idiomatic way of doing this?

crates/libcontainer/src/container/container_kill.rs
```diff
@@ -95,7 +103,14 @@ impl Container {
...
             log::debug!("kill signal {} to {}", signal, pid);
-            signal::kill(pid, signal)
+            let res = signal::kill(pid, signal);
+            match res {
+                Err(nix::errno::Errno::ESRCH) => {
+                    /* the process does not exist, which is what we want */
+                    Ok(())
+                }
+                _ => res,
+            }
```
YJDoc2, line 113: Now thinking, I'm not sure if we need this, as we are freezing the cgroup before getting all pids, so whatever pids we get, the process must exist, as freezing cgroup will suspend all processes in the cgroup

utam0k (commented):
May I ask you to add this test case into our integration test?

YJDoc2:
> May I ask you to add this test case into our integration test?

I'm not sure what you mean by this. If we want to add a new test to validate this, I'm not sure how would I go about doing this, because this is basically a race condition. The problem here was that between we getting list of processes and we sending the kill signal, a process from list exits on its own, making the kill call fail. As we want to kill the process anyways, and it has exited, here we just ignore that specific failure. In my opinion, it would be difficult to consistently recreate this race condition in a test.

Let me know if I'm getting this wrong

utam0k:
> > May I ask you to add this test case into our integration test?
> 
> I'm not sure what you mean by this. If we want to add a new test to validate this, I'm not sure how would I go about doing this, because this is basically a race condition. The problem here was that between we getting list of processes and we sending the kill signal, a process from list exits on its own, making the kill call fail. As we want to kill the process anyways, and it has exited, here we just ignore that specific failure. In my opinion, it would be difficult to consistently recreate this race condition in a test.
> 
> Let me know if I'm getting this wrong

You are right!

utam0k (approved):

PR #1347, opened 2022-11-15
Fix TestContainerNoBinaryExists test, by making create behaviour similar to runc

YJDoc2:
When creating, this will check if the binary to run as container process exists and if it has proper permissions in the init process.
If it doesn't, then generate an error in create.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1347"></a>

### DIFF
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -15,7 +15,6 @@ use nix::unistd::setsid;
 use nix::unistd::{self, Gid, Uid};
 use oci_spec::runtime::{LinuxNamespaceType, Spec, User};
 use std::collections::HashMap;
-use std::os::unix::fs::PermissionsExt;
 use std::os::unix::io::AsRawFd;
 use std::{
     env, fs,
@@ -395,8 +394,8 @@ pub fn container_init_process(
         }
     }
 
-    // this checks if the binary to run actually exists and if we have permissions to
-    // run it. Taken from https://github.com/opencontainers/runc/blob/main/libcontainer/standard_init_linux.go#L195-L206
+    // this checks if the binary to run actually exists and if we have permissions to run it.
+    // Taken from https://github.com/opencontainers/runc/blob/25c9e888686773e7e06429133578038a9abc091d/libcontainer/standard_init_linux.go#L195-L206
     if let Some(args) = proc.args() {
         let path_var = {
             let mut ret: &str = "";
@@ -405,7 +404,7 @@ pub fn container_init_process(
                     ret = var;
                 }
             }
-            ret.trim_start_matches("PATH=")
+            ret
         };
         let executable_path = utils::get_executable_path(&args[0], path_var);
         match executable_path {
@@ -414,12 +413,7 @@ pub fn container_init_process(
                 args[0]
             ),
             Some(path) => {
-                let metadata = path.metadata()?;
-                let permissions = metadata.permissions();
-                // we have to check if it a file, as dir have there executable bit set,
-                // and check if x in rwx is unset. Logic is taken
-                // from https://docs.rs/is_executable/latest/src/is_executable/lib.rs.html#90
-                if !metadata.is_file() || permissions.mode() & 0o001 == 0 {
+                if !utils::is_executable(&path)? {
                     bail!("file {:?} does not have executable permission set", path);
                 }
             }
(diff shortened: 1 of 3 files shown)


YJDoc2:
@utam0k I validated it with runc, and that too fails the test. The problem behind it is that we are validating the binary to be run in the init process exists, but the tests does not setup the binary at all. If we want to follow runc, we can skip this test, or we can choose to ignore the test in containerd validation. Unfortunately we have to make a call here. :cry: :sob:

utam0k:
@YJDoc2 Thanks for the great investigation. I vote for skipping it

utam0k:
@kzys May I ask you to help us? Is make integration not used in containerd? It appears that there is a test that fails even with runc. Is there something we missed?
https://containers.github.io/youki/developer/containerd_integration_test_using_youki.html

kzys:
Let me take a look this week. We are at least using `make integration`.

utam0k:
> Let me take a look this week. We are at least using `make integration`.

:pray: Thanks.

YJDoc2:
@utam0k after checking runc repo, I don't think they are using oci-integration tests anywhere in their CI or makefile. they seem to be using their self-written tests for integration.

utam0k:
> Let me take a look this week. We are at least using `make integration`.

@kzys Sorry, I thought we were talking about containerd, but we were talking about something else. I was mistaken🙇 Never mind.

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -394,6 +395,37 @@ pub fn container_init_process(
         }
     }
 
+    // this checks if the binary to run actually exists and if we have permissions to
+    // run it. Taken from https://github.com/opencontainers/runc/blob/main/libcontainer/standard_init_linux.go#L195-L206
```
utam0k, line 399:
```suggestion
    // run it. Taken from https://github.com/opencontainers/runc/blob/25c9e888686773e7e06429133578038a9abc091d/libcontainer/standard_init_linux.go#L195-L206
```

scripts/oci_integration_tests.sh
```diff
@@ -54,7 +54,7 @@ test_cases=(
   "linux_seccomp/linux_seccomp.t"
   "linux_sysctl/linux_sysctl.t"
   "linux_uid_mappings/linux_uid_mappings.t"
-  "misc_props/misc_props.t"
+  #"misc_props/misc_props.t" runc also fails this, check out https://github.com/containers/youki/pull/1347#issuecomment-1315332775
```
utam0k, line 57:
```suggestion
  # "misc_props/misc_props.t" runc also fails this, check out https://github.com/containers/youki/pull/1347#issuecomment-1315332775
```

crates/libcontainer/src/utils.rs
```diff
@@ -331,6 +331,20 @@ pub fn get_temp_dir_path(test_name: &str) -> PathBuf {
     std::env::temp_dir().join(test_name)
 }
 
+pub fn get_executable_path(name: &str, path_var: &str) -> Option<PathBuf> {
```
utam0k, line 334: May I ask you to add a unit test about `get_executable_path`?

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -394,6 +395,37 @@ pub fn container_init_process(
...
+            None => bail!(
+                "executable '{}' for container process does not exist",
+                args[0]
+            ),
+            Some(path) => {
+                let metadata = path.metadata()?;
+                let permissions = metadata.permissions();
+                // we have to check if it a file, as dir have there executable bit set,
+                // and check if x in rwx is unset. Logic is taken
+                // from https://docs.rs/is_executable/latest/src/is_executable/lib.rs.html#90
```
utam0k, line 421:
```suggestion
                // from https://docs.rs/is_executable/1.0.1/src/is_executable/lib.rs.html#90
```

utam0k: I feel this comment #L421 doesn't need.

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -394,6 +395,37 @@ pub fn container_init_process(
...
+                if !metadata.is_file() || permissions.mode() & 0o001 == 0 {
+                    bail!("file {:?} does not have executable permission set", path);
+                }
+            }
+        }
```
utam0k, line 426: How about dividing these lines up as the method to make it easy to understand?

YJDoc2: So you want me to split the finding path variable part and finding if it is executable into two functions, or some other splits?

utam0k: By making it a function, I ask you to name this part of the process. Instead of a comment :pray:

utam0k (approved):
Thanks!

PR #1349, opened 2022-11-17
Fixed set capability fail.

higuruchi:
Signed-off-by: higuruchi <fumiya2324@gmail.com>

If you use caps::all function, caps::drop function fails unexpectedly because caps::all function get capabilities that the old kernel does not support. 
Therefore, Avoid unintended failures by use caps read function that get only the kernel's corresponding capabilities.

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1349"></a>

### DIFF
--- a/crates/libcontainer/src/syscall/linux.rs
+++ b/crates/libcontainer/src/syscall/linux.rs
@@ -174,7 +174,7 @@ impl Syscall for LinuxSyscall {
             // so we do it differently
             CapSet::Bounding => {
                 // get all capabilities
-                let all = caps::all();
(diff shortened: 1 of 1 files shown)


utam0k:
> If you use caps::all function, caps::drop function fails unexpectedly because caps::all function get capabilities that the old kernel does not support.

I just am interested in which version it became available?

utam0k (approved):

higuruchi:
https://man7.org/linux/man-pages/man7/capabilities.7.html 

CAP_CHECKPOINT_RESTORE validate since Linux 5.9.
So newer than Linux5.9 become available.

PR #1352, opened 2022-11-18
Detect architecture by uname command

udzura:
Sorry for that I have only tested this in aarch64 machine. But this may be useful

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1352"></a>

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -10,7 +10,7 @@ usage_exit() {
 }
 
 VERSION=debug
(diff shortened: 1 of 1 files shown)


YJDoc2:
Currently we don't have a way to test this on not non-64 bit machines, but till we get that, this not breaking current working is good enough.
Thanks for your contribution!

YJDoc2 (approved):
lgtm :+1:

PR #1376, opened 2022-11-27
Add hostname test

chermehdi:
The test is ported from https://github.com/opencontainers/runtime-tools/blob/2802ff9ff5/validation/hostname/hostname.go

The intent is to make sure that the hostname set in the spec is actually set within the container.

Signed-off-by: chermehdi <mehdi.cheracher@gmail.com>

<a href="https://gitpod.io/#https://github.com/containers/youki/pull/1376"></a>

### DIFF
--- a/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
@@ -7,12 +7,9 @@ fn create_spec(hostname: &str) -> Spec {
     SpecBuilder::default()
         .hostname(hostname)
         .linux(
+            // Need to reset the read-only paths
             LinuxBuilder::default()
-                .readonly_paths(vec![
-                    "/proc/bus".to_string(),
-                    "/proc/fs".to_string(),
-                    "/proc/sys".to_string(),
-                ])
+                .readonly_paths(vec![])
                 .build()
                 .expect("error in building linux config"),
         )
--- a/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
@@ -1,0 +1,53 @@
+use anyhow::{Context, Result};
+use oci_spec::runtime::{LinuxBuilder, ProcessBuilder, Spec, SpecBuilder};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::utils::test_inside_container;
+
+fn create_spec(hostname: &str) -> Result<Spec> {
+    Ok(SpecBuilder::default()
+        .hostname(hostname)
+        .linux(
+            LinuxBuilder::default()
+                .readonly_paths(vec![
+                    "/proc/bus".to_string(),
+                    "/proc/fs".to_string(),
+                    "/proc/sys".to_string(),
+                ])
+                .build()
+                .expect("should build linux config"),
+        )
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["runtimetest".to_string()])
+                .build()
+                .expect("can create process"),
+        )
+        .build()
+        .context("Failed to build hostname spec")?)
+}
+
+fn hostname_test() -> TestResult {
+    let cases = vec!["hostname-specific", ""];
+    for case in cases {
+        let spec = create_spec(case).expect("should create spec");
+        let test_result = test_inside_container(spec, &|_| {
+            // As long as the container is created, we expect the hostname to be determined
+            // by the spec, so nothing to prepare prior.
+            Ok(())
+        });
+        // fail fast
+        if let TestResult::Failed(e) = test_result {
+            return TestResult::Failed(e);
+        }
+    }
+    TestResult::Passed
+}
+
+pub fn get_hostname_test() -> TestGroup {
+    let mut test_group = TestGroup::new("set_host_name");
+    let hostname_test = Test::new("set_host_name_test", Box::new(hostname_test));
+    test_group.add(vec![Box::new(hostname_test)]);
+
+    test_group
+}
--- a/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
@@ -1,11 +1,10 @@
-use anyhow::{Context, Result};
 use oci_spec::runtime::{LinuxBuilder, ProcessBuilder, Spec, SpecBuilder};
 use test_framework::{Test, TestGroup, TestResult};
 
 use crate::utils::test_inside_container;
 
-fn create_spec(hostname: &str) -> Result<Spec> {
-    Ok(SpecBuilder::default()
+fn create_spec(hostname: &str) -> Spec {
+    SpecBuilder::default()
         .hostname(hostname)
         .linux(
             LinuxBuilder::default()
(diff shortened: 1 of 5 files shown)


tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+
+fn create_spec(hostname: &str) -> Result<Spec> {
+    Ok(SpecBuilder::default()
+        .hostname(hostname)
+        .linux(
+            LinuxBuilder::default()
+                .readonly_paths(vec![
+                    "/proc/bus".to_string(),
+                    "/proc/fs".to_string(),
+                    "/proc/sys".to_string(),
```
YJDoc2, line 15: Why are we setting these specifically to ro? Is there anything in these tests that needs this?

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+                ])
+                .build()
+                .expect("should build linux config"),
```
YJDoc2, line 18:
```suggestion
                .expect("error in building linux config"),
```

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+        )
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["runtimetest".to_string()])
+                .build()
+                .expect("can create process"),
```
YJDoc2, line 24:
```suggestion
                .expect("error in creating process config"),
```

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+        )
+        .build()
+        .context("Failed to build hostname spec")?)
```
YJDoc2, line 27:
```suggestion
        .context("failed to build hostname spec")?)
```
The error messages should always be lowercase :sweat_smile: 
Also, we can just unwrap here, as failing this should fail the case anyways ( and realistically this should never fail)

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+    TestResult::Passed
+}
+
+pub fn get_hostname_test() -> TestGroup {
+    let mut test_group = TestGroup::new("set_host_name");
+    let hostname_test = Test::new("set_host_name_test", Box::new(hostname_test));
+    test_group.add(vec![Box::new(hostname_test)]);
+
+    test_group
+}
```
YJDoc2, line 53: It is preferable that we don't loop over test cases as in the original ones. Instead, please make two separate functions for each and add them to test group. It sometimes get pretty confusion in the original source (oci-tests) to see what tests exactly are there, and what are their expected outcomes. Also separating functions will make it easier if we ever want to modify the case later.

Also, after that we can simply return the result of `test_inside_container` instead of checking for error

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+                .expect("can create process"),
+        )
+        .build()
+        .context("Failed to build hostname spec")?)
+}
+
+fn hostname_test() -> TestResult {
+    let cases = vec!["hostname-specific", ""];
+    for case in cases {
+        let spec = create_spec(case).expect("should create spec");
```
YJDoc2, line 33:
```suggestion
        let spec = create_spec(case).expect("error in creating spec");
```
Alternatively unwrap in the `create_spec` as suggested in other comment

YJDoc2 (changes requested):
Hey, I have suggested some changes, please take a look

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
chermehdi: Without setting at least one read-only file, the test fails because of [this](https://github.com/containers/youki/blob/main/tests/rust-integration-tests/runtimetest/src/tests.rs#L14)

And if I leave the default configuration which is exactly the files I added and `/proc/sysrq-trigger` which when tried on my machine fails with a permission error [on](https://github.com/containers/youki/blob/main/tests/rust-integration-tests/runtimetest/src/tests.rs#L29-L32).

I am not sure if this is related to my setup (My machine is running a recent ubuntu version with a `5.15` kernel), and I am also not sure why there is a test to enforce the existence of read-only paths (I understand why do we need them, but it would've been nicer to be able to set the read_only_paths to `vec![]`).

chermehdi: To be exact, this is the error I get when running with the default configuration.

```
container stderr was not empty, found : in readonly paths, error in testing read access for path /proc/sysrq-trigger : Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
```

YJDoc2: Hmm, this seems to be issue with my implementation of ro paths tests then... I'll take a look at this later and update back. Thanks for the detailed info of errors :)

YJDoc2: Hey, so I checked on my system, and removing lines 9->18 works fine, I think there might be some issue with setup of sysreq-trigger on your side. Can you check the permissions of that file / if the file exists or not?

YJDoc2:
@chermehdi I have updated my comment, please take a look

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
YJDoc2: Hey, I didn't add any vagrant-like base because primarily :
- I wanted these tests to run without needing extra stuff, and having minimal dependencies on env (not a good thing in hindsight)
- Up till now, there was no such issue reported by anyone who did update the tests. I personally have ran this on plain ubuntu and Pop OS and both worked fine. (again, not the best thing in hindsight)

That all said, if there is no issue with your setup, and the error is still present, we would implement a vagrant based solution, as already done for containerd tests :+1:

chermehdi: @YJDoc2 update: the file does exist with permissions `-w------- | 200`, so it is not readable, the test will always fail on my machine regardless.

YJDoc2: Interestingly, it now also fails on my machine, so I suspect I was using old build of tests when I reported that it works. However, you are correct in this, and can I ask you to do following changes?
- In readonly paths test, silently return instead of showing error when no paths are set (in both cases)
- in hostname test, set ro paths to either None or empty vector

I feel this is a better solution than setting some specific ro paths in a hostname test and explaining via comment, as well as will also help in future tests
WDYT?

YJDoc2:
Hey, mostly it is correct, but instead of removing the condition altogether, can you just return from the if-block? It will also indicate that we meant to return if no paths were given in the test instead of a potentially ambiguous the case of empty paths is not considered if someone has to update the that test later.

Also, please amend the last commit instead of adding new commit for this :smiley:

chermehdi:
That's what I had initially, but it's moot, since the loop below wouldn't run if the ro_paths is empty so no extra tests would run, the behavior is as expected with _less_ code.

about commits, I don't really put too much attention on amending commits as I assume we will squash all of them at the end of the day when merging, It just gives a better history line when looking at the PR for what changes happened and why/when.

YJDoc2:
> That's what I had initially, but it's moot, since the loop below wouldn't run if the ro_paths is empty so no extra tests would run, the behavior is as expected with less code.

I can understand that the loop will not run if empty array, but we should have some sign that we have considered the empty array case, and the behavior is intentional. Also, we usually don't squash commits, but in this case, sure I'll squash this PR for merge

YJDoc2 (approved):
lgtm :+1: Thanks for your contribution!

PR #1383, opened 2022-11-29
Change targets in the makefile

### LINKED ISSUE #1267: Make scripts should be more selective
Currently there are following make targets :
- build
- release-build
- test-all
- test
- oci-integration-test
- integration-test
- validate-rust-tests
- clean

One of the big issue is that the build and release-build scripts currently compile 3 binaries : youki, integration-test (rust) and runtimetest. 
This not only takes time, but is completely unnecessary in many cases : for release, tests should not be compiled at all, and unless one is developing/adding integration-tests, most of the times they are not needed until very end.

Unfortunately the problem is worse due to runtimetest. It has to be compiled as a statically linked binary with some compile flags, and because of that the cached intermediate files have to be redone for runtimetest and again for next youki compilation.

Ideally we should have separate atomic scripts to compile youki, integration tests and runtime tests, and the make targets should also be separated, where `all` target will build everything, but otherwise we can select which binary to build.

While doing this, also consider changing git url from ssh to https to resolve one of the issues faced in #1212 .

YJDoc2:
Should resolve #1267


YJDoc2:
!github easy-benchmark

.gitignore
```diff
@@ -18,4 +19,4 @@ tags.temp
 *~
 
 /bundle.tar.gz
-/test.log
+/test.log
```
utam0k, line 22:
```suggestion
/test.log

```

scripts/rust_integration_tests.sh
```diff
@@ -24,7 +24,7 @@ if [ ! -f ${ROOT}/bundle.tar.gz ]; then
 fi
 touch ${LOGFILE}
 
-sudo YOUKI_LOG_LEVEL="error" ${ROOT}/integration_test run --runtime "$RUNTIME" --runtimetest ${ROOT}/runtimetest > $LOGFILE
+sudo YOUKI_LOG_LEVEL="error" ${ROOT}/integration_test run --runtime "$RUNTIME" --runtimetest ${ROOT}/runtimetest
```
YJDoc2, line 27: This should be reverted and kept as original :sweat_smile:

Makefile
```diff
@@ -1,33 +1,51 @@
...
+
+test-all: unittest featuretest oci-tests containerd-test # currently not doing rust-oci here
+
+# Misc
 
+lint:
+	cargo clippy --all-targets --all-features
+
+clean:
+	./scripts/clean.sh $(ROOT)
```
utam0k, line 51:
```suggestion
	./scripts/clean.sh $(ROOT)
	
```

Makefile
```diff
@@ -1,33 +1,51 @@
...
-	./scripts/build.sh -o $(ROOT) -r
+youki: youki-dev # helper
 
-test-all: test oci-integration-test integration-test features-test
+youki-dev:
+	./scripts/build.sh -o $(ROOT) -c youki
 
-test: build
-	cd crates && cargo test
+youki-release:
```
utam0k, line 10: I think most people expect the `build` subcommand and it builds youki itself. Why not have an alias as a helper? I'd rather just have `build`.
How about using an environment variable to switch between release and non-release?

YJDoc2: Would it be fine to have build similar to how youki is here? that way `make build` , `make youki` and `make youki-dev` would do the same thing. Alternatively, we can alias `build` to `youki-release` as well. WDYT?

YJDoc2:
@utam0k PTAL now

utam0k (approved):

PR #1389, opened 2022-12-01
Check capabilities in youki info subcommand

udzura:
Shows supported capabilities that started to be supported in recent kernel release.

At first I tried to call `caps::drop()` actually for each newly added caps, but it required privilege. So I just check these caps from default `CapSet::Bounding`.

### DIFF
--- a/crates/youki/src/commands/info.rs
+++ b/crates/youki/src/commands/info.rs
@@ -4,7 +4,6 @@ use std::collections::HashSet;
 use std::{fs, path::Path};
 
 use anyhow::Result;
-use caps::Capability;
 use clap::Parser;
 use libcontainer::rootless;
 use procfs::{CpuInfo, Meminfo};
--- a/crates/youki/src/commands/info.rs
+++ b/crates/youki/src/commands/info.rs
@@ -224,17 +224,18 @@ pub fn print_namespaces() {
     }
 }
 
+#[inline]
+fn get_cap_supported(caps: &caps::CapsHashSet, cap: caps::Capability) -> &'static str {
+    if caps.contains(&cap) {
+        "available"
+    } else {
+        "unsupported"
+    }
+}
+
 pub fn print_capabilities() {
     println!("Capabilities");
     if let Ok(current) = caps::read(None, caps::CapSet::Bounding) {
-        fn get_cap_supported(caps: &caps::CapsHashSet, cap: caps::Capability) -> &'static str {
-            if caps.contains(&cap) {
-                "available"
-            } else {
-                "unsupported"
-            }
-        }
-
         println!(
             "{:<17} {}",
             "CAP_BPF",
--- a/crates/youki/src/commands/info.rs
+++ b/crates/youki/src/commands/info.rs
@@ -225,11 +225,11 @@ pub fn print_namespaces() {
 }
 
 #[inline]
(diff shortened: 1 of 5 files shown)


YJDoc2:
Hey, the CI is failing due to a lint error
Also, is there any related issue/reasoning behind this? Particularly if we are printing the caps, why check only the newly supported ones for printing, why not all that are supported?

crates/youki/src/commands/info.rs
```diff
@@ -223,6 +225,37 @@ pub fn print_namespaces() {
...
+pub fn print_capabilities() {
+    println!("Capabilities");
+    if let Ok(current) = caps::read(None, caps::CapSet::Bounding) {
+        fn get_cap_supported(caps: &caps::CapsHashSet, cap: caps::Capability) -> &'static str {
+            if caps.contains(&cap) {
+                "available"
+            } else {
+                "unsupported"
+            }
+        }
```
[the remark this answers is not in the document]
udzura: It make sense! I'll move it to top level.

YJDoc2:
>Also, is there any related issue/reasoning behind this? Particularly if we are printing the caps, why check only the newly supported ones for printing, why not all that are supported?

Hey @udzura apologies for the delay, but can you clarify the above? Apart from that this PR is good to go :+1:

udzura:
@YJDoc2 

Apologies, just find the questions!

> is there any related issue/reasoning behind this?

No, but if anything, I just encountered the same error as referred in https://github.com/containers/youki/pull/1349 . If `youki info` shows these capabilities, we can fix this kind of errors with much more ease.

> why not all that are supported

I've considered this in advance, but I find there are so many capability flags to show in solo command. Maybe a special flag such as `youki info --capabilities` is required.

crates/youki/src/commands/info.rs
```diff
@@ -223,6 +224,38 @@ pub fn print_namespaces() {
     }
 }
 
+#[inline]
+fn get_cap_supported(caps: &caps::CapsHashSet, cap: caps::Capability) -> &'static str {
+    if caps.contains(&cap) {
+        "available"
+    } else {
+        "unsupported"
```
YJDoc2, line 232: Can we  have either `supported` and `unsupported` or `available` and `unavailable`? I thought on it and it feels it'd be better if we have one of these two

crates/youki/src/commands/info.rs
```diff
@@ -223,6 +224,38 @@ pub fn print_namespaces() {
     }
 }
 
+#[inline]
+fn get_cap_supported(caps: &caps::CapsHashSet, cap: caps::Capability) -> &'static str {
```
YJDoc2, line 228: can we rename this to `is_cap_supported` or`is_cap_available`  depending on how we are wording the output?

YJDoc2 (changes requested):
Hey, Apologies, can I ask you for two small changes?

YJDoc2:
Hey @udzura , would it be possible for you to make the changes? If you are busy, I'd be happy to push a commit in this PR and merge. Please let me know :)

udzura:
@YJDoc2 Very sorry for being late! I've applied suggestions.

YJDoc2 (approved):
lgtm :+1: Thanks for your contribution!

PR #1402, opened 2022-12-10
Add support for wasmtime

Furisto:
This adds wasmtime as a new supported wasm runtime. For test steps [check the docs](https://containers.github.io/youki/user/webassembly.html) and compile youki with `feature=wasm-wasmtime`.

Fix: https://github.com/containers/youki/issues/1278

### DIFF
--- a/tools/wasm-sample/src/main.rs
+++ b/tools/wasm-sample/src/main.rs
@@ -1,0 +1,11 @@
+fn main() {
+    println!("Printing args");
+    for arg in std::env::args().skip(1) {
+        println!("{}", arg);
+    }
+
+    println!("Printing envs");
+    for envs in std::env::vars() {
+        println!("{:?}", envs);
+    }
+}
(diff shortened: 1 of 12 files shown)


tools/wasm-sample/src/main.rs
```diff
@@ -0,0 +1,11 @@
+fn main() {
```
utam0k, line 1: May I ask you to update or create the docs for this tool? How can I use it?

Furisto: Added [README](https://github.com/containers/youki/blob/84ab1655ae49c1bc09ccd1c266cd811f58ca9998/tools/wasm-sample/README.md)

utam0k: Thanks :heart_eyes:

utam0k (approved):
Awesome

PR #1403, opened 2022-12-11
libcontainer: Make the workloads injectable

utam0k:
It allows flexibility by allowing libcontainer users to create and inject workloads.
For example, users of libcontainer can add their favorite wasm runtime and not use the normal workloads.

### DIFF
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -1,3 +1,4 @@
+use crate::workload::{Executor, ExecutorManager};
 use crate::{syscall::Syscall, utils::PathBufExt};
 use anyhow::{anyhow, Context, Result};
 use std::path::PathBuf;
@@ -18,6 +19,8 @@ pub struct ContainerBuilder<'a> {
     pub(super) console_socket: Option<PathBuf>,
     /// File descriptors to be passed into the container process
     pub(super) preserve_fds: i32,
+    /// TODO: Comment
+    pub(super) executor_manager: ExecutorManager,
 }
 
 /// Builder that can be used to configure the common properties of
@@ -28,8 +31,13 @@ pub struct ContainerBuilder<'a> {
 /// ```no_run
 /// use libcontainer::container::builder::ContainerBuilder;
 /// use libcontainer::syscall::syscall::create_syscall;
+/// use libcontainer::workload::default::DefaultExecutor;
 ///
-/// ContainerBuilder::new("74f1a4cb3801".to_owned(), create_syscall().as_ref())
+/// ContainerBuilder::new(
+///     "74f1a4cb3801".to_owned(),
+///     create_syscall().as_ref(),
+///     vec![Box::new(DefaultExecutor::default())],
+/// )
 /// .with_root_path("/run/containers/youki").expect("invalid root path")
 /// .with_pid_file(Some("/var/run/docker.pid")).expect("invalid pid file")
 /// .with_console_socket(Some("/var/run/docker/sock.tty"))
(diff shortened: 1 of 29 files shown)


utam0k:
@Furisto May I ask you to review this one?

utam0k:
@Furisto May I ask you to review this PR?

crates/libcontainer/src/container/builder.rs
```diff
@@ -18,6 +22,8 @@ pub struct ContainerBuilder<'a> {
     pub(super) console_socket: Option<PathBuf>,
     /// File descriptors to be passed into the container process
     pub(super) preserve_fds: i32,
+    /// TODO: Comment
```
Furisto, line 25: :point_up:

crates/youki/src/commands/run.rs
```diff
@@ -4,16 +4,22 @@ use anyhow::{Context, Result};
...
-        .with_console_socket(args.console_socket.as_ref())
-        .with_root_path(root_path)?
-        .with_preserved_fds(args.preserve_fds)
-        .as_init(&args.bundle)
-        .with_systemd(systemd_cgroup)
-        .build()?;
+    let mut container = ContainerBuilder::new(
+        args.container_id.clone(),
+        syscall.as_ref(),
+        default_executors(),
```
Furisto, line 14: I would simplify this. The idea behind the `new` method is that we specify all the arguments that are required and then use the `with` methods on the builder to customize optional behavior. A consumer of the API should not know what this executor thing is unless they specifically seek out to change them. So remove the executors from new and add a `with_executors` method that would replace the default executor with the specified executors.

crates/libcontainer/src/workload/mod.rs
```diff
@@ -1,51 +1,32 @@
...
+impl ExecutorManager {
+    pub fn exec(&self, spec: &Spec) -> Result<()> {
+        for executor in self.executors.iter() {
+            if executor.can_handle(spec)? {
+                return executor.exec(spec).context("execution failed");
+            }
         }
-
-        DefaultExecutor::exec(spec).context("default execution failed")
+        bail!("cannot find executor")
```
Furisto, line 30:
```suggestion
        bail!("cannot find an executor that satisfies all requirements")
```

crates/libcontainer/src/workload/mod.rs
```diff
@@ -1,51 +1,32 @@
...
+pub struct ExecutorManager {
+    pub executors: Vec<Box<dyn Executor>>,
+}
 
-        #[cfg(feature = "wasm-wasmtime")]
-        if WasmtimeExecutor::can_handle(spec)? {
-            return WasmtimeExecutor::exec(spec).context("wasmtime execution failed");
+impl ExecutorManager {
+    pub fn exec(&self, spec: &Spec) -> Result<()> {
+        for executor in self.executors.iter() {
```
Furisto, line 25: Check if there are zero executors and return with an error message.

crates/libcontainer/src/process/args.rs
```diff
@@ -35,4 +36,6 @@ pub struct ContainerArgs<'a> {
     pub cgroup_manager: Box<dyn CgroupManager>,
     /// If the container is to be run in detached mode
     pub detached: bool,
+    /// TODO: Comment
```
Furisto, line 39: :point_up:

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -43,6 +44,8 @@ pub(super) struct ContainerBuilderImpl<'a> {
     pub preserve_fds: i32,
     /// If the container is to be run in detached mode
     pub detached: bool,
+    /// TODO: Comment
```
Furisto, line 47: :point_up:

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -442,7 +441,8 @@ pub fn container_init_process(
     }
 
     if proc.args().is_some() {
-        ExecutorManager::exec(spec)
+        args.executor_manager.exec(spec)?;
+        unreachable!("process image should have been replaced after exec");
```
Furisto, line 445: The image will not have been replaced if you are using wasm to execute the workload and youki will crash in that case.

utam0k:
@Furisto PTAL

Furisto (approved):

PR #1406, opened 2022-12-12
skip cgroup v2 test of oci-tools

utam0k:
Ignore the failure because it cannot be tested in a cgroup v2 environment. For example, it fails with ubuntu 22.04.

### DIFF
--- a/scripts/oci_integration_tests.sh
+++ b/scripts/oci_integration_tests.sh
@@ -84,6 +84,11 @@ check_environment() {
         return 1
     fi
   fi
+  if [[ $test_case == "delete_only_create_resources/delete_only_create_resources.t" ]]; then
+    if [[ ! -e "/sys/fs/cgroup/pids/cgrouptest/tasks" ]]; then
+        return 1
+    fi
+  fi
 }
 
 if [[ ! -e $RUNTIME ]]; then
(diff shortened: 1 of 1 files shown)


scripts/oci_integration_tests.sh
```diff
@@ -116,8 +121,12 @@ for case in "${test_cases[@]}"; do
...
   sudo RUST_BACKTRACE=1 RUNTIME=${RUNTIME} ${OCI_TEST_DIR}/validation/$case >$logfile 2>&1 || (cat $logfile && exit 1)
   if [ 0 -ne $(grep "not ok" $logfile | wc -l ) ]; then
-      cat $logfile
-      exit 1
+    if [ 0 -eq $(grep "# cgroupv2 is not supported yet " $logfile | wc -l ) ]; then
+      echo "Skip $case bacause oci-runtime-tools doesn't support cgroup v2"
+      continue;
+    fi
+    cat $logfile
+    exit 1
```
YJDoc2, line 129: Hey, I'm not sure I understand the purpose of this part. If we are checking for the test in `check_environment` function above, why do we need this here? Also, shouldn't the `0 -eq ...` part be `0 -ne ...` instead? If the grep finds occurrence, then wc will output a non-zero number, in which case we should be entering the condition, right?

utam0k: The logic is that if the test fails and the cause of the failure is not supported by the test in cgroup v2, the failure is ignored. I could have used check_environment, but I added this because of the uniform error messages.

PR #1398, opened 2022-12-06
Support recursive mount attrs by using mount_setattr(2).

higuruchi:
Implement "rro" mount option that implemented in runc(https://github.com/opencontainers/runc/pull/3272) and crun(https://github.com/containers/crun/pull/778).

This mount option "rro" makes the mount point recursively read-only, by calling mount_setattr(2) with MOUNT_ATTR_RDONLY and AT_RECURSIVE.

Signed-off-by: higuruchi <fumiya2324@gmail.com>

PR #1428, opened 2022-12-19
Added recursive mount attr test

higuruchi:
Added test of rro mount option that implemented in #1398.
`/tmp/` is used as the bind mount path used for testing.

Now, this PR implement rro option test only.
And if this test is correct, add other option test(rrw, rnosuid...).

- [x] rro
- [ ] rrw
- [x] rnosuid
- [ ] rsuid
- [ ] rnodev
- [ ] rdev
- [x] rnoexec
- [ ] rexec
- [ ] rnodiratime
- [ ] rdiratime
- [ ] rrelatime
- [ ] rnorelatime
- [ ] rnoatime
- [ ] ratime
- [ ] rstrictatime
- [ ] rnostrictatime
- [ ] rnosymfollow
- [ ] rsymfollow

Signed-off-by: higuruchi <fumiya2324@gmail.com>

### DIFF
--- a/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
@@ -15,7 +15,7 @@ fn create_spec(hostname: &str) -> Spec {
         )
         .process(
             ProcessBuilder::default()
-                .args(vec!["runtimetest".to_string()])
+                .args(vec!["runtimetest".to_string(), "set_host_name".to_string()])
                 .build()
                 .expect("error in creating process config"),
         )
--- a/tests/rust-integration-tests/runtimetest/src/main.rs
+++ b/tests/rust-integration-tests/runtimetest/src/main.rs
@@ -1,28 +1,9 @@
 mod tests;
 mod utils;
 
-use clap::Parser;
 use oci_spec::runtime::Spec;
 use std::path::PathBuf;
-
-#[derive(Parser, Debug)]
-#[clap(version = "0.0.1", author = "youki team")]
-struct Opts {
-    #[clap(subcommand)]
-    command: SubCommand,
-}
-
-#[derive(Parser, Debug)]
-enum SubCommand {
-    #[clap(name = "readonly_paths")]
-    ReadonlyPaths,
-
-    #[clap(name = "set_host_name")]
-    SetHostName,
-
-    #[clap(name = "mounts_recursive")]
-    MountsRecursive,
-}
+use std::env;
 
 const SPEC_PATH: &str = "/config.json";
 
@@ -38,12 +19,17 @@ fn get_spec() -> Spec {
 }
 
 fn main() {
-    let opts: Opts = Opts::parse();
     let spec = get_spec();
-
-    match opts.command {
-        SubCommand::ReadonlyPaths => tests::validate_readonly_paths(&spec),
-        SubCommand::SetHostName => tests::validate_hostname(&spec),
-        SubCommand::MountsRecursive => tests::validate_mounts_recursive(&spec),
+    let args: Vec<String> = env::args().collect();
+    let execute_test = match args.get(1) {
+        Some(execute_test) => execute_test.to_string(),
+        None => return eprintln!("error due to execute test flag not found"),
     };
+    
+    match &*execute_test {
+        "readonly_paths" => tests::validate_readonly_paths(&spec),
+        "set_host_name" => tests::validate_hostname(&spec),
+        "mounts_recursive" => tests::validate_mounts_recursive(&spec),
+        _ => eprintln!("error due to unexpected execute test flag: {}", execute_test),
+    }
 }
--- a/tests/rust-integration-tests/runtimetest/src/main.rs
+++ b/tests/rust-integration-tests/runtimetest/src/main.rs
@@ -23,13 +23,13 @@ fn main() {
     let args: Vec<String> = env::args().collect();
     let execute_test = match args.get(1) {
         Some(execute_test) => execute_test.to_string(),
-        None => return eprintln!("error due to execute test flag not found"),
+        None => return eprintln!("error due to execute test name not found"),
     };
     
     match &*execute_test {
         "readonly_paths" => tests::validate_readonly_paths(&spec),
         "set_host_name" => tests::validate_hostname(&spec),
         "mounts_recursive" => tests::validate_mounts_recursive(&spec),
-        _ => eprintln!("error due to unexpected execute test flag: {}", execute_test),
+        _ => eprintln!("error due to unexpected execute test name: {}", execute_test),
     }
 }
--- a/tests/rust-integration-tests/runtimetest/src/main.rs
+++ b/tests/rust-integration-tests/runtimetest/src/main.rs
@@ -1,8 +1,28 @@
 mod tests;
 mod utils;
 
+use clap::Parser;
 use oci_spec::runtime::Spec;
 use std::path::PathBuf;
+
+#[derive(Parser, Debug)]
+#[clap(version = "0.0.1", author = "youki team")]
+struct Opts {
+    #[clap(subcommand)]
+    command: SubCommand,
+}
+
+#[derive(Parser, Debug)]
+enum SubCommand {
+    #[clap(name = "readonly_paths")]
+    ReadonlyPaths,
+
+    #[clap(name = "set_host_name")]
+    SetHostName,
+
+    #[clap(name = "mounts_recursive")]
+    MountsRecursive,
+}
 
 const SPEC_PATH: &str = "/config.json";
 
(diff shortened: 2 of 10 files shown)


tests/rust-integration-tests/runtimetest/src/tests.rs
```diff
@@ -76,3 +79,51 @@ pub fn validate_hostname(spec: &Spec) {
...
+            do_test_mounts_recursive(d.path().as_path(), test_fn)?;
+        }
+
+        if f_type.is_file() {
+            test_fn(d.path().as_path())?;
+        }
+    }
+
+    Ok(())
+}
```
YJDoc2, line 99: 1. don't use doc-comment for function, just use `//`
2. please improve naming,  eg, use `dirs` instead of `rd`, `dir` instead of `d` and so on

tests/rust-integration-tests/runtimetest/src/tests.rs
```diff
@@ -76,3 +79,51 @@ pub fn validate_hostname(spec: &Spec) {
...
+
+pub fn validate_mounts_recursive(spec: &Spec) {
+    if let Some(mounts) = spec.mounts() {
+        for m in mounts {
+            if let Some(options) = m.options() {
+                for o in options {
+                    match o.as_str() {
```
YJDoc2, line 106: Similarly improve naming here as well

tests/rust-integration-tests/runtimetest/src/tests.rs
```diff
@@ -76,3 +79,51 @@ pub fn validate_hostname(spec: &Spec) {
...
+                        "rro" => {
+                            if let Err(e) =
+                                do_test_mounts_recursive(m.destination(), &|test_file_path| {
+                                    if utils::test_write_access(test_file_path.to_str().unwrap())
+                                        .is_ok()
+                                    {
+                                        // Return Err if writeable
+                                        bail!("Unexpectedl writeable");
```
YJDoc2, line 114:
```suggestion
                                        bail!("path {:?} expected to be read-only, found writable",test_file_path);
```

tests/rust-integration-tests/runtimetest/src/tests.rs
```diff
@@ -76,3 +79,51 @@ pub fn validate_hostname(spec: &Spec) {
...
+                                    }
+                                    Ok(())
+                                })
+                            {
+                                eprintln!("mounts_recursive rro error: {}", e);
```
YJDoc2, line 119:
```suggestion
                                eprintln!("error in testing rro recursive mounting : {}", e);
```

tests/rust-integration-tests/runtimetest/src/tests.rs
```diff
@@ -76,3 +79,51 @@ pub fn validate_hostname(spec: &Spec) {
...
+                            }
+                        }
+                        "rrw" => { /* TODO... */ }
```
YJDoc2, line 122: Otherwise test looks fine, so please add other tests in this PR as well :+1:

tests/rust-integration-tests/runtimetest/src/main.rs
```diff
@@ -1,9 +1,29 @@
...
+enum SubCommand {
+    #[clap(name = "readonly_paths")]
+    ReadonlyPaths,
+
+    #[clap(name = "set_host_name")]
+    SetHostName,
+
+    #[clap(name = "mounts_recursive")]
+    MountsRecursive,
+}
```
YJDoc2, line 25: What is the reason for adding this? I'm not saying this is a bad idea, but why do we need this?
Also, rather than adding clap just to parse this, how about , for not, `match` on argv[1], and directly compare with string? That way we won't need this whole library.

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -15,7 +15,7 @@ fn create_spec(hostname: &str) -> Spec {
         )
         .process(
             ProcessBuilder::default()
-                .args(vec!["runtimetest".to_string()])
+                .args(vec!["runtimetest".to_string(), "set_host_name".to_string()])
```
YJDoc2, line 18: Why this change (same question as adding clap in runtimetest)?

YJDoc2 (commented):
Hey, overall seems fine, there are some changes suggested and a couple of questions. Please take a look, and add tests for remaining mount types as well.

tests/rust-integration-tests/runtimetest/src/main.rs
higuruchi: >What is the reason for adding this? I'm not saying this is a bad idea, but why do we need this?

As the number of test items increases, the number of functions executed in the container also increases. I added it because I thought it would be better to execute only a specific function in order to finish the test quickly.
If this is not the correct way to test, revert this to its original state.

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
higuruchi: Same above comment.

tests/rust-integration-tests/runtimetest/src/tests.rs
higuruchi: fixed

tests/rust-integration-tests/runtimetest/src/tests.rs
higuruchi: fixed

tests/rust-integration-tests/runtimetest/src/tests.rs
higuruchi: fixed

tests/rust-integration-tests/runtimetest/src/tests.rs
higuruchi: fixed

YJDoc2:
Hey @higuruchi sorry I haven't looked at this in a while. Is there any issue you are facing? If you are busy and cannot work on this, it is ok too. We can add the partial tests and add another task for remaining.

If there is any issue, feel free to ping or mention!

tests/rust-integration-tests/runtimetest/src/main.rs
YJDoc2: Hey @higuruchi , the idea to run specific function is great,  and we should definitely do it. However, the `runtimetest` binary is statically compiled and links all linking dependencies directly into the binary. Which means if we add clap, all of clap's linking dependencies will be statically compiled into the binary, which will make it larger in size.

As we just need to match the string, and not do any fancy parsing, can we instead just do
```rust
match argv[1]{
"readonly_paths" =>{/*...*/}
"set_host_name" =>{/*...*/}
"mount_recursive" =>{/*...*/}
_ => {panic!(....)}
}
```
That way we won't need clap

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
YJDoc2: Makes sense, just see reply to the above.

tests/rust-integration-tests/runtimetest/src/main.rs
higuruchi: That's good.
thank you for your advice

higuruchi:
@YJDoc2 
Also, I am unlikely to be able to work on it for a few weeks.
After fixing the clap problem, can I issue the rest?

tests/rust-integration-tests/runtimetest/src/main.rs
```diff
@@ -38,12 +19,17 @@ fn get_spec() -> Spec {
...
     let spec = get_spec();
-
-    match opts.command {
-        SubCommand::ReadonlyPaths => tests::validate_readonly_paths(&spec),
-        SubCommand::SetHostName => tests::validate_hostname(&spec),
-        SubCommand::MountsRecursive => tests::validate_mounts_recursive(&spec),
+    let args: Vec<String> = env::args().collect();
+    let execute_test = match args.get(1) {
+        Some(execute_test) => execute_test.to_string(),
+        None => return eprintln!("error due to execute test flag not found"),
```
YJDoc2, line 26: nitpick : instead of `test flag` can we make it `test name` as it is not technically a flag . Same for the error message below this.

YJDoc2:
Hey @higuruchi No worries. If you are fine with it , we can merge this now, and when you  get time you can come back and work on the rest if you want. In the meantime, if anyone else picks up the missing cases, we can progress that way too.
Thanks for setting up the basic test and doing all this work :heart:

higuruchi:
@YJDoc2 
> If you are fine with it , we can merge this now
Please merge if there is no problem.
And i will add more tests as time permits.

YJDoc2:
Hey @higuruchi , sure, the PR seems mostly fine , its just that I didn't have time to take a final look.  I'll try to merge this today.
>Please merge if there is no problem.

YJDoc2 (approved):
lgtm :+1:

YJDoc2:
Hey @higuruchi , I'm merging this. Thank you for the contribution !

higuruchi:
@YJDoc2 May I raise an issue showing the rest of the tests?

Issue #1462, opened 2023-01-09
libcontainer fails to compile with musl


krisnova:
Compile with musl

```
cargo install --target x86_64-unknown-linux-musl
```

Cargo.toml

```
libcontainer = "0.0.4"
```

Output

```
error[E0308]: mismatched types
   --> /home/nova/.cargo/registry/src/github.com-1ecc6299db9ec823/libcontainer-0.0.4/src/syscall/linux.rs:228:44
    |
228 |         let res = unsafe { libc::setrlimit(rlimit.typ() as u32, rlim) };
    |                            --------------- ^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u32`
    |                            |
    |                            arguments to this function are incorrect
    |
note: function defined here
   --> /home/nova/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.139/src/unix/linux_like/linux/musl/mod.rs:715:12
    |
715 |     pub fn setrlimit(resource: ::c_int, rlim: *const ::rlimit) -> ::c_int;
    |            ^^^^^^^^^
help: you can convert a `u32` to an `i32` and panic if the converted value doesn't fit
    |
228 |         let res = unsafe { libc::setrlimit((rlimit.typ() as u32).try_into().unwrap(), rlim) };
    |                                            +                   +++++++++++++++++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `libcontainer` due to previous error
warning: build failed, waiting for other jobs to finish...
```

YJDoc2:
Hey, @krisnova Can you specify your system setup and how can I reproduce this? I am working on x86 gnu c system, and don't have much exp in cross-compiling rust :sweat_smile: 
Once I can get this compiling with this error, I can try fixing it :)

PR #1472, opened 2023-01-11
Compile libcontainer with musl static build

krisnova:
Hey @utam0k 🦀 😄 I love your beautiful work.

Can we please discuss #1462 and https://github.com/aurae-runtime/aurae/issues/251 in this pull request? 

This small patch will address the issue I am seeing when I compile with MUSL, however I use `.unwrap()` which is not the best practices. 

Do you have thoughts on implementation? Can I provide more detail on how to replicate this on your machine?

```bash 
git clone git@github.com:aurae-runtime/aurae.git
cd aurae
cargo install --target x86_64-unknown-linux-musl --path ./auraed
```

Arch Linux with 6.1.1 kernel on X86 architecture because I am basic.

```bash
uname -r
6.1.1-arch1-1

rustc --version
rustc 1.66.0 (69f9c33d7 2022-12-12)

cargo --version
cargo 1.66.0 (d65d197ad 2022-11-15)

musl-gcc --version
gcc (GCC) 12.2.0

musl-clang --version
clang version 14.0.6
Target: x86_64-pc-linux-gnu
```

There are other dependencies that `libcontainer` tries to depend on from the gnu collection. Primarily:

 - `libdbus`
 - `libseccomp`

Both of which prevent me from linking with musl. If possible I would like to work towards finding a way for `libcontainer` to build without gnu collection? Otherwise I might need to fork youki 😞 

Signed-off-by: Kris Nóva <kris@nivenly.com>

### DIFF
--- a/crates/libcontainer/src/syscall/linux.rs
+++ b/crates/libcontainer/src/syscall/linux.rs
@@ -374,7 +374,7 @@ impl Syscall for LinuxSyscall {
             rlim_cur: rlimit.soft(),
             rlim_max: rlimit.hard(),
         };
-        let res = unsafe { libc::setrlimit(rlimit.typ() as u32, rlim) };
+	let res = unsafe { libc::setrlimit((rlimit.typ() as u32).try_into().unwrap(), rlim) };
         if let Err(e) = Errno::result(res).map(drop) {
             bail!("Failed to set {:?}. {:?}", rlimit.typ(), e)
         }


crates/libcontainer/src/syscall/linux.rs
```diff
@@ -374,7 +374,7 @@ impl Syscall for LinuxSyscall {
             rlim_cur: rlimit.soft(),
             rlim_max: rlimit.hard(),
         };
-        let res = unsafe { libc::setrlimit(rlimit.typ() as u32, rlim) };
+	let res = unsafe { libc::setrlimit((rlimit.typ() as u32).try_into().unwrap(), rlim) };
```
krisnova, line 377: Suggestion:

```rust
    let res = {
        #[cfg(target_env = "musl")]
        unsafe {
            libc::setrlimit(rlimit.typ() as i32, rlim)
        }
        #[cfg(not(target_env = "musl"))]
        unsafe {
            libc::setrlimit(rlimit.typ() as u32, rlim)
        }
    };
```

YJDoc2:
Hey @krisnova format check is failing, please run `cargo fmt` once and re push :)

Furisto:
Hi @krisnova, there is already a feature flag for managing cgroups with systemd (which pulls the the libdbus dependency because we need dbus to communicate with systemd). We could introduce another feature flag for seccomp. With both feature flags disabled this should solve your problem. What do you think?

krisnova:
Hey @Furisto the `libseccomp` feature flag is introduced in #1484

PR #1483, opened 2023-01-17
Introduce libcgroup examples and README updates

krisnova:
As I have been working on Aurae project I would have found value in some README and examples. I created the things that would have been helpful for me, in the hopes they will make it easy for others to adopt Youki.

Signed-off-by: Kris Nóva <kris@nivenly.com>


utam0k:
@krisnova Thanks for your contribution! I love your idea, but it looks like that CI got failed. May I ask you to check and fix it? If you need help, please ask me. You may leave the bpf example off for the moment. In that case, it would be helpful if you could create an issue.

krisnova:
Yes sounds good I will send more follow up commits. I will also add more README files where I work and offer other small improvements as I work if this is acceptable to Youki.

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -4,14 +4,15 @@ use crate::syscall::Syscall;
...
-    rootless::Rootless, seccomp, tty, utils,
+    rootless::Rootless, tty, utils,
 };
 use anyhow::{bail, Context, Ok, Result};
 use nix::mount::MsFlags;
 use nix::sched::CloneFlags;
 use nix::sys::stat::Mode;
 use nix::unistd::setsid;
 
+use log::warn;
```
YJDoc2, line 15: Hey, it seems this is making clippy fail. Can you remove this and check if it fixes it?

YJDoc2:
Hey @krisnova , The examples and docs are something that we definitely need, but can we not make code changes in this PR then, and keep it just for docs / comment docs?
#1484 seems to extend this in some way (?) or at least a lot of code in that is also present in this. Can we separate the PRs such that one is only for docs and the other for feature intro?

krisnova:
Yes makes sense I will get it cleaned up.

PR #1500, opened 2023-01-22
Relax the version of some crates we dependent

utam0k:
Relaxing the version simplifies the introduction of crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -812,11 +812,32 @@ dependencies = [
 
 [[package]]
 name = "derive_builder"
+version = "0.11.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


YJDoc2 (approved):
lgtm :+1:

@utam0k , can you give me more context on this? For eg, where this will be useful or what is breaking because of this?

utam0k:
@YJDoc2 For example, suppose a user who wants to use libcontainer is already using oci-spec-rs and its version is 0.5.5.
The user cannot use libcontainer because libcontainer requires 0.6.0 of oci-spec-rs, or the version of oci-spec-rs must be increased.
This is prevented.

YJDoc2:
makes sense. I'll go ahead and merge this. Thanks :)

PR #1511, opened 2023-01-23
Add descriptors.json when creating checkpoint

adrianreber:
runc as well as crun create a file called descriptors.json in the checkpoint directory. descriptors.json contains a list of how the FDs 0, 1, 2 are connected before checkpointing:

$ cat descriptors.json # created by runc in this case ["/dev/null","pipe:[230688]","pipe:[230689]"]

With this information the FDs can be reconnected correctly during restore.

With this commit is it possible to do:

$ youki run container
$ youki checkpointt container
$ runc restore container

Now the checkpoint is in a format that can be handled by other container runtimes.


utam0k (approved):
Useful!

PR #1516, opened 2023-01-27
Update to rust 1.67

Furisto:


Furisto:
@utam0k The checks do not pass because it expects that integration tests run with 1.65.0. Looks like this is coming from branch protection rules: https://github.com/orgs/community/discussions/26698. I tried to configure the rules (see [docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule)) but I do not see that option. Do you have permissions to change the rules?

utam0k:
@Furisto Sorry for the late! I updated.

YJDoc2 (approved):
lgtm :+1:

PR #1529, opened 2023-01-31
refactor lifcycle integration test to create, delete and kill

yeongjukang:
This is refactoring lifecycle integration test to createm delete and kill, following #361.
1. I tried to keep old functions in lifecycle module but changed visibility to prepare full split.
2. README and Makefile don't seem to work properly. What I did was like below. Can I update README instruction like this?
```bash
# at project root
youki build
cp tests/rust-integration-tests/integration_test/bundle.tar.gz target/debug
cd target/debug

# at target/debug
sudo RUST_BACKTRACE=1 ./integration_test run --runtime ./youki --runtimetest ./runtimetest
```
3. Integration test results seem inconsistent. I didn't change other conditions but just ran twice.
- first try example
```
# Start group kill
1 / 5 : kill a created container : not ok
        Created container should have been killed.
2 / 5 : kill a deleted container : ok
3 / 5 : kill a started container : not ok
        Started container should have been killed.
4 / 5 : kill attempt without id : ok
5 / 5 : kill non-existing container : ok
```
- second try example
```
# Start group create
1 / 3 : create without id : ok
2 / 3 : create_with_same_id : ok
3 / 3 : normal create : ok
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2700,9 +2700,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.92"
+version = "1.0.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7434af0dc1cbd59268aa98b4c22c131c0584d2232f6fb166efb993e2832e896a"
+checksum = "cad406b69c91885b5107daf2c29572f6c8cdb3c66826821e286c533490c0bc76"
 dependencies = [
  "itoa",
  "ryu",


YJDoc2:
Hey @yeongjukang , as for the issue with running the tests, you can simply run `make rust-oci-test` , after creating the bundle dir, it should not give any issue. Also can you provide more info on flaky tests? I have ran the tests several times, that these tests pass as expected.
Also,
1. Can you rebase the PR to solve conflicts
2. What is the reasoning behind this PR?

yeongjukang:
thanks for reply. @YJDoc2 

for 2, as you and others discussed at #361, refactoring integration test as its own module,
covering create, delete and kill. 
cases in each contains similar content with [runtime-tools](https://github.com/opencontainers/runtime-tools/tree/master/validation).

I will follow your guide and update these soon.

PR #1563, opened 2023-02-12
Make youki a cargo subcommand

### LINKED ISSUE #1562: Enable running youki with cargo
Since youki is now published on crates.io (#586) it is possible to run `cargo install youki` to install youki into cargo's bin dir, but since it is called `youki` not `cargo-youki` one is unable to run it as a cargo sub command `cargo youki ...`

Will-Shanks:
By adding the `[[bin]]` sections when a user runs `cargo install youki` both `youki` and `cargo-youki` will be installed into their cargo bin dir. This means a user can run `cargo youki` to execute youki. The bin section for `youki` was included so that if this dir is in a users path they can also simply run `youki`, and not need to run it via cargo.

This should be all that is necessary to Close #1562.

I believe this is all that is needed to make this work, but I'm not sure how to actually test it.

### DIFF
--- a/crates/youki/Cargo.toml
+++ b/crates/youki/Cargo.toml
@@ -11,6 +11,14 @@ edition = "2021"
 build = "build.rs"
 keywords = ["youki", "container"]
 
+[[bin]]
+name = "youki"
+path = "src/main.rs"
+
+[[bin]]
+name = "cargo-youki"
+path = "src/main.rs"
+
 [features]
 systemd = ["libcgroups/systemd", "libcontainer/systemd", "v2"]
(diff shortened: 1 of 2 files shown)


utam0k:
@Will-Shanks Thanks for your first contribution. I'm not sure what benefit came from this PR. Please tell me a specific case.

Will-Shanks:
One can install and run youki through cargo, making it easier to get.

With this pr getting started with youki is as simple as `cargo install youki && cargo youki -h` no other setup is needed.

YJDoc2:
Hey @Will-Shanks , let us continue the discussion about reasoning behind this in the corresponding issue you have opened, I will comment there :)

PR #1578, opened 2023-02-16
libcontainer: make device creation interfaces public

ipuustin:
I'm investigating how to implement device file support to runwasi in this PR: https://github.com/containerd/runwasi/pull/68. In order to use libcontainer functions for device file creation, they would need to be public for use from other crates. I would be happy to hear feedback regarding whether or not this is the right way to approach this problem.

Another question has to do with the libcontainer test mocks. The libcontainer device creating functions, when compiled in test configuration, know how to mock system calls. Is there a way for exposing that to users outside of the crate too? Maybe having a different libcontainer feature selected in dev-dependencies?

### DIFF
--- a/crates/libcontainer/src/rootfs/mod.rs
+++ b/crates/libcontainer/src/rootfs/mod.rs
@@ -5,7 +5,10 @@
 pub(crate) mod rootfs;
 pub use rootfs::RootFS;
 
-pub(super) mod device;
+pub mod device;
+pub use device::Device;
+
 pub(super) mod mount;
 pub(super) mod symlink;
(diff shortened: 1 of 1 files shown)


Furisto:
Making Device public seems ok. I am hesitant add another feature flag to the crate as we already have quite a few and I do not want complicate it further without having a strong demand for this. How about we add a `new_with_syscall()` function that allows specifying the syscall implementation?

ipuustin:
I think `new_with_syscall()` would be a great solution!

Furisto (approved):

PR #1581, opened 2023-02-17
buffer read and write

wlsnx:
https://github.com/serde-rs/json/blob/0ebeede28a0d5f0f07f18124078f55d62b180a2d/src/ser.rs#L1540
https://github.com/serde-rs/json/blob/0ebeede28a0d5f0f07f18124078f55d62b180a2d/src/read.rs#L194

before:
write(3, "{", 1)                        = 1
write(3, "\n", 1)                       = 1
write(3, "  ", 2)                       = 2
write(3, "\"", 1)                       = 1
write(3, "ociVersion", 10)              = 10
write(3, "\"", 1)                       = 1
write(3, ": ", 2)                       = 2
write(3, "\"", 1)                       = 1
write(3, "1.0.2-dev", 9)                = 9
write(3, "\"", 1)                       = 1
...

read(3, "{", 1)                         = 1
read(3, "\n", 1)                        = 1
read(3, " ", 1)                         = 1
read(3, " ", 1)                         = 1
read(3, "\"", 1)                        = 1
read(3, "o", 1)                         = 1
read(3, "c", 1)                         = 1
read(3, "i", 1)                         = 1
read(3, "V", 1)                         = 1
read(3, "e", 1)                         = 1
read(3, "r", 1)                         = 1
read(3, "s", 1)                         = 1
...

after:
read(3, "{\n  \"ociVersion\": \"1.0.2-dev\",\n "..., 8192) = 3293
write(3, "{\n  \"ociVersion\": \"1.0.2-dev\",\n "..., 3282) = 3282

### DIFF
--- a/crates/youki/src/commands/spec_json.rs
+++ b/crates/youki/src/commands/spec_json.rs
@@ -7,6 +7,7 @@ use oci_spec::runtime::{
 };
 use serde_json::to_writer_pretty;
 use std::fs::File;
+use std::io::{BufWriter, Write};
 use std::path::Path;
 use std::path::PathBuf;
 
@@ -89,7 +90,10 @@ pub fn spec(args: liboci_cli::Spec) -> Result<()> {
     };
 
     // write data to config.json
-    to_writer_pretty(&File::create("config.json")?, &spec)?;
+    let file = File::create("config.json")?;
+    let mut writer = BufWriter::new(file);
+    to_writer_pretty(&mut writer, &spec)?;
+    writer.flush()?;
     Ok(())
 }
 
@@ -106,7 +110,10 @@ mod tests {
         let spec = get_rootless()?;
         let tmpdir = create_temp_dir("test_spec_json").expect("failed to create temp dir");
         let path = tmpdir.path().join("config.json");
-        to_writer_pretty(&File::create(path)?, &spec)?;
+        let file = File::create(path)?;
+        let mut writer = BufWriter::new(file);
+        to_writer_pretty(&mut writer, &spec)?;
+        writer.flush()?;
         Ok(())
     }
 }
--- a/crates/libcontainer/src/config.rs
+++ b/crates/libcontainer/src/config.rs
@@ -1,5 +1,6 @@
 use std::{
     fs,
+    io::{BufReader, BufWriter, Write},
     path::{Path, PathBuf},
 };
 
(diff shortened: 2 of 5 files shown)


YJDoc2:
Hey, if I understand correctly, this changes normal read/write to buffered, so that number of syscalls are reduced?

wlsnx:
> Hey, if I understand correctly, this changes normal read/write to buffered, so that number of syscalls are reduced?

Yes,serde_json read content from reader byte by byte,if without buffered,there are huge number of read syscalls.Similarly,serde_json will call write for every element.

crates/youki/src/commands/spec_json.rs
```diff
@@ -106,7 +110,9 @@ mod tests {
         let spec = get_rootless()?;
         let tmpdir = create_temp_dir("test_spec_json").expect("failed to create temp dir");
         let path = tmpdir.path().join("config.json");
-        to_writer_pretty(&File::create(path)?, &spec)?;
+        let file = File::create(path)?;
+        let writer = BufWriter::new(file);
+        to_writer_pretty(writer, &spec)?;
```
YJDoc2, line 115: Hey, I know this is a test, but should we not also add a `writer.flush()` at the end, like we have done at other places?

wlsnx: https://doc.rust-lang.org/nightly/std/io/struct.BufWriter.html
It is critical to call [flush](https://doc.rust-lang.org/nightly/std/io/struct.BufWriter.html#method.flush) before BufWriter<W> is dropped. Though dropping will attempt to [flush](https://doc.rust-lang.org/nightly/std/io/struct.BufWriter.html#method.flush) the contents of the buffer, any errors that happen in the process of dropping will be ignored. Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations.

YJDoc2: Yes, so we should add a flush call after `to_writer_pretty` like other places right?

wlsnx: Yes.

YJDoc2: So can you update and push it, so we can go ahead and merge this?

wlsnx: Updated.

YJDoc2 (approved):
lgtm :+1:

YJDoc2:
Thanks a lot for making this improvement !

PR #1602, opened 2023-02-28
feat Add container id validate

lengrongfu:
issue: https://github.com/containers/youki/issues/1580

### DIFF
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -1,5 +1,5 @@
 use crate::{syscall::Syscall, utils::PathBufExt};
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use std::path::PathBuf;
 
 use super::{init_builder::InitContainerBuilder, tenant_builder::TenantContainerBuilder};
@@ -50,7 +50,6 @@ impl<'a> ContainerBuilder<'a> {
     /// ```
     pub fn new(container_id: String, syscall: &'a dyn Syscall) -> Self {
         let root_path = PathBuf::from("/run/youki");
-
         Self {
             container_id,
             root_path,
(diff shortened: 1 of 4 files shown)


crates/libcontainer/src/container/builder.rs
```diff
@@ -37,6 +37,9 @@ pub struct ContainerBuilder<'a> {
 /// .build();
 /// ```
 impl<'a> ContainerBuilder<'a> {
+    /// ERR_INVALID_ID is container id validate error
+    const ERR_INVALID_ID: &str = "invalid container ID format";
```
yihuaf, line 41: Can you please use the `anyhow` error and result, following how the rest of the code is doing? This is actually a good example where we should use `thiserror`, but we have not done this yet.

crates/libcontainer/src/container/builder.rs
```diff
@@ -50,7 +53,7 @@ impl<'a> ContainerBuilder<'a> {
     /// ```
     pub fn new(container_id: String, syscall: &'a dyn Syscall) -> Self {
         let root_path = PathBuf::from("/run/youki");
-
+        Self::validate_id(container_id.clone()).unwrap();
```
yihuaf, line 56: I like the idea of validating the ID, but perhaps this check should happen before getting to the builder.

crates/libcontainer/src/container/builder.rs
```diff
@@ -225,6 +261,27 @@ mod tests {
             .context("build container")?;
         assert_eq!(path_builder.pid_file, Some(cwd.join("not/existing/path")));
 
+        // validate container_id
+        let result =
```
yihuaf, line 265: These tests should be in their own testcases.

crates/libcontainer/src/container/builder.rs
```diff
@@ -61,6 +64,39 @@ impl<'a> ContainerBuilder<'a> {
...
+    /// - uppercase (A-Z) and lowercase (a-z) Latin letters;
+    /// - digits (0-9);
+    /// - underscore (_);
+    /// - plus sign (+);
+    /// - minus sign (-);
+    /// - period (.).
+    ///
+    /// In addition, IDs that can't be used to represent a file name
+    /// (such as . or ..) are rejected.
+    fn validate_id(container_id: String) -> Result<(), &'static str> {
```
yihuaf, line 84: This is good.

crates/libcontainer/src/container/builder.rs
lengrongfu: Already using `anyhow` error and Result, I can submit pr later to use `thiserror`.

crates/libcontainer/src/container/builder.rs
lengrongfu: Please take a look, it has been modified to perform verification before creating the builder.

crates/libcontainer/src/container/builder.rs
lengrongfu: done

yihuaf (approved):
Much better :)

Issue #1601, opened 2023-02-27
Use `clone(2)` instead of `fork(2)`


utam0k:
# Why?

We have to need to use `CLONE_PARENT` but unfortunately `fork(2)` doesn't support it.
Without CLONE_PARENT, the init process becomes a zombie process and no one can `wait(2)`.
For example, it is a problem when trying to implement the exec and wait commands.

utam0k:
ref: https://github.com/containers/youki/pull/217

utam0k:
ref: https://github.com/opencontainers/runc/blob/5a0642d6fd0a7c217c1cd6fd472b72b36820caff/libcontainer/nsenter/nsexec.c#L389-L398

utam0k:
ref: https://stackoverflow.com/questions/33404939/how-to-call-fork-with-flags-or-clone-without-the-first-argument-fn

yihuaf:
Wanna me to take a look at this?

yihuaf:
@utam0k I thought a bit more on this subject and here is my understanding. Currently, your observation is correct that the container init process is orphaned to pid 1 when the youki intermediate process exits. This is normal behavior, but this means no one will wait on the init process.

Here is what I think how it should work. Consider we have the `caller of youki`, `youki main process`, `intermediate` and `container init` process. The end result we want to see is that the container init process is re-parent to the `caller of youki` instead of pid 1 in the detached mode. In the foreground mode, we want `youki main process` to wait on the `container init` process.

If we agree that the the above is the result we want, there are a few ways to achieve this. The first one, without using CLONE_PARENT, is for the youki main process to set itself as the `PR_SET_CHILD_SUBREAPER`. In this way, when the intermediate process terminates, the container init process is re-parented to the youki main process.

Another way is to clone the intermediate process as the child of the youki main process. The intermediate process can then clone the container init process as a sibling process with CLONE_PARENT. Then, youki main process becomes the parent of the container init process. There is no need to use `PR_SET_CHILD_SUBREAPER`.

Here are some other edge cases. I don't think we should clone both the intermediate process and the container init using CLONE_PARENT. The end result will be both the intermediate process and container init process becomes children of the `caller of youki`. Based on what I read, the `caller of youki` (containerd-shim or conmon) is expected to to use `PR_SET_CHILD_SUBREAPER`. Or same as `runc`, if caller doesn't set the sub-reaper flag, the container init process will reparent to pid 1 after youki main process exits.

[...]

PR #1610, opened 2023-03-01
Implement the container_clone using CLONE_PARENT

yihuaf:
This is the first step into fixing #1601 which implements a `container_clone` call that uses clone and `CLONE_PARENT`. Since this function uses `CLONE_PARENT` which creates sibling process instead of child process, we need to re-examine the overall logic to make sure all the `waitpid` calls are still correct, since a process can only wait on its immediate child process.

### DIFF
--- a/crates/libcontainer/src/process/fork.rs
+++ b/crates/libcontainer/src/process/fork.rs
@@ -1,6 +1,52 @@
-use anyhow::Result;
-use nix::unistd;
+use anyhow::{Context, Result};
+use libc::SIGCHLD;
 use nix::unistd::Pid;
+
+// Fork/Clone a sibling process that shares the same parent as the calling
+// process. This is used to launch the container init process so the parent
+// process of the calling process can receive ownership of the process. If we
+// clone a child process as the init process, the calling process (likely the
+// youki main process) will exit and the init process will be re-parented to the
+// process 1 (system init process), which is not the right behavior of what we
+// look for.
+pub fn container_clone_sibling<F: FnOnce() -> Result<i32>>(cb: F) -> Result<Pid> {
+    let mut clone = clone3::Clone3::default();
+    // Note: normally, an exit signal is required, but when using
+    // `CLONE_PARENT`, the `clone3` will return EINVAL if an exit signal is set.
+    // The older `clone` will not return EINVAL in this case. Instead it ignores
+    // the exit signal bits in the glibc wrapper.
+    clone.flag_parent();
+
+    container_clone(cb, clone).with_context(|| "failed to clone sibling process")
+}
+
+// A simple clone wrapper to clone3 so we can share this logic in different
+// fork/clone situations. We decided to minimally support kernel version >= 5.4,
+// and `clone3` requires only kernel version >= 5.3. Therefore, we don't need to
+// fall back to `clone` or `fork`.
+fn container_clone<F: FnOnce() -> Result<i32>>(
+    cb: F,
+    mut clone_cmd: clone3::Clone3,
+) -> Result<Pid> {
+    // Return the child's pid in case of parent/calling process, and for the
+    // cloned process, run the callback function, and exit with the same exit
+    // code returned by the callback. If there was any error when trying to run
+    // callback, exit with -1
+    match unsafe { clone_cmd.call().with_context(|| "failed to run clone3")? } {
+        0 => {
+            // Inside the cloned process
+            let ret = match cb() {
+                Err(error) => {
+                    log::debug!("failed to run child process in clone: {:?}", error);
+                    -1
+                }
+                Ok(exit_code) => exit_code,
+            };
+            std::process::exit(ret);
+        }
+        pid => Ok(Pid::from_raw(pid)),
+    }
+}
 
 // Execute the cb in another process. Make the fork works more like thread_spawn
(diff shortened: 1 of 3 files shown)


utam0k:
@yihuaf How about using `clone3(2)`?

yihuaf:
> @yihuaf How about using `clone3(2)`? https://github.com/utam0k/youki/blob/d1def1119b159c6916ed5ffb0da31da73c918c80/crates/libcontainer/src/process/fork.rs#L34-L63

Good point. Let me take a look tomorrow.

yihuaf:
@utam0k I looked into `clone3` and I am a little hesitant to use it for this specific usecase. First, `clone3` doesn't add anything extra compared to `clone` syscall for this usecase. It is a nicer interface, but using both `clone` or `clone3` can get the job done. Second, the crate used for the `clone3` is not merged into the `libc` crate, which raises some concern. If we decide to use `clone3`, I would recommend vendoring the crate into our code base to be maintained. Lastly, should we fall back to `clone` in the case where `clone3` fails or not supported? If so, then I would argue that it will be easier to just use clone for all cases. `clone` being older, is definitely more mature and more well documented.

I don't have a strong preference. If you don't have a strong preference, I would recommend us to stay with `clone`. Given the abstraction we have built around `fork.rs`, we can easily make the switch at a later point. However, if you feel strongly about `clone3`, then I am happy to update this PR. Your call :)

yihuaf:
Plus, I am hitting this with specifying both CLONE_PARENT and exit signal (SIGCHLD) with `clone3`. https://www.spinics.net/lists/linux-man/msg23564.html

yihuaf:
Aha, following this thread: https://github.com/checkpoint-restore/criu/issues/926

It looks like we should not pass in exit signal when CLONE_PARENT is used. There is more context in the referenced thread.

yihuaf:
Updated [68511cf](https://github.com/containers/youki/pull/1610/commits/68511cf3316f09c298d745f5a683ff19bc091b0a) with `clone3`. Let me know what you think.

Furisto:
I am in favor of using clone3. We have decided that we will support kernels >= 5.4 only and clone3 has been available since kernel 5.3 so we should not need a fallback.

yihuaf:
> I am in favor of using clone3. We have decided that we will support kernels >= 5.4 only and clone3 has been available since kernel 5.3 so we should not need a fallback.

Sounds good. I updated the comments with what has been discussed above. PTAL. Thanks.

utam0k:
> I am in favor of using clone3. We have decided that we will support kernels >= 5.4 only and clone3 has been available since kernel 5.3 so we should not need a fallback.

:100: Since they are too new, there are some options we have to be careful to use.

crates/libcontainer/src/process/fork.rs
```diff
@@ -1,37 +1,77 @@
...
+// or clone, so it is easier to reason. Compared to clone call, fork is easier
+// to use since fork will magically take care of all the variable copying. If
+// using clone, we would have to manually make sure all the variables are
+// correctly send to the new process, especially Rust borrow checker will be a
+// lot of hassel to deal with every details.
+pub fn container_fork<F: FnOnce() -> Result<i32>>(cb: F) -> Result<Pid> {
+    // Using `clone3` to mimic the effect of `fork`.
+    let mut clone = clone3::Clone3::default();
+    let mut pidfd = -1;
+    clone.flag_pidfd(&mut pidfd).exit_signal(SIGCHLD as u64);
```
utam0k, line 62: Do you have the plan to use pidfd?

Furisto:
Can you test what the performance impact of this change is?
`hyperfine --prepare 'sudo sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' --warmup 10 --min-runs 100 'sudo ./youki create -b tutorial a && sudo ./youki start a && sudo ./youki delete -f a'`

crates/libcontainer/src/process/fork.rs
yihuaf: I don't see a need for this, so I can remove them for now, until we need it.

yihuaf:
> hyperfine --prepare 'sudo sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' --warmup 10 --min-runs 100 'sudo ./youki create -b tutorial a && sudo ./youki start a && sudo ./youki delete -f a'

With this PR.
```
Benchmark 1: sudo ./youki create -b tutorial a && sudo ./youki start a && sudo ./youki delete -f a
  Time (mean ± σ):     233.5 ms ±  13.9 ms    [User: 7.5 ms, System: 8.0 ms]
  Range (min … max):   197.7 ms … 265.5 ms    100 runs
```

Without the PR:
```
Benchmark 1: sudo ./youki create -b tutorial a && sudo ./youki start a && sudo ./youki delete -f a
  Time (mean ± σ):     231.5 ms ±  14.5 ms    [User: 7.4 ms, System: 7.9 ms]
  Range (min … max):   198.0 ms … 261.5 ms    100 runs
```

I believe the result matches the theoretical assumption that the performance different should be minimal since the underlying syscall is the same. The average with clone3 seems to be a little slower (about 1 ms) difference, but this should be within the standard deviation.

Furisto (approved):

utam0k (approved):

utam0k:
Thanks, @yihuaf

PR #1637, opened 2023-03-08
Fixed container init process not re-parent to youki main process

### LINKED ISSUE #1601: Use `clone(2)` instead of `fork(2)`

yihuaf:
Fix #1601 

In this PR, I decided to clone `intermediate` process as child and then clone the container init process as sibling of the `intermediate` process and child of `youki main` process. In this way, when intermediate process exits, the container init process will not re-parent to pid 1 and become zombie. The caller of youki can decide to use subreaper flag to adopt the container init process when youki main process exits.

This method cleaner because youki main process is guaranteed to be the parent of both intermediate process and container init process at all time. There is no extra re-parenting.

This PR is only part of the fix. We do need to revisit the detached and foreground mode. Not everything is fixed, but this PR should maintain the old behavior on anything that it does not fix.

### DIFF
--- a/crates/libcontainer/src/container/init_builder.rs
+++ b/crates/libcontainer/src/container/init_builder.rs
@@ -90,10 +90,15 @@ impl<'a> InitContainerBuilder<'a> {
             notify_path,
             container: Some(container.clone()),
             preserve_fds: self.base.preserve_fds,
-            detached: false, // TODO this should be set properly based on how the command is given
+            // TODO: This should be set properly based on how the command is
+            // given. For now, set the detached to true because this is what
+            // `youki create` defaults to.
+            detached: true,
         };
 
-        builder_impl.create()?;
+        // TODO: Fix waiting on this pid (init process) when detached = false.
+        let _pid = builder_impl.create()?;
+
         container.refresh_state()?;
 
         Ok(container)
--- a/crates/libcontainer/src/process/container_main_process.rs
+++ b/crates/libcontainer/src/process/container_main_process.rs
@@ -1,9 +1,6 @@
 use crate::{
     container::ContainerProcessState,
-    process::{
-        args::{ContainerArgs, ContainerType},
-        channel, container_intermediate_process, fork,
-    },
+    process::{args::ContainerArgs, channel, container_intermediate_process, fork},
     rootless::Rootless,
     seccomp, utils,
 };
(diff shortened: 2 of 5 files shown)


yihuaf:
See this comment for more discussion: https://github.com/containers/youki/issues/1601#issuecomment-1459499447

crates/libcontainer/src/container/init_builder.rs
```diff
@@ -90,10 +90,15 @@ impl<'a> InitContainerBuilder<'a> {
             notify_path,
             container: Some(container.clone()),
             preserve_fds: self.base.preserve_fds,
-            detached: false, // TODO this should be set properly based on how the command is given
+            // TODO: This should be set properly based on how the command is
+            // given. For now, set the detached to true because this is what
+            // `youki create` defaults to.
+            detached: true,
```
utam0k, line 96: Please create the issue about this comment.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -110,13 +96,22 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, Pi
...
-    // intermediate pid is required in the case when we call exec, as we need to wait for the
-    // intermediate process to exit, which itself waits for child process (the exec process) to exit
-    // in order to get the proper exit code. We cannot simply wait for the init_pid , that is the actual container
-    // process, as it is not (direct) child of our process
-    Ok((intermediate_pid, init_pid))
+    // Before the main process returns, we want to make sure the intermediate
+    // process is exit and reaped. By this point, the intermediate process
+    // should already exited successfully. If intermediate process errors out,
+    // the `init_ready` will not be sent.
+    match waitpid(intermediate_pid, None)? {
```
utam0k, line 103: Does this affect the exec command?

crates/libcontainer/src/container/init_builder.rs
yihuaf: I will do this. I was planning to fix this right after since this will be required to enable foreground mode. I want this PR to make all the detached mode situation work and `youki create` defaults to detached but `youki run` has a flag to switch between foreground and background mode.

yihuaf: Actually https://github.com/containers/youki/issues/317 and https://github.com/containers/youki/issues/628 covers the situation fairly well.

crates/libcontainer/src/process/container_main_process.rs
yihuaf: No. This is reaping the intermediate process, so `create`, `run`, `exec` all have to reap this process correctly, which is expected to always exit and the `youki main` process should wait for the intermediate process to exit and reap it.

utam0k: :ok:

utam0k (approved):
:100:

PR #1649, opened 2023-03-09
fix container delete error

lengrongfu:
fixes https://github.com/containers/youki/issues/1640

### DIFF
--- a/crates/libcontainer/src/container/container_delete.rs
+++ b/crates/libcontainer/src/container/container_delete.rs
@@ -34,37 +34,36 @@ impl Container {
         log::debug!("container status: {:?}", self.status());
         if self.can_delete() {
             if self.root.exists() {
-                let config = YoukiConfig::load(&self.root).with_context(|| {
-                    format!("failed to load runtime spec for container {}", self.id())
-                })?;
-                log::debug!("config: {:?}", config);
+                if let Ok(config) = YoukiConfig::load(&self.root) {
+                    log::debug!("config: {:?}", config);
+
+                    // remove the cgroup created for the container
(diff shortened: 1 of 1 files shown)


crates/libcontainer/src/config.rs
```diff
@@ -47,7 +47,11 @@ impl<'a> YoukiConfig {
 
     pub fn load<P: AsRef<Path>>(path: P) -> Result<Self> {
         let path = path.as_ref();
-        let file = fs::File::open(path.join(YOUKI_CONFIG_NAME))?;
+        let config_path = path.join(YOUKI_CONFIG_NAME);
+        if let Err(e) = config_path.try_exists() {
+            return Err(anyhow!("file not exists {}", e));
```
yihuaf, line 52: Is the only reason to have the extra step to return a better error message? `File::open` will fail when file doesn't exist, so I am not sure the purpose of doing the `try_exist` here.

crates/libcontainer/src/container/container_delete.rs
```diff
@@ -34,37 +34,37 @@ impl Container {
         log::debug!("container status: {:?}", self.status());
         if self.can_delete() {
             if self.root.exists() {
-                let config = YoukiConfig::load(&self.root).with_context(|| {
-                    format!("failed to load runtime spec for container {}", self.id())
-                })?;
-                log::debug!("config: {:?}", config);
+                let result_config = YoukiConfig::load(&self.root);
+                if let Ok(config) = result_config {
```
yihuaf, line 38:
```suggestion
                if let Ok(config) = YoukiConfig::load(&self.root) {
```

yihuaf (changes requested):

yihuaf (approved):

Issue #1640, opened 2023-03-08
youki delete bug


lengrongfu:
I executed `youki run test` using a user with non-root privileges, and the execution failed, then executed `youki list` can find `test` this container, but executed `youki delete test` fail.

- youki list
```shell
ubuntu@myubuntu:~$ youki list
ID    PID  STATUS   BUNDLE                  CREATED  CREATOR
test       Stopped  /home/ubuntu/container
```

- youki delete
```shell
ubuntu@myubuntu:~$ youki delete test
[ERROR crates/youki/src/main.rs:138] 2023-03-08T19:11:04.971051412+08:00 error in executing command: failed to delete container test

Caused by:
    0: failed to load runtime spec for container test
    1: No such file or directory (os error 2)
Error: failed to delete container test

Caused by:
    0: failed to load runtime spec for container test
    1: No such file or directory (os error 2)
```

I found that there is only state.json file in the container data directory, not youki_ Config.json file, which is the root cause of deletion failure.

```shell
ubuntu@myubuntu:~$ tree /run/user/1000/youki
/run/user/1000/youki
└── test
    └── state.json

1 directory, 1 file
```

Currently I think there are two solutions:
1. After `youki run` fails, you should clear the operation.
2. In `youki delete`, it is compatible with the situation where `youki_config.json` does not exist.

need to discuss.

yihuaf:
I suspect that this is a duplicate of this fix: https://github.com/containers/youki/pull/1556

PR #1654, opened 2023-03-10
Refactor youki delete to match runc/crun.

yihuaf:
The discrepancy in deleting a `created` status container without `force` flag is first brought up in #1556. See the PR for detailed discussion. In this PR, we fix the discrepancy and refactored the delete code to be more readable. In this way, it is clear that we first check the condition if container can be deleted. Then perform the delete. We also add a debug assert to make the the invariant in not broken when deleting.


YJDoc2:
Hey, I'll take a loot at this later today :+1:

yihuaf:
> Hey, I'll take a loot at this later today 👍

Thank you.

crates/libcontainer/src/container/container_delete.rs
```diff
@@ -27,14 +27,48 @@ impl Container {
...
+                // `runc` and `crun` allows deleting `created`. Therefore we
+                // decided to follow `runc` and `crun`. In the case where
+                // container is in `created` status, we first kill the container
+                // init process which is waiting on start signal. Since only a
+                // single container process (the init process) is created, we do
+                // not need to send signals to all in this case.
+                self.do_kill(signal::Signal::SIGKILL, false)?;
+                self.set_status(ContainerStatus::Stopped).save()?;
+            }
+            ContainerStatus::Creating | ContainerStatus::Running | ContainerStatus::Paused => {
```
YJDoc2, line 49: Should the `Creating` state also handled along with `Created` ? If I'm correct, then even at the time of `Creating` state, there should be only one process in the container's cgroup. If that is correct, then we don't get any advantage / have any reason to use kill all with force. I checked the spec, and it specifies that states other than `stopped` should generate error, but we are going according to runc anyways.

Please correct me If my understanding was wrong.

crates/libcontainer/src/container/container_delete.rs
```diff
@@ -58,20 +92,22 @@ impl Container {
...
-                fs::remove_dir_all(&self.root).with_context(|| {
-                    format!("failed to remove container dir {}", self.root.display())
-                })?;
+                Err(err) => {
+                    // There is a brief window where the container state is
+                    // created, but the container config is not yet generated
+                    // from the OCI spec. In this case, we assume as if we
+                    // successfully deleted the config and moving on.
+                    log::warn!("skipping loading youki config due to: {err:?}, continue to delete");
+                }
```
YJDoc2, line 101: Do we need to worry about race condition? For eg, if we are deleting a container that is being created, then there might be a window where the config is not generated, and we are treating it as deleted. If that does get created after this moment, then should we be concerned about it or not?

If we are treating the delete root after this as catch all, then I guess this is fine

YJDoc2 (commented):
Hey, I have left a couple of comments, but overall, looks great! Really nice commenting :+1:

crates/libcontainer/src/container/container_delete.rs
yihuaf: You asked a very good question and here is what I think. First of all, the actual bug is addressed in (https://github.com/containers/youki/pull/1649) and this PR only adds a warning and an explanation, so the context of the bug is in #1649 and #1640. In short, if rootless::New errors for whatever reason, we are stuck in a state where container state is created but not the youki config. This will cause the delete to not able to clean up because it is missing the youki config.

Here is the complex part about the race condition and you are right that there is a potential race here. If process 1 is calling `youki create` and process 2 is calling `youki delete --force`, we have a window where process 1 has not launched the process yet, but process 2 is going ahead to delete. There is no way for process 2 to signal process 1 to stop creating the container processes with force flag. What will happen is that process 2 will delete the container root and process 1 will continue to start the process wand wait on the exec.fifo, which is a file (named pipe). So when container root is deleted, the exec.fifo will be gone and the process waiting on it will error out. Therefore, both should be clean up correctly. Or process 1 at any point trying to save the state and fails. So in theory, the race condition will not cause leakage in terms of process or files on the system. I believe this the best we can guarantee. I think this covers all the situations, but please double check my work here.

crates/libcontainer/src/container/container_delete.rs
yihuaf: Technically 2 process, container init and intermediate process before it exits. With that being said, it is possible I am over optimizing in this case. The goal here is to kill all the process in the container cgroup, so I should possibly just call kill all here for either cases.

crates/libcontainer/src/container/container_delete.rs
YJDoc2: Hey, I think this makes sense :+1:
I think this should solve the issue and not cause any leakage of resources.
If possible, give me a day's time, I'll try to do a manual verification check by launching parallel processes and checking that resources are cleaned up.

In either case, I think this implementation is good enough :)

crates/libcontainer/src/container/container_delete.rs
YJDoc2: Hey, If there are two processes, then this logic makes perfect sense.
Also, I don't think we should merge `Created` with rest 3 (if that was what you meant) , because as per runc and crun, we need to allow deleting `Created` without force flag, however need force flag for others.

If there are going to be (even technically) two processes, in `Creating` state, it is better to handle it by killing all in cgroup, just to be safe.

crates/libcontainer/src/container/container_delete.rs
yihuaf: > Hey, I think this makes sense 👍 I think this should solve the issue and not cause any leakage of resources. If possible, give me a day's time, I'll try to do a manual verification check by launching parallel processes and checking that resources are cleaned up.
> 
> In either case, I think this implementation is good enough :)

Take you time please :) I am curious as well.

YJDoc2: Hey @yihuaf Unfortunately, I didn't get time in last week to do this :sweat:  I'll try to get this done soon :pray:

yihuaf: Another option is to merge this PR as is, since the PR is fixing a specific issue. We can continue to test the edge case on main and file issue as needed.

Issue #1650, opened 2023-03-09
youki does not pass cargo test under certain circumstances


unknowndevQwQ:
At the time I ran into this problem, the git commit was fe58002c3.
Run cargo test --frozen --release and after waiting a while you will see the following:
```
test syscall::linux::tests::test_close_range_userspace ... ok
test seccomp::tests::test_basic ... ok
test syscall::linux::tests::test_close_range_native ... ok
test seccomp::tests::test_seccomp_notify ... ok
test syscall::linux::tests::test_get_open_fds ... ok
error: test failed, to rerun pass `-p libcontainer --lib`
```

YJDoc2:
I have seen this once or twice before, but the issue with this is  when retried, even immediately, the tests pass, so it is either some race condition that occurs within tests or some cargo test assumptions which our tests somehow break.

One way I can think to proceed here is to make a script looping ~1000 times, running tests, seeing how many times they fail. Then annotating each test  to run serially  (or cargo might have a flag to do this) and see how many times we get failure there.

That way we can potentially rule out if it is some race condition in our tests or what.

A note is that running tests ~1000 times will take considerable time + resources , so whoever is doing , be careful!

unknowndevQwQ:
> 

Wait, how should I test only one item and not all of them?
Before trying 1000 tests that will fail, I need to avoid testing tests that will definitely succeed at the same time, because that is too time consuming.

Overflow0xFFFF:
I've been seeing a similar issue for a little bit now. With instructions from [this article](https://fdeantoni.medium.com/running-tests-sequentially-in-rust-eed7566f63f0), I was able to run tests sequentially with `cargo test -- --test-threads 1`:

```
test seccomp::tests::test_basic ... ok
test seccomp::tests::test_moby ... ok
test seccomp::tests::test_seccomp_notify ... ok
test signal::tests::test_conversion_from_string ... ok
test signal::tests::test_conversion_from_string_should_be_failed ... ok
test syscall::linux::tests::test_close_range_native ... ok
test syscall::linux::tests::test_close_range_userspace ... ok
test syscall::linux::tests::test_get_open_fds ... ok
test tty::tests::test_setup_console ... error: test failed, to rerun pass `-p libcontainer --lib`
```

I'm running rustc 1.67.0 (fc594f156 2023-01-24) on Fedora 36.

Overflow0xFFFF:
[...]

PR #1676, opened 2023-03-16
fix(libcontainer): Run test_is_executable with a more common file

Overflow0xFFFF:
This test was failing on some popular Linux distros, like Fedora, because `/boot/initrd.img` doesn't exist. This change alters the test so that a more common file, `/etc/hosts`, is used.

After searching on the internet, there doesn't seem to be any one file that is always guaranteed to exist, other than the root filesystem path `/`. But `/etc/hosts` is very, very common, so I'm hopeful that this change means that the Youki build environment is more likely to build and pass its tests on more systems.

Relates to #1650.

### DIFF
--- a/crates/libcontainer/src/utils.rs
+++ b/crates/libcontainer/src/utils.rs
@@ -556,10 +556,12 @@ mod tests {
     #[test]
     fn test_is_executable() {
         let executable_path = PathBuf::from("/bin/sh");
-        let directory_path = PathBuf::from("/tmp");
-        // a file guaranteed to be on linux and not executable
-        let non_executable_path = PathBuf::from("/boot/initrd.img");
+        let directory_path =
+            create_temp_dir("test_is_executable").expect("create temp directory for test");
+        let non_executable_path = directory_path.join("non_executable_file");
         let non_existent_path = PathBuf::from("/some/non/existent/path");
+
+        File::create(non_executable_path.as_path()).unwrap();
 
         assert!(is_executable(&non_existent_path).is_err());
         assert!(is_executable(&executable_path).unwrap());


yihuaf:
Is it possible to detect the distro (Ubuntu vs. Fedora) in the test and choose the path accordingly? In this way, we can provide a warning to distro we did not explicitly test.

I will let @YJDoc2 review the specifics in this PR :)

YJDoc2:
Hey @Overflow0xFFFF Thanks for the  PR, I'll take a look at this by today :+1:

Overflow0xFFFF:
> Drive by comment: Is it possible to detect the distro (Ubuntu vs. Fedora) in the test and choose the path accordingly? In this way, we can provide a warning to distro we did not explicitly test.
> 
> I will let @YJDoc2 review the specifics in this PR :)

I think so. Most distros ship some kind of variant on the LSB Release file and have a utility / script that can parse that information. `cargo test` doesn't allow printing any messages without the `--nocapture` flag passed from the CLI, but that's not hard to fix. However, running the tests in parallel, the user might miss the message output in a sea of other messages.

If our goal is to notify the user, we could probably just add that logic to the Makefile, though. This would be easy to maintain in shell script format.

A third alternative would be to create a non-executable file ourselves and avoid parsing any distro-specific info anyway! I need to brush up some more on the context for the test, though. I was merely fixing a test that was broken in my workspace, so creating a file as part of the test suite could either be normal or overkill. Some software developers also don't like touching the filesystem as part of unit tests. If that's a philosophy that Youki's developers share, I don't want to intrude. 😄

Separately, would there be any interest in adding Rust builds on other OSes to the GitHub Actions workflow? And does Youki have certain supported distros in mind, or is it too early to ask that question?

YJDoc2:
>I think so. Most distros ship some kind of variant on the LSB Release file and have a utility / script that can parse that information. cargo test doesn't allow printing any messages without the --nocapture flag passed from the CLI, but that's not hard to fix. However, running the tests in parallel, the user might miss the message output in a sea of other messages.

I don't think warning users in test is particularly effective, I might suggest simply marking the test as passed + making sure that it always runs in the CI (i.e.  whatever we use is SURE to be present in CI) so that we don't miss it in PRs.
I think we already check for OS info in the `youki info` , we might be able to get the distro from there.


>A third alternative would be to create a non-executable file ourselves and avoid parsing any distro-specific info anyway!

This would be much preferable than parsing distro info, as long as we can guarantee that the file generated is cleaned up afterwards, irrespective of pass/fail. Youki's test already need mutable access to system in order to test certain cgroup stuff, so creating a file (as long as it is cleaned up) is not that big of an issue. I believe we have this kind of always-clean up logic for cgroups tests, we can work with that.

>Separately, would there be any interest in adding Rust builds on other OSes to the GitHub Actions workflow?

I think we would like to validate support to other distros in CI, but we will need more discussion on this, specifically with regard to how to do it, is it worth to do it (as we mostly deal with linux , not specifically any distro related stuff), and if so to what extent we want to validate (just build, build+unit tests , build+integration tests)

YJDoc2:
@Overflow0xFFFF , Overall this PR is good, I have added a comment please take a look.

If you think creating a file and testing it is better, and willing to take a try at it in this PR, we can go that way ; or else you'd rather have this merged, and not work  on that / work on it later , I can approve and merge this.

Let me know, and thanks a lot for the contribution!

YJDoc2:
Hey @Overflow0xFFFF , thanks for the update. I think a bit better approach for temp file wold be to use `TempDir` structure, and create a file in that. That dir deletes itself in `Drop` , so we can be assured it gets cleaned up.

It is defined in https://github.com/containers/youki/blob/main/crates/libcontainer/src/utils.rs#L279 and you can see the example usage in https://github.com/containers/youki/blob/main/crates/libcontainer/src/rootless.rs#L450

Can you take a look and use it instead? Also can I ask you to edit the last commit and force push so we have a nice commit history?
Thanks a lot!

Overflow0xFFFF:
@YJDoc2, thank you for the links and advice! I have put TempDir to good use, and I've updated the test to rely on that instead. If there's anything else that's worth updating, changing, or tweaking, let me know!

Overflow0xFFFF:
No worries about the nitpicks! I completely understand, and I thank you for your patience and all the extra review time you've put in!

I grepped the codebase for other examples of `create_temp_dir()` so that this test creates a temporary directory in the same way. This hopefully looks like any other test in the code base!

YJDoc2 (approved):
lgtm :+1:
Thanks for your contributions!

PR #1743, opened 2023-03-29
[Trivial] Fix makefile targets to use PHONY

yihuaf:
Not all makefile target in Youki is using `.PHONY` target. We don't use any file targets, so fixing it properly for all makefile.

On a different note, there has been a lot of popularity in using `justfile` as command runner to invoke different command, especially for Rust projects. Since we use Makefile only as a command runner,  I want to bring `justfile` up to see if there is any interest. Note, Makefile is definitely good enough for us. The upside is that `justfile` is a little more ergonomic being used as a command runner. The downside is that we have to install `justfile` as part of the dependency, whereas Makefile is more readily available. Food for thought :)

Reference: https://github.com/casey/just

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,53 +2,71 @@ ROOT = $(shell git rev-parse --show-toplevel)
 
 # builds
 
+.PHONY:build
 build: youki-release
 
+.PHONY: youki
 youki: youki-dev # helper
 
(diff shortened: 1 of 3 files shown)


utam0k (approved):
I didn't know case/just. I'll check it. Thanks

PR #1744, opened 2023-03-29
[Trivial] Remove the metadata semvar causing a warning.

yihuaf:


YJDoc2:
Hey, I have seen the warning from this,  but the crate itself adds this metadata, as we can see in install section https://crates.io/crates/libbpf-sys . I think the second part maps to the libbpf against which this is built.
Would it be a good idea to remove (not that this should change anything compilation-wise, as cargo ignores it anyways)

yihuaf:
> Hey, I have seen the warning from this, but the crate itself adds this metadata, as we can see in install section https://crates.io/crates/libbpf-sys . I think the second part maps to the libbpf against which this is built. Would it be a good idea to remove (not that this should change anything compilation-wise, as cargo ignores it anyways)

I filed https://github.com/libbpf/libbpf-sys/issues/58 to clarify with the lib author. If I am understanding correctly, the metadata version is used as a human readable indicator for the version of upstream of libbpf used. So it should be safe to remove from the cargo.toml since as you mentioned cargo ignores it.

yihuaf:
@YJDoc2 Confirmed with the `libbpf` author that the metadata can be ignored safely in the cargo.toml. Reference: https://github.com/libbpf/libbpf-sys/issues/58#issuecomment-1490917692

YJDoc2 (approved):
lgtm :+1:

PR #1777, opened 2023-04-05
Address ECHILD

utam0k:
Assume the case where intermediate_process is waitpid by someone and ignore ECHILD in that case.
This is safe. This is safe because intermediate_process and main_process check if the process is finished by piping instead of exit code.

### DIFF
--- a/crates/libcontainer/src/process/container_main_process.rs
+++ b/crates/libcontainer/src/process/container_main_process.rs
@@ -4,7 +4,7 @@ use crate::{
     rootless::Rootless,
     utils,
 };
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use nix::sys::wait::{waitpid, WaitStatus};
 use nix::unistd::Pid;
 
@@ -12,8 +12,9 @@ use nix::unistd::Pid;
 use crate::seccomp;
 #[cfg(feature = "libseccomp")]
 use nix::{
+    errno::Errno,
     sys::socket::{self, UnixAddr},
-    unistd::{self},
+    unistd,
 };
 #[cfg(feature = "libseccomp")]
 use oci_spec::runtime;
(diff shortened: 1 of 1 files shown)


rumpl (approved):
Nice!

ipuustin (commented):
LGTM. The only thing we need to check is that there won't be a race (where libcontainer succeeds in reaping the process before the other waitpid user does it). But that's not really youki's concern.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -107,15 +108,22 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
             log::warn!("intermediate process failed with exit status: {s}");
         }
-        WaitStatus::Signaled(_, sig, _) => {
+        Ok(WaitStatus::Signaled(_, sig, _)) => {
             log::warn!("intermediate process killed with signal: {sig}")
         }
-        _ => (),
+        Ok(_) => (),
+        Err(err) => {
+            if err == Errno::ECHILD {
```
yihuaf, line 121: Nit: Can this be turned into `Err(Errno::ECHILD)` case?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -107,15 +108,22 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+        Ok(WaitStatus::Exited(_, s)) => {
             log::warn!("intermediate process failed with exit status: {s}");
         }
-        WaitStatus::Signaled(_, sig, _) => {
+        Ok(WaitStatus::Signaled(_, sig, _)) => {
             log::warn!("intermediate process killed with signal: {sig}")
         }
-        _ => (),
+        Ok(_) => (),
+        Err(err) => {
```
yihuaf, line 120: Can you add the comment in the PR explaining it is safe ignoring ECHILD to here to the explanation is close to the code?

yihuaf (approved):
LGTM

crates/libcontainer/src/process/container_main_process.rs
utam0k: I think we don't need to do it because `waitpid` returns  [Result](https://doc.rust-lang.org/nightly/core/result/enum.Result.html)<T, [Errno](https://docs.rs/nix/latest/nix/errno/enum.Errno.html)>
https://docs.rs/nix/latest/nix/type.Result.html

crates/libcontainer/src/process/container_main_process.rs
utam0k: Sure!

PR #1781, opened 2023-04-06
Fix path to youki binary in dockerd command

kemkemG0:
### Update README.md with correct --add-runtime path

Hi, I noticed that the README.md file still contains the old target path for the --add-runtime flag.

Specifically, this part:
```
$ dockerd --experimental --add-runtime="youki=$(pwd)/target/x86_64-unknown-linux-gnu/debug/youki"
```
I think this doesn't match the updated target path in the Makefile.

So I update the path to ```$(pwd)/youki```

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ $ ../youki run rootless-container   # will create and run a container with rootl
 Start the docker daemon.
 
 ```console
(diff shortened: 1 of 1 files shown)


YJDoc2:
Hey @kemkemG0 Thanks for this update! Can I ask you to also update the same in docs?

YJDoc2 (approved):
Thanks!

Issue #1579, opened 2023-02-16
Build test error:  expected `u8`, found `i8` for bpf


sou-chon:
Hi I am trying to run `make unittest` and `make featuretest` but I am getting the following errors

<img width="895" alt="Screenshot 2023-02-16 at 15 56 54" src="https://user-images.githubusercontent.com/35537528/219419706-248ee29c-4480-47ad-872d-4470d306f28f.png">

YJDoc2:
Hey @sou-chon , can you share the details of the system you are building youki on? I tried running these on my system, and they seem to work without any error. Thanks :)

kemkemG0:
I just made a PR #1792 that seems to be related.

utam0k:
If you haven't resolved this issue yet, @sou-chon, please don't hesitate to reopen it.

utam0k:
Thanks, @kemkemG0 💯

PR #1792, opened 2023-04-07
Modify pointer type from i8 to c_char

kemkemG0:
## [Environment Info]

```
root ➜ /workspaces/youki  $ uname -a
Linux 80f9ee1dbb2e 5.15.49-linuxkit  aarch64  GNU/Linux
```

## Description

When running the command: 

```
$ make unittest
```

I encountered the following error:

```
error[E0308]: mismatched types
  --> crates/libcgroups/src/v2/devices/bpf.rs:39:22
   |
39 |             log_buf: ptr::null_mut::<i8>(),
   |                      ^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
   |
   = note: expected raw pointer `*mut u8`
              found raw pointer `*mut i8`
```
I found a similar error in another repository, which can be found here: https://github.com/remacs/remacs/issues/1393#issuecomment-475803562

My understanding is that the pointer type on ARM and x86 is defined differently. Following the advice on the issue, I used the c_char type instead of i8.

I also noticed that while the issue uses libc::c_char, this repository uses ::std::os::raw::c_char in some other parts. After reading this thread: https://users.rust-lang.org/t/libc-c-char-vs-std-os-c-char/21198, I decided to use ::std::os::raw::*.

Therefore, I used ::std::os::raw::c_char in my modifications.

### DIFF
--- a/crates/libcgroups/src/v2/devices/bpf.rs
+++ b/crates/libcgroups/src/v2/devices/bpf.rs
@@ -36,7 +36,7 @@ pub mod prog {
         let insns = insns as *const _ as *const bpf_insn;
         let opts = libbpf_sys::bpf_prog_load_opts {
             kern_version: 0,
-            log_buf: ptr::null_mut::<i8>(),
+            log_buf: ptr::null_mut::<::std::os::raw::c_char>(),
             log_size: 0,
             ..Default::default()
         };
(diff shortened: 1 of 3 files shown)


kemkemG0:
Well, I think I made a mistake on the GitHub operation... I'll fix it soon ;(

kemkemG0:
Now I realized that #1579 is related.

Furisto:
Hey @kemkemG0 changes looks good. Can you please rebase to get rid of the merge commits? The integration test is failing because of a clippy lint. Would you like to fix that as well or should we do it separately?

kemkemG0:
Hi, @Furisto thanks for your review. I will rebase to get rid of the merge commits.

However, I don't see why the clippy lint fails as I didn't modify the part clippy is a warning.

Should I fix that part as it suggests?

I apologize if I misunderstood the point ;)


Screenshot 2023-04-09 at 5 07 39 PM

kemkemG0:
I apologize for messing up the commit logs .. :(
I am not used to rebasing or dealing with forked repo...
I'll fix it soon

Furisto (approved):

Furisto:
Thanks @kemkemG0

PR #1793, opened 2023-04-09
Update version check in validate_spec to support 1.X.Y version. 

utam0k:
Previously, the validate_spec function only supported runtime spec versions that started with "1.0".
This commit updates the function to support all versions starting with "1." (e.g., 1.X.Y), making it compatible with a broader range of runtime spec versions.

https://github.com/containerd/runwasi/pull/78#issuecomment-1499597585


yihuaf (approved):

utam0k:
@yihuaf Thanks for your review 🙇

PR #1818, opened 2023-04-14
youki exec should not clean up on error

### LINKED ISSUE #1810: Youki exec a does not exist process will stop running container
# Steps to reproduce:

## Step 1
use `youki run -d {container_id}` command to start a container in background mode and keep the container running continuously.
and the container status will become `Running` after execution.

## Step 2
Use `youki exec -it {container_id} /bin/bash` to enter the container. If `/bin/bash` does not exist in the container, it will fail to enter the container, and the container status will become `Stopped` after execution.

I use runc to test, there is no such problem.

The following is the error exec log:
```
[DEBUG crates/libcontainer/src/capabilities.rs:135] 2023-04-13T14:32:53.549633813+08:00 dropping bounding capabilities to Some({AuditWrite, Kill, NetBindService})
[DEBUG crates/libcontainer/src/process/fork.rs:40] 2023-04-13T14:32:53.550891240+08:00 failed to run child process in clone: executable '/bin/bash' for container process does not exist
[INFO crates/libcgroups/src/common.rs:208] 2023-04-13T14:32:53.552311362+08:00 cgroup manager V2 will be used
[DEBUG crates/libcgroups/src/v2/manager.rs:132] 2023-04-13T14:32:53.552426219+08:00 remove cgroup "/sys/fs/cgroup/test"
exec failed : failed to create container
```

yihuaf:
Fix #1810

### DIFF
--- a/crates/libcontainer/src/container/builder_impl.rs
+++ b/crates/libcontainer/src/container/builder_impl.rs
@@ -52,10 +52,17 @@ impl<'a> ContainerBuilderImpl<'a> {
     pub(super) fn create(&mut self) -> Result<Pid> {
         match self.run_container().context("failed to create container") {
             Ok(pid) => Ok(pid),
-            Err(outer) => {
+            Err(outer) if matches!(self.container_type, ContainerType::InitContainer) => {
+                // Only the init container should be cleaned up in the case of
+                // an error.
                 if let Err(inner) = self.cleanup_container() {
                     return Err(outer.context(inner));
                 }
+                Err(outer)
+            }
+            Err(outer) => {
+                // In the cases of tenant containers, we should not clean up.
+                // Simply returns the error here.
                 Err(outer)
             }
         }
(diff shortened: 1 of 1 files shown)


YJDoc2:
Hey @yihuaf , what are your thoughts on moving the if-check inside match, instead of using it as a match guard ? There is no particular benefit to doing one instead of other, but I feel that it might more sense to move it inside like :
```rust
Ok(pid)=>...
Err(outer)=>{
     if matches!(....){
          // cleanup, if that errors , return outer.context(inner)
     }
     Err(outer)
}
```

I don't know if this will be better, so giving lgtm, because looks good :+1:

YJDoc2 (approved):
lgtm :+1:

utam0k (commented):
May I ask you to write the integration test for this case?

yihuaf:
> May I ask you to write the integration test for this case?

Yes I was thinking about the same.

utam0k:
@yihuaf 👍

Issue #443, opened 2021-10-30
Support for Intelrdt


utam0k:
https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#intelrdt

ipuustin:
I'm working on this. Hoping to have a PR soon!

PR #1822, opened 2023-04-17
Add support to Intel RDT.

ipuustin:
Add support to Intel RDT.

Fixes: #443.

This PR implements also the suggested additions to OCI runtime spec here: https://github.com/opencontainers/runtime-spec/pull/1196


utam0k:
@ipuustin Thanks for your PR. Since I need some time to learn Interl RDT, I'll check this next weekend.

ipuustin:
Note that this PR is needed to oci-spec-rs in order for the `closID` spec setting to have any effect: https://github.com/containers/oci-spec-rs/pull/128

crates/libcontainer/Cargo.toml
```diff
@@ -42,6 +42,8 @@ serde_json = "1.0"
 syscalls = "0.6.9"
 rust-criu = "0.4.0"
 clone3 = "0.2.3"
+regex = "1.7.3"
+lazy_static = "1.4.0"
```
utam0k, line 46: Since we are using `once_cell`, please use it instead of `lazy_static` :pray:

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -69,6 +72,7 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
     // The intermediate process will send the init pid once it forks the init
     // process.  The intermediate process should exit after this point.
     let init_pid = main_receiver.wait_for_intermediate_ready()?;
+    let mut need_to_clean_up_intel_rdt_subdirectory: bool = false;
```
utam0k, line 75: nits
```suggestion
    let mut need_to_clean_up_intel_rdt_subdirectory = false;
```

utam0k (changes requested):
Wait for updating oci-spec-rs

ipuustin:
@utam0k Thanks for reviewing! I implemented the suggested changes.

utam0k:
@ipuustin Thanks 😍 May I ask you to give me the instructions to confirm the behavior you have implemented?

ipuustin:
The documentation of the kernel resctrl interface (which this PR uses) is here: https://docs.kernel.org/x86/resctrl.html

Briefly, you mount the resctrl filesystem (`mount -t resctrl resctrl /sys/fs/resctrl`) and then you should be set. The code doesn't really concern itself with what the resctrl actually does: it just maps the OCI config values to the `/sys/fs/resctrl/<container_id>/schemata` file, with the instructions speficied in the OCI spec. In case you specify the `closID` value, the changes are done to `/sys/fs/resctrl/<clos_id>/schemata` file instead (if the directory doesn't exist already, otherwise the proposed changes will just be compared with the existing schemata file).

Which changes can you do to the schemata files depend on the computer system you have in use. A good approximation would be to see what the kernel has prepopulated to the schemata files and try to modify those values.

The OCI config spec has this example for the config.json file:

    "linux": {
        "intelRdt": {
            "closID": "guaranteed_group",
            "l3CacheSchema": "L3:0=7f0;1=1f",
            "memBwSchema": "MB:0=20;1=70"
        }
    }

ipuustin:
Rebased to fix conflict in Cargo.lock.

utam0k (approved):
:100:

PR #1868, opened 2023-05-01
Refactor the lifecycle test

yihuaf:
As previously discussed in #1827 , the `TestResult` is meant to be used in the test cases and not utility functions. Before this PR, the lifecycle test calls other tests using `TestResult` which is not inline with our intention. This PR fix this by refactoring all tests to not call other tests in test cases.

### DIFF
--- a/tests/rust-integration-tests/integration_test/src/tests/lifecycle/container_create.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/lifecycle/container_create.rs
@@ -42,29 +42,36 @@ impl ContainerCreate {
                 let _ = delete::delete(&self.project_path, &self.container_id);
                 TestResult::Passed
             }
-            Err(_) => TestResult::Failed(anyhow::anyhow!(
-                "Container should have been created with valid id, but was not created."
-            )),
+            Err(err) => {
+                TestResult::Failed(err.context(
+                    "container should have been created with valid id, but was not created.",
+                ))
+            }
         }
     }
 
     // runtime should not create container with is that already exists
     fn create_duplicate_id(&self) -> TestResult {
         let id = generate_uuid().to_string();
-        let _ = create::create(&self.project_path, &id);
-        match create::create(&self.project_path, &id) {
-            Ok(()) => {
-                let _ = kill::kill(&self.project_path, &id);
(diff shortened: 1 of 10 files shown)


tests/rust-integration-tests/integration_test/src/tests/lifecycle/container_create.rs
```diff
@@ -26,39 +26,45 @@ impl ContainerCreate {
...
-        match temp {
-            TestResult::Passed => TestResult::Failed(anyhow::anyhow!(
-                "Container should not have been created with same id, but was created."
-            )),
-            TestResult::Failed(_) => TestResult::Passed,
-            TestResult::Skipped => TestResult::Skipped,
+        match create::create(&self.project_path, &id) {
+            Ok(()) => {
+                let _ = kill::kill(&self.project_path, &id);
+                let _ = delete::delete(&self.project_path, &id);
```
YJDoc2, line 58: This is duplicated in both arms, so we should extract it before the match.

tests/rust-integration-tests/integration_test/src/tests/lifecycle/container_create.rs
```diff
@@ -26,39 +26,45 @@ impl ContainerCreate {
...
-            delete::delete(&self.project_path, &self.container_id);
+        match create::create(&self.project_path, &self.container_id) {
+            Ok(_) => {
+                let _ = kill::kill(&self.project_path, &self.container_id);
+                let _ = delete::delete(&self.project_path, &self.container_id);
+                TestResult::Passed
+            }
+            Err(_) => TestResult::Failed(anyhow::anyhow!(
+                "Container should have been created with valid id, but was not created."
+            )),
```
YJDoc2, line 47: Maybe we should bubble the actual error up too (through the `anyhow!`), so if this ever fails, we can get the error

YJDoc2 (changes requested):
Hey, overall its fine, just a few nits.

As you had said this makes much more sense that what we currently have. Thanks for taking initiative and making this change!

yihuaf:
@YJDoc2 PTAL

YJDoc2 (approved):
lgtm :+1:

Issue #1377, opened 2022-11-27
Investigate to use `thiserror` in our crates


utam0k:
# Introduction

Now we are using `anyhow` everywhere. But It would be better to provide some codes as crates(e.g. libcontainer, libcgroup...). Their crates in their current state are not very user-friendly of error handling.

# Goal

Investigate and implement to use of `thiserror` instead of `anyhow` in crates

# Tracking

- [ ] [liboci-cli](https://github.com/containers/youki/tree/main/crates/liboci-cli)
- [ ] [libcgroups](https://github.com/containers/youki/tree/main/crates/libcgroups)
- [ ] [libcontainer](https://github.com/containers/youki/tree/main/crates/libcontainer)

utam0k:
@chermehdi I'm happy if you have interested in this issue. But this issue is not easy and take a lof of time to achieve better than other good first issues.

chermehdi:
Sounds good, let us do them in that order, the first one should give us a pretty good idea about how will the others look like, and how much effort is required as well.

utam0k:
@chermehdi That's right, libcontainer is going to be quite large task, so if there is a way to make it divisible that would be nice too. May I assign you? There is no deadline, etc. Of course, I can help you whenever.

chermehdi:
@utam0k yeah sounds good, just assing me.

chermehdi:
@utam0k `liboci-cli` only contains arg definitions for the cli implementation. I don't see any behavior that requires error handling, I am not sure why it was included as part of this issue. Can you clarify?

utam0k:
> @utam0k `liboci-cli` only contains arg definitions for the cli implementation. I don't see any behavior that requires error handling, I am not sure why it was included as part of this issue. Can you clarify?

Ops, sorry. You don't need to take it. Let's look into libcgroups.

chermehdi:
@utam0k Where do we want to use thiserror? I suggest to minimize the changes we only introduce thiserror usage right at the edge of the library and keep error handling internally using anyhow/context. this will minimize the number of changes we have to make.

Another thing that I want us to think about is what is our API boundary, and how we want this to be used as a library.

* Do we want to provide libcgroups as it's own crate? or will we provide Youki as a library as a whole with the API like `createFromSpec`, `start`, `stop`?

utam0k:
@chermehdi Sorry for the late reply.

:100: 
[...]

PR #1872, opened 2023-05-02
Adopt `thiserror` for libcgroups

squili:
As prescribed in #1377

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -60,60 +60,60 @@ pub enum AnyManagerError {
     V2(#[from] v2::manager::V2ManagerError),
 }
 
-pub enum AnyManager {
+pub enum AnyCgroupManager {
     Systemd(systemd::manager::Manager),
     V1(v1::manager::Manager),
     V2(v2::manager::Manager),
 }
 
-impl CgroupManager for AnyManager {
+impl CgroupManager for AnyCgroupManager {
     type Error = AnyManagerError;
 
     fn add_task(&self, pid: Pid) -> Result<(), Self::Error> {
         match self {
-            AnyManager::Systemd(m) => Ok(m.add_task(pid)?),
-            AnyManager::V1(m) => Ok(m.add_task(pid)?),
-            AnyManager::V2(m) => Ok(m.add_task(pid)?),
+            AnyCgroupManager::Systemd(m) => Ok(m.add_task(pid)?),
+            AnyCgroupManager::V1(m) => Ok(m.add_task(pid)?),
+            AnyCgroupManager::V2(m) => Ok(m.add_task(pid)?),
         }
     }
 
     fn apply(&self, controller_opt: &ControllerOpt) -> Result<(), Self::Error> {
         match self {
-            AnyManager::Systemd(m) => Ok(m.apply(controller_opt)?),
-            AnyManager::V1(m) => Ok(m.apply(controller_opt)?),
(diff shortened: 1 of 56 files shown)


crates/libcgroups/src/stats.rs
```diff
@@ -317,76 +362,107 @@ pub fn parse_nested_keyed_data(file_path: &Path) -> Result<HashMap<String, Vec<S
     Ok(stats)
 }
 
-/// Parses a file that is structed according to the nested keyed format
-/// # Example
-/// ```
-/// use libcgroups::stats::parse_device_number;
```
yihuaf, line 323: Do we need to keep the example here?

yihuaf (approved):
LGTM

crates/libcgroups/src/common.rs
```diff
@@ -92,39 +161,103 @@ pub struct ControllerOpt<'a> {
...
+    #[error("failed to write {data} to {path}: {err}")]
+    Write {
+        err: std::io::Error,
+        path: PathBuf,
+        data: String,
+    },
+    #[error("failed to read {path}: {err}")]
+    Read { err: std::io::Error, path: PathBuf },
+    #[error("failed to create dir {path}: {err}")]
+    CreateDir { err: std::io::Error, path: PathBuf },
```
utam0k, line 177: I love this WrappedIoError idea ❤️ 
If the dirs cannot be created in the libcgroup crates, does this result in the failure to create a new cgroup?
I believe it is more user-friendly to display a message such as "failed to create the cgroup" or something similar. WDYT?

crates/libcgroups/src/v1/controller_type.rs
```diff
@@ -1,6 +1,6 @@
 use std::fmt::Display;
 
-#[derive(Hash, PartialEq, Eq, Debug, Clone)]
+#[derive(Hash, PartialEq, Eq, Debug, Clone, Copy)]
```
utam0k, line 3: Why we need `Copy`?

crates/youki/src/commands/mod.rs
```diff
@@ -4,7 +4,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use libcgroups::common::CgroupManager;
+use libcgroups::common::AnyManager;
```
utam0k, line 7: I believe that the name "AnyManager" may cause confusion. Would "CgroupManager" or "AnyCgroupManager" be better options?

utam0k:
It looks like that lint said the error. Please check them 🙏

crates/libcgroups/src/stats.rs
squili: The example was triggering doctets, which then failed because the function was made crate-only

crates/libcgroups/src/common.rs
squili: That's what errors higher up in the chain are for - they can be added to the higher-level error adts with more user-friendly messages

crates/libcgroups/src/v1/controller_type.rs
squili: This enum is essentially just a fancy number, so the performance of passing by reference and copying are the same

squili: Specifically, this allows us to do `*controller_type` instead of `controller_type.clone()` to turn a reference into an owned value

utam0k (approved):
💯 
I can't believe that you make this PR in such a short period of time 😲

PR #1876, opened 2023-05-03
Implement `thiserror` for libcontainer - Part 1

yihuaf:
This is a first attempt to add `thiserror` to `libcontainer` and all comments are welcome. Note, I decide to break this effort into multiple parts because a single PR would be too large to reasonably review. The aim is to start at the bottom turning the errors into `thiserror` while keeping the top level `anyhow::Error` until we finish this effort. By keeping the top level as `anyhow::Error` for the moment, we can easily turn `thiserror` into an `anyhow::Error`, but not vice versa.

A few notes. I tried my best to create appropriate error, but they are not perfect in any means. I would like in this PR to focus on the broader effort to make the transition, instead of focusing on the specifics of a single error. We can always make additional PR to fix.

I am not sure if I am using the `libcgroups` error correctly. For now, I just turned them into string messages. We may need to spend more effort since it currently only returns a `std::error::Error + 'static` trait, rather than a concrete type where to can expose structurally.


crates/libcontainer/src/syscall/mod.rs
```diff
@@ -8,3 +8,109 @@ pub mod syscall;
 pub mod test;
 
 pub use syscall::Syscall;
+#[derive(Debug, thiserror::Error)]
+pub enum SyscallError {
```
utam0k, line 12: 👍

utam0k (approved):

utam0k:
@squili May I ask you to review this PR?

squili (approved):
This is great! I love how you structured the error ADTs!

utam0k:
Thanks @squili

PR #1881, opened 2023-05-04
Implemented more `thiserror` for libcontainer (Part 2)

yihuaf:
Implemented more `thiserror` for the libcontainer. This PR covers most of the low hanging fruits with smaller crates. Note, I decided to add a new syscall error wrapper that wraps the `nix::Error`, `std::io::Error`, and our own syscall crate. This aims to simplify where we call syscalls from 3 different sources and can simplify/unify many errors.

CC @squili PTAL

Part of #1377


utam0k (approved):

utam0k:
@squili 
Could you review this PR?

crates/libcontainer/src/notify_socket.rs
```diff
@@ -8,29 +7,67 @@ use std::path::{Path, PathBuf};
 
 pub const NOTIFY_FILE: &str = "notify.sock";
 
+#[derive(Debug, thiserror::Error)]
+pub enum NotifyListenerError {
+    #[error("failed to chdir when create notify socket")]
```
squili, line 12: maybe `"failed to chdir while creating notify socket"` would be more clear

squili (changes requested):
Looks great - just a minor nitpick

squili (approved):

PR #1884, opened 2023-05-06
Add easy way to test with K8s

utam0k:
You can confirm this script: 
```console
$ make test/k3s
[...]

$ make test/k3s/clean
```


tests/k8s/deploy.yaml
```diff
@@ -0,0 +1,27 @@
...
+      labels:
+        app: nginx
+    spec:
+      runtimeClassName: youki
+      containers:
+      - name: nginx
+        image: nginx:1.16.1
+        ports:
+        - containerPort: 80
+      automountServiceAccountToken: false
```
utam0k, line 27: Deleting this line, unfortunately, youki gives an error 😭

Makefile
```diff
@@ -60,6 +64,55 @@ test-oci: oci-tests rust-oci-tests
...
+	sudo cp /var/lib/rancher/k3s/agent/etc/containerd/config.toml /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl && \
+	echo 'default_runtime_name = "youki"' | sudo tee -a /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl && \
+	echo '[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.youki]' | sudo tee -a /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl && \
+	echo '  runtime_type = "io.containerd.runc.v2"' | sudo tee -a /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl && \
+	echo '  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.youki.options]' | sudo tee -a /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl && \
+	echo '    BinaryName = "$(PWD)/youki"' | sudo tee -a /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl && \
+	echo "CONTAINERD_NAMESPACE='default'" | sudo tee /etc/systemd/system/k3s-runwasi.service.env && \
+	echo "NO_PROXY=192.168.0.0/16" | sudo tee -a /etc/systemd/system/k3s-runwasi.service.env && \
+	sudo systemctl daemon-reload && \
+	sudo systemctl restart k3s-youki && \
```
yihuaf, line 78: In general, I recommend separate the install step from the actual test step. We can create a `prepare` target that sets up the k3s environment for the test, then `make test` should only check the dependency and run the test. In this way, the test can run multiple times without repeating the install step.

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -204,7 +204,7 @@ pub fn container_init_process(
         // before pivot_root is called. This runs in the container namespaces.
         if let Some(hooks) = hooks {
             hooks::run_hooks(hooks.create_container().as_ref(), container)
-                .context("Failed to run create container hooks")?;
+                .context("failed to run create container hooks")?;
```
yihuaf, line 207: We don't need this since these will be replaced with `thiserror`. Same for the others as well.

tests/k8s/deploy.yaml
yihuaf: Do you have an idea why? We should track this as an issue.

yihuaf (approved):
Great! One this to note is that this is more `makefile`. @YJDoc2 How is the progress on the `justfile` migration? Should we wait?

YJDoc2:
@yihuaf The work is almost done, however I will take today/tomorrow to update the PR. In any case this should not be stopped for justfile. I'll update my PR for these changes as needed.

Once this is ready to go, please merge without waiting on Justfile PR :+1:

crates/libcontainer/src/process/container_init_process.rs
utam0k: Prioritize the merge for subsequent work 🙇

Makefile
utam0k: @YJDoc2 Since I can't take any more time today, but I don't want to interrupt your work, can I ask you to fix it with just PR?

Issue #1883, opened 2023-05-06
Clean up and unify the use of tempdir and tempfile


yihuaf:
Currently, we are rolling our own implementation of tempdir and duplicated the code in multiple places. In addition, I don't like the struct implements using `option<path>` and the `path()` has a `expect()` which can panic if called after remove. The proposed solution is to use `tempfile` crate. https://github.com/Stebalien/tempfile

- [x] libcontainer #1887 
- [x] youki #1887 
- [x] integration test #1891 
- [x] libcgroup #1888

utam0k:
👍

PR #1887, opened 2023-05-08
Migrate to `tempfile` for `libcontainer` and `youki` crate

yihuaf:
As explained in #1883, we would like to move away from rolling out our own tempdir implementation. This is the first PR to migrate `libcontainer` and `youki` crate. The `libcgroup` and `integrate_test` will follow.

A few notes:

First, I remove using test name to create unique tempdir. Since the tempdir are designed to go away once out of the scope, using random names is enough to create unique tempdir. The directory will be removed regardless of test successful or not, so there is no reason to use human readable names.

Second, a number of test is partly re-written so it works better with the `tempdir` crate. Notably, we introduced `scopeguard` and `defer!()` to recreate the `Drop` delete a specific path that is not inside a tempdir logic. I find this to communicate the intent of the test more clear.


crates/libcontainer/src/rootless.rs
```diff
@@ -7,6 +7,42 @@ use std::path::Path;
 use std::process::Command;
 use std::{env, path::PathBuf};
 
+// Wrap the uid/gid path function into a struct for dependency injection. This
+// allows us to mock the id mapping logic in unit tests.
+#[derive(Debug, Clone, Default)]
+pub struct RootlessIDMapper {
+    pub path: Option<PathBuf>,
```
utam0k, line 14: Since I feel this name `path` is too abstract, it was difficult to determine what field it was. Can we name it something more specific? I like the mechanism itself 👍

crates/youki/src/main.rs
```diff
@@ -147,195 +139,208 @@ fn main() -> Result<()> {
     cmd_result
 }
 
-fn determine_root_path(root_path: Option<PathBuf>) -> Result<PathBuf> {
```
utam0k, line 150: Is it possible to have this cut out into a separate file? 🙏

crates/youki/src/main.rs
```diff
@@ -147,195 +139,208 @@ fn main() -> Result<()> {
...
-    let uid = getuid().as_raw();
+mod rootpath {
```
utam0k, line 142: Is it possible to have this cut out into a separate file? 🙏

crates/youki/src/main.rs
```diff
@@ -147,195 +139,208 @@ fn main() -> Result<()> {
...
+    use anyhow::{bail, Result};
+    use libcontainer::rootless::rootless_required;
+    use libcontainer::utils::create_dir_all_with_mode;
+    use nix::libc;
+    use nix::sys::stat::Mode;
+    use nix::unistd::getuid;
+    use std::fs;
+    use std::path::{Path, PathBuf};
+
+    pub fn determine(root_path: Option<PathBuf>) -> Result<PathBuf> {
```
utam0k, line 152: It looks like nothing has changed, but is that right?

yihuaf: Yes, nothing has changed, except I moved them into a separate mod.

yihuaf:
@utam0k Addressed the review. PTAL :)

PR #1891, opened 2023-05-08
Migrate integration test to use tempfile

yihuaf:
This is the last part of switching to the `tempfile` crate. The integration test migration is mostly straight forward. Previously, we use container id as the bundle name. Now we use a random name from the `tempfile` as the bundle path. There is no hard requirement that the bundle path name has to be the container id (uuid), so we are fine here. Also, since the bundle path prepared using `tempfile`, it will be cleaned up regardless if test success or fails. So we are safe to make the change.

There are some minor edits to the tests I tagged along. It is small enough that I am too lazy to separate out as a different PR :)


utam0k (approved):
Good work

Issue #1348, opened 2022-11-15
OpenTelemetry tracing support to extend observability


saschagrunert:
I think this project could outstandingly shine when featuring tracing support. We have a implementation in conmon-rs which supports exporting data to the OpenTelemetry exporter (OTLP): https://github.com/containers/conmon-rs/blob/6cfba5a/conmon-rs/server/src/telemetry.rs

Everything is based on the [tracing crate](https://docs.rs/tracing/latest/tracing), which is also used for logging.

What do you think about that?

utam0k:
@YJDoc2 @Furisto @yihuaf I'd like to hear your input.

utam0k:
https://kubernetes.io/blog/2022/12/01/runtime-observability-opentelemetry/

saschagrunert:
@alban mentioned in https://social.linux.pizza/@albcr/109437558144011212 that we could use an annotation to pass the trace ID around. This way we can link youki/runc/crun with higher level runtimes like CRI-O.

YJDoc2:
Hey, I haven't yet looked into this in depth, but from what I have seen, and read this seems like a good idea if we can support. We might not need to make too many changes for it, however I think we must feature-gate this, and disable it by default. As stated in the blog, this will cause some performance degradation, and given that _usually_ a user will need some additional tools to take full advantage of the spans and stuff, it is better left to the user to decide if to opt-in or not.

I'll try taking a more detailed look and maybe add details on how we could go on adding the support.

lengrongfu:
> @alban mentioned in https://social.linux.pizza/@albcr/109437558144011212 that we could use an annotation to pass the trace ID around. This way we can link youki/runc/crun with higher level runtimes like CRI-O.

For low-level runtimes such as `youki` or `runc`, they are more called by `containerd` high-level runtime. I think we should first communicate with upstream `containerd` on how to transmit traceID and spandID.

yihuaf:
Introducing `tracing` to `youki` would be a great idea. We can also clean up the logging libs we have, since logging becomes a special case of tracing. This seems to be where the rust ecosystem is moving to anyway, so we should embrace it. I will take a jab at this after I migrate the libcontainer to `thiserror`.

utam0k:
@yihuaf 👍

utam0k:
@saschagrunert Do you have an idea how to get the tracing id from the CRI layer? container id?

utam0k:
[...]

PR #1899, opened 2023-05-10
convert youki to use tracing

yihuaf:
This PR converts youki to use `tracing` crate for logging. I did the minimal to make the transition, but `tracing` crate allows for much richer logging and instrumentation. This is related to #1348 in that this would be a pre-requirement to implement open-telemetry.

In addition, the reason I decide to do this now because error handling can be difficult without tracing/logging support. In general with error handling, the audience are either the user/operator (human) or for code to handle. It is important to log as much as possible of the context of an error. There are usually 2 choices. Either we store the context of the error into the error itself, or we log the context and keep the error size small. There is no right answer and usually we need a combination of both. `tracing` crate allows us to minimize the context we have to store inside the error itself, and instead opting for rich logging.

Note, I tried to keep the change minimal, but switching logging infrastructure does have a wide impact. Hopefully this is not too disruptive to people.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1756,7 +1756,6 @@ dependencies = [
  "flate2",
  "libcgroups",
  "libcontainer",
- "log",
  "nix",
  "num_cpus",
  "oci-spec",
@@ -1770,6 +1769,8 @@ dependencies = [
  "tar",
  "tempfile",
  "test_framework",
+ "tracing",
+ "tracing-subscriber",
  "uuid",
  "which",
 ]
@@ -1930,7 +1931,6 @@ dependencies = [
  "fixedbitset",
  "libbpf-sys",
  "libc",
- "log",
  "mockall",
  "nix",
  "oci-spec",
@@ -1942,6 +1942,7 @@ dependencies = [
  "serial_test",
  "tempfile",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
@@ -1959,7 +1960,6 @@ dependencies = [
  "libc",
  "libcgroups",
  "libseccomp",
- "log",
  "mio",
  "nix",
  "oci-spec",
@@ -1977,6 +1977,7 @@ dependencies = [
  "syscalls",
  "tempfile",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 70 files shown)


yihuaf:
I am able to reproduce this now. First, the old behavior is the same as the new, with the log line being slightly different, but the overall length is similar. In fact, the tracing in this PR produced a shorter line since the file path and line number is not turned on by default. Therefore, I believe this is a separate issue not related to logging or tracing.

To resolve this, it seems that bpftrace is omitting the long lines. I can confirm that my shell window will wrap the line correctly. So bpftrace is cutting off the lines.

yihuaf:
The integration test is failing and I suspect this has to do with the tempfile change. I will investigate.

yihuaf:
> The container runtime does not work with this timescale.

This is a good point.

> bpftrace is probably currently the most usable way to fix bugs in kubernetes. I am not sure how to do it with other ways, because it is too deep in the hierarchy from kubelet, and it is difficult to do.

Oh, this is how you are using `bpftrace`. This makes sense.

yihuaf:
I turned off the timestamp for the text format to stderr. This should produce a shorter log line with log level, log target, and the actual log line. I left the other format unchanged for now. Let me know if you think we should remove timestamp for other cases as well (JSON and logging to file).

> I don't think it would be desirable to make the logs longer simply from the standpoint of readability.

Normally, this is not too much of an issue because we can make a long log line to be more human readable with nice formatting. In fact, the tracing crate has a really nice `pretty` formatter to make the logs more readable. It breaks the long line into multiple lines and adds coloring to the log line.

However, in this specific case, we are using `bpftrace`. From this perspective, the log line is a single `write` syscall with a long line. It will also cut off at 225 ish char due to internal limits. This limits makes sense because `bpftrace` is not designed to read logs. It is designed to capture the syscall trace. In another word, you are hacking `bpftrace` as a log collector for youki, but I can see this is a nice way to debug youki under kubelet. Please correct me if I am misunderstanding you.

In my view, assume debugging is your only concern, the long term solution is to use something more appropriate to view tracing/logging. For example, we can configure the trace/log to write to journald or syslog. Then, by observing syslog or journald or using `journalctl`, we can view the logs from `youki` in a similar way on how you use `bpftrace` right now. Of course, there are other tracing viewers as well but they all work in similar ways. I believe these days, we can even ship the tracing to something like datadog if we really wants.

For now, let's keep it simple and see this the current setup works for us. But, the real super power of tracing is to log as much information as possible along with metrics. It can be pretty powerful beyond just logging and the amount of logs/data entry generated from tracing will become too much for `bpftrace` to handle at some point. We are not there now, but we slowly will.

utam0k:
@yihuaf I agree with you, logging with bpftrace is pretty hacky. It should be turned off. It would be best if it could be turned off in the future, but for now this debugging method is necessary.
However, if OpenTelemetry tracing could be placed anywhere and the sampling rate could be set to 100%, I think it could be turned off.

utam0k (approved):
Thanks a lot 🙏

PR #1900, opened 2023-05-10
deps: do not use chrono default-features.

ipuustin:
Make `chrono` dependency not use default features. This is because that's the recommended workaround for a `time` crate [vulnerability](https://rustsec.org/advisories/RUSTSEC-2020-0071).


YJDoc2:
Hey @ipuustin Looks good overall, but can you run the build locally and also commit the `cargo.lock` file that gets generated? Thanks!

YJDoc2 (approved):
lgtm :+1:
Thanks !

PR #1908, opened 2023-05-11
[Trivial] fix dependency for fedora

yihuaf:
Not sure if this is a typo or package name is changed, but `libclang-dev` does not exist on dns packages. Change to the correct name and verified on Fedora 38.


YJDoc2 (approved):
Thanks!

PR #1912, opened 2023-05-14
Implemented `thiserror` for libcontainer - Part 4

yihuaf:
This is the second to last PR. I converted most of the module in crate libcontainer to use `thiserror`. The missing 2 are `container` and `workload`. Both of these 2 requires a bit more thinking. The `workload` requires us to define the Executor type with an Error associated type. The container crate will be the main interface exposed by the `libcontainer` crate. It is easier to review these two changes in a different PR.


utam0k (approved):
👍 👍 👍

PR #1937, opened 2023-05-17
Finally, remove `anyhow` from the libcontainer dependency.

### LINKED ISSUE #1377: Investigate to use `thiserror` in our crates

yihuaf:
This is the final PR that implements the `thiserror` in libcontainer. Fix #1377

### DIFF
--- a/crates/libcontainer/src/error.rs
+++ b/crates/libcontainer/src/error.rs
@@ -59,8 +59,6 @@ pub enum LibcontainerError {
     State(#[from] crate::container::state::StateError),
     #[error("oci spec error")]
     Spec(#[from] oci_spec::OciSpecError),
-    // #[error("cgroups error: {0}")]
-    // Cgroups(String),
     #[error(transparent)]
     MainProcess(#[from] crate::process::container_main_process::ProcessError),
     #[error(transparent)]
--- a/crates/libcontainer/src/error.rs
+++ b/crates/libcontainer/src/error.rs
@@ -75,9 +75,9 @@ pub enum LibcontainerError {
     // Catch all errors that are not covered by the above
     #[error("syscall error")]
     OtherSyscall(#[source] nix::Error),
-    #[error("IO error")]
+    #[error("io error")]
     OtherIO(#[source] std::io::Error),
-    #[error("Serialization error")]
+    #[error("serialization error")]
     OtherSerialization(#[source] serde_json::Error),
     #[error("{0}")]
     OtherCgroup(String),
--- a/crates/libcontainer/src/error.rs
+++ b/crates/libcontainer/src/error.rs
@@ -26,9 +26,11 @@ pub enum MissingSpecError {
 #[derive(Debug, thiserror::Error)]
 pub enum LibcontainerError {
     #[error("failed to perform operation due to incorrect container status")]
-    IncorrectContainerStatus,
+    IncorrectStatus,
     #[error("container already exists")]
-    ContainerAlreadyExists,
+    Exist,
+    #[error("container state directory does not exist")]
+    NoDirectory,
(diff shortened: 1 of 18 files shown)


crates/libcontainer/src/error.rs
```diff
@@ -57,18 +59,28 @@ pub enum LibcontainerError {
...
+    CgroupCreate(#[from] libcgroups::common::CreateCgroupSetupError),
+    #[error(transparent)]
+    CgroupGet(#[from] libcgroups::common::GetCgroupSetupError),
 
     // Catch all errors that are not covered by the above
     #[error("syscall error")]
     OtherSyscall(#[source] nix::Error),
     #[error("IO error")]
     OtherIO(#[source] std::io::Error),
+    #[error("Serialization error")]
```
utam0k, line 80:
```suggestion
    #[error("serialization error")]
```

crates/youki/src/workload/wasmedge.rs
```diff
@@ -1,19 +1,18 @@
...
 
 const EXECUTOR_NAME: &str = "wasmedge";
 
 #[derive(Default)]
 pub struct WasmEdgeExecutor {}
 
-impl Executor for WasmEdgeExecutor {
-    fn exec(&self, spec: &Spec) -> Result<()> {
+impl WasmEdgeExecutor {
+    fn exec_inner(spec: &Spec) -> anyhow::Result<()> {
```
utam0k, line 15: Why don't we use this error here?

yihuaf: This is mostly to minimize the changes since the original code uses anyhow. The implementation of the 3 wasm executor is implemented in `youki`, not `libcontainer`, so there is no hard requirement to remove `anyhow`. I was not sure how best to approach this so I wrapped the original implementation into the `exec_inner` and only `map_err` once to meet the trait requirement of `Executor`.

I also wasn't sure how best to implement the `Executor` trait and the interface `libcontainer` should provide. This PR does the minimal so there are still rooms for improvement. I have a vague plan to implement the `Executor` trait with an `Error` associated type, so the implementation can provide whatever error type they wish. However, this proves to be tricky because we need to store the `Executor` in `ContainerBuilder` and `ContainerArgs`. I could not figure out how to store a trait object with unknown associated type yet. Perhaps the new Generic Associated Type can be of some help. In the end, I decided to keep things simple for now to get this PR merged, and then we can figure out how best to approach this.

On a different note, I suspect that `Executor` interface can become a great thing for youki and libcontainer, because it opens up a lot of possibility. Currently, we switching the the default `execvp` calls for the wasm calls. I can also switching the default `execvp` to run an arbitrary rust function. This would allow us to better and easier testing the `libcontainer` directly, without the full setup of the oci bundle.

This is a long winded way to answer your question :) But I can certainly remove the `anyhow` from these executor as well.

yihuaf: Done. Fixed these wasm executor not using anyhow.

utam0k (approved):
:tada::tada::tada::tada::tada::tada:

utam0k:
@yihuaf Thanks a lot 🙏

PR #1948, opened 2023-05-19
Fix the test to not use sigkill

yihuaf:
Sigkill is not the right signal to use because it kills the process without going through the handler. Use sigint instead.

Also in this PR, I decided to refactor the `test_in_child_process` function into a common crate (along with the channel dependency). The `test_in_child_process` function is useful for us because of the process isolation we can use in the test. Cargo runs all tests in the same process, so many of our tests would require process isolation. For example, here we also refactored the `tracing` init related tests using the `test_in_child_process` because we can only call init tracing once.

I called the new crate `common`. I am open to other names. I tried my best separating the crate into `test_utils` and the default feature with the channel implementation, so we can have different dependencies in tests.


crates/common/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "common"
```
utam0k, line 2: All crates must be registered on crates.io in order to register on crates.io.
At least this name cannot be used. I am also not in favor of having a utility-like crate in terms of having to register with crates.io. Why not just use the libcontainer one as public?

yihuaf: > All crates must be registered on crates.io in order to register on crates.io.

Good point.

Originally, the `test_utils` is meant to only be used as part of the test with `#[cfg(test)]`. Therefore, I can't share the functions outside of libcontainer with `youki` crate. Perhaps I can create a new `test_utils` feature for the `libcontainer`. Let me try that.

yihuaf: I implemented the `test_utils` as a feature for the libcontainer and set it to `youki` crate's dev dependency. Let me know what you think.

crates/libcontainer/src/test_utils.rs
```diff
@@ -0,0 +1,140 @@
+use nix::sys::wait;
```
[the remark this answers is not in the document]
yihuaf: Yes, I set the feature in the `lib.rs`

```
#[cfg(feature = "test_utils")]
pub mod test_utils;
```

I don't think I need to further specify things in the file itself, but I may be wrong.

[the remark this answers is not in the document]
utam0k: Oh, sorry, I missed it 😭

utam0k (approved):

PR #1949, opened 2023-05-20
Simplified syscall error

yihuaf:
- Refactored the syscall errors. There is no need to create an error for each syscall. Only the errno is what is required. Other context can be logged at the site of the errors.
- Removed the `UnifiedSyscallError`. It was not well thought out at the moment.
- Refactored close_range and mount_set_attr to use `libc::syscall` directly.


utam0k:
> Refactored the syscall errors. There is no need to create an error for each syscall. Only the errno is what is required. Other context can be logged at the site of the errors.

I agree with you! It would be wiser for them to see man than for us to give them the message from us.

utam0k (approved):

utam0k:
I've been working on this area lately, and thiserror makes it easy to identify errors. Thanks.

PR #1963, opened 2023-05-24
Add some clean up that improves coverage

yihuaf:
Remove some code that are no longer used.


crates/libcontainer/src/workload/default.rs
```diff
@@ -38,7 +38,7 @@ impl Executor for DefaultExecutor {
             ExecutorError::Execution(err.into())
         })?;
 
-        // After do_exec is called, the process is replaced with the container
```
utam0k, line 41: Sorry, I forgot to delete them 😭

utam0k (approved):

PR #1973, opened 2023-05-27
do not log error in the syscall crate

yihuaf:
Related to #1974. The integration tests uses `stderr` to check if there are any errors. However, there are a number of cases where we want to mount with ENOTDIR or ENOENT error but we ignore these errors. Therefore, we do not want to log these errors. Otherwise the integration tests would be broken.

tag along: Bump rust version to 1.70 and fix some comments lint in integration test. As a side effect, this will trigger the integration validation test, so we can test the fix is actually working in CI.

### DIFF
--- a/crates/libcontainer/src/rootfs/rootfs.rs
+++ b/crates/libcontainer/src/rootfs/rootfs.rs
@@ -38,7 +38,7 @@ impl RootFS {
         bind_devices: bool,
         cgroup_ns: bool,
     ) -> Result<()> {
-        tracing::debug!("Prepare rootfs: {:?}", rootfs);
+        tracing::debug!(?rootfs, "prepare rootfs");
         let mut flags = MsFlags::MS_REC;
         let linux = spec.linux().as_ref().ok_or(MissingSpecError::Linux)?;
 
@@ -52,20 +52,34 @@ impl RootFS {
         }
 
         self.syscall
-            .mount(None, Path::new("/"), None, flags, None)?;
+            .mount(None, Path::new("/"), None, flags, None)
+            .map_err(|err| {
+                tracing::error!(
+                    ?err,
+                    ?flags,
+                    "failed to change the mount propagation type of the root"
+                );
+
+                err
+            })?;
 
         let mounter = Mount::new();
 
         mounter.make_parent_mount_private(rootfs)?;
 
         tracing::debug!("mount root fs {:?}", rootfs);
-        self.syscall.mount(
-            Some(rootfs),
-            rootfs,
-            None,
-            MsFlags::MS_BIND | MsFlags::MS_REC,
-            None,
-        )?;
+        self.syscall
+            .mount(
+                Some(rootfs),
+                rootfs,
+                None,
+                MsFlags::MS_BIND | MsFlags::MS_REC,
+                None,
+            )
+            .map_err(|err| {
+                tracing::error!(?rootfs, ?err, "failed to bind mount rootfs");
+                err
+            })?;
 
         let global_options = MountOptions {
             root: rootfs,
@@ -108,9 +122,16 @@ impl RootFS {
         };
 
         if let Some(flags) = flags {
-            tracing::debug!("make root mount {:?}", flags);
             self.syscall
-                .mount(None, Path::new("/"), None, flags, None)?;
+                .mount(None, Path::new("/"), None, flags, None)
+                .map_err(|err| {
+                    tracing::error!(
+                        ?err,
+                        ?flags,
+                        "failed to adjust the mount propagation type of the root"
+                    );
+                    err
+                })?;
         }
 
         Ok(())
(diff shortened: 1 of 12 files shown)


crates/libcontainer/src/syscall/linux.rs
```diff
@@ -489,10 +490,20 @@ impl Syscall for LinuxSyscall {
         flags: MsFlags,
         data: Option<&str>,
     ) -> Result<()> {
```
utam0k, line 492: How about returning the type of error in `OK`?

utam0k (changes requested):
What about tracing at the caller? It is beyond the role of the syscall mod to determine the type of error.

yihuaf:
> What about tracing at the caller? It is beyond the role of the syscall mod to determine the type of error.

I made the change for `mount` only in this PR to unblock the CI. We can extend this to all syscalls in this mod, but I'd like to do this in another PR.

utam0k:
>  We can extend this to all syscalls in this mod, but I'd like to do this in another PR.

I agree with you. we should do it.

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -489,13 +490,7 @@ impl Syscall for LinuxSyscall {
...
-        mount(source, target, fstype, flags, data).map_err(|err| {
-            tracing::error!(
```
utam0k, line 493: Is there any need to bring this tracing to the callers?

yihuaf: Yes, 47bc85926f55c27cefd1925df7622a0b51910d75

yihuaf:
@utam0k since this PR hasn't been merged, I went ahead and clean up the syscall crate as well. All the simpler wrappers no longer log error and logs only at the caller site. I kept the other more complex functions such as `pivot_root` and `close_range` alone, since it makes sense to log the error there.

yihuaf:
@utam0k Can we get a review if you have some time? I would like to merge this since this will unblock the integration validation test. I would also ask if it is possible to merge this in as is and address any reviews in a follow up PR since the integration validation CI is failing.

YJDoc2:
Hey @yihuaf I think utam0k is unwell, so I'll take a look at this instead. The changes look ok, and as you said, if there are any issues, we can fix them in follow up PR. The only question I  have is that this was opened because  we logged errors in `mount` calls which makes the tests fail. However, I still see that we are doing tracing::error in the mount calls' `map_err` . I think you had removed is in one of the commits, but it got re-added for some reason. Can you please correct me if I'm wrong? Otherwise this is fine, I'll approve this then so it can unblock rest of the work.

yihuaf:
> Hey @yihuaf I think utam0k is unwell, so I'll take a look at this instead. The changes look ok, and as you said, if there are any issues, we can fix them in follow up PR. The only question I have is that this was opened because we logged errors in `mount` calls which makes the tests fail. However, I still see that we are doing tracing::error in the mount calls' `map_err` . I think you had removed it in one of the commits, but it got re-added for some reason. Can you please correct me if I'm wrong? Otherwise this is fine, I'll approve this then so it can unblock rest of the work.

@YJDoc2 Thank you for taking a look. So issue that caused the integration failure is the `syscall::mount` logging error. It is removed in this PR (https://github.com/containers/youki/pull/1973/files#diff-b7916a1b6e425d0953faba26141b604a4b71e65782a58c78dd6f59f620916dccR470). The right thing to do is logging at the caller of the `syscall::mount`. There are a few places where the `syscall::mount` will return `ENOTDIR` or `ENOENT`, but the caller ignores these error because it is expected. This happens for example in the masked_path (https://github.com/containers/youki/pull/1973/files#diff-9773882fbf80be1f21b71393425530b52a9fc276cd89e5f0910d2435ca45fbb5R217). In this case, `ENOENT` is considered successful because a non-existent path is considered to be masked already (not accessible).

I believe all the tracing add you see are at the caller of `syscall::mount`. In most cases, the mount call failure indicates an actual failure, so we log them, unless the error case is handled.

I will also add one more commit to trigger the `integration_validation` test, so we can be complete in this PR.

crates/libcontainer/src/rootfs/rootfs.rs
```diff
@@ -52,20 +52,34 @@ impl RootFS {
...
-        )?;
+        self.syscall
+            .mount(
+                Some(rootfs),
+                rootfs,
+                None,
+                MsFlags::MS_BIND | MsFlags::MS_REC,
+                None,
+            )
+            .map_err(|err| {
```
yihuaf, line 79: For example, this call to `syscall::mount` is inside the `rootfs` module. Here, mount call failure indicates an actual failure, so we log the error here.

YJDoc2:
Hey @yihuaf Thanks for the detailed explanation! I think I got the actual call site and implementation mixed up originally. This looks good then, and once the CI pass, I'll approve and merge. Thanks a lot!

PR #1975, opened 2023-05-27
Implemented sending logs to systemd-journald

yihuaf:
@utam0k mentioned a usecase where when test `youki` inside k8s, it is hard to track logs of each `youki` implementation. A workaround was using `bpf` tracing the syscalls. However, it would be much better to see the actual `youki` logs. This is where `journald` comes in. It is one of the common ways for linux services and application to collect logs into a central place. In this PR, we implemented a different tracing layer that sends `youki` logs to `journald`. All logs are tagged as `youki` and we can obtain the logs with `journalctl -t youki`.

Note. I am not sure how best to enable this feature. I decided to implement this as a flag `--systemd-log` on the command line. However, this is not part of the OCI spec, so I added as a youki extension in the struct. However, this may not be the best way.

Another way to implement this is through `features`.

Tag along in this PR is a small refactor of the existing logging code. This refactors the logging into using different layers where we can get ready for the opentelelmetry work.

### DIFF
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -15,6 +15,14 @@ use crate::commands::info;
 
 use liboci_cli::{CommonCmd, GlobalOpts, StandardCmd};
 
+// Additional options that are not defined in OCI runtime-spec, but are used by Youki.
+#[derive(Parser, Debug)]
+struct YoukiExtendOpts {
+    /// Enable logging to systemd-journald
+    #[clap(long)]
+    pub systemd_log: bool,
+}
+
 // High-level commandline option definition
 // This takes global options as well as individual commands as specified in [OCI runtime-spec](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md)
 // Also check [runc commandline documentation](https://github.com/opencontainers/runc/blob/master/man/runc.8.md) for more explanation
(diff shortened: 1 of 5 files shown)


utam0k (commented):
Could you write how to test?

yihuaf:
> Could you write how to test?

In one terminal, run `journalctl -f -t youki`. The `-t` means tag. We tag all the logs using `youki` tag, which we hardcoded in this PR. The `-f` means follow, so you wait on the logs as they arrive. On another terminal, run normal youki commands with `--systemd-log` flag, such as `./youki --debug --systemd-log create -b tutorial test`. You should see logs shows up in systemd log through `journalctl`.

Currently, we are exposing this feature using a flag `--systemd-log`. We may consider exposing this as a feature flag at compile time, so for k8s testing, no runtime flag needs to be added.

Give it a try first to see how you like how this prototype works currently. Note, `journalctl` itself has a lot of features, so we don't do much other than shipping the log/trace to it. For example, `journalctl` can present the log as json format, so our implementation don't need to do it.

crates/youki/src/main.rs
```diff
@@ -24,6 +32,9 @@ struct Opts {
     #[clap(flatten)]
     global: GlobalOpts,
 
+    #[clap(flatten)]
+    youki_extend: YoukiExtendOpts,
```
utam0k, line 36: I love this idea ❤️

utam0k (approved):

PR #1984, opened 2023-05-29
deprecate crossbeam since it is merged with std

yihuaf:
- crossbeam is merged into std since 1.67
- clean up a few unused dependencies


YJDoc2 (approved):
lgtm. All tests are passing, and integration tests check for hooks, so lgtm

PR #1996, opened 2023-06-01
Rust 1.70

yihuaf:
1.70 is released.

Note the fix to the new lint, `///!` is not a valid comment. `//!` is valid. Now clippy complains it. Reference: https://doc.rust-lang.org/reference/comments.html


yihuaf:
Pending #1973 to fix the integration test

YJDoc2:
bookkeeping : done in https://github.com/containers/youki/pull/1973

PR #2012, opened 2023-06-06
Refactor CI

### LINKED ISSUE #1974: Need to add integration test to the regular CI runs, not just integration_validation
@utam0k @YJDoc2 Would you please double check my thinking here.

This is triggered by #1967 failing the integration test because the extra error logging I added. Upon investigation, I think the integration test is either not triggered correctly, or we should add integration test into the main CI workflow.

Currently, the `integration_test_validation.yaml` is only triggered when the integration test code are changed. The workflow runs `runc` and `youki` to make sure that the integration tests runs for both. This make sense because we want to make sure we don't introduce breaking changes so the same validation tests can be used for `runc` as well.

However, other than this occasion, the integration tests are not triggered at all. This means code changes to `youki` and/or `libcontainer` alone do not trigger integration test. This is bad because we can easily break integration test without knowing.

To resolve this, we should either:

1. add integration test into the `main.yaml` and use the same trigger. Likely, we should trigger the integration as a different job so that we can parallelize the integration test and the existing tests (unit tests, clippy and etc.)
2. OR, add a new workflow that specifically triggers the integration.

I would prefer (1) so that we can share the same trigger.

yihuaf:
- attempt to standardized all CI to use Ubuntu 22 (the integration validation using `runc` is currently failing on Ubuntu 22, but others are successful)
- refactored the justfile to be more ergonomics
- moved the rust integration test to e2e
- moved the oci validation go version to e2e
- change the file filter action to `tj-actions/changed-files`
- refactored the dependencies installation to `just prepare` recipe
- update readme to replace `make` related calls.

Fix #1974 

CI should all pass. Note the `integration_test` is in a `Expected` state. This is due to branch protection rules and not directly related to this PR. In this PR, we renamed the integration_test to `oci-validation-go`. When moving to `e2e.yaml`, the name `integration_test` is way too ambiguous. We will fix the branch protection rule once the PR is merged into main or right after approve of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,13 @@ Here is why we are writing a new container runtime in Rust.
   <summary>Details about the benchmark</summary>
 
   - A command used for the benchmark
-    ```console
-    $ hyperfine --prepare 'sudo sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' --warmup 10 --min-runs 100 'sudo ./youki create -b tutorial a && sudo ./youki start a && sudo ./youki delete -f a'
+
+    ```bash
+    hyperfine --prepare 'sudo sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' --warmup 10 --min-runs 100 'sudo ./youki create -b tutorial a && sudo ./youki start a && sudo ./youki delete -f a'
     ```
+
   - Environment
+
     ```console
     $ ./youki info
     Version           0.0.1
@@ -91,6 +94,7 @@ Here is why we are writing a new container runtime in Rust.
     spec: 1.0.0
     +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
     ```
+
   </details>
 
 - The development of [railcar](https://github.com/oracle/railcar) has been suspended. This project was very nice but is no longer being developed. This project is inspired by it.
@@ -143,6 +147,8 @@ For other platforms, please use the [Vagrantfile](#setting-up-vagrant) that we h
 
 ## Dependencies
 
+To install `just`, follow the instruction [here](https://github.com/casey/just#installation).
+
 ### Debian, Ubuntu and related distributions
 
 ```console
@@ -172,11 +178,11 @@ $ sudo dnf install          \
 
 ## Build
 
-```console
-$ git clone git@github.com:containers/youki.git
-$ cd youki
-$ make youki-dev # or youki-release
-$ ./youki -h # you can get information about youki command
+```bash
+git clone git@github.com:containers/youki.git
+cd youki
+just youki-dev # or youki-release
+./youki -h # you can get information about youki command
 ```
 
 ## Tutorial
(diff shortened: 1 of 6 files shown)


yihuaf:
@YJDoc2 This is ready :)

YJDoc2 (changes requested):
Hey, one of the issue I am facing is with the prepare recipe. Here we have moved all dependency install in that, and run it to setup ci env. However, we only check for plain ubuntu and fedora ; which is too restrictive. For example, this fails on Pop OS, which  is ubuntu derivative. What I think would be better is to leave the general setup instructions as they are right now (so readme will still show apt/dnf install commands), and make a recipe tailored to ubuntu which will be use only to setup CI env.

Also, maybe rename it to 'setup-ci-env' instead of prepare...

README.md
```diff
@@ -143,40 +147,25 @@ For other platforms, please use the [Vagrantfile](#setting-up-vagrant) that we h
...
-      libdbus-glib-1-dev  \
-      build-essential     \
-      libelf-dev          \
-      libseccomp-dev      \
-      libclang-dev        \
-      libssl-dev
+Install `just`:
+
+```bash
+curl --proto '=https' --tlsv1.2 -sSf <https://just.systems/install.sh> | bash -s
```
YJDoc2, line 153: Link to just's github repo instead of giving specific instructions, so users can choose whatever option suites them best.

yihuaf:
@YJDoc2 Thanks for the quick review. I understand the concern you raised regarding to the setup. I will restore the instruction back for now and link to the `just` git repo. I will propose another PR so this point won't block the rest of the PR.

There are two approaches when it comes to setting up environment. One, as you suggested, is to link information in the readme and/or other source of readings, and let users decide how to proceed. The other approach is to make it as close to a single button click setup as possible using automation. In this approach, the goal is that users don't have to figure out what packages to install and where to download `just`. This is what a `prepare` target tries to achieve. This has the added benefit of single point of truth, where adding one more dependencies doesn't require changing a bunch of other places, like we do today.

Here is what I propose. We declare that Ubuntu and Fedora are the official dev environment we support. We will create automation for setting up the environment for these two OS. Eventually, we would even make CI a matrix strategy running both Ubuntu and Fedora.

For all other dev environment, we can point them to the Ubuntu/Fedora scripts and let people figure out what needs to be done for their specific environment. This probably provides a balance between automation in the fast path while do not neglect the slow path.

yihuaf:
@YJDoc2 PTAL

YJDoc2:
Hey @yihuaf , I think the idea of officially supported distributions and whether to have a one-step setup process or not needs more discussion, as they are pretty big changes for all users/developers who want to build/setup youki. I think what you have suggested definitely has a merit and if I remember correctly it was once discussed when issue numbers were in ~300 ...
So for now, I'll merge this PR in its current state as it is lgtm and does the work of refactoring the ci as intended. I'll open a new Issue later today to discuss about the official supported distros etc. Given that it can be a big impacting decision (to make certain distros "officially" supported) , I think we should definitely wait for utam0k to get better and hear his opinion on it as well.

YJDoc2 (approved):
lgtm :+1:

Thanks!

yihuaf:
> Hey @yihuaf , I think the idea of officially supported distributions and whether to have a one-step setup process or not needs more discussion, as they are pretty big changes for all users/developers who want to build/setup youki. I think what you have suggested definitely has a merit and if I remember correctly it was once discussed when issue numbers were in ~300 ... So for now, I'll merge this PR in its current state as it is lgtm and does the work of refactoring the ci as intended. I'll open a new Issue later today to discuss about the official supported distros etc. Given that it can be a big impacting decision (to make certain distros "officially" supported) , I think we should definitely wait for utam0k to get better and hear his opinion on it as well.

I agree. Btw, thank you for pointing this out. I do realize now that this has a bigger impact than I originally anticipated. I am used to making dev environment as automated and consistent as possible, because that is how it is done when you work for a company. The priorities are how to onboard someone quickly and keeping things as consistent as possible. Here, the priorities are different where we should accommodate to a wide number of audiences with many different dev setups. We will continue the discussion in the other issue as you suggest.

YJDoc2:
>I agree. Btw, thank you for pointing this out. I do realize now that this has a bigger impact than I originally anticipated. I am used to making dev environment as automated and consistent as possible, because that is how it is done when you work for a company. The priorities are how to onboard someone quickly and keeping things as consistent as possible. Here, the priorities are different where we should accommodate to a wide number of audiences with many different dev setups. We will continue the discussion in the other issue as you suggest.


No worries! Even I noticed it because it was discussed once before, and I had read discussions similar to this in some other open source s/w , where they wanted to decide what should be their official supported platforms :) Also, I can understand the ease of having an autometed setup and consistent base ; its just that as you said correctly, we also need to consider that our audience is wider, and hacks stuff with varied setup. (I think there was recently an issue opened for dind and capabilites)

fyi, I have updated the branch protection rules for the new ci name as you have suggested.

I'll open the separate issue for this discussion later today, will also link the old issue I mentioned and some other pov I have in mind.

PR #2018, opened 2023-06-07
Using `typos-cli` to catch typos + fixes for existing typos

### LINKED ISSUE #1929: Investigate the use of `typos` to automatically catch all the typos.
This is a tool I have used previously: https://github.com/crate-ci/typos

We can investigate setting this up so the CI can automatically catch all the typos.

yihuaf:
Fix #1929 

`typos-cli` is a tool I have used in the past to automatically catch typos. Feel free to checkout the details here https://github.com/crate-ci/typos

In addition to running this as part of the CI, I do want users the ability to check the typos as well on their local dev setup, so, I provided the `just spellcheck`. A small note is that we have to install `typos-cli`. Since this is installed using the `cargo`, I went ahead to add `just dev-prepare` where we install the tool using `cargo`. Since we assume the users should already setup cargo and rust for their dev environment, `cargo` should make this process OS agnostic. In the future, this recipe can be used to install tools such as `cargo-bloat` or `cargo-udep` and etc.

For CI, the github action takes care of setting the cli up and checking.


YJDoc2 (approved):
lgtm :+1:
Thanks a lot!

PR #2027, opened 2023-06-09
Add the test with kind to github action

utam0k:


justfile
```diff
@@ -1,29 +1,32 @@
...
-ROOT := `git rev-parse --show-toplevel`
+KIND_CLUSTER_NAME := 'youki'
 
 # build
 
 # build all binaries
 build-all: youki-release rust-oci-tests-bin runtimetest
 
 # build youki in dev mode
+[no-cd]
```
yihuaf, line 12: May I ask what are we trying to achieve with the `[no-cd]` attribute? I don't think we need this attribute here because this is a justfile at the project root.

justfile
```diff
@@ -1,29 +1,32 @@
...
 youki-dev:
-    ./scripts/build.sh -o {{ ROOT }} -c youki
+    ./scripts/build.sh -o . -c youki
```
yihuaf, line 14: I think using relative path like `.` here can be a footgun down the road. We should use absolute path here, so the intent is explicit. We can either use the old `ROOT` or use `{{ justfile_directory() }}`?

justfile
```diff
@@ -43,21 +46,47 @@ featuretest:
...
 # run containerd integration tests
 containerd-test: youki-dev
-	VAGRANT_VAGRANTFILE=Vagrantfile.containerd2youki vagrant up
-	VAGRANT_VAGRANTFILE=Vagrantfile.containerd2youki vagrant provision --provision-with test
+    VAGRANT_VAGRANTFILE=Vagrantfile.containerd2youki vagrant up
+    VAGRANT_VAGRANTFILE=Vagrantfile.containerd2youki vagrant provision --provision-with test
+
+[private]
+kind-cluster: bin-kind
+    if ! {{ path_exists('tests/k8s/_out') }}; then mkdir -p tests/k8s/_out/; fi
```
yihuaf, line 69: I would suggest turning this into a bash script by adding:

```
    #!/usr/bin/env bash
    set -euo pipefail
```

yihuaf (commented):
With regarding the path issue in justfile, we usually keep track of two path. One is the project root through `git rev-parse` and another is the justfile is located, through `justfile_directory`.

```
TOP := `git rev-parse --show-toplevel`
CWD := {{ justfile_directory() }}
```

Usually these two location is all we need to run all relative path into an absolute path. Usually, we want the justfile executed at the location of the justfile. Where we invoke the `just` command should not matter. In this way, all command has a fixed and predictable path.

justfile
utam0k: I forgot to delete them.

utam0k:
> With regarding the path issue in justfile, we usually keep track of two path. One is the project root through `git rev-parse` and another is the justfile is located, through `justfile_directory`.
> 
> ```
> TOP := `git rev-parse --show-toplevel`
> CWD := {{ justfile_directory() }}
> ```
> 
> Usually these two location is all we need to run all relative path into an absolute path. Usually, we want the justfile executed at the location of the justfile. Where we invoke the `just` command should not matter. In this way, all command has a fixed and predictable path.
> 
> Otherwise LGTM.

I do not want to leave the case for using git. This is because of the dependence on git. We want to reduce the dependency. Also, I feel it is difficult to judge how to use with `justfile_directory`. I honestly don't know much of the difference. can you tell me when ROOT is needed?
```console
$ just -f ../justfile youki-dev
/home/utam0k/ghq/src/github.com/utam0k/youki/scripts/build.sh -o /home/utam0k/ghq/src/github.com/utam0k/youki -c youki
* FEATURES:
* features:
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
```

yihuaf:
> I do not want to leave the case for using git. This is because of the dependence on git. We want to reduce the dependency.

I see. Is it because you are running the `justfile` inside the container so `git` is required?

> Also, I feel it is difficult to judge how to use with justfile_directory. I honestly don't know much of the difference. can you tell me when ROOT is needed?

`ROOT` is needed when you want to reference a specific directory relative to the `youki` project root, especially if inside subdirectory justfiles. `git` is the only way to deterministically obtain the project root from anywhere inside. We can certainly workaround not using `git`, but it will probably requires us to keep track of relative path of where justfiles are inside the subdir and using `../` to get to project root.

I think if the current PR works as it stands, we should merge this and refactor justfile in follow ups. No need to block this PR further on jsutfile refactors since we will do those anyway.

yihuaf (approved):
For the nits with regarding to `justfile`, we will deal with those in future refactoring. Let's get the test into CI first.

utam0k:
> I see. Is it because you are running the justfile inside the container so git is required?

I had imaged the case in GH actions.

PR #2029, opened 2023-06-09
Bump the oci-spec-rs to 0.6.1 to resolve seccomp rule issue

### LINKED ISSUE #2022: youki need seccomp unconfined when runc/crun don't
As the title said, for some reason youki need `seccomp unconfined`, two containers that I tested was mariadb and jellyfin.

Jellyfin log
```
jellyfin_server  | Failed to create CoreCLR, HRESULT: 0x80070008
jellyfin_server exited with code 137
```

Mariadb log
```
documize-db-1  | 2023-06-08 13:56:14+07:00 [Note] [Entrypoint]: Entrypoint script for MariaDB Server 1:10.11.2+maria~debsid started.
documize-db-1  | 2023-06-08 13:56:14+07:00 [ERROR] [Entrypoint]: mysqld failed while attempting to check config
documize-db-1  | 	command was: mariadbd --character-set-server=utf8mb4 --collation-server=utf8mb4_bin --verbose --help --log-bin-index=/tmp/tmp.axLXmY9ftc
documize-db-1  | 	Can't initialize timers
documize-db-1 exited with code 0

```

Runtime:
Kernel 6.3
Ubuntu Jammy
Docker 24.0.2

Compiler
Rust 1.70

yihuaf:
Fix #2022 

The detail of the cause of the issue and why 0.6.1 fixes the issue are explained in #2022. 

However, the 0.6.1 introduced the `time` namespace into the runtime spec. For this PR, we set the `time` namespace to be unsupported. Both `libc` and `nix` crate have not provided the support. We should ideally wait for the support to land in either of this crate first before we support it. As a result, there are a number of functions needs to propagate errors now. This is a good add because in the future I imagine we will run into similar situations when new namespaces is introduced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2335,9 +2335,9 @@ dependencies = [
 
 [[package]]
 name = "oci-spec"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "214b837f7dde5026f2028ead5ae720073277c19f82ff85623b142c39d4b843e7"
+checksum = "cf77d2eace1f4909b081d231d1ad3570ba3448ae95290ab701314faaee1b611a"
 dependencies = [
  "derive_builder",
  "getset",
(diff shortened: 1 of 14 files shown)


utam0k:
> However, the 0.6.1 introduced the time namespace into the runtime spec. For this PR, we set the time namespace to be unsupported.

👍

utam0k (approved):

PR #2031, opened 2023-06-10
Don't create a file when it already exists when mounting with bind

utam0k:
For example, it is more likely that the /etc/hosts file already exists. In that case, it fails because it tries to open the file with write permission while it is RO.

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -497,7 +497,7 @@ impl Mount {
                 err
             })?;
 
-            if src.is_file() {
+            if src.is_file() && !dest.exists() {
                 OpenOptions::new()
                     .create(true)
                     .write(true)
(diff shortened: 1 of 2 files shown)


crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -497,7 +497,7 @@ impl Mount {
                 err
             })?;
 
-            if src.is_file() {
+            if src.is_file() && !src.exists() {
```
yihuaf, line 500: I don't think this is correct. This is the source of the bind mount, not the destination. The source will have to exist for bind mount to happen. The destination on the other hand can exist or not exist.

utam0k: 🤦You are right. Thanks for your great reviewing. I have updated it.

yihuaf (approved):

PR #2036, opened 2023-06-12
Introduce a `log-level` flag.

### LINKED ISSUE #2028: [RFC] Introducing a new `--log-level` flag
Currently, there are a few ways we can manage the log level in `youki`.

- Through the `YOUKI_LOG_LEVEL` env variable.
I believe this copying what traditionally rust logging library does. Both the `env-logger` we used previously and `tracing` has native interface to parse log level from the env variable.

- Through the `debug` build.
Currently, if `youki` is build with debug build, the default log level is automatically to `debug`

-Through the `--debug` flag.
Currently, both `runc` and us have a `debug` flag where we turn the log level to `debug`.

Up until now, these 3 options have served us well. However, I would like to introducing a new level `trace` that is more verbose than `debug`. For example, currently I am trying to debug `seccomp` and I need to log a lot of information on exactly what is going on with each seccomp rules addes and translated. Docker by default passes in 200+ seccomp rules in the profiles, so the logs can become very spammy, even by `debug` level standards. I don't think it is good devx to turn these spammy logs on by default. So I believe putting these logs behind `trace` level makes sense.

With this new level, the only way to switch the level during runtime is through the `YOUKI_LOG_LEVEL`. Both `--debug` flag and debug build will not help in this case. However, managing env variable is a pain when using `youki` under docker or containerd, because the config files for those only specify the binary path and args, not env. We can certainly write our own wrapper to add the env variable, but that is a pain and the original inspiration for us to add the `--debug` flag.

One potential solution:

We introduce a runtime `--log-level` flag. This should be easier to use compared to all existing options and more flexible. Also, a `log-level` flag is common is all kinds of CLI.

The potential question is what should we do with the existing options. Should we create a hierarchy where we clearly states one option will override the other option. Or we can reject setting `--debug` and `--log-level` together. Or we can remove the `--debug` flag and in favor of the new `--log-level debug`.

@containers/youki-maintainers Any opinions/ideas/comments?

Another potentially viable option is to turn the default level to `trace`, but I am not too happy with this because potentially as we add more `trace` level logging, it will become way too spammy.

yihuaf:
Fix #2028 

As proposed in #2028, we introduce a `--log-level` flag to allow us to easily set log level through runtimes. To avoid confusion, we also deprecated setting log level through `env` variable. The `--debug` flag is kept for compatibility reasons.

Note, the integration test was not changed in this PR. If this PR is accepted, we will make the changes to the rust integration test to keep things consistent.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -7,7 +7,6 @@ on:
   pull_request:
     branches:
       - main
-  workflow_dispatch:
 
 jobs:
   youki-build:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   youki-build:
--- a/docs/src/user/basic_usage.md
+++ b/docs/src/user/basic_usage.md
@@ -171,3 +171,13 @@ cd ..
 ./youki list
 ./youki delete rootless_container
 ```
+
+#### Log level
+
+`youki` defaults the log level to `error` in the release build. In the debug
+build, the log level defaults to `debug`. The `--log-level` flag can be used to
+set the log-level. For least amount of log, we recommend using the `error` log
+level. For the most spammy logging, we have a `trace` level.
+
+For compatibility with `runc` and `crun`, we have a `--debug` flag to set the
+log level to `debug`. This flag is ignored if `--log-level` is also set.
--- a/scripts/rust_integration_tests.sh
+++ b/scripts/rust_integration_tests.sh
@@ -24,7 +24,7 @@ if [ ! -f ${ROOT}/bundle.tar.gz ]; then
 fi
 touch ${LOGFILE}
 
-sudo YOUKI_LOG_LEVEL="error" ${ROOT}/integration_test run --runtime "$RUNTIME" --runtimetest ${ROOT}/runtimetest > $LOGFILE
+sudo ${ROOT}/integration_test run --runtime "$RUNTIME" --runtimetest ${ROOT}/runtimetest > $LOGFILE
 
 if [ 0 -ne $(grep "not ok" $LOGFILE | wc -l ) ]; then
     cat $LOGFILE
--- a/.github/workflows/integration_tests_validation.yaml
+++ b/.github/workflows/integration_tests_validation.yaml
@@ -20,7 +20,7 @@ jobs:
         with:
           files: |
             .github/workflows/integration_tests_validation.yaml
-            integration_test/tests/rust-integration-tests/**
+            tests/rust-integration-tests/**
           files_ignore: |
             **.md
       - name: List all changed files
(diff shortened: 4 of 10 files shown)


scripts/rust_integration_tests.sh
```diff
@@ -24,7 +24,7 @@ if [ ! -f ${ROOT}/bundle.tar.gz ]; then
 fi
 touch ${LOGFILE}
 
-sudo YOUKI_LOG_LEVEL="error" ${ROOT}/integration_test run --runtime "$RUNTIME" --runtimetest ${ROOT}/runtimetest > $LOGFILE
+sudo ${ROOT}/integration_test run --runtime "$RUNTIME" --runtimetest ${ROOT}/runtimetest > $LOGFILE
```
YJDoc2, line 27: We should be also passing the --log-level flag here, set to error, right?

yihuaf: No, this script is calling the integration_test binary, which don't take the `log-level` flag. At least I did not make the change in this PR. The actual code that calls the `runtime` is changed to add the `log-level` flag. The integration_test binary uses `YOUKI_INTEGRATION_LOG_LEVEL` env variable. https://github.com/containers/youki/blob/1a6d1f4bd7553e971d6d787698a9732836188444/tests/rust-integration-tests/integration_test/src/logger.rs#L6

It is possible that the original intent is `YOUKI_INTEGRATION_LOG_LEVEL`?

.github/workflows/e2e.yaml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
```
utam0k, line 10: Why do we need `workflow_dispatch`?

docs/src/user/basic_usage.md
```diff
@@ -171,3 +171,13 @@ cd ..
 ./youki list
 ./youki delete rootless_container
 ```
+
+#### Log level
```
utam0k, line 175: 👍

.github/workflows/e2e.yaml
utam0k: Can we use this button?
image

yihuaf: You can, but you have to create a PR upfront with all CI triggered. And this only do so with already ran CI. I can't just push to a branch (without creating a PR) and trigger 1 specific CI run to test things quickly. What is your main concern for adding this? I can remove it from this PR if you have no issue with the rest, so this one is not blocked.

utam0k: > What is your main concern for adding this?

The debugging of CI in the upstream repositories is to occur. It should do in each fork branch.

yihuaf: I agree that the debugging should be done in the forked branch. However, I was not able to use the CI in the forked repo unless the workflow yaml specify that the the `workflow_dispatch` trigger. I can't trigger the CI in the forked repo unless the workflow yaml file in the main branch contains that trigger. I can change the branch matching rule to trigger on every push in the forked repo on my own branch, but I can't add a new trigger without merging this into main.

yihuaf:
@utam0k removed the `workflow_dispatch`. PTAL.

utam0k (approved):
💯 Thanks a lot 🙏

PR #2060, opened 2023-06-18
Fix the feature test and turn on in CI

### LINKED ISSUE #2046: libcgroup build is broken for individual feature v1, v2 or systemd
Currently, we default the libcgroup feature to [v1, v2, systemd], so this went under our radar. However, if we compile libcontainer with individual features, the `libcgroups` fails to compile.

For example, this will fail.
```
cargo build --no-default-features --package libcontainer --features v2
```

yihuaf:
The feature test has been broken for unknown amount of time. The libcgroups and libcontainer has been tested with all v1, v2, and systemd feature turn on. However, individual features are not working. This PR fix this by introduce a stub implementation of libcgroup implementation if the feature is not turned on. Due to how the code is structured, we can't completely remove the reference when a feature is turned off. So a stub is required.

Fix #2046 

Other fixes:

- fix justfile path to use absolute path
- turn the feature test in CI
- the `musl` test are separated out to a different test script outside of the feature test, since the goal for the test less about features, but more about targeting `musl`. The two test cases are passing, but I am following up with another PR to put these tests into CI.
- removed two justfile in the subdir, which are not currently used.


Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
utam0k, line 2: I think it doesn't need because this has been the default resolver since Rust 2021 edition.

crates/libcgroups/src/v1_stub/manager.rs
```diff
@@ -0,0 +1,43 @@
+use crate::common::{AnyCgroupManager, CgroupManager};
```
utam0k, line 1: Is there any reason to separate the `v1` and `v1_stub` dir? I'd like to put them together into the `v1` directory like that:

```
v1
|-- manager.rs
L-- stub_manger.rs(or stub.rs)
```

scripts/build.sh
```diff
@@ -13,6 +12,7 @@ VERSION=debug
 TARGET="$(uname -m)-unknown-linux-gnu"
 CRATE="youki"
 RUNTIMETEST_TARGET="$ROOT/runtimetest-target"
+features=""
```
utam0k, line 15: Why do we need this change?

crates/libcgroups/src/v1_stub/manager.rs
utam0k: Or are all stub files put together into one dir?

Cargo.toml
yihuaf: It produces warning. This is the workspace `cargo.toml`, so it doesn't have the `Rust 2021 edition` definition. Therefore the `resolver = "2"2` is required here to be explicit. I don't think this changes anything for us other than making the warning goes away, because all of our crates are 2021 edition. However this warning is geared to people having a mixed editions for different crates in the workspace.

Specifically here are the a few reference for this issue:

This reference points to we need to explicitly set resolver in the workspace cargo.toml: https://github.com/rust-lang/edition-guide/pull/267/files#diff-e24ffee7de69b814b2c59439c6cfc9b8f9978b864175fd4425911b897391d845R17

One issue I found for the specific problem: https://github.com/rust-lang/cargo/issues/9956

The specific lint rule is added with this issue: https://github.com/rust-lang/cargo/issues/10112

crates/libcgroups/src/v1_stub/manager.rs
yihuaf: > Is there any reason to separate the v1 and v1_stub dir?

Because I need these to be different module, so I can use feature switch to control which module is listed. This minimizes the amount of `#[cfg(not(feature = "..."))]` we do. The attribute only applies to a single item in the next line, or I will have to group them into a new mod.

> I'd like to put them together into the v1 directory like that

As mentioned above, we want to put stub vs. not stub into different modules. Likely we will want to maintain the same `pub mod` across the board to keep things easier to find. For example, the `manager.rs` is exposed as `pub mod manager` in both stubs, so it is easy to locate where the files are.

> Or are all stub files put together into one dir?

This can work. I will make the changes.

scripts/build.sh
yihuaf: Because `features` as a variable is not set nor has a default value. Bash complaints when I turned on the `-u`, because `features` is unbounded if no opts is specified. Later in the script, we test for features as a unbounded variable with if [ -n ], but we should be explicit with `features = ""` so that it is bounded. There is a more strict set of bash options I've recently started to put in all my bash scripts:

```bash
    #!/usr/bin/env bash
    set -euo pipefail
```

crates/libcgroups/src/v1_stub/manager.rs
yihuaf: Implemented in 894d424ac43cbf0c860161a83114577197b0b30d

utam0k: 👍

utam0k (approved):

PR #2068, opened 2023-06-20
re-export oci-spec in libcontainer

### LINKED ISSUE #2066: [RFC] re-export the oci-spec from `libcontainer`
As a user of `libcontainer`, having to match `libcontainer` version with the `oci-spec-rs` version is a pain. The `oci-spec-rs` can introduce breaking changes. For example, with the introduction of time namespace, the libcontainer is broken and require code change. In this case, the users need to spend efforts trying to understand what happens.

Making oci-spec-rs backward compatible does not make much sense. Therefore I propose we re-export the oci-spec-rs that `libcontainer` uses. In this way, to use `libcontainer`, users don't have to import `oci-spec-rs` into their own dependencies.

Ref: https://www.lurklurk.org/effective-rust/re-export.html

yihuaf:
Fix #2066 

- re-export the oci-spec crate used by libcontainer as part of the libcontainer api
- fix the oci-spec version, because oci-spec can have breaking changes with version bump.
- remove oci-spec from `youki` crate dependency (this is the key here)

### DIFF
--- a/crates/libcgroups/Cargo.toml
+++ b/crates/libcgroups/Cargo.toml
@@ -22,7 +22,7 @@ cgroupsv2_devices = ["rbpf", "libbpf-sys", "errno", "libc"]
 [dependencies]
 nix = "0.26.2"
 procfs = "0.15.1"
-oci-spec = { version = "^0.6.1", features = ["runtime"] }
+oci-spec = { version = "0.6.1", features = ["runtime"] }
 dbus = { version = "0.9.7", optional = true }
 fixedbitset = "0.4.2"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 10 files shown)


crates/libcgroups/Cargo.toml
```diff
@@ -22,7 +22,7 @@ cgroupsv2_devices = ["rbpf", "libbpf-sys", "errno", "libc"]
 [dependencies]
 nix = "0.26.2"
 procfs = "0.15.1"
-oci-spec = { version = "^0.6.1", features = ["runtime"] }
+oci-spec = { version = "0.6.1", features = ["runtime"] }
```
utam0k, line 25:
```suggestion
oci-spec = { version = "~0.6.1", features = ["runtime"] }
```

crates/libcontainer/src/lib.rs
```diff
@@ -18,3 +18,7 @@ pub mod test_utils;
 pub mod tty;
 pub mod utils;
 pub mod workload;
+
+// Because the `libcontainer` api uses the oci_spec who resides in a different
+// crate, we re-export the version of oci_spec this crate uses.
```
utam0k, line 23: Please add the issue link.

utam0k (approved):
Thanks for the great suggestion.

PR #2069, opened 2023-06-20
turn on musl test in CI

yihuaf:
Turn the musl target testing in CI. At the moment, this is only for `libcontainer` since there is an actual usecase from `aurae` project to compile libcontainer statically with `musl`. I choose to put the musl in a separate job because of the unique requirement for the nightly toolchain. While it is easy for us to switch toolchains in the dev box, it proves to be not possible to mix toolchains with the existing rust github action setup. It is easier to create a new job with the specific toolchains required.


justfile
```diff
@@ -32,17 +32,21 @@ rust-oci-tests-bin:
...
+test-all: unittest test-features oci-tests containerd-test # currently not doing rust-oci here
 
 # run cargo unittests
 unittest:
     cd ./crates
     LD_LIBRARY_PATH=${HOME}/.wasmedge/lib cargo test --all --all-targets --all-features
 
 # run permutated feature compilation tests
-featuretest:
+test-features:
```
utam0k, line 43: 👍

.github/workflows/main.yml
```diff
@@ -68,7 +68,34 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v3
+      # We do not use `actions-rust-lang/setup-rust-toolchain` here because we
+      # want to override the default toolchain file to use nightly toolchains.
+      # The `setup-rust-toolchain` action will always choose toolchain file with
+      # no way to override. 
+      - name: Setup Rust toolchain and cache
+        uses: dtolnay/rust-toolchain@stable
```
utam0k, line 88: Do we need to specify `stable`? I wonder because we have had `rust-toolchain.toml`.

yihuaf: Good catch! I mis-read the documentation and thought the `stable` referred to the github action version... I will make the change.

utam0k (approved):
🚀

PR #2081, opened 2023-06-23
Implement `clone3` fallback to `clone` when ENOSYS returns

### LINKED ISSUE #2030: [RFC] Implement a fallback when `clone3` returns ENOSYS
During the recent research on seccomp while trying to resolve #2022, I learned that `clone3` returning ENOSYS or EPERM is actually an issue in the security context, not as simple as whether the kernel has a minimal version. Specifically, this reminded me of #1861. In this issue, the `clone3` returned ENOSYS, but the kernel version is 5.15, which should definitely support `clone3`. I suspect that it is because there are rules on the host system that blocks `clone3` but returns ENOSYS instead. Normally, these blocked calls would return EPERM, but returning EPERM for `clone3` would break glibc and therefore almost all applications. Therefore, it is very likely that `clone3` is blocked as `ENOSYS` and the application should just fallback to fork or clone.

However, we choose to not implement a fallback. I believe we were naive at the time thinking that as long as we mandate a minimum kernel version, this should not be an issue. However, since now there are real evidence that real system has these behaviors,  we may want to reconsider the assumptions we had at the time. Falling back to `clone` from a implementation perspective is not hard. We already have the change, twice, LOL.

On the other hand, `clone3` is the only way to use new features such as the new time namespace. I am not sure about the implication on this just yet.

@containers/youki-maintainers comments?

A few note for references. This is PR #1610 where we decided to use `clone3`. We also have to use `clone` because fork doesn't let us control create a sibling process.

yihuaf:
Fix #2030 

Potentially controversial:

- Removed the use of `clone3` crate. Our usecase for the clone3 syscall is very specific and do not need to support everything. I can easily replace the clone3 crate with a few lines of code, so I did it in this PR. We usually favors safe syscall wrappers like `nix` over `libc`, but I believe the save is worth it here.
- We use the raw `clone` syscall instead of `glibc` wrapper. Details are in the comments.
- Using raw syscalls for both clone and clone3 actually makes the code well organized. For our usecase, they share the input types.

Otherwise, this PR does what the RFC outline. Unit tests using seccomp profile to simulate clone3 being blocked are added.

Also, I implemented fallback for `ENOSYS` only. Technically, we can add `EPERM` to the list as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,7 +34,7 @@ version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "once_cell",
  "version_check",
 ]
@@ -182,7 +182,7 @@ checksum = "233d376d6d185f2a3093e58f283f60f880315b6c60075b01f36b3b85154564ca"
 dependencies = [
  "addr2line",
  "cc",
- "cfg-if 1.0.0",
+ "cfg-if",
  "libc",
  "miniz_oxide 0.6.2",
  "object",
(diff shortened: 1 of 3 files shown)


yihuaf:
Holding off on this one. It seems we should not be calling raw syscall directly without glibc for clone or clone3...

PR #2096, opened 2023-06-29
Quick install guide

utam0k:


utam0k:
This docs page is lined from https://docs.docker.com/engine/alternative-runtimes/#examples. We need to maintain it more.

YJDoc2:
@utam0k , overall this looks good, but in my opinion the `$` signs in console code  are redundant, as most of the times it is understood that the install instructions are to be run in a shell. Also it is problematic when copying the lines, as many times when selecting the command to be run to copy, the `$` sign gets copied as well, and it gives error in shell.

I'll approve this for now if you want to get it updated quick, and we can discuss the `$` sign later if you want.

YJDoc2 (approved):

utam0k:
@YJDoc2 Thanks for your review.
Without a signature at the beginning, it will be difficult to tell when there is a set of commands and outputs.
Also, for the sake of uniformity, if you want to delete them, delete them from this repository all together.

utam0k:
However, I understand your feelings. But using $ is common. I think the only solution would be to implement a copy button.
https://go.dev/doc/install

YJDoc2:
Hey, I agree that we should make sure of uniformity, and whatever we choose, to keep or not to keep `$` , we should do it in all the places. I'll think on this more, as well as check out the link you have added.

For now, this is good to merge :+1: , I'll go ahead and merge it.

PR #2121, opened 2023-07-04
Implemented clone fallback when clone3 returns ENOSYS

### LINKED ISSUE #2030: [RFC] Implement a fallback when `clone3` returns ENOSYS

yihuaf:
Fix #2030 

Apologize for the large change. This PR implements the clone fallback when clone3 returns ENOSYS. `clone` interface is messier to use compared to the clone3 kernel syscall, so a large chunk of the changes is to deal with that. For example, the `container_args` now implements `Clone` so we can clearly pass the ownership of the memory to the new process.

A few other changes to make this happen:

- Delay the creation of the syscall struct and only pass around a enum signaling the type of syscall.
- Remove the use the executor and use a function pointer instead. Now the executor is much more explicit about the order of how each executor is tried. Instead of a list of Executor, now executors are composable to create new executor.
- Delay the creation of cgroup manager until the intermediate process.
- Changed some of the cgroup functions to take &Path instead of PathBuf.

### DIFF
--- a/crates/libcontainer/src/channel.rs
+++ b/crates/libcontainer/src/channel.rs
@@ -18,12 +18,13 @@ pub enum ChannelError {
     #[error("channel connection broken")]
     BrokenChannel,
 }
-
+#[derive(Clone)]
 pub struct Receiver<T> {
     receiver: RawFd,
     phantom: PhantomData<T>,
 }
 
+#[derive(Clone)]
 pub struct Sender<T> {
     sender: RawFd,
     phantom: PhantomData<T>,
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -32,12 +32,11 @@ pub struct ContainerBuilder {
 ///
 /// ```no_run
 /// use libcontainer::container::builder::ContainerBuilder;
-/// use libcontainer::syscall::syscall::create_syscall;
-/// use libcontainer::workload::default::DefaultExecutor;
+/// use libcontainer::syscall::syscall::SyscallType;
 ///
 /// ContainerBuilder::new(
 ///     "74f1a4cb3801".to_owned(),
-///     create_syscall().as_ref(),
+///     SyscallType::default(),
 /// )
 /// .with_root_path("/run/containers/youki").expect("invalid root path")
 /// .with_pid_file(Some("/var/run/docker.pid")).expect("invalid pid file")
@@ -53,12 +52,11 @@ impl ContainerBuilder {
     ///
     /// ```no_run
     /// use libcontainer::container::builder::ContainerBuilder;
-    /// use libcontainer::syscall::syscall::create_syscall;
-    /// use libcontainer::workload::default::DefaultExecutor;
+    /// use libcontainer::syscall::syscall::SyscallType;
     ///
     /// let builder = ContainerBuilder::new(
     ///     "74f1a4cb3801".to_owned(),
-    ///     create_syscall().as_ref(),
+    ///     SyscallType::default(),
     /// );
     /// ```
     pub fn new(container_id: String, syscall: SyscallType) -> Self {
@@ -115,12 +113,11 @@ impl ContainerBuilder {
     ///
     /// ```no_run
     /// # use libcontainer::container::builder::ContainerBuilder;
-    /// # use libcontainer::syscall::syscall::create_syscall;
-    /// # use libcontainer::workload::default::DefaultExecutor;
+    /// # use libcontainer::syscall::syscall::SyscallType;
     ///
     /// ContainerBuilder::new(
     ///     "74f1a4cb3801".to_owned(),
-    ///     create_syscall().as_ref(),
+    ///     SyscallType::default(),
     /// )
     /// .as_tenant()
     /// .with_container_args(vec!["sleep".to_owned(), "9001".to_owned()])
@@ -136,12 +133,11 @@ impl ContainerBuilder {
     ///
     /// ```no_run
     /// # use libcontainer::container::builder::ContainerBuilder;
-    /// # use libcontainer::syscall::syscall::create_syscall;
-    /// # use libcontainer::workload::default::DefaultExecutor;
+    /// # use libcontainer::syscall::syscall::SyscallType;
     ///
     /// ContainerBuilder::new(
     ///     "74f1a4cb3801".to_owned(),
-    ///     create_syscall().as_ref(),
+    ///     SyscallType::default(),
     /// )
     /// .as_init("/var/run/docker/bundle")
     /// .with_systemd(false)
@@ -157,12 +153,11 @@ impl ContainerBuilder {
     ///
     /// ```no_run
     /// # use libcontainer::container::builder::ContainerBuilder;
-    /// # use libcontainer::syscall::syscall::create_syscall;
-    /// # use libcontainer::workload::default::DefaultExecutor;
+    /// # use libcontainer::syscall::syscall::SyscallType;
     ///
     /// ContainerBuilder::new(
     ///     "74f1a4cb3801".to_owned(),
-    ///     create_syscall().as_ref(),
+    ///     SyscallType::default(),
     /// )
     /// .with_root_path("/run/containers/youki").expect("invalid root path");
     /// ```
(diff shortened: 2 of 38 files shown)


crates/libcontainer/src/container/builder.rs
```diff
@@ -253,74 +243,66 @@ impl<'a> ContainerBuilder<'a> {
...
-    pub fn with_executor(
-        mut self,
-        executors: Vec<Box<dyn Executor>>,
-    ) -> Result<Self, LibcontainerError> {
-        if executors.is_empty() {
-            return Err(LibcontainerError::NoExecutors);
-        };
-        self.executor_manager = ExecutorManager { executors };
+    pub fn with_executor(mut self, executor: Executor) -> Result<Self, LibcontainerError> {
+        self.executor = executor;
```
YJDoc2, line 256: Does this still needs to return result? we are simply setting the executor and returning, so can simply return self instead of a result

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -130,22 +123,31 @@ impl<'a> ContainerBuilderImpl<'a> {
             prctl::set_dumpable(false).unwrap();
         }
 
+        // Need to create the notify socket before we pivot root, since the unix
+        // domain socket used here is outside of the rootfs of container. During
+        // exec, need to create the socket before we enter into existing mount
+        // namespace. We also need to create to socket before entering into the
+        // user namespace in the case that the path is located in paths only
+        // root can access.
+        let notify_listener =
```
YJDoc2, line 132: very minor nitpick, but we can directly name this `notify_socket` , so we don't have to specify when building ContainerArgs below

crates/libcontainer/src/channel.rs
```diff
@@ -18,12 +18,13 @@ pub enum ChannelError {
...
+#[derive(Clone)]
 pub struct Receiver<T> {
     receiver: RawFd,
     phantom: PhantomData<T>,
 }
 
+#[derive(Clone)]
 pub struct Sender<T> {
     sender: RawFd,
     phantom: PhantomData<T>,
```
YJDoc2, line 30: Given that this contains RawFd, is it fine to allow clone on this? Also, should we impose Clone on T as well (even if we are not storing that) ?

YJDoc2 (commented):
Hey, haven't gone through the all changes or conceptual changes, just had a couple of specific questions. If any changes are required, I'd suggest to keep them and bundle with any other changes from other review.

crates/libcontainer/src/container/builder.rs
yihuaf: No. Good catch.

crates/libcontainer/src/container/builder_impl.rs
yihuaf: Yes

crates/libcontainer/src/channel.rs
yihuaf: > Clone RawFd

This is just a c_int, so this is safe to do. It is like pass the RawFd by value with copy. When across the fork/clone boundary, the file descriptor table is duplicated so existing fds works exactly the same.

> T for Clone

I don't think we have to impose `Clone` to T. In fact, it is possible that T does not have Clone trait. As long as T is serializable, we can use the channel to send and receive. So T can be serializable but not clone-able. The `PhantomData` does not store T as you pointed out and it is simply there to signal ownership of T.

crates/libcontainer/src/container/builder_impl.rs
yihuaf: b220267

crates/libcontainer/src/container/builder.rs
yihuaf: b220267

yihuaf:
I am leaving the commits for easier time to review, but I should probably squash before we merge.

utam0k:
I can't find time for this PR unless it's on a Saturday or Sunday, so wait a minute 🙏

yihuaf:
Also, some of the rust clone can be mitigated using `Rc`, especially to manage the cost of duplicating structures. I am working on a follow up PR next to address the issue in case there is a concern on this.

utam0k (changes requested):
This is too big a change to mix in with this change. libcontainer has people using it. We need to contact them. Please leave this out of this PR 🙏 

> Remove the use the executor and use a function pointer instead. Now the executor is much more explicit about the order of how each executor is tried. Instead of a list of Executor, now executors are composable to create new executor.

crates/libcontainer/src/container/builder.rs
```diff
@@ -1,18 +1,18 @@
...
 
-pub struct ContainerBuilder<'a> {
+pub struct ContainerBuilder {
     /// Id of the container
     pub(super) container_id: String,
     /// Root directory for container state
     pub(super) root_path: PathBuf,
     /// Interface to operating system primitives
-    pub(super) syscall: &'a dyn Syscall,
+    pub(super) syscall: SyscallType,
```
utam0k, line 15: Since this is a breaking change for the users of libcontainer, please create another PR.

yihuaf:
@utam0k maintaining backward compatibility is going to be hard and I don't think we have thought enough about the libcontainer api to support it. I hope there is a way to can make breaking changes to the interface. With that being said, I will break the change into smaller pieces and let see how to work through this. I do believe the importance of making this PR to go in as I have great concerns with us using clone3 syscall without libc directly as the only path.

crates/libcontainer/src/container/builder.rs
yihuaf: I will make these changes into a series of PR so we can these changes out.

yihuaf:
> This is too big a change to mix in with this change. libcontainer has people using it. We need to contact them. Please leave this out of this PR 🙏
> 
> > Remove the use the executor and use a function pointer instead. Now the executor is much more explicit about the order of how each executor is tried. Instead of a list of Executor, now executors are composable to create new executor.

Can you give me a sense of who are the big consumers? I can help take a look and gauge the amount of effort to fix.

yihuaf:
Converting this PR to draft. I will break this PR into smaller piece and use this as a reference.

utam0k:
@yihuaf I don't mean to deny destructive change. If they are necessary, let's include them. However, I do think that disruptive changes that are not related to 1PR will cause confusion among users. Runwasi is probably the biggest user. Thanks for understanding🙇🙇

yihuaf:
> @yihuaf I don't mean to deny breaking change. If they are necessary, let's include them. However, I do think that breakrng changes that are not related to 1PR will cause confusion among users. Runwasi is probably the biggest user. Thanks for understanding🙇🙇

Understand. I will parse into multiple PRs and the destructive changes should be in their own PR. I will also fix `runwasi` when I introduce the executor change.

yihuaf:
Actually, since I will break this PR into multiple PRs, we can close this to avoid confusion.

utam0k:
@yihuaf Thanks a lot🙏

PR #2122, opened 2023-07-04
Add one label to generate release notes

utam0k:
Make the heaviest of release tasks, the generation of release notes, much easier.

### DIFF
--- a/.github/workflows/label.yaml
+++ b/.github/workflows/label.yaml
@@ -1,0 +1,17 @@
+name: 🏷️ Pull Request Labels
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 1
+          labels: "kind/feature, kind/bug, kind/documentation, kind/test, kind/cleanup, dependencies"
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,21 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot
+  categories:
+    - title: 💪 Improvements
+      labels:
+        - kind/feature
+    - title: 🐛 Bug Fixes
+      labels:
+        - kind/bug
+    - title: 📖 Documentation improvements
+      labels:
+        - kind/documentation
+    - title: 🧪 Test improvements and Misc Fixes
+      labels:
+        - kind/test
+        - kind/cleanup
+    - title: Other Changes
+      labels:
+        - "*"


YJDoc2 (approved):
lgtm :sparkle: This is amazing!

YJDoc2:
We should also add the auto-merge PR workflow to add label dependabot on the PRs, else their workflow will fail

.github/workflows/label.yaml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 1
+          labels: "kind/feature, kind/bug, kind/documentation, kind/test, kind/cleanup, dependabot"
```
YJDoc2, line 16: Should the last one be dependencies instead? That is the one dependabot automatically adds

yihuaf:
Is there a tag or a way to exclude small fixes and trivial changes? Or we should lump these under the `other` section?

YJDoc2:
Maybe having a `kind/miscellaneous` would be a good way to collect those miscellaneous changes?

utam0k:
I thought about `kind/misc` too. But this time I decided not to. I thought about it but decided not to do it this time because relying on it would eventually burden the release notes.
Perhaps something similar might be added in the future, but I didn't think it was necessary to include it from the beginning. What do you think?

YJDoc2:
I think the way that is there currently in PR for grouping from labels is fine, I was thinking to add `kind/misc ` as a catchall for label checking CI. Think of it as chore-type of commits where we update CI or fix typos and so on. For them, currently the most applicable label is cleanup, but it doesn't seem correct. So if we allow `kind/misc` in PR checking CI, that will allow it to pass ; but we still count those in the 'other' section which is  the catch-all for generating release notes.

utam0k:
@YJDoc2 I think this case you said works fine using `kind/cleanup` following Kubernetes.
https://github.com/kubernetes/kubernetes/labels?q=kind%2F
https://github.com/kubernetes/kubernetes/pull/118186

utam0k:
I think `misc` is a meaningless label that categorizes nothing. Then I think we should allow no labels.

YJDoc2:
> @YJDoc2 I think this case you said works fine using kind/cleanup following Kubernetes.


> I think misc is a meaningless label that categorizes nothing. Then I think we should allow no labels.

Now I'm not sure either. In the end it is our call , so we can use cleanup for the typos and minor fixes as k8s does ;  or we can allow a PR to be without label. I think just for the sake of having all important PRs in the release notes and not missing them due to missing label ; as well as for reducing the task of manual writing of release notes ; I'll suggest we go ahead for now with the cleanup label approach, and if it has any difficulties in future, we can come back and re-discuss .  wdyt @yihuaf ?

utam0k:
@YJDoc2 Thanks for your understanding.

utam0k:
Once, I will merge this PR, but I welcome your feedback.

yihuaf:
cleanup is fine with me.

PR #2133, opened 2023-07-05
[Trivial] exclude the oci-runtime-test from the typos

yihuaf:
The oci runtime test in Go is maintained separately and used as a submodule. For us, there is no reason to check the typos.


YJDoc2 (approved):
lgtm :+1:

PR #2143, opened 2023-07-06
Increate musl CI test timeout to 20

YJDoc2:
Recently there were several cancellations in CI due to musl tests getting timed out. This increases timeout from 15->20 mins, as in most of those cases the test running is underway, and our unit tests should be finished within the increases 5 minutes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,7 @@ jobs:
     needs: [changes]
     if: needs.changes.outputs.any_modified == 'true'
     runs-on: ubuntu-22.04
-    timeout-minutes: 15
+    timeout-minutes: 20
     steps:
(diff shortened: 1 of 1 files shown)


yihuaf:
We can try this, but I am suspecting that the `musl` test has a bug somewhere that cause the test to hang.

yihuaf (approved):

YJDoc2:
Let us try increasing the timeout, if it still fails, then we can investigate more :)

YJDoc2:
It seems you were correct, there is some kind of bug or flake which is causing the musl test to timeout...

yihuaf:
> It seems you were correct, there is some kind of bug or flake which is causing the musl test to timeout...

See https://github.com/containers/youki/issues/2144 for details

PR #2149, opened 2023-07-07
Liboci additional flags and subcommands, as required by ociplex

c3d:
These are a few changes in the `liboci-cli` crate that are necessary for some additional commands that I am implementing in `ociplex`.

### DIFF
--- a/crates/liboci-cli/README.md
+++ b/crates/liboci-cli/README.md
@@ -16,6 +16,7 @@ Interface](https://github.com/opencontainers/runtime-tools/blob/master/docs/comm
 | checkpoint |            |                   | ✅   | ✅   |       |
 | events     | ✅         |                   | ✅   |      | ✅    |
 | exec       | ✅         |                   | ✅   | ✅   | ✅    |
+| features   | ✅         |                   | 🏴   |     |      |
 | list       | ✅         |                   | ✅   | ✅   | ✅    |
 | pause      | ✅         |                   | ✅   | ✅   | ✅    |
 | ps         | ✅         |                   | ✅   | ✅   | ✅    |
@@ -24,3 +25,6 @@ Interface](https://github.com/opencontainers/runtime-tools/blob/master/docs/comm
 | run        | ✅         |                   | ✅   | ✅   | ✅    |
 | spec       | ✅         |                   | ✅   | ✅   | ✅    |
 | update     |            |                   | ✅   | ✅   |       |
+
+🏴 The `features` subcommand is [implemented](https://github.com/opencontainers/runc/pull/3296)
+in `runc`, but not documented.
--- a/crates/liboci-cli/README.md
+++ b/crates/liboci-cli/README.md
@@ -16,7 +16,7 @@ Interface](https://github.com/opencontainers/runtime-tools/blob/master/docs/comm
 | checkpoint |            |                   | ✅   | ✅   |       |
 | events     | ✅         |                   | ✅   |      | ✅    |
 | exec       | ✅         |                   | ✅   | ✅   | ✅    |
-| features   | ✅         |                   | 🏴   |     |      |
+| features   | ✅         |                   | ✅   |      |       |
 | list       | ✅         |                   | ✅   | ✅   | ✅    |
 | pause      | ✅         |                   | ✅   | ✅   | ✅    |
 | ps         | ✅         |                   | ✅   | ✅   | ✅    |
@@ -25,6 +25,3 @@ Interface](https://github.com/opencontainers/runtime-tools/blob/master/docs/comm
 | run        | ✅         |                   | ✅   | ✅   | ✅    |
 | spec       | ✅         |                   | ✅   | ✅   | ✅    |
 | update     |            |                   | ✅   | ✅   |       |
-
-🏴 The `features` subcommand is [implemented](https://github.com/opencontainers/runc/pull/3296)
-in `runc`, but not documented.
--- a/crates/liboci-cli/src/exec.rs
+++ b/crates/liboci-cli/src/exec.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use clap::Parser;
 
 /// Execute a process within an existing container
+/// Reference: https://github.com/opencontainers/runc/blob/main/man/runc-exec.8.md
 #[derive(Parser, Debug)]
 pub struct Exec {
     /// Unix socket (file) path , which will receive file descriptor of the writing end of the pseudoterminal
@@ -20,6 +21,12 @@ pub struct Exec {
     /// Environment variables that should be set in the container
     #[clap(short, long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
     pub env: Vec<(String, String)>,
+    /// Run the command as a user
+    #[clap(short, long, value_parser = parse_colon_separated_pair::<u32, u32>)]
+    pub user: Option<(u32, Option<u32>)>,
+    /// Add additional group IDs. Can be specified multiple times
+    #[clap(long, short = 'g', number_of_values = 1)]
+    pub additional_gids: Vec<u32>,
     /// Prevent the process from gaining additional privileges
     #[clap(long)]
     pub no_new_privs: bool,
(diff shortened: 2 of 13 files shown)


crates/liboci-cli/src/exec.rs
```diff
@@ -49,3 +76,19 @@ where
         .ok_or_else(|| format!("invalid KEY=value: no `=` found in `{s}`"))?;
     Ok((s[..pos].parse()?, s[pos + 1..].parse()?))
 }
+
+fn parse_colon_separated_pair<T, U>(
```
yihuaf, line 80: This is to parse the user string. I would suggest to rename this function to parse user or something along the line. Then have a comment explaining the format of the user and why we parse it as colon separated pairs.

yihuaf:
Also, you can run `just lint` and `just spellcheck` locally to check for these misc things.

crates/liboci-cli/README.md
```diff
@@ -24,3 +25,6 @@ Interface](https://github.com/opencontainers/runtime-tools/blob/master/docs/comm
 | run        | ✅         |                   | ✅   | ✅   | ✅    |
 | spec       | ✅         |                   | ✅   | ✅   | ✅    |
 | update     |            |                   | ✅   | ✅   |       |
+
+🏴 The `features` subcommand is [implemented](https://github.com/opencontainers/runc/pull/3296)
+in `runc`, but not documented.
```
utam0k, line 30: This is defined by OCI Runtime Spec.
https://github.com/opencontainers/runtime-spec/blob/07dd1f41412c2c0b8f8d535c8aeaa9ced8dc3f5d/features-linux.md

c3d: Thanks for the link, @utam0k. This is probably because [when I wrote the code (March 9)](https://github.com/containers/youki/commit/947b1ee1c0a82e379997a29983c80e9ee1666245) the documentation was [not committed yet (March 21)](https://github.com/opencontainers/runtime-spec/commit/689874fc76bf43d5a79cb733bcc7bcf7756bdb0e). I will fix it.

crates/liboci-cli/src/exec.rs
c3d: > This is to parse the user string. I would suggest to rename this function to parse user or something along the line. Then have a comment explaining the format of the user and why we parse it as colon separated pairs.

This is easily done, but this seems to be inconsistent with existing usage: `parse_key_val` describes the parsing format (key/value pair), not what it is used for (environment variables). I suspect we might see other use cases for a colon-separated pair (e.g. for `docker` or `podman`, the `-v` option takes such a colon-separated pair).

I'll submit this as a separate commit on the same branch, also changing the `parse_key_val` to `parse_env`. You decide whether you want to pick that commit or not.

c3d: @yihuaf Implemented in ec24d0a

crates/liboci-cli/README.md
c3d: @utam0k Implemented in commit 4381887

utam0k:
@c3d May I ask you to fix the error from clippy? If you need, please ignore it

crates/liboci-cli/src/exec.rs
yihuaf: I see. You raised a good point. Thank you for doing this. We can revert to `parse_key_val` and `parse_colon_separated_pair`. Other than the clippy, LGTM.

c3d:
> @c3d May I ask you to fix the error from clippy? If you need, please ignore it

I saw an error about variants size pertaining to the subcommands. 

```
error: large size difference between variants
Error:   --> crates/youki/src/main.rs:48:1
   |
48 | / enum SubCommand {
49 | |     // Standard and common commands handled by the liboci_cli crate
50 | |     #[clap(flatten)]
51 | |     Standard(liboci_cli::StandardCmd),
   | |     --------------------------------- the second-largest variant contains at least 104 bytes
52 | |     #[clap(flatten)]
53 | |     Common(liboci_cli::CommonCmd),
   | |     ----------------------------- the largest variant contains at least 320 bytes
...  |
57 | |     Completion(commands::completion::Completion),
58 | | }
   | |_^ the entire enum is at least 320 bytes
```

There is indeed a lot of variations between the subcommands, since this branch adds a lot of command-specific options. I think that this is expected. Adding boxing is probably of little value for command-line arguments (they won't use that much memory to start with).

Is there a way to tell the rust linter to ignore this specific case?

yihuaf:
> > @c3d May I ask you to fix the error from clippy? If you need, please ignore it
> 
> I saw an error about variants size pertaining to the subcommands.
> 
> ```
> error: large size difference between variants
> Error:   --> crates/youki/src/main.rs:48:1
>    |
> 48 | / enum SubCommand {
> 49 | |     // Standard and common commands handled by the liboci_cli crate
> 50 | |     #[clap(flatten)]
> 51 | |     Standard(liboci_cli::StandardCmd),
>    | |     --------------------------------- the second-largest variant contains at least 104 bytes
> 52 | |     #[clap(flatten)]
> 53 | |     Common(liboci_cli::CommonCmd),
>    | |     ----------------------------- the largest variant contains at least 320 bytes
> ...  |
> 57 | |     Completion(commands::completion::Completion),
> 58 | | }
>    | |_^ the entire enum is at least 320 bytes
> ```
> 
> There is indeed a lot of variations between the subcommands, since this branch adds a lot of command-specific options. I think that this is expected. Adding boxing is probably of little value for command-line arguments (they won't use that much memory to start with).
> 
> Is there a way to tell the rust linter to ignore this specific case?

Try:

```
#[derive(Parser, Debug)]
#[allow(clippy::large_enum_variant)]
enum SubCommand {
...
}
```

The main issue is that this gets triggered in `youki` but the affected changes comes from `liboci-cli` which are suppose to be consumed as an individual library. We can use the above `allow` to suppress `youki` but is there any other consumer for `liboci-cli` out there?

c3d:
> > Is there a way to tell the rust linter to ignore this specific case?
> 
> Try:
> 
> ```
> #[derive(Parser, Debug)]
> #[allow(clippy::large_enum_variant)]
> enum SubCommand {
> ...
> }
> ```
> 
> The main issue is that this gets triggered in `youki` but the affected changes comes from `liboci-cli` which are suppose to be consumed as an individual library. We can use the above `allow` to suppress `youki` but is there any other consumer for `liboci-cli` out there?

Well, I'm co-developing the "other" user at the moment (a project called `ociplex`), which is the reason why this was put in a separate crate to start with.

> 
> Alternatively, you can box up a few of the large structure such as `Exec`.

I'll give this approach a try. My initial thinking is that this was more overhead than the difference in size, knowing that we don't have that many command structures to start with in a normal application, so the extra cost of heap management might not be worth it.

c3d:
@yihuaf I added a [separate commit](https://github.com/containers/youki/pull/2149/commits/bcabc067f538c497bdfb5c506353b11bea411242). Do you see this as an improvement? I can't say I like it much, but it does eliminate the `just lint` warning.

yihuaf:
> @yihuaf I added a [separate commit](https://github.com/containers/youki/pull/2149/commits/bcabc067f538c497bdfb5c506353b11bea411242). Do you see this as an improvement? I can't say I like it much, but it does eliminate the `just lint` warning.

Yea I don't like this as well, especially since this breaks up the liboci-cli apis to make each sub command argument inconsistent. The user now have to pay attention on which is box and which is not. How about we box the SubCommand in the application? In this way, each users of the liboci-cli can decide how to proceed or if to enforce the lint rule at all. I find this to be a better alternative. In the following diff, I boxed up the `enum` from liboci-cli in the youki application.

```diff
[eng@stardust youki]$ git diff
diff --git a/crates/liboci-cli/src/lib.rs b/crates/liboci-cli/src/lib.rs
index 6dafaf44..89c48a6d 100644
--- a/crates/liboci-cli/src/lib.rs
+++ b/crates/liboci-cli/src/lib.rs
@@ -52,7 +52,7 @@ pub enum StandardCmd {
 pub enum CommonCmd {
     Checkpointt(Checkpoint),
     Events(Events),
-    Exec(Box<Exec>),
+    Exec(Exec),
     Features(Features),
     List(List),
     Pause(Pause),
diff --git a/crates/youki/src/main.rs b/crates/youki/src/main.rs
index 09374a41..6a92be8d 100644
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -48,9 +48,9 @@ struct Opts {
 enum SubCommand {
     // Standard and common commands handled by the liboci_cli crate
     #[clap(flatten)]
-    Standard(liboci_cli::StandardCmd),
+    Standard(Box<liboci_cli::StandardCmd>),
     #[clap(flatten)]
-    Common(liboci_cli::CommonCmd),
+    Common(Box<liboci_cli::CommonCmd>),
 
     // Youki specific extensions
     Info(info::Info),
@@ -106,7 +106,7 @@ fn main() -> Result<()> {
     let systemd_cgroup = opts.global.systemd_cgroup;
 
     let cmd_result = match opts.subcmd {
-        SubCommand::Standard(cmd) => match cmd {
+        SubCommand::Standard(cmd) => match *cmd {
             StandardCmd::Create(create) => {
                 commands::create::create(create, root_path, systemd_cgroup)
             }
@@ -115,12 +115,12 @@ fn main() -> Result<()> {
             StandardCmd::Delete(delete) => commands::delete::delete(delete, root_path),
             StandardCmd::State(state) => commands::state::state(state, root_path),
         },
-        SubCommand::Common(cmd) => match cmd {
+        SubCommand::Common(cmd) => match *cmd {
             CommonCmd::Checkpointt(checkpoint) => {
                 commands::checkpoint::checkpoint(checkpoint, root_path)
             }
             CommonCmd::Events(events) => commands::events::events(events, root_path),
-            CommonCmd::Exec(exec) => match commands::exec::exec(*exec, root_path) {
+            CommonCmd::Exec(exec) => match commands::exec::exec(exec, root_path) {
                 Ok(exit_code) => std::process::exit(exit_code),
                 Err(e) => {
                     eprintln!("exec failed : {e}");
[eng@stardust youki]$ 
```

yihuaf:
I will help you to get this merged since this is the last outstanding items.

yihuaf (approved):

PR #2150, opened 2023-07-08
disable musl test for now

yihuaf:
The musl test is too flaky for now (hangs) and should be resolved before merging. ~~I introduced a new CI to park the code so we can easily re-enable the test once we are done.~~

Removed the test and we can revert this PR when the test become stable again.

Details of the problem in https://github.com/containers/youki/issues/2144

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,31 +70,32 @@ jobs:
       - name: Run feature tests
         run: just test-features
 
-  # musl target testing is religated to a separate job because our current build
-  # for musl requires nightly. It is difficult to mix stable and nightly
-  # toolchains in the same job.
-  musl:
-    needs: [changes]
-    if: needs.changes.outputs.any_modified == 'true'
-    runs-on: ubuntu-22.04
-    timeout-minutes: 20
-    steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/parked.yml
```diff
@@ -0,0 +1,52 @@
+# This CI action is intended for experimental, currently broken, or flaky
```
utam0k, line 1: May I ask you to add the name of this CI.

yihuaf: After thinking about this more, I think it makes more sense to just remove the `musl` section of the test. This PR can be served as the record in case we want to revert the changes when the test becomes stable again.

YJDoc2:
@yihuaf Sorry for the nitpick, but can I ask you to simply comment the test in CI file instead of completely removing it? Its kind of a personal preference, but I feel that way it serves as a reminder that this test is temp removed, and needs to be fixed ; rather than completely removing it. Apologies for late review :sweat:

yihuaf:
> @yihuaf Sorry for the nitpick, but can I ask you to simply comment the test in CI file instead of completely removing it? Its kind of a personal preference, but I feel that way it serves as a reminder that this test is temp removed, and needs to be fixed ; rather than completely removing it. Apologies for late review 😓

No worries :) Done.

YJDoc2 (approved):
lgtm

PR #2155, opened 2023-07-10
Use syscall type to delay the creation of syscall struct.

yihuaf:
Pulling out changes from #2121 

Included #2150 to disable musl test. Will rebase once #2150 is merged.

Details:

In this PR we propose to store a SyscallType instead of syscall structure. The syscall structure itself is am empty struct, so we can delay the creation. In this way, we don't have to pass this reference through the fork/clone boundary. The eventual goal is not to use lifetime reference in the `container_args` structure.

### DIFF
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -1,18 +1,19 @@
 use crate::error::{ErrInvalidID, LibcontainerError};
+use crate::syscall::syscall::SyscallType;
+use crate::utils::PathBufExt;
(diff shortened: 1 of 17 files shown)


utam0k (approved):
The code looks good, but I made the kind label more appropriate `kind/cleanup` .

utam0k:
@yihuaf 
Thanks for resolving my tech debut 🙏

yihuaf:
> @yihuaf Thanks for resolving the tech debut I made🙏

No worries :) More incoming.

PR #2168, opened 2023-07-14
Refactor the libcgroups interface

yihuaf:
- Refactored the cgroup creation into a struct instead of individual parameters. This reduce the amount of container_args field we have to carry across the process boundry.
- Refactored the use of &Path vs. PathBuf in the interface.
- Tag along since it is too small. Minor updates to the notify listener comments and variable name. Also makes it clone-able.

I have 3 more PR.

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -316,17 +316,22 @@ pub enum CreateCgroupSetupError {
     Systemd(#[from] systemd::manager::SystemdManagerError),
 }
 
-pub fn create_cgroup_manager<P: Into<PathBuf>>(
-    cgroup_path: P,
-    systemd_cgroup: bool,
-    container_name: &str,
+#[derive(Clone)]
+pub struct CgroupConfig {
+    pub cgroup_path: PathBuf,
+    pub systemd_cgroup: bool,
+    pub container_name: String,
+}
+
+pub fn create_cgroup_manager(
+    config: CgroupConfig,
 ) -> Result<AnyCgroupManager, CreateCgroupSetupError> {
     let cgroup_setup = get_cgroup_setup().map_err(|err| match err {
         GetCgroupSetupError::WrappedIo(err) => CreateCgroupSetupError::WrappedIo(err),
         GetCgroupSetupError::NonDefault => CreateCgroupSetupError::NonDefault,
         GetCgroupSetupError::FailedToDetect => CreateCgroupSetupError::FailedToDetect,
     })?;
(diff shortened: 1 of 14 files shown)


crates/libcgroups/src/common.rs
```diff
@@ -316,65 +316,70 @@ pub enum CreateCgroupSetupError {
...
-pub fn create_cgroup_manager<P: Into<PathBuf>>(
-    cgroup_path: P,
-    systemd_cgroup: bool,
-    container_name: &str,
+#[derive(Clone)]
+pub struct CgroupConfig {
+    pub cgroup_path: PathBuf,
+    pub systemd_cgroup: bool,
+    pub container_name: String,
+}
```
utam0k, line 324: What's the benefit of this change?

utam0k:
This is a breaking change to libcgroup, not libcontainer, right?

crates/libcgroups/src/common.rs
yihuaf: This reduces the number of variables I have to store inside container_args. This PR moves the creation of the cgroup manager to the first thing in intermediate process, so we have to store the variables needed to call cgroup create manager. I can either store 3 or store a 1 struct. Also, these variables are easily shown to be clone-able.

crates/libcontainer/src/notify_socket.rs
```diff
@@ -93,6 +96,15 @@ impl NotifyListener {
...
 }
 
+impl Clone for NotifyListener {
+    fn clone(&self) -> Self {
+        let fd = self.socket.as_raw_fd();
+        // This is safe because we just duplicate a valid fd.
+        let socket = unsafe { UnixListener::from_raw_fd(fd) };
+        Self { socket }
+    }
+}
```
utam0k, line 106: 👍 but may I ask you to add the test to ensure the same fs?

yihuaf:
> This is a breaking change to libcgroup, not libcontainer, right?

Yes.

crates/libcontainer/src/notify_socket.rs
yihuaf: Added a unit test to make sure the cloned listener still functions correctly.

crates/libcgroups/src/common.rs
utam0k: I didn't say enough. I think making it a structure is a good change. However, I doubt that the type is changing🤔

yihuaf: > I doubt that the type is changing

I am not sure I follow what you are saying here? Can you rephrase/elaborate?

PR #2186, opened 2023-07-19
refactored executor and executor manager

yihuaf:
Refactored the executor and executor manager to be clone-able and more ergonomics.

Removed the use of executor manager with a vector of executor. Instead we favor composing executors into a new executor. Composing also allows the implementer to be precise about the execution orders. The old executor manager with its vector implementation is not explicit enough about the order when multiple executor is involved.

Re-implement the executor to be a function pointer instead of the Fn family traits. Traits will type erase and difficult to deal with when implementing clone.

Refactored the wasm related exectors to use the new scheme.

### DIFF
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -1,8 +1,7 @@
 use crate::error::{ErrInvalidID, LibcontainerError};
 use crate::syscall::syscall::SyscallType;
 use crate::utils::PathBufExt;
-use crate::workload::default::DefaultExecutor;
-use crate::workload::{Executor, ExecutorManager};
+use crate::workload::{self, Executor};
 use std::path::PathBuf;
 
 use super::{init_builder::InitContainerBuilder, tenant_builder::TenantContainerBuilder};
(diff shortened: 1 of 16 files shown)


yihuaf:
I will be responsible for fixing any breakage with runwasi if required.

crates/libcontainer/src/container/builder.rs
```diff
@@ -23,7 +22,7 @@ pub struct ContainerBuilder {
     pub(super) preserve_fds: i32,
     /// Manage the functions that actually run on the container
     /// Default executes the specified execution of a generic command
```
utam0k, line 24: Please update this comment 🙏

yihuaf:
> Can you add Developer Documentation on this so we can use another PR or just create an issue? https://containers.github.io/youki/

I added a section in the libcontainer.md. I don't think there is any previous mentions of executor. Let me know if this is good enough for now.

utam0k:
> > Can you add Developer Documentation on this so we can use another PR or just create an issue? https://containers.github.io/youki/
> 
> I added a section in the libcontainer.md. I don't think there is any previous mentions of executor. Let me know if this is good enough for now.

Can you post an example of how to use it, just another PR?

utam0k (approved):
Thansks!

PR #2193, opened 2023-07-21
Make container_args clone-able

yihuaf:
Make the container_args clone-able and remove the use of references. Container_args is passed to the intermediate and init process, so we need to make sure it can across the clone process boundry safely by allowing passing of the container_args ownership to the new process.


crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -28,7 +28,7 @@ pub(super) struct ContainerBuilderImpl<'a> {
     /// Id of the container
     pub container_id: String,
     /// OCI compliant runtime spec
-    pub spec: &'a Spec,
+    pub spec: Rc<Spec>,
```
utam0k, line 31: 👍

utam0k (approved):

PR #2230, opened 2023-08-03
Refactored the Executor interface yet again

yihuaf:
I apologize having to refactor this interface yet again. Previously, we introduced the executor to be a function pointer. This works out nicely because the function pointer in rust can be clone-ed without hassel. However, I realized that using function pointer is way too restrictive for our users. The executor may wish to include additional context when calling the exec function. The function pointer limited the input only `oci spec`. The new constraint is that the struct implementing executor must be clone-able, which we need to carry this struct across process boundry.

### DIFF
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -22,7 +22,7 @@ pub struct ContainerBuilder {
     pub(super) preserve_fds: i32,
     /// The function that actually runs on the container init process. Default
     /// is to execute the specified command in the oci spec.
-    pub(super) executor: Executor,
+    pub(super) executor: Box<dyn Executor>,
 }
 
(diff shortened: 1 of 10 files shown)


YJDoc2 (approved):
looks good :+1:
I'm going to squash and merge so that the doc test and spell fixing commits get merged together.
If I'm not wrong, then this is basically the original executor scheme except for the executor manager and executor lists, right?

yihuaf:
> If I'm not wrong, then this is basically the original executor scheme except for the executor manager and executor lists, right?

Yes. I also removed the `can_handle` and the executor name. The executor name is removed because this doesn't apply to the new scheme. The `can_handle` part of the logic is moved into `exec` with an error kind `can't handle`. With that being said, there are some talks to add the `can_handle` or a `validate` function back, so we can abstract the validation part of the logic out for each executor.

PR #2233, opened 2023-08-06
Fix podman tests to properly run

YJDoc2:
Currently almost all of the podman tests are failing to run. The CI doesn't fail because we are intentionally ignoring the failures. Many of the tests use skopeo to download certain container image for the test. Unfortunately the default version on ubuntu for skopeo is 1.4.0, however a certain cmd option used by the tests to download the images only exist in newer versions. 
Thus now we clone the 1.13.1 release and build the skopeo in the CI itself. Moreover we use the makefile target `localsystem` instead of directly running bats command. This also adds a `workflow_dispatch` for podman CI so we can run them easily on specific branches that we want.

As it stands now, the current main have ~76 failing tests in rootful mode. Several of them are due to incorrect configuration of env, for eg some tests expect an ssh key to be configured etc. However, these kind of failing tests can be ignored, as they are not invoking youki. The rest of the failing tests might indicate something wrong in youki-podman interfacing, and should be fixed.

On the other hand onrootless mode, currently ~279 tests are failing with current main and on my branch where rootless is fixed. The main branch failures were expected, and have the expected error message, however I need to figure out and fix the them when I make my rootless PR if possible.

In long term, I think we should fix all youki-related failures, ignore others, and make it so that any failures will fail CI. Then we can have a label based trigger, where after all review and testing is done (for big PRs only) we add a lable like `run-podman-tests` and it will trigger CI for that , after which it can be merged. (That is not done in this PR)

### DIFF
--- a/.github/workflows/podman_tests.yaml
+++ b/.github/workflows/podman_tests.yaml
@@ -3,33 +3,53 @@ name: Test for podman
 on:
   schedule:
   - cron: "0 0 * * *"
+  workflow_dispatch:
 
 jobs:
   podman-tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
       - name: Install just
         uses: taiki-e/install-action@just
       - name: Install requirements
         run: sudo env PATH=$PATH just ci-prepare
-      - run: just youki-dev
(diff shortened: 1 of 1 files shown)


utam0k (approved):
Thansk!

PR #2257, opened 2023-08-12
Rename the rootless struct  to UserNamespaceConfig

YJDoc2:
The current `Rootless` struct has more to do with creating a new user namespace that rootless containers. For example, it is possible and valid that a root user will run a container with new user ns, and in such case the current code treats it as "running in rootless" . This PR simply re-names stuff, but does not actually change any functionality.  This is roughly 1st of 3 parts for making podman rootless work.

The main changes here are changing the `rootless.rs` file to `user_ns.rs` , and changing the struct names 
- Rootless -> UserNamespaceConfig
- RootlessIDMapper -> UserNamespaceIDMapper
- RootlessError -> UserNamespaceError

Few of the struct members are also renamed to reflect these. Apart from this file, changes in other files are simply to update for these re-naming. If you have any better idea for the names, feel free to comment.

### DIFF
--- a/MirgationGuide.md
+++ b/MirgationGuide.md
@@ -1,0 +1,14 @@
+This contains information for migrating library versions.
+
+## V0.1.0 -> v0.2.0
+
+### libcontainer
+
+- The `Rootless` struct has been re-named as `UserNamespaceConfig` , `RootlessIDMapper` has been re-named to `UserNamespaceIDMapper` , and correspondingly the `RootlessError` has been re-named to `UserNamespaceError` . This is due to the fact that the structure was to be used for containers when a new user namespace is to be created, and that is not strictly only for rootless uses. Accordingly, the fields of various structs has been updated to reflect this change :
+    - rootless (module name) -> user_ns
+    - Rootless::rootless_id_mapper -> UserNamespaceConfig::id_mapper
+    - LibcontainerError::Rootless -> LibcontainerError::UserNamespace
+    - ContainerBuilderImpl::rootless -> ContainerBuilderImpl::user_ns_config
+    - ContainerArgs::rootless -> ContainerArgs::user_ns_config
+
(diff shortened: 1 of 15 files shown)


MirgationGuide.md
utam0k, line 1: Could you create an issue for releasing `v0.2.0` instead of this file?

yihuaf (approved):

MirgationGuide.md
YJDoc2: Hey, I created this file, as a lot of libraries when releasing breaking versions usually have some info on what has changed and what changes wold need to be done for upgrading to new version (separate from release notes). That way anyone upgrading can see that and make changes accordingly.

I can make an issue for the release, but  do you still want to remove this file?

utam0k: Thanks for your response!

We are responsible for not only libcontainer but also youki itself. I'm worried that this filename makes youki's user confusing.
> Hey, I created this file, as a lot of libraries when releasing breaking versions usually have some info on what has changed

I want to know why we need to separate the release note and this file.

YJDoc2: Fair enough, It can be confusing for youki users, as there is no upgrade for youki as a whole, only lib*
>I'm worried that this filename makes youki's user confusing.

Usually from what I have seen for other libraries, in release notes, they usually link PRs and important changes, and then have a separate .MD file, or a section in readme which lists what changes in code users of the library will need to do in order to upgrade to new version. I thought we should have something similar for lib* . So for example, the release note will only contain `Changed executor interface for composibility and ease-of-use` , whereas the other file will list steps one might take or considerations one might have for changing their own executor implementation. It will be a more detailed and longer description than release notes. Currently I have only listed structure re-names, as that is only changed in this PR.
> I want to know why we need to separate the release note and this file.

utam0k (approved):
👍

PR #2279, opened 2023-08-19
Change rootless required function and privilege decision

YJDoc2:
This fixes various checks that we do with regard to root user, in order to make rootless containers work properly. 
Primarily, we not only check that the uid of user is 0, but along with that we also check the uid map file to see if available range is max. If the range is not, that means we are in a new user namespace, and thus might not be actual root user, and thus we should not do things that requires root user permissions. This is 2nd out of 3 PRs for making youki work with podman in rootless. Even though even with this PR youki still does not work with podman in rootless mode (as dbus connection is still incorrect), I have merged this branch with my dbus->zbus branch and tested that with correct dbus connection logic, youki works with podman in rootless context. The main validation for this right now is that this PR does not change/break any existing functionality.

---

A side note that I noticed while making this : not all logs and errors that we are generating are actually getting displayed when using youki. For example, as of now, we verify the binary before running the container, and error if it does not exist. However, the podman log simply errors out, not showing anything, whereas with runc proper log of missing executable is shown : 
```sh
# Youki
podman run --runtime $PWD/youki busybox /bin/bash
DEBUG youki: started by user 0 with ArgsOs { inner: ["...../youki/youki", "delete", "--force", "9c937d43d590695b71556382fd6d9f4baa51c9780c7969b94e7c7753ad542487"] }
DEBUG youki::commands::delete: start deleting 9c937d43d590695b71556382fd6d9f4baa51c9780c7969b94e7c7753ad542487
Error: container create failed (no logs from conmon): EOF
```

```sh
# runc
podman run --runtime runc busybox /bin/bash
Error: runc create failed: unable to start container process: exec: "/bin/bash": stat /bin/bash: no such file or directory: OCI runtime attempted to invoke a command that was not found
```

I'm pretty sure that we are generating similar error and propagating that upwards. This is similarly observed in actual `journalctl` logs (after always enabling them for debug in this PR) , that not all of tracing logs are shown. cCan this be because with namespace change that we do in intermediate/init process, those logs are not getting forwarded properly?

### DIFF
--- a/crates/libcontainer/src/container/init_builder.rs
+++ b/crates/libcontainer/src/container/init_builder.rs
@@ -13,7 +13,7 @@ use crate::{
     error::{ErrInvalidSpec, LibcontainerError, MissingSpecError},
     notify_socket::NOTIFY_FILE,
     process::args::ContainerType,
-    tty, user_ns,
+    tty, user_ns, utils,
 };
 
 use super::{
@@ -87,6 +87,11 @@ impl InitContainerBuilder {
         };
 
         let user_ns_config = UserNamespaceConfig::new(&spec)?;
+
+        if utils::rootless_required() && !utils::is_in_new_userns() && user_ns_config.is_none() {
+            return Err(LibcontainerError::NoUserNamespace);
+        }
+
         let config = YoukiConfig::from_spec(&spec, container.id(), user_ns_config.is_some())?;
         config.save(&container_dir).map_err(|err| {
             tracing::error!(?container_dir, "failed to save config: {}", err);
--- a/crates/libcontainer/src/container/init_builder.rs
+++ b/crates/libcontainer/src/container/init_builder.rs
@@ -88,10 +88,6 @@ impl InitContainerBuilder {
 
         let user_ns_config = UserNamespaceConfig::new(&spec)?;
 
-        if utils::rootless_required() && !utils::is_in_new_userns() && user_ns_config.is_none() {
-            return Err(LibcontainerError::NoUserNamespace);
-        }
-
         let config = YoukiConfig::from_spec(&spec, container.id(), user_ns_config.is_some())?;
         config.save(&container_dir).map_err(|err| {
             tracing::error!(?container_dir, "failed to save config: {}", err);
@@ -196,6 +192,8 @@ impl InitContainerBuilder {
                 }
             }
         }
+
+        utils::validate_spec_for_new_user_ns(spec)?;
 
         Ok(())
     }
--- a/crates/libcontainer/src/utils.rs
+++ b/crates/libcontainer/src/utils.rs
@@ -10,6 +10,10 @@ use std::path::{Component, Path, PathBuf};
 use nix::sys::stat::Mode;
 use nix::sys::statfs;
 use nix::unistd::{Uid, User};
+use oci_spec::runtime::Spec;
+
+use crate::error::LibcontainerError;
+use crate::user_ns::UserNamespaceConfig;
 
 #[derive(Debug, thiserror::Error)]
 pub enum PathBufExtError {
@@ -273,10 +277,28 @@ pub fn rootless_required() -> bool {
     is_in_new_userns()
 }
 
+/// checks if given spec is valid for current user namespace setup
+pub fn validate_spec_for_new_user_ns(spec: &Spec) -> Result<(), LibcontainerError> {
+    let config = UserNamespaceConfig::new(spec)?;
+
+    // In case of rootless, there are 2 possible cases :
+    // we have a new user ns specified in the spec
+    // or the youki is launched in a new user ns (this is how podman does it)
+    // So here, we check if rootless is required,
+    // but we are neither in a new user ns nor a new user ns is specified in spec
+    // then it is an error
+    if rootless_required() && !is_in_new_userns() && config.is_none() {
+        return Err(LibcontainerError::NoUserNamespace);
+    }
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::test_utils;
     use anyhow::{bail, Result};
+    use serial_test::serial;
 
     #[test]
     pub fn test_get_unix_user() {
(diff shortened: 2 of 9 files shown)


crates/libcontainer/src/container/init_builder.rs
```diff
@@ -87,6 +87,11 @@ impl InitContainerBuilder {
         };
 
         let user_ns_config = UserNamespaceConfig::new(&spec)?;
+
+        if utils::rootless_required() && !utils::is_in_new_userns() && user_ns_config.is_none() {
+            return Err(LibcontainerError::NoUserNamespace);
+        }
```
utam0k, line 93: Can we cover this case in the test?

utam0k: Since I feel  `validate_spec` has a role to check that a spec is valid, how about moving there?

crates/youki/src/observability.rs
```diff
@@ -73,7 +73,13 @@ where
...
     let log_format = detect_log_format(config.log_format.as_deref())
         .with_context(|| "failed to detect log format")?;
-    let systemd_journald = if config.systemd_log {
+
+    #[cfg(debug_assertions)]
+    let journald = true;
+    #[cfg(not(debug_assertions))]
+    let journald = config.systemd_log;
+
+    let systemd_journald = if journald {
```
utam0k, line 82: @yihuaf May I ask you to review around there?

yihuaf: This should be moved in an issue to track. I need to understand how conmon is handling the logs.

yihuaf: https://github.com/containers/youki/issues/2281

crates/youki/src/observability.rs
```diff
@@ -73,7 +73,13 @@ where
     let log_level_filter = tracing_subscriber::filter::LevelFilter::from(level);
     let log_format = detect_log_format(config.log_format.as_deref())
         .with_context(|| "failed to detect log format")?;
-    let systemd_journald = if config.systemd_log {
+
+    #[cfg(debug_assertions)]
```
yihuaf, line 77: Is this intended to be merged with this PR?

crates/libcontainer/src/container/init_builder.rs
```diff
@@ -87,6 +87,11 @@ impl InitContainerBuilder {
         };
 
         let user_ns_config = UserNamespaceConfig::new(&spec)?;
+
+        if utils::rootless_required() && !utils::is_in_new_userns() && user_ns_config.is_none() {
```
yihuaf, line 91: Nit: can we move this an individual function and add some comments for this if statement? It will be easier to read this if conditions.

yihuaf (approved):
Looks good.

crates/libcontainer/src/container/init_builder.rs
YJDoc2: If you mean to only move the if condition in a function, I'm not sure if that would be much useful, as that is only used twice, at the places where we construct new UserNamespaceConfig. Can you please let me know if you meant to move anything else as well?
I'll definitely add comment to the condition explaining it in either case :+1:

crates/youki/src/observability.rs
YJDoc2: I added this because it helped with debugging youki when running via podman, because a lot of logs were not getting displayed by podman. I can remove this in this PR and we can come back to this in the issue you have opened for the logs, this is not necessary for the rootless changes.

crates/youki/src/observability.rs
YJDoc2: Thanks :+1: I'll mark this comment as resolved for now

crates/libcontainer/src/container/init_builder.rs
YJDoc2: Hey, sure :+1: I'll add tests to cover this case, as well as look into moving it into the `validate_spec` function instead . Thanks!

utam0k: Thanks @YJDoc2

crates/libcontainer/src/container/init_builder.rs
yihuaf: I mean `utils::rootless_required() && !utils::is_in_new_userns() && user_ns_config.is_none()` into a function. It will take a reader a good second to parse out all correct conditions to understand what this if statement does. I am fine with having some comments to explain.

crates/youki/src/observability.rs
yihuaf: I see. We can keep it.

utam0k (approved):

utam0k:
Feel free to merge it after reacting to yihuaf 's comment.

YJDoc2:
> Feel free to merge it after reacting to yihuaf 's comment.

Hey sorry, I was busy with some other stuff. I still have to respond to that and add tests as you had suggested, will try to get it done in this weekend

YJDoc2:
Hey @utam0k , can you help me out regarding to tests that you have suggested? I was trying to write a unit test that would check for various conditions of the if-statement that I have added. However to do that, I'll need to basically setup the whole container dir with the intended spec and the bundle. Setting up the spec is straightforward, but I'm not sure how I should setup the bundle dir in unit tests. I couldn't find any other existing tests which do that either.

I can add individual unit tests for testing `is_in_new_userns` and `rootless_required` by doing fork and test ; but that is a little tedious too. Can you suggest me some approach to add tests for this?

YJDoc2:
Hey, actually I thought a bit more, and I think I have figured out a cleaner way, which also makes writing the tests easier, so I'll do that and push for review.

crates/libcontainer/src/utils.rs
```diff
@@ -379,4 +401,38 @@ mod tests {
...
+            assert!(validate_spec_for_new_user_ns(&rootful_spec).is_ok());
+
+            let rootless_spec = Spec::rootless(1000, 1000);
+            // following should succeed irrespective if we're in user ns or not
+            assert!(validate_spec_for_new_user_ns(&rootless_spec).is_ok());
+            Ok(())
+        })?;
+
+        Ok(())
+    }
```
YJDoc2, line 437: This still doesn't cover all possible permutations, but as we should run the tests themselves as non-root, we cannot test cases where the user is actually root, and both specs pass.

crates/libcontainer/src/container/init_builder.rs
YJDoc2: Hey, actually doing this helped me to test that function cleanly too! Thanks for the suggestion :+1: I have also added corresponding comment explaining the conditions.

crates/libcontainer/src/container/init_builder.rs
YJDoc2: Done :+1:

crates/libcontainer/src/utils.rs
```diff
@@ -379,4 +401,36 @@ mod tests {
 
         Ok(())
     }
+
+    // the following test is marked as serial because
+    // we are doing unshare of user ns and fork, so better to run in serial,
+    #[test]
+    #[serial]
+    fn test_userns_spec_validation() -> Result<(), test_utils::TestError> {
```
utam0k, line 409: 👍

utam0k (approved):

YJDoc2:
@utam0k Is it fine if I merge this now? I have also addressed yihuaf's comment in the last commit

PR #2280, opened 2023-08-20
Move the bin verification into executor

yihuaf:
With libcontainer supporting usecases other than OCI container, we need accomodate the usecase where executor such as from wasm may not have the same requirements for binary validation. Therefore, we should not verify binary in the common path in container_init_process.

We move the binary verification logic into the OCI default executor. In this way, this logic is only run against the OCI container. Other executor are free to implement their own binary verification.

Note, this diviates from how `runc` in the case where binary is not valid. Before, we fail in the `youki create` step. Now, we fail at the executor or `youki start` step. This likely should not cause any issue for docker, podman, or k8s since they run `runc create` and `runc start` together. Nonethless, there should be documented.

### DIFF
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -76,72 +76,6 @@ pub enum InitProcessError {
 
 type Result<T> = std::result::Result<T, InitProcessError>;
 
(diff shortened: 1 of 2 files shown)


yihuaf:
Never mind. This fails the containerd integration tests.

PR #2308, opened 2023-08-27
Return an error when passing unsupported mount options

utam0k:
Relate: https://github.com/containers/youki/issues/2307

https://github.com/opencontainers/runtime-spec/blob/main/config.md#linux-mount-options

### DIFF
--- a/crates/libcontainer/src/rootfs/utils.rs
+++ b/crates/libcontainer/src/rootfs/utils.rs
@@ -190,6 +190,7 @@ mod tests {
 
     use super::*;
 
+    use anyhow::Result;
     use oci_spec::runtime::MountBuilder;
 
     #[test]
@@ -205,7 +206,7 @@ mod tests {
     }
 
     #[test]
-    fn test_parse_mount() -> anyhow::Result<()> {
+    fn test_parse_mount() -> Result<()> {
         let mount_option_config = parse_mount(
             &MountBuilder::default()
                 .destination(PathBuf::from("/proc"))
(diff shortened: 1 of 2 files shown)


crates/libcontainer/src/rootfs/utils.rs
```diff
@@ -198,15 +205,14 @@ mod tests {
     }
 
     #[test]
-    fn test_parse_mount() {
+    fn test_parse_mount() -> anyhow::Result<()> {
```
Furisto, line 208:
```suggestion
    fn test_parse_mount() -> Result<()> {
```
Add a `use anyhow`

crates/libcontainer/src/rootfs/utils.rs
```diff
@@ -118,7 +159,12 @@ pub fn parse_mount(m: &Mount) -> MountOptionConfig {
                 "norelatime" => Some((true, MsFlags::MS_RELATIME)),
                 "strictatime" => Some((true, MsFlags::MS_STRICTATIME)),
                 "nostrictatime" => Some((true, MsFlags::MS_STRICTATIME)),
-                _ => None,
+                unknown => {
```
Furisto, line 162: Wouldn't this silently ignore unknown mount options that are not idmap? I think we should always return an error here.

crates/libcontainer/src/rootfs/utils.rs
utam0k: 👍

crates/libcontainer/src/rootfs/utils.rs
utam0k: For example, `mode=755` should be passed these codes, but we don't need to set `MsFlags`.
https://github.com/opencontainers/runtime-spec/blob/main/config.md#linux-mount-options
> Runtimes SHOULD pass unknown options to [mount(2)](http://man7.org/linux/man-pages/man2/mount.2.html) via the fifth argument (const void *data).

`idmap` and `ridmap` are defined in OCI Runtime Spec, So users may think these options will affect their mounts and pass them on. But we do not support them and would like to return an error to be safe in this case.

Furisto (approved):

PR #2345, opened 2023-09-10
Allow our integration test to capture the stdout

utam0k:
The method of writing test results to standard output is rather convenient.

### DIFF
--- a/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
@@ -1,5 +1,5 @@
 use oci_spec::runtime::{LinuxBuilder, ProcessBuilder, Spec, SpecBuilder};
-use test_framework::{Test, TestGroup, TestResult};
+use test_framework::{assert_result_eq, Test, TestGroup, TestResult};
 
 use crate::utils::test_inside_container;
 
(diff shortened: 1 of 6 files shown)


utam0k:
@yihuaf @YJDoc2 May I ask you to review this PR?🙏

YJDoc2:
Hey @utam0k , sorry I was busy last week, so couldn't get to this.  So this change intends to print the test result to stdout instead of returning pass/fail? I am a bit confused :sweat_smile: 
Also what was the reason for hostname test changes?

tests/rust-integration-tests/integration_test/src/tests/hostname/mod.rs
```diff
@@ -24,21 +24,24 @@ fn create_spec(hostname: &str) -> Spec {
...
-        // by the spec, so nothing to prepare prior.
-        Ok(())
-    })
+    let hostname = "hostname-specific";
+    let spec = create_spec(hostname);
+    match test_inside_container(spec, &|_| Ok(())) {
+        Err(e) => TestResult::Failed(e),
+        Ok(output) => {
+            let stdout = String::from_utf8_lossy(&output.stdout);
+            let mut expected = hostname.to_string();
```
yihuaf, line 33: Nit: user `format!("{hostname}\n")` for create the expected string?

yihuaf (approved):
May be also get a +1 from @YJDoc2

utam0k:
> Hey @utam0k , sorry I was busy last week, so couldn't get to this. So this change intends to print the test result to stdout instead of returning pass/fail? I am a bit confused 😅 Also what was the reason for hostname test changes?

I was thinking that it would be simpler to use stdout instead of using the runtime in the case of simple. But I gave it a thought again, and I found this way not good, so I'll close it.
The runtime method is difficult at first, but I thought it wiser to have documentation and examples for it than to have multiple testing methods.

Issue #2317, opened 2023-08-28
Update nix to 0.27.*


YJDoc2:
ref : https://github.com/containers/youki/pull/2312
Nix has released a breaking version, and we need to do some corresponding code changes and verify that nothing is breaking.

anti-entropy123:
Hi, I'm also working on resolving this issue. #2369. 
Feel free to offer suggestions.

YJDoc2:
@anti-entropy123 please comment on the issue before you make the PRs especially if it is assigned to someone else already ... :slightly_smiling_face: 
@lengrongfu are you also working on this? Is it fine if I re-assign this?

lengrongfu:
Of course, please redistribute it @YJDoc2 .thank you for your contribution @anti-entropy123 .

YJDoc2:
this is closed via #2369 
Thanks @anti-entropy123 !

PR #2369, opened 2023-09-19
update nix to 0.27.1

anti-entropy123:
I'm currently working on issue #2317.

**note: the following content is now outdated**

In the code changes I've made, there are two things that might raise questions:
1. The use of `std::mem::forget()`.
2. The use of an `unsafe` function `OwnedFd::from_raw_fd()`.

I use `forget()` because some functions in the `nix` library now return an `OwnedFd`. This helps avoid accidentally closing the file descriptor too soon by preventing its automatic drop. Maybe fully relying on `OwnedFd` to manage the fd's lifetime instead of manually calling close on RawFd is a better approach, but I think this could require quite a bit of code changes.

I also use `OwnedFd::from_raw_fd()` because of the `nix` library. The `nix::sched::setns()` function now requires an `Fd: AsFd` as one of its parameters, and I couldn't find a better way to convert an `i32` into `Fd: AsFd`.

Additionally, this is my first time contributing code to an open-source repository. If there are any important details I've missed, please let me know.

### DIFF
--- a/crates/libcontainer/src/channel.rs
+++ b/crates/libcontainer/src/channel.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     io::{IoSlice, IoSliceMut},
     marker::PhantomData,
-    os::unix::prelude::RawFd,
+    os::{fd::AsRawFd, unix::prelude::RawFd},
 };
 
 #[derive(Debug, thiserror::Error)]
@@ -210,10 +210,14 @@ where
 
 // Use socketpair as the underlying pipe.
 fn unix_channel() -> Result<(RawFd, RawFd), ChannelError> {
-    Ok(socket::socketpair(
+    let (f1, f2) = socket::socketpair(
         socket::AddressFamily::Unix,
         socket::SockType::SeqPacket,
         None,
         socket::SockFlag::SOCK_CLOEXEC,
-    )?)
+    )?;
+    let f1 = std::mem::ManuallyDrop::new(f1);
+    let f2 = std::mem::ManuallyDrop::new(f2);
+
+    Ok((f1.as_raw_fd(), f2.as_raw_fd()))
 }
--- a/crates/libcontainer/src/channel.rs
+++ b/crates/libcontainer/src/channel.rs
@@ -216,6 +216,8 @@ fn unix_channel() -> Result<(RawFd, RawFd), ChannelError> {
         None,
         socket::SockFlag::SOCK_CLOEXEC,
     )?;
+    // It is not straightforward to share the OwnedFd across forks, so we 
+    // treat them as i32. We use ManuallyDrop to keep the connection open. 
     let f1 = std::mem::ManuallyDrop::new(f1);
     let f2 = std::mem::ManuallyDrop::new(f2);
 
--- a/crates/libcontainer/src/channel.rs
+++ b/crates/libcontainer/src/channel.rs
@@ -216,8 +216,8 @@ fn unix_channel() -> Result<(RawFd, RawFd), ChannelError> {
         None,
         socket::SockFlag::SOCK_CLOEXEC,
     )?;
-    // It is not straightforward to share the OwnedFd across forks, so we 
-    // treat them as i32. We use ManuallyDrop to keep the connection open. 
+    // It is not straightforward to share the OwnedFd across forks, so we
+    // treat them as i32. We use ManuallyDrop to keep the connection open.
     let f1 = std::mem::ManuallyDrop::new(f1);
     let f2 = std::mem::ManuallyDrop::new(f2);
 
--- a/crates/libcontainer/src/process/fork.rs
+++ b/crates/libcontainer/src/process/fork.rs
@@ -1,4 +1,4 @@
-use std::{ffi::c_int, num::NonZeroUsize};
+use std::{ffi::c_int, fs::File, num::NonZeroUsize};
 
 use libc::SIGCHLD;
 use nix::{
@@ -164,12 +164,14 @@ fn clone(cb: CloneCb, flags: u64, exit_signal: Option<u64>) -> Result<Pid, Clone
     // do not use MAP_GROWSDOWN since it is not well supported.
     // Ref: https://man7.org/linux/man-pages/man2/mmap.2.html
     let child_stack = unsafe {
-        mman::mmap(
+        // Since nix = "0.27.1", `mmap()` requires a generic type `F: AsFd`.
+        // `::<File>` doesn't have any meaning because we won't use it.
+        mman::mmap::<File>(
             None,
             NonZeroUsize::new(default_stack_size).ok_or(CloneError::ZeroStackSize)?,
             mman::ProtFlags::PROT_READ | mman::ProtFlags::PROT_WRITE,
             mman::MapFlags::MAP_PRIVATE | mman::MapFlags::MAP_ANONYMOUS | mman::MapFlags::MAP_STACK,
-            -1,
+            None,
             0,
         )
         .map_err(CloneError::StackAllocation)?
--- a/crates/libcontainer/src/process/seccomp_listener.rs
+++ b/crates/libcontainer/src/process/seccomp_listener.rs
@@ -5,7 +5,7 @@ use nix::{
     unistd,
 };
 use oci_spec::runtime;
-use std::{io::IoSlice, path::Path};
+use std::{io::IoSlice, os::fd::AsRawFd, path::Path};
 
 use super::channel;
 
@@ -76,7 +76,7 @@ fn sync_seccomp_send_msg(listener_path: &Path, msg: &[u8], fd: i32) -> Result<()
         );
         SeccompListenerError::UnixOther(err)
     })?;
-    socket::connect(socket, &unix_addr).map_err(|err| {
+    socket::connect(socket.as_raw_fd(), &unix_addr).map_err(|err| {
         tracing::error!(
             ?err,
             ?listener_path,
@@ -91,15 +91,19 @@ fn sync_seccomp_send_msg(listener_path: &Path, msg: &[u8], fd: i32) -> Result<()
     let iov = [IoSlice::new(msg)];
     let fds = [fd];
     let cmsgs = socket::ControlMessage::ScmRights(&fds);
-    socket::sendmsg::<UnixAddr>(socket, &iov, &[cmsgs], socket::MsgFlags::empty(), None).map_err(
-        |err| {
-            tracing::error!(?err, "failed to write container state to seccomp listener");
-            SeccompListenerError::UnixOther(err)
-        },
-    )?;
+    socket::sendmsg::<UnixAddr>(
+        socket.as_raw_fd(),
+        &iov,
+        &[cmsgs],
+        socket::MsgFlags::empty(),
+        None,
+    )
+    .map_err(|err| {
+        tracing::error!(?err, "failed to write container state to seccomp listener");
+        SeccompListenerError::UnixOther(err)
+    })?;
     // The spec requires the listener socket to be closed immediately after sending.
-    let _ = unistd::close(socket);
-
+    drop(socket);
     Ok(())
 }
 
--- a/tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
@@ -4,7 +4,14 @@ use nix::{
     sys::socket::{self, UnixAddr},
     unistd,
 };
-use std::{io::IoSliceMut, os::unix::prelude::RawFd, path::Path};
+use std::{
+    io::IoSliceMut,
+    os::{
+        fd::{AsFd, AsRawFd},
+        unix::prelude::RawFd,
+    },
+    path::Path,
+};
 
 const DEFAULT_BUFFER_SIZE: usize = 4096;
 
@@ -24,14 +31,14 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
         None,
     )
     .context("failed to create seccomp listener socket")?;
-    socket::bind(socket, &addr).context("failed to bind to seccomp listener socket")?;
+    let socket = socket;
+    socket::bind(socket.as_raw_fd(), &addr).context("failed to bind to seccomp listener socket")?;
     // Force the backlog to be 1 so in the case of an error, only one connection
     // from clients will be waiting.
-    socket::listen(socket, 1).context("failed to listen on seccomp listener")?;
-    let conn = match socket::accept(socket) {
+    socket::listen(&socket.as_fd(), 1).context("failed to listen on seccomp listener")?;
+    let conn = match socket::accept(socket.as_raw_fd()) {
         Ok(conn) => conn,
         Err(e) => {
-            let _ = unistd::close(socket);
             bail!("failed to accept connection: {}", e);
         }
     };
@@ -47,15 +54,13 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
         Ok(msg) => msg,
         Err(e) => {
             let _ = unistd::close(conn);
-            let _ = unistd::close(socket);
             bail!("failed to receive message: {}", e);
         }
     };
 
     // We received the message correctly here, so we can now safely close the socket and connection.
     let _ = unistd::close(conn);
-    let _ = unistd::close(socket);
-
+    drop(socket);
     // We are expecting 1 SCM_RIGHTS message with 1 fd.
     let cmsg = msg
         .cmsgs()
--- a/tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
+++ b/tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
@@ -31,7 +31,7 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
         None,
     )
     .context("failed to create seccomp listener socket")?;
-    let socket = socket;
+
     socket::bind(socket.as_raw_fd(), &addr).context("failed to bind to seccomp listener socket")?;
     // Force the backlog to be 1 so in the case of an error, only one connection
     // from clients will be waiting.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1747,7 +1747,7 @@ dependencies = [
  "flate2",
  "libcgroups",
  "libcontainer",
- "nix",
+ "nix 0.27.1",
  "num_cpus",
  "oci-spec",
  "once_cell",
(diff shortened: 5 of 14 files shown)


YJDoc2:
Hey, thanks for the PR. I think there was someone else also assigned to this you should also ping them .
As there are a lot of changes, I'll need some more time to review this, probably over the weekend.

utam0k:
I'll take a look at this PR tomorrow.

crates/libcontainer/src/channel.rs
```diff
@@ -198,18 +201,20 @@ where
...
+        receiver: os_receiver.as_raw_fd(),
         phantom: PhantomData,
     };
     let sender = Sender {
-        sender: os_sender,
+        sender: os_sender.as_raw_fd(),
         phantom: PhantomData,
     };
+    std::mem::forget(os_sender);
+    std::mem::forget(os_receiver);
```
utam0k, line 212: > I use forget() because some functions in the nix library now return an OwnedFd. This helps avoid accidentally closing the file descriptor too soon by preventing its automatic drop. Maybe fully relying on OwnedFd to manage the fd's lifetime instead of manually calling close on RawFd is a better approach, but I think this could require quite a bit of code changes.

What should we do to support `OwnedFd`? I feel it is too dangerous for low-level container runtime because we must take care of the security risk.

crates/libcontainer/src/container/init_builder.rs
```diff
@@ -85,6 +87,14 @@ impl InitContainerBuilder {
...
             None
         };
+        let csocketfd = csocketfd.map(|sockfd| match sockfd {
+            Some(sockfd) => {
+                let fd = sockfd.as_raw_fd();
+                forget(sockfd);
+                fd
+            }
+            None => -1,
+        });
```
utam0k, line 97: What happens when it returns `-1`?

utam0k: Apparently, we can combine L90-L94 into L81-89.  Is there a reason why you separate them?

crates/libcontainer/src/process/fork.rs
```diff
@@ -164,12 +164,12 @@ fn clone(cb: CloneCb, flags: u64, exit_signal: Option<u64>) -> Result<Pid, Clone
     // do not use MAP_GROWSDOWN since it is not well supported.
     // Ref: https://man7.org/linux/man-pages/man2/mmap.2.html
     let child_stack = unsafe {
-        mman::mmap(
+        mman::mmap::<File>(
```
utam0k, line 167: May I ask you to write a comment that `File` doesn't have any meaning because we won't use it.

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -117,6 +118,11 @@ impl TenantContainerBuilder {
         // if socket file path is given in commandline options,
         // get file descriptors of console socket
         let csocketfd = self.setup_tty_socket(&container_dir)?;
+        let csocketfd = csocketfd.map(|sockfd| {
+            let fd = sockfd.as_raw_fd();
+            forget(sockfd);
+            fd
+        });
```
utam0k, line 125:
```suggestion
        let csocketfd = self.setup_tty_socket(&container_dir)?.map(|sockfd| {
            let fd = sockfd.as_raw_fd();
            forget(sockfd);
            fd
        });
```

crates/libcontainer/src/process/seccomp_listener.rs
```diff
@@ -91,14 +91,17 @@ fn sync_seccomp_send_msg(listener_path: &Path, msg: &[u8], fd: i32) -> Result<()
...
     let fds = [fd];
     let cmsgs = socket::ControlMessage::ScmRights(&fds);
-    socket::sendmsg::<UnixAddr>(socket, &iov, &[cmsgs], socket::MsgFlags::empty(), None).map_err(
-        |err| {
-            tracing::error!(?err, "failed to write container state to seccomp listener");
-            SeccompListenerError::UnixOther(err)
-        },
-    )?;
-    // The spec requires the listener socket to be closed immediately after sending.
-    let _ = unistd::close(socket);
```
utam0k, line 101: Why we can remove `close()`

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
```diff
@@ -47,14 +50,12 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
         Ok(msg) => msg,
         Err(e) => {
             let _ = unistd::close(conn);
-            let _ = unistd::close(socket);
```
utam0k, line 50: We can remove `unisted::close` because `OwnedFd` is automatically closed by `Drop`, can't we?

utam0k: Is there a way to check if `fd` is leaking?

utam0k (changes requested):
Thanks for your first contribution, and congrats your first OSS contribution! Awesome.
I left some comments.

crates/libcontainer/src/process/fork.rs
anti-entropy123: There's no specific reason. it's related to the change in the function signature of `nix::sys::mman::mmap()`. The current signature of the `mmap()` function is:
```rs
pub unsafe fn mmap<F: AsFd>(
    addr: Option<NonZeroUsize>, 
    length: NonZeroUsize, 
    prot: ProtFlags, 
    flags: MapFlags, 
    f: Option<F>, 
    offset: off_t) -> Result<*mut c_void>
```
If you remove `::<File>`, it will result in a compilation error:
> type annotations needed; cannot infer type of the type parameter `F` declared on the function `mmap`

Is there any other way to avoid the compilation error?

crates/libcontainer/src/container/init_builder.rs
anti-entropy123: > What happens when it returns `-1`?

It doesn't seem to make any difference compared to the past.
The previous `setup_console_socket()` would return `Ok(-1)` under certain conditions. Now, under the same conditions, it returns `Ok(None)`.

anti-entropy123: > Apparently, we can combine L90-L94 into L81-89. Is there a reason why you separate them?

There's nothing special. I can make the modification as you suggested.

crates/libcontainer/src/process/seccomp_listener.rs
anti-entropy123: I think `OwnedFd` will automatically call `libc::close()` when it goes out of function body.

crates/libcontainer/src/channel.rs
anti-entropy123: For `Sender` and `Receiver`, the problem is relatively easy. I noticed that `Sender::close()` will closes the fd.

I don't have a solid plan for fully supporting `OwnedFd`. I can only say that I've tried to make my PR as compatible as possible with the previous implementation that used `rawfd`, and reuse the code that previously managed file descriptors.

The main difficulty right now is that if we changes entirely to `OwnedFd`, the `console_socket` in `struct ContainerArgs` (crates/libcontainer/src/process/args.rs:20) won't satisfy the `Clone trait`. Maybe need to introduce `Rc<T>`.

anti-entropy123: Should I avoid `forget()` in this PR, or address it in a separate issue? I'd like to hear your opinion.

crates/libcontainer/src/process/fork.rs
utam0k: I understood it, but it's unclear to the reader of this code. May I ask you to write a comment about it?

crates/libcontainer/src/process/seccomp_listener.rs
utam0k: > The spec requires the listener socket to be closed immediately after sending.

We need to close it as soon as possible.

crates/libcontainer/src/channel.rs
utam0k: > Should I avoid forget() in this PR, or address it in a separate issue? I'd like to hear your opinion.

Thanks for your detailed investigation. This problem relates to the security issue. So I can't merge it before removing `forget()`.

crates/libcontainer/src/process/seccomp_listener.rs
anti-entropy123: L106 is the return statement, so `sendmsg()` will be immediately followed by the drop of the socket. 
However, considering that the spec emphasizes this point, I believe it's necessary to explicitly close the socket. I plan to call `drop(socket)` on L105. Do you think that's okay?

yihuaf:
I will take a look this week.

crates/libcontainer/src/channel.rs
anti-entropy123: I agree with this cautious decision. I have some ideas for removing `forget()`, but it will take some time to implement.

utam0k:
@yihuaf Thanks for taking your time 🙏 I want you to review it, too.

crates/libcontainer/src/channel.rs
utam0k: Thanks for understanding 🙏 

> but it will take some time to implement.

NP 👍 Let's give it a try!

anti-entropy123:
Hi, I've resubmitted the new code, and I've removed all `forget()` (except for one in a test function, crates/libcontainer/src/tty.rs:211 ). 
In my downstream repository, all CI tests have already passed. 
Please review it again when you have time. Thank you very much. @utam0k

utam0k:
@anti-entropy123 May I ask you to resolve the conflict? 🙏

crates/libcontainer/src/channel.rs
utam0k, line 1: Use thiserror instead of `unwrap`.

anti-entropy123:
> May I ask you to resolve the conflict?

:ok_hand:  Done. @utam0k

utam0k:
@anti-entropy123 Thanks. Please tell me why we can't use `try_clone()`?
https://doc.rust-lang.org/nightly/std/os/fd/struct.OwnedFd.html#method.try_clone

anti-entropy123:
> Please tell me why we can't use `try_clone()`? 

Hi, thanks for your advice.

Honestly, I didn't initially think of trying `try_clone()`. But I think using `try_clone()` might cause the file descriptor to be accidentally closed.

For example, in the `container_main_process()` (crates\libcontainer\src\process\container_main_process.rs:49), the channels are copied into closures on the heap. For the parent process, this closure will be destructed after `clone()` finishes, causing all the channels within it to be closed.

Is my understanding correct?

utam0k:
@anti-entropy123 I assume that `try_clone()` creates a new file descriptor with `DUPFD`. so there is no problem I think...

anti-entropy123:
> I assume that `try_clone()` creates a new file descriptor with `DUPFD`. so there is no problem I think...

@utam0k  Indeed, you are right. So, do you think we shouldn't use `Rc`?  :thinking: 

Although you might already be aware, I'd like to mention that `try_clone()` might also result in the need to use `close()` to cleanup those old FDs after the child process is created.

utam0k:
@anti-entropy123 I think Rc is one of a complex category. I would avoid using it if possible.

utam0k:
I wish I had noticed that first. I'm sorry it took me so long to notice.

anti-entropy123:
> I would avoid using it if possible.

@utam0k  Okey, I will update the relevant code.

YJDoc2:
Hey, I took a look at the changes and the conversation that is happening, and 
- I don't think at this point youki can completely switch from RawFd to OwnedFd, because at several places we are opening sockets or connections, and then passing them around, or sharing them across forks, for which (as of now) we need them to be treated as i32
- Given that till now we were using RawFds, we have taken care to make sure everything that was opened is closed properly. I think in the upgrade to nix 0.27.1, we should only update types so that they compile and propagate properly, and while  doing that we can see where RawFds are being used, and check that they are properly being closed. Then in a separate PR, we can do the change from RawFd to (maybe) OwnedFd , and see how complex that becomes.
- I agree that using Rc/Arc would not be a good idea, they have their own handling complexity, and I would prefer not to use ref-counted things in this context as much as possible due to potential overhead they have.
- I think that instead of using `std::mem::forget` , using `std::mem::ManuallyDrop` would be a better idea, as that fits here better semantically. https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html
- I have opened a PR on my fork https://github.com/YJDoc2/youki/pull/260 , where I have only done type changes and used ManuallyDrop, so you can see how that looks for reference.

@utam0k :eyes:

utam0k:
@anti-entropy123 Sorry for confusing you, but I agree with @YJDoc2 . We should take care of `OwnedFd` in another PR if possible.

anti-entropy123:
> We should take care of OwnedFd in another PR if possible.

I understand. I referenced @YJDoc2 's fork,  and using `ManuallyDrop` seems to be a better approach.
Please re-review again. @utam0k

utam0k (approved):
@anti-entropy123 May I ask you to create the issue to fully support `OwnedFd` in another PR

crates/libcontainer/src/channel.rs
```diff
@@ -210,10 +210,14 @@ where
...
-    Ok(socket::socketpair(
+    let (f1, f2) = socket::socketpair(
         socket::AddressFamily::Unix,
         socket::SockType::SeqPacket,
         None,
         socket::SockFlag::SOCK_CLOEXEC,
-    )?)
+    )?;
+    let f1 = std::mem::ManuallyDrop::new(f1);
+    let f2 = std::mem::ManuallyDrop::new(f2);
```
utam0k, line 220: nits: Could you leave a comment on why we need to ManuallyDrop?

anti-entropy123:
> May I ask you to create the issue to fully support `OwnedFd` in another PR

No problem. :ok_hand: @utam0k

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
```diff
@@ -24,14 +31,14 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
         None,
     )
     .context("failed to create seccomp listener socket")?;
-    socket::bind(socket, &addr).context("failed to bind to seccomp listener socket")?;
+    let socket = socket;
```
YJDoc2, line 34: Why is this assignment `let socket = socket;` necessary? Something left from previous iterations?

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
```diff
@@ -24,14 +31,14 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
...
     // from clients will be waiting.
-    socket::listen(socket, 1).context("failed to listen on seccomp listener")?;
-    let conn = match socket::accept(socket) {
+    socket::listen(&socket.as_fd(), 1).context("failed to listen on seccomp listener")?;
+    let conn = match socket::accept(socket.as_raw_fd()) {
         Ok(conn) => conn,
         Err(e) => {
-            let _ = unistd::close(socket);
             bail!("failed to accept connection: {}", e);
         }
```
YJDoc2, line 43: We should also add a `drop(socket)` here, now that we have removed the manual close? Or we are relying on the destructor to be run when function returns via `bail!` ?

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
```diff
@@ -47,15 +54,13 @@ pub fn recv_seccomp_listener(seccomp_listener: &Path) -> SeccompAgentResult {
...
-            let _ = unistd::close(socket);
             bail!("failed to receive message: {}", e);
         }
     };
 
     // We received the message correctly here, so we can now safely close the socket and connection.
     let _ = unistd::close(conn);
-    let _ = unistd::close(socket);
-
+    drop(socket);
```
YJDoc2, line 63: Can we add a message for why we are manually dropping the socket here in comment, so we don't accidentally remove it in future?

YJDoc2 (commented):
Hey, some minor comments, but nothing blocking. So if everything else is ok, we can go ahead and merge, and address these later

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
anti-entropy123: I think L61 has already explained the reason, so is it necessary to add more comments?

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
anti-entropy123: My mistake. Initially, there was `let socket = std::mem::ManuallyDrop::new(socket);`, and later, I removed `ManuallyDrop::new`. 
I will fix here. Thanks for your review!

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
YJDoc2: No, I think I got confused with something else. L61 Should be good enough :+1:

tests/rust-integration-tests/integration_test/src/tests/seccomp_notify/seccomp_agent.rs
anti-entropy123: I think we can rely on `bail!` to close `socket`, because `bail!` is equivalent to `return`. Moreover, there will likely not be a significant increase in code, so manual `drop()` is not necessary.

YJDoc2: fair enough :+1:

YJDoc2 (approved):
LGTM :+1:
nits by utam0k have been implemented, so going ahead and merging this.

YJDoc2:
@anti-entropy123 thank you for following on this and implementing :pray: Apologies for the back-and-forth that was done from our side, and you had to change the PR :sweat: 
Congrats on your first contribution to Youki ! :tada: :tada:

anti-entropy123:
@YJDoc2  You're welcome, and I greatly appreciate your reviews and suggestions. I've learned a lot from them 👍.  I'm looking forward to making more contributions to Youki ! 

By the way, I have a question. I noticed that there are many individual commits in the main branch after merging this PR. Why not use 'Squash and merge'? Or should I aim to keep the number of commits as small as possible?

YJDoc2:
> I'm looking forward to making more contributions to Youki !

:raised_hands: 

>By the way, I have a question. I noticed that there are many individual commits in the main branch after merging this PR. Why not use 'Squash and merge'? Or should I aim to keep the number of commits as small as possible?

So there is no hard-and-fast rule regarding this. We usually try to keep the main branch clean, and keep the commits in PR, so we can refer back to specific commit in history if needed. However, if there are lots of commits, or many commits which are just fixing lints / minor changes, we ask dev to rebase and squash, or merge the PR in squash-and-merge fashion. For this particular PR, 5 commits are not too much, and  even though there are couple of lint/comment fix commits, I felt that would be fine.

Overall, if possible one should aim to keep number of commits small, so the main branch stays clean ; but if it is required to have large number of commits (eg step-wise changes, separating each in a commit would make review easier) then one can request on PR to sqash as well (or we'd do so when we merge).

PR #2370, opened 2023-09-19
Feat/podman rootless

YJDoc2:
Ref : https://github.com/containers/youki/issues/2208 , https://github.com/containers/youki/issues/719 , https://github.com/containers/youki/pull/1171

This switches from dbus-rs to natively implemented dbus connection implementation, in order to make rootless invocation with podman work. ~This is almost done but still few final touches are remaining, so marked as WIP.~

This PR can be split into 4 parts :

- dbus_native implementation : commits https://github.com/containers/youki/pull/2370/commits/b67833777e8d2169905a507434ceef1d5cf5eae8 , https://github.com/containers/youki/pull/2370/commits/e3c6f57d527c6658338590f8976710794be70280 , https://github.com/containers/youki/pull/2370/commits/c36983a3426f394644fd5efeccfd60c7036d45f5
- Changing rest of stuff (tests, types etc) to use dbus_native : commits https://github.com/containers/youki/pull/2370/commits/982ec4b24bb2efe194b03833cb5603caeb475428 , https://github.com/containers/youki/pull/2370/commits/82add24a9acbf2bb9746d0641c3db767f7cf9373 , https://github.com/containers/youki/pull/2370/commits/1e88851d6d47270ef929e4a98fddbcf5fa705b3a
- Implementing the original dbus interface code into dbus native : commits https://github.com/containers/youki/pull/2370/commits/e45d4837abc394784d3d0b079fc42f73e8780d74 , https://github.com/containers/youki/pull/2370/commits/a4169bae34e742ea8fcd1833d6762fdf71b73b77 , https://github.com/containers/youki/pull/2370/commits/66fea6f190ea5bf0b6cc50df8c56667e2edcc606
- misc and documentation : commits https://github.com/containers/youki/pull/2370/commits/8f67d300578f2b830e09d1d7d065f44e8f903165 , https://github.com/containers/youki/pull/2370/commits/a6505143fd413564e9c1841d70334647a209205c

Apologies for such large PR, but to validate that the code works, I have to completely switch over to dbus_native, and test. If required, I can split the PR into multiple ones according to above splits, but then the individual PRs may not be testable.

There are several places where I'm not particularly happy with my implementation, so any and all comments, suggestions and feedback is welcome to improve this!

Thank you :)


crates/libcgroups/src/systemd/cpuset.rs
```diff
@@ -44,20 +44,28 @@ impl CpuSet {
...
 
         if let Some(mems) = cpu.mems() {
-            let mems_mask = to_bitmask(mems).map_err(SystemdCpuSetError::MemoryNodesBitmask)?;
-            properties.insert(ALLOWED_NODES, Box::new(mems_mask));
+            let mems_mask: Vec<_> = to_bitmask(mems)
+                .map_err(SystemdCpuSetError::MemoryNodesBitmask)?
+                .into_iter()
+                .map(|v| v as u64)
+                .collect();
+            properties.insert(ALLOWED_NODES, Variant::ArrayU64(mems_mask));
```
YJDoc2, line 68: In tests as well as other uses, the original code expected u64, so changed this to type u64 as well, instead of casting there

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -0,0 +1,555 @@
...
+/// which message is for which request etc etc
+// Client is a wrapper providing higher level API and abatraction around dbus.
+// For more information see https://www.freedesktop.org/wiki/Software/systemd/dbus/
+pub struct DbusConnection {
+    /// Is the socket system level or session specific
+    system: bool,
+    /// socket fd
+    socket: i32,
+    /// name id assigned by dbus for the connection
+    id: Option<String>,
```
YJDoc2, line 27: it is not necessary to send ID, but I have added it as sending it is a better practice

crates/libcgroups/src/systemd/dbus_native/message.rs
YJDoc2, line 1: Unit tests in this test the combined functionality of message serialization and individual type serialization. The "original" message in all tests are taken from zbus calls, so we know that those are valid messages.

YJDoc2:
With these changes, the failing tests in podman rootless are 136, 4 more that current main (132). I'll check and fix them in a separate PR, this only fixes the running with podman issues.
The binary size is also slightly large, but not much, which I think is acceptable.

The primary test for validating is the following command should run with and without root permissions, both
```sh
podman create --runtime $PWD/youki --name test hello-world
podman start -ia test # this should print the hello world message on terminal
podman rm test

podman run -it fedora /bin/bash # verify you can run commands inside the container
```

docs/src/user/basic_setup.md
```diff
@@ -20,7 +20,6 @@ To compile and run, Youki itself depends on some underlying libraries being inst
 $ sudo apt-get install    \
       pkg-config          \
       libsystemd-dev      \
-      libdbus-glib-1-dev  \
```
utam0k, line 23: 😍

utam0k: @YJDoc2 May I ask you to write a migration guide to let the users know that `libdbus` can be deleted?

docs/src/user/dbus-native.md
```diff
@@ -0,0 +1,10 @@
+# Dbus Native
```
utam0k, line 1: The `Dbus` module is one of the libcgroups modules. So how about it be included in libcgroups section?

utam0k (commented):
Awesome.
Especially, I think it is good to have a good unit tests Why not divert some unit tests and add doc test?

.github/workflows/e2e.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           go-version: '1.20'
           cache: true
       - run: sudo apt-get -y update
-      - run: sudo apt-get install -y pkg-config libsystemd-dev libdbus-glib-1-dev libelf-dev libseccomp-dev btrfs-progs libbtrfs-dev
```
yihuaf, line 45: Hooray!

yihuaf (approved):

yihuaf:
This is technically a breaking change. I would include this change in a minor version bump.

YJDoc2:
> This is technically a breaking change. I would include this change in a minor version bump.

Yes, I will be adding a migration guide for this as @utam0k has suggested, just haven't  bumped the version in this PR, as we usually do that in the release PRs when we do a new release.

docs/src/user/dbus-native.md
YJDoc2: Yep, that makes more sense. I'll  move this accordingly , Thanks!

docs/src/user/basic_setup.md
YJDoc2: Yes, I'll update the migration guide with changes in this commit.

YJDoc2:
>  Especially, I think it is good to have a good unit tests Why not divert some unit tests and add doc test?

@utam0k , I'm not sure what you mean by this? Do you want me to change some of the current unit tests that I have added to doc tests instead, or do you want to keep unit tests as they are, and add more tests as doc tests?

utam0k:
> > Especially, I think it is good to have a good unit tests Why not divert some unit tests and add doc test?
> 
> @utam0k , I'm not sure what you mean by this? Do you want me to change some of the current unit tests that I have added to doc tests instead, or do you want to keep unit tests as they are, and add more tests as doc tests?

May I ask you to use [documentation tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) in simple util functions? For example:
https://github.com/containers/youki/pull/2370/files#diff-9347c2507177cda90306312df09422eccd4d3a259bc837ecb73a78c963a5c0f5R81

YJDoc2:
Hey @utam0k apologies for the delay. I have updated the migration guide and docs as you commented.
I tried converting some utils tests to doctests, but as the dbus_native module is private, they cannot be imported in the doctest.

> When compiling a crate for use in doctests (with --test option), rustdoc will set #[cfg(doctest)]. Note that they will still link against only the public items of your crate; if you need to test private items, you need to write a unit test.

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests

So for now I have kept the unit tests as they were.
Please take a look : )

docs/src/user/dbus-native.md
YJDoc2: done

docs/src/user/basic_setup.md
YJDoc2: done

utam0k (approved):
May I ask you to add e2e test for rootless podmain in another PR if possible?

utam0k:
@orimanabu May I ask you to give it a try?

YJDoc2:
> May I ask you to add e2e test for rootless podmain in another PR if possible?

Yes, I'll do that :+1:

PR #2381, opened 2023-09-23
Update the developer documentation for the e2e tests.

utam0k:


docs/src/developer/e2e/runtimetest.md
```diff
@@ -0,0 +1,50 @@
...
+./youki: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=...., for GNU/Linux 3.2.0, with debug_info, not stripped`
+```
+
+This output indicates that the binary is dynamically linked, thus cannot be run inside the container process
+
+```console
+./runtimetest: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=...., for GNU/Linux 3.2.0, with debug_info, not stripped
+```
+
+This output indicates that the binary is statically linked, and can be run inside the container process
```
YJDoc2, line 44: Can we have the lines `This output indicates...` before their respective console examples? This way can be a bit confusing.

YJDoc2 (approved):
overall lgtm, a minor nitpick

YJDoc2:
@utam0k may I ask you to do the naming changes for `oci-tests` -> `test-oci` and similar, that are done after https://github.com/containers/youki/pull/2330 merge?

docs/src/developer/e2e/runtimetest.md
utam0k: Thanks for your comment, But it is only copied from the original. We have to take care of it in another PR.
https://containers.github.io/youki/developer/runtimetest.html

YJDoc2: ohk, we can do it separately then :+1:

PR #2382, opened 2023-09-24
Add the documentation for debugging

utam0k:


yihuaf (approved):

yihuaf:
Just to add that another way is to send the youki logs to systemd. This will let systemd capture all the logs from different processes in the case that stdout fails. `youki` as a binary supports this out of box, but `libcontainer` needs caller to configure this correctly,

YJDoc2 (approved):
lgtm, and as yihuaf said, it'd good to add the systemd info as well

utam0k:
> Just to add that another way is to send the youki logs to systemd. This will let systemd capture all the logs from different processes in the case that stdout fails. `youki` as a binary supports this out of box, but `libcontainer` needs caller to configure this correctly.

@yihuaf 
You have more knowledge about it than me around using systemd info so could you add the documentation? Of course, I'm looking forward to the doc you wrote to know another way to debug.

PR #2406, opened 2023-10-05
Add initial tests for rootless podman

YJDoc2:
This adds a simple script for validating rootless podman works with youki, now that #2370 is merged. This is not a "proper" test, just a basic bash script which will error if any of the steps error out ; but I'm planning out a small refactoring of rust integration tests, and these tests will be easier to implement in rust once that is done. So  I'll be porting the script to proper crate once that is done. This, in the meantime, will make sure we don't accidentally regress the rootless support.


YJDoc2:
@utam0k May I ask you to take a quick look at this, so I can continue with the actual test refactor?

utam0k (approved):

PR #2411, opened 2023-10-07
feat: allow customize cgroup root path

fengxsong:
https://github.com/containers/youki/issues/2392

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -9,7 +9,6 @@ use std::{
 use nix::{
     sys::statfs::{statfs, CGROUP2_SUPER_MAGIC, TMPFS_MAGIC},
     unistd::Pid,
-    NixPath,
 };
 use oci_spec::runtime::LinuxResources;
 #[cfg(any(feature = "cgroupsv2_devices", feature = "v1"))]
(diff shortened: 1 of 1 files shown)


fengxsong:
@utam0k PTAL

utam0k:
@fengxsong May I ask you to make it ready?

crates/libcgroups/src/common.rs
```diff
@@ -335,10 +339,17 @@ pub struct CgroupConfig {
...
+pub fn create_cgroup_manager_with_root(
+    root_path: &Path,
     config: CgroupConfig,
 ) -> Result<AnyCgroupManager, CreateCgroupSetupError> {
-    let cgroup_setup = get_cgroup_setup().map_err(|err| match err {
+    let root = if root_path.is_empty() {
+        Path::new(DEFAULT_CGROUP_ROOT)
+    } else {
+        root_path
+    };
```
utam0k, line 350: How about using `Option`?
https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default

utam0k (changes requested):
Thanks for your PR. I left a comment.

crates/libcgroups/src/common.rs
fengxsong: good idea. I've changed it.

crates/libcgroups/src/common.rs
```diff
@@ -335,10 +338,16 @@ pub struct CgroupConfig {
     pub container_name: String,
 }
 
-pub fn create_cgroup_manager(
+pub fn create_cgroup_manager_with_root(
```
utam0k, line 341: Please comment on the behavior of passing None

fengxsong: done

utam0k (approved):
Thanks for your contribution 😍

PR #2418, opened 2023-10-09
[WIP] Put clone3 behind a feature

YJDoc2:
DNM, just a test PR for https://github.com/containerd/runwasi/issues/347
This puts clone3 behind a feature flag, which is off by default.

### DIFF
--- a/crates/libcontainer/Cargo.toml
+++ b/crates/libcontainer/Cargo.toml
@@ -18,6 +18,7 @@ systemd = ["libcgroups/systemd", "v2"]
 v2 = ["libcgroups/v2"]
 v1 = ["libcgroups/v1"]
 cgroupsv2_devices = ["libcgroups/cgroupsv2_devices"]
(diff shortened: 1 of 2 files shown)


jprendes:
`process::fork::test::test_clone_fallback` is failing, which is expected.
`tty::tests::test_setup_console` is also failing, which (to me) is not expected.

jprendes:
The other failure is on `process_rlimits/process_rlimits.t`.
When doing `mmap` to allocate the stack, it returns `ENOMEM`.

Reducing the allocated stack size to 4M makes the test pass for me, but admitedly 4M is just a made up number.

jprendes:
''process_rlimits` sets the stack limits to
```go
g.AddProcessRlimits("RLIMIT_STACK", 10*gigaBytes, 9*gigaBytes)
```

so, yeah, I see why `mmap` doesn't like it.

PR #2419, opened 2023-10-09
Add a `no-clone3` feature to disable `clone3`

jprendes:
This PR adds a `no-clone3` feature that disables the use of `clone3` in `libcontainer`.

This can be used to counteract the undefined behaviour introduced by using a raw `sysall` for `clone3`, which leads to problem in `musl` and on `glibc` when the main process uses threads.

See https://github.com/containerd/runwasi/issues/347

### DIFF
--- a/crates/libcontainer/Cargo.toml
+++ b/crates/libcontainer/Cargo.toml
@@ -18,6 +18,7 @@ systemd = ["libcgroups/systemd", "v2"]
 v2 = ["libcgroups/v2"]
 v1 = ["libcgroups/v1"]
 cgroupsv2_devices = ["libcgroups/cgroupsv2_devices"]
+no-clone3 = []
 
 [dependencies]
 bitflags = "2.4.0"
--- a/crates/libcontainer/src/process/fork.rs
+++ b/crates/libcontainer/src/process/fork.rs
@@ -25,15 +25,13 @@ pub enum CloneError {
 }
 
 /// The callback function used in clone system call. The return value is i32
-/// which is consistent with C functions return code. The trait has to be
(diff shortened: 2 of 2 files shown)


YJDoc2:
Hey, actually I was also trying the same here https://github.com/containers/youki/pull/2418 which either takes clone3 or clone depending on feature without any fallback. Its failing some tests, so need to see that there.

YJDoc2:
Hey, there's one conflict due to recent merge in master, can you rebase? Also I saw your comment on my PR regarding failing test, I'll try to get to that, or if this is done first/ has more preferable approach, we can go ahead with this.

Also a nit, personal preference : can the feature name be "positive" i.e. instead of `no-clone3` , can we have `clone3`, and invert the checks instead?

jprendes:
> can the feature name be "positive" i.e. instead of no-clone3 , can we have clone3, and invert the checks instead?

I thought about that and my concern is that when we request `no-clone3`, we really don't want `clone3`.
Due to how cargo resolves features (the union of all features requested), a feature can be enabled by another crate even if your crate didn't enable it. The same doesn't apply to disabling features (you can't disable features other than default).
Naming it `clone3` means that some other crate can accidentally enable it.
Naming it `no-clone3` means that once we enable it, it can't be accidentally disabled.

jprendes:
> can you rebase?

Rebased :-)

jprendes:
Converting to draft as this doesn't fix the sporadic failures in https://github.com/containerd/runwasi/issues/347

utam0k:
@jprendes Can we close this PR?

PR #2420, opened 2023-10-09
Use static build of wasmedge

jprendes:
WasmEdge supports static linking, which avoid having to use `LD_LIBRARY_PATH`.
This PR enables that feature.


utam0k (approved):
👍

PR #2425, opened 2023-10-10
Use raw syscalls to avoid sporadic hangs

jprendes:
The `setgroups`/`setresgid`/`setresuid` functions in `nix`/`libc` are not safe to call after `clone`/`clone3`.
These function use a blocking mechanism that synchronises across all threads in the process.
To do this `libc` (`glibc`/`musl`) needs to do some bookkeeping, which breaks on the new process after `clone`.
See the discussion in https://github.com/containerd/runwasi/issues/347 for more details.

This PR worksaround that issue by using raw syscalls for `setgroups`/`setresgid`/`setresuid`.

This is generally not safe, as the raw syscalls only affect the calling thread and not the whole process.
It is safe to do so in this case because at the point when they are called in libcontainer the init process only has one thread.

### DIFF
--- a/crates/libcontainer/src/syscall/linux.rs
+++ b/crates/libcontainer/src/syscall/linux.rs
@@ -7,8 +7,7 @@ use nix::{
     mount::{mount, umount2, MntFlags, MsFlags},
     sched::{unshare, CloneFlags},
     sys::stat::{mknod, Mode, SFlag},
-    unistd,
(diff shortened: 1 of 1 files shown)


utam0k:
@jprendes I appreciate your investigation and PR. Have you checked if this patch worked fine for runwasi?

utam0k (approved):
LGTM

PR #2486, opened 2023-10-29
Name the test tools `contest`

utam0k:
Until now our testing tools have not had names, but we will give them names to make them easier to understand.


YJDoc2:
Hey, I'll take a look at this around the weekend. In the meantime, May I ask you to check if we need any changes in our documentation for this?

.gitmodules
```diff
@@ -1,5 +1,3 @@
-[submodule "integration_tests/oci-runtime-tests"]
+[submodule "tests/oci-runtime-tests/src/github.com/opencontainers/runtime-tools"]
 	path = tests/oci-runtime-tests/src/github.com/opencontainers/runtime-tools
 	url = https://github.com/opencontainers/runtime-tools.git
-	ignore = dirty
```
YJDoc2, line 4: I think we should still keep the `ignore = dirty` part ? Is it not needed?

YJDoc2 (commented):
I feel moving the test_framework outside the contest, i.e. under the path `tests/test_framework` would be better, as then we can re-use it for other tests such as podman rootless and wasm tests when we add them.
---

@utam0k Overall looks nice, just have few comments, please take a look at them.

YJDoc2:
@utam0k :eyes:

utam0k:
Sorry I don't find out the time to address it 🙏

YJDoc2:
@utam0k , overall looks ok, but can you take a look at the remaining comments above when you get time? Thanks!

.gitmodules
utam0k: I have already addressed it.

utam0k:
@YJDoc2 Thanks for your quick review 🙏  Is there anything I can address?

YJDoc2 (approved):

PR #2495, opened 2023-10-31
add schedule entity

lengrongfu:
issue https://github.com/containers/youki/issues/1706


utam0k:
@lengrongfu May I ask you to add the e2e test?
https://containers.github.io/youki/developer/e2e/rust_oci_test.html

YJDoc2:
Hey @lengrongfu , if there is any issue with the e2e test addition, or any other problems that you might be facing, feel free to ping me!

YJDoc2:
Hey @lengrongfu , are you still working on this? In case you're busy with something else, and cannot add the tests right now, can you post a comment with what the changes are and how one would approach testing them? That way we can either make and push tests to this branch ourselves, or merge this for now, and get another PR for the test. Let me know, and thanks for the work you have already done :)

lengrongfu:
> Hey @lengrongfu , are you still working on this? In case you're busy with something else, and cannot add the tests right now, can you post a comment with what the changes are and how one would approach testing them? That way we can either make and push tests to this branch ourselves, or merge this for now, and get another PR for the test. Let me know, and thanks for the work you have already done :)

I'm so sorry.

YJDoc2:
> I'm so sorry. please take look.

Hey, no worries, thanks for getting back :+1:

A couple of quick questions :

I don't think the comment `////////// ANCHOR: create_spec policy: u32` is needed in the test, can it be removed?

Also in the integration tests, we are only checking that the containers are getting created / running successfully. Is there a way to check specifically that the policy we specified is getting applied correctly? That would be a better test.

Another thing is that this adds dependency of nc to do a raw syscall ; does nix or libc libraries does not support this? And if that is the case, can you add a comment in the code, with` TODO` saying to replace it when those library supports those?

Thanks!

lengrongfu:
> Also in the integration tests, we are only checking that the containers are getting created / running successfully. Is there a way to check specifically that the policy we specified is getting applied correctly? That would be a better test.

By calling the `sched_getattr` function, you can obtain the scheduling strategy of the current process and compare it with the expected value. What do you think of this solution?

> does nix or libc libraries does not support this?

What I know is that nix does not have it. I need to check libc.

YJDoc2:
> By calling the sched_getattr function, we can obtain the scheduling strategy of the current process and compare it with the expected value. What do you think of this solution?

This seems like a good idea. This will be done from inside the container right (i.e. in runtimetests) ? We can set the sched strategy in the integration tests, maybe with some specific values, and check that inside the container.

>What I know is that nix does not have it. I need to check libc.

:+1:

lengrongfu:
i am in run `just validate-rust-oci-runc` command, how to watch `container_init_process.rs` file print log, can you help me?
@YJDoc2

YJDoc2:
> i am in run just validate-rust-oci-runc command, how to watch container_init_process.rs file print log, can you help me?

Hey, the `validate-rust-oci-runc` runs the rust oci test code (i.e. the ones we are changing here) on `runc` , not youki. So the `container_init_process` from youki is not involved at all. If the tests are failing, then is is possible that something we are expecting in the test is not supported / implemented differently by runc. What issue are you facing?

YJDoc2:
Can you once check if sched policy is supported by runc yet or not? If it is not, then the issue is not with your code.

Also, don't mind the pending CI checks, that is due to changes from some other PR.

lengrongfu:
runc is having this function，but not sure it is valid. i will try again.

YJDoc2:
@lengrongfu Hey! Let me know if you need any help with this :+1:
Also please rebase on the main to make the CI pass properly!

lengrongfu:
> @lengrongfu Hey! Let me know if you need any help with this 👍 Also please rebase on the main to make the CI pass properly!

current only runc integration_test can't run success.

YJDoc2:
> current only runc integration_test can't run success.

Thanks for rebasing and updating! Did you check if the runc release we are using in CI supports the sched entity as expected? If that is the case, I'll try to take a more detailed look on why this might be failing.

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -741,6 +750,59 @@ fn set_io_priority(syscall: &dyn Syscall, io_priority_op: &Option<LinuxIOPriorit
...
+            sched_priority: sc.priority().unwrap_or(0) as u32,
+            sched_runtime: sc.runtime().unwrap_or(0),
+            sched_deadline: sc.deadline().unwrap_or(0),
+            sched_period: sc.period().unwrap_or(0),
+            sched_util_min: 0,
+            sched_util_max: 0,
+        };
+        // TODO when nix or libc support this function, replace nx crates.
+        unsafe {
+            let result = nc::sched_setattr(1, &mut a, 0);
```
YJDoc2, line 793: Hey, I was looking through the runc implementation for this, [here](https://github.com/opencontainers/runc/pull/4025/files#diff-84fffb149f8b86c903e09613cbebad12de596f4139d54e288388b1458e08b8ec) , and one difference I see is that in their setattr call, the first param, i.e. pid is set to 0, whereas here we are setting it to 1. Maybe that is causing some differences, resulting in failure. Any idea why the 0 v/s 1 difference is there?

tests/runtimetest/src/tests.rs
```diff
@@ -307,3 +310,52 @@ pub fn validate_sysctl(spec: &Spec) {
...
+        sched_nice: 0,
+        sched_priority: 0,
+        sched_runtime: 0,
+        sched_deadline: 0,
+        sched_period: 0,
+        sched_util_min: 0,
+        sched_util_max: 0,
+    };
+    unsafe {
+        match nc::sched_getattr(0, &mut get_sched_attr, size, 0) {
```
YJDoc2, line 332: Also, here we are checking the attr of pid 0, while in youki code, we are doing setattr for pid 1 ; is that correct?

YJDoc2:
@lengrongfu I checked this, and I think the issue is even though the sched support PR is merged in runc, its still not in any release. I tested this with main branch of runc, and the tests are passing with it.
I'll discuss and decide what should be done in such case, but for now can you take a look at my previous comments regarding the pid arg to set/get attr calls?
Thanks!

crates/libcontainer/src/process/container_init_process.rs
lengrongfu: Here it is 0, it was 0 at first, but the runc test failed, so I tried to change this parameter to 1, but forgot to change it back.

YJDoc2:
@lengrongfu may I ask you to rebase this to current master, and for the failing runc test, for now do this :
keep everything the same, just comment out the addition of test in the main.rs . This will skip the test for now, but I'll do a fix in a separate PR to make it work for both.
Thanks a lot :pray:

tests/contest/contest/src/main.rs
```diff
@@ -123,6 +125,7 @@ fn main() -> Result<()> {
     tm.add_test_group(Box::new(mounts_recursive));
     tm.add_test_group(Box::new(intel_rdt));
     tm.add_test_group(Box::new(sysctl));
+    tm.add_test_group(Box::new(scheduler));
```
YJDoc2, line 128: Just comment out this line (don't remove) , and if the linter complains, add the ignore lint annotation to the place where it is complaining.

tests/contest/runtimetest/src/main.rs
```diff
@@ -35,6 +35,8 @@ fn main() {
         "mounts_recursive" => tests::validate_mounts_recursive(&spec),
         "seccomp" => tests::validate_seccomp(&spec),
         "sysctl" => tests::validate_sysctl(&spec),
+        // "scheduler_policy_other" => tests::validate_scheduler_policy(&spec),
```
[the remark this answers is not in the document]
YJDoc2: No you can keep runtime test, as unless we provide "scheduler_policy_other" arg, it won't execute right?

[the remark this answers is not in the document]
lengrongfu: ok

lengrongfu:
@YJDoc2 Do you know why this ci error? https://github.com/containers/youki/actions/runs/7538727425/job/20519819284?pr=2495
```
Run just runtimetest rust-oci-tests-bin
error: Justfile does not contain recipe `rust-oci-tests-bin`.
Error: Process completed with exit code 1.
```

YJDoc2:
>Do you know why this ci error?

We recently changed the file structure and the test tool name in a PR. If you rebased on recent main, you might have missed that change in it. Make sure that 1. just file is correctly picked from the main ; and 2. the test you added here is in correct dir after the rebase.

YJDoc2:
Also,  you can ref https://github.com/containers/youki/pull/1544/commits/16dedc2cfebba3b52101965430704c33ed28d6a2 for commenting out the test I mentioned

YJDoc2:
@lengrongfu , apologies, the failing CI was an issue from our side, we missed a change in that PR. I have pused a fix, can you merge/rebase master and see :sweat:

lengrongfu:
maybe need wait this pr merge https://github.com/containers/youki/pull/2615

YJDoc2:
Hey @lengrongfu , I updated the nice value checking condition, similar to how you did it on [runc](https://github.com/opencontainers/runc/pull/4151/files) . I'll wait for a bit and merge this after sometime, as otherwise this is ok. In case you think there is some mistake, please let me know ; even after merge, so we can correct it accordingly. Thanks!

YJDoc2 (approved):
lgtm :+1:

PR #2496, opened 2023-11-01
Grouping patch updates in dependabot.

utam0k:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


YJDoc2 (approved):

YJDoc2:
@utam0k It seems that `update-type` strictly requires an array, so check is failing on master : https://github.com/containers/youki/runs/18288323874

PR #2497, opened 2023-11-01
Specify the protobuf crate because of the rust-criu crate

utam0k:
Workaround: https://github.com/checkpoint-restore/rust-criu/issues/19

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1958,6 +1958,7 @@ dependencies = [
  "once_cell",
  "prctl",
  "procfs",
+ "protobuf",
  "quickcheck",
  "rand",
  "regex",
--- a/crates/libcontainer/Cargo.toml
+++ b/crates/libcontainer/Cargo.toml
@@ -48,6 +48,7 @@ libseccomp = { version = "0.3.0", optional = true }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 rust-criu = "0.4.0"
+protobuf = "= 3.2.0" # https://github.com/checkpoint-restore/rust-criu/issues/19
 regex = "1.10.2"
 thiserror = "1.0.50"
 tracing = { version = "0.1.40", features = ["attributes"] }


YJDoc2:
Hey @utam0k is there anything urgent blocking on this? I think this dependency bug should be handled in upstream itself, rather than we adding the constraint here. I'll approve this PR in case something urgently needs this patch in youki, but otherwise I think we should wait for the crate itself to fix it properly.

YJDoc2 (approved):

utam0k:
> Hey @utam0k is there anything urgent blocking on this? I

Thanks for your review. We can't build youki from scratch at lease in my environment 😭

Mossaka:
hey there, I would like to understand more about this situation, as `runwasi` depends on both `youki` and `containerd-shim`, which in turn depend on `protobuf`. I would not like to see a transititive dependency version conflict if both crates specify a **exact version** for protobuf.

If you have encountered a breaking change in the latest version `protobuf`, I'd suggest not to use exact version pinning, but more relexed measure like this one: https://github.com/containerd/rust-extensions/pull/322

utam0k:
> hey there, I would like to understand more about this situation, as `runwasi` depends on both `youki` and `containerd-shim`, which in turn depend on `protobuf`. I would not like to see a transititive dependency version conflict if both crates specify a **exact version** for protobuf.
> 
> If you have encountered a breaking change in the latest version `protobuf`, I'd suggest not to use exact version pinning, but more relexed measure like this one: [containerd/rust-extensions#322](https://github.com/containerd/rust-extensions/pull/322)

Hi, @Mossaka. As far as my memory, it caused flaky CI as the comment on https://github.com/checkpoint-restore/rust-criu/issues/19.  How about specifying the bottom version of `protobuf` crate in youki?

> This is fatal as this WILL [bring build errors](https://github.com/terrapkg/packages/actions/runs/6691482351/job/18178806142#step:7:3271) on anything that depends on your crate once a newer version of protobuf is available.

Mossaka:
What does bottom version mean?

utam0k:
> What does bottom version mean?

It means the range for the crate's version. Like `^0.3.2`. Is it enough for runwasi?

utam0k:
Or it may now be stable. Let me see for a while if the CI would be unstable.

Mossaka:
Yeah, runwasi resolved this issue by updating containerd-shim-protos to the latest version, which I made the change to relex the version requirement for `protobuf`. `^0.3.2` is better than `=0.3.2` but still not great.

PR #2524, opened 2023-11-12
fix(libcgroups): report CPU throttling stats in 'libcgroups::v2'

### LINKED ISSUE #2513: libcgroups v2 manager stats doesn't set cpu throttling field
https://github.com/containers/youki/blob/77d31666bbf59a741c3bac5ad94ea4bf22480a8c/crates/libcgroups/src/v2/cpu.rs#L65

we can see the v2 manager `stats` doesn't set cpu throttling field, user can't get the `cpu.max` data when using v2 manager

xiaoyang-sde:
fix #2513 

This pull request adds logic to read relevant fields from `cpu.stat` file to populate the `CpuThrottling` struct for the `stats` method of `libcgroups::v2::Cpu`.

Reference: https://www.kernel.org/doc/Documentation/admin-guide/cgroup-v2.rst

```rst
CPU Interface Files
~~~~~~~~~~~~~~~~~~~

All time durations are in microseconds.

  cpu.stat
	A read-only flat-keyed file.
	This file exists whether the controller is enabled or not.

	It always reports the following three stats:

	- usage_usec
	- user_usec
	- system_usec

	and the following five when the controller is enabled:

	- nr_periods
	- nr_throttled
	- throttled_usec
	- nr_bursts
	- burst_usec
```

### DIFF
--- a/crates/libcgroups/src/v2/cpu.rs
+++ b/crates/libcgroups/src/v2/cpu.rs
@@ -1,12 +1,11 @@
 use std::{
     borrow::Cow,
-    num::ParseIntError,
     path::{Path, PathBuf},
 };
 
 use crate::{
     common::{self, ControllerOpt, WrappedIoError},
-    stats::{self, CpuStats, StatsProvider},
+    stats::{self, CpuStats, ParseFlatKeyedDataError, StatsProvider},
 };
 
 use oci_spec::runtime::LinuxCpu;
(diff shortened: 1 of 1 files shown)


xiaoyang-sde:
It seems that the `stats` method of `libcgroup::v1::Cpu` is using `parse_flat_keyed_data` to parse the `cpu.stats` file:

https://github.com/containers/youki/blob/7b5e8f238e45497fe041c9f622a9a4e0b3f0e7e4/crates/libcgroups/src/v1/cpu.rs#L66-L90

Should we use `parse_flat_keyed_data` in `libcgroup::v2::Cpu`?

xiaoyang-sde:
> Thanks for your PR!
> 
> > Should we use parse_flat_keyed_data in libcgroup::v2::Cpu?
> 
> I prefer to use it if possible.

✅

It turns out that the following macro appears at least 3 times in the code base. This macro makes it easier to extract fields from the parsed flat keyed data. Should we move this common logic to `crates/libcgroups/src/stats.rs`?

```rs
        macro_rules! get {
            ($name: expr => $field1:ident.$field2:ident) => {
                stats.$field1.$field2 =
                    *stats_table
                        .get($name)
                        .ok_or_else(|| V2CpuStatsError::MissingField {
                            field: $name,
                            path: stats_path.clone(),
                        })?;
            };
```

utam0k:
> It turns out that the following macro appears at least 3 times in the code base. This macro makes it easier to extract fields from the parsed flat keyed data. Should we move this common logic to crates/libcgroups/src/stats.rs?

Go ahead 👍 But please send out a new PR about it. When you will do it, please set me as a reviewer.

utam0k (approved):
LGTM

PR #2530, opened 2023-11-14
fix(libcgroups): clean up 'libcgroups::v1::manager'

xiaoyang-sde:
This pull request cleans up the 'libcgroups::v1::manager' module.
- Remove the redundant `use crate::v1::ControllerType` statement
- Replace `for subsys in &self.subsystems` with `for (ctrl_type, cgroup_path) in &self.subsystems` (because `self.subsystems` is a `HashMap`)
- Remove redundant invocations of `Result::map_err()`


utam0k (approved):
💯

PR #2531, opened 2023-11-15
test(integration_test): port 'runtime-tools/validation/linux_seccomp'

xiaoyang-sde:
#361

This pull request ports the [`runtime-tools/validation/linux_seccomp`](https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_seccomp/linux_seccomp.go) test case to `tests/integration_test/src/tests/seccomp/mod.rs`.

### DIFF
--- a/tests/integration_test/src/tests/seccomp/mod.rs
+++ b/tests/integration_test/src/tests/seccomp/mod.rs
@@ -28,7 +28,6 @@ fn seccomp_test() -> TestResult {
     let spec = create_spec(
         LinuxSeccompBuilder::default()
             .default_action(LinuxSeccompAction::ScmpActAllow)
-            .architectures(vec![Arch::ScmpArchX86_64])
             .syscalls(vec![LinuxSyscallBuilder::default()
                 .names(vec![String::from("getcwd")])
                 .action(LinuxSeccompAction::ScmpActErrno)
--- a/tests/integration_test/src/tests/seccomp/mod.rs
+++ b/tests/integration_test/src/tests/seccomp/mod.rs
@@ -1,5 +1,5 @@
 use oci_spec::runtime::{
-    Arch, LinuxBuilder, LinuxSeccomp, LinuxSeccompAction, LinuxSeccompBuilder, LinuxSyscallBuilder,
+    LinuxBuilder, LinuxSeccomp, LinuxSeccompAction, LinuxSeccompBuilder, LinuxSyscallBuilder,
     ProcessBuilder, Spec, SpecBuilder,
 };
 use test_framework::{Test, TestGroup, TestResult};
--- a/tests/integration_test/src/tests/seccomp/mod.rs
+++ b/tests/integration_test/src/tests/seccomp/mod.rs
@@ -1,0 +1,49 @@
+use oci_spec::runtime::{
+    Arch, LinuxBuilder, LinuxSeccomp, LinuxSeccompAction, LinuxSeccompBuilder, LinuxSyscallBuilder,
+    ProcessBuilder, Spec, SpecBuilder,
+};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::utils::test_inside_container;
+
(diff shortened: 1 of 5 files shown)


YJDoc2:
Hey, the CI is passing, and overall seems ok. I'll take a more detailed look a bit later, but should be fine with one-two nitpicks.
Thanks a lot for contributing!

xiaoyang-sde:
> Hey, the CI is passing, and overall seems ok. I'll take a more detailed look a bit later, but should be fine with one-two nitpicks. Thanks a lot for contributing!

No problem! 👍

tests/integration_test/src/tests/seccomp/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+        )
+        .build()
+        .unwrap()
+}
+
+fn seccomp_test() -> TestResult {
+    let spec = create_spec(
+        LinuxSeccompBuilder::default()
+            .default_action(LinuxSeccompAction::ScmpActAllow)
+            .architectures(vec![Arch::ScmpArchX86_64])
```
YJDoc2, line 31: Do we need to setup the arch specifically, or because we are using the build pattern, we can not specify it and let the builder take default one?

tests/runtimetest/src/tests.rs
```diff
@@ -268,6 +269,21 @@ pub fn validate_mounts_recursive(spec: &Spec) {
     }
 }
 
+pub fn validate_seccomp(spec: &Spec) {
+    let linux = spec.linux().as_ref().unwrap();
+    if linux.seccomp().is_some() {
+        if let Err(errno) = getcwd() {
+            if errno != Errno::EPERM {
+                eprintln!("'getcwd()' failed with unexpected error code 'EPERM', found '{errno}'");
```
YJDoc2, line 277:
```suggestion
                eprintln!("'getcwd()' failed with unexpected error code '{errno}', expected  'EPERM'");
```

tests/integration_test/src/tests/seccomp/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+            .syscalls(vec![LinuxSyscallBuilder::default()
+                .names(vec![String::from("getcwd")])
+                .action(LinuxSeccompAction::ScmpActErrno)
+                .build()
+                .unwrap()])
+            .build()
+            .unwrap(),
+    );
```
YJDoc2, line 39: Do we not need the iterating check like https://github.com/opencontainers/runtime-tools/blob/master/cmd/runtimetest/main.go#L926-L940 , where we specifically check if the syscall name is `getcwd` and only ?

My thought is because we have setup where we specifically run a runtimetest test according to the commandline arg provided to it, we do not need that. Can you confirm that it is so?

YJDoc2 (changes requested):
Hey, overall good, just minot nitpicks, good to go after those are fixed :+1:

tests/integration_test/src/tests/seccomp/mod.rs
xiaoyang-sde: I think your thought is correct.

Based on https://github.com/opencontainers/runtime-tools/blob/master/cmd/runtimetest/main.go#L1253, the `runtimetest` executable will execute all validators regardless of the specific integration test case. Therefore, it needs to check if the syscall name is `getcwd`.
We don't need that check because the `validate_seccomp` function will only execute in the `seccomp` test case.

tests/integration_test/src/tests/seccomp/mod.rs
xiaoyang-sde: It seems that we don't need to explicitly set this option in the builder.

tests/integration_test/src/tests/seccomp/mod.rs
YJDoc2: Yep, thanks!

YJDoc2 (approved):
LGTM! Thanks for your contributions : )

PR #2532, opened 2023-11-15
Fix emulated cgroups v1 subsystem when running docker-in-docker

### LINKED ISSUE #2528: `youki` fails to run in `docker-in-docker` with `cgroups v1`
Reproduction:
```bash
cargo install cross --git https://github.com/cross-rs/cross
git clone --branch dind git@github.com:jprendes/youki.git
cd youki
cross build --features systemd,v1,v2 --bin youki
./dind.sh
```

Removing `--runtime=youki` from `dind.sh`, the example runs fine, but with youki it fails.

<details>
  <summary>Old error message</summary>
The error I receive is

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: unable to retrieve OCI runtime error (open /var/run/docker/containerd/daemon/io.containerd.runtime.v2.task/moby/278419ddb31fde6701841c5b302a632751de2a5a6aa051a62530b12ee8ab4763/log.json: no such file or directory): /youki did not terminate successfully: exit status 1: unknown.
ERRO[0003] error waiting for container:
```
  
This originates from tying to run runwasi on docker-in-docker. In that case it works fine on cgroups v2 but fails on cgroups v1, and the error is while mounting cgroups.
</details>

New error message

After workind around the issue with journald, I get an error when using `youki` with `dind` + `cgroups v1`.
```
libcontainer::rootfs::mount: failed to canonicalize "/sys/fs/cgroup/systemd/docker/af0c557cb9806654e1a9eac1de4a12afda57bf9f60952e4c3db016668c554ea6": No such file or directory (os error 2)
```
See https://github.com/containers/youki/issues/2528#issuecomment-1809985629 for log context.

jprendes:
Fixes #2528.

Normally the pid-1 cgroup path is empty, but this is not the case when running inside a container.
This needs to be accounted on for the cgroups subsystem emulation.
This PR fixes the source of the mountpoints for the emulated cgroups subsystem by stripping the cgroup paths of pid-1 from the mountpoint source path.

I also makes the failure to setup tracing to syslog a warning instead of fatal, as that fails with the `docker` image used for _docker-in-docker_.

I would like to add testing for this based on the reproduction steps in #2528, but that would be better done by building `youki` as a static binary using `musl` (since the `docker` image is based on `alpine`). I would like to wait until we merge https://github.com/containers/youki/pull/2498 to add `musl` support and add a test.

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -87,7 +87,7 @@ impl Mount {
                         panic!("libcontainer can't run in a Legacy or Hybrid cgroup setup without the v1 feature");
                         #[cfg(feature = "v1")]
                         self.mount_cgroup_v1(mount, options).map_err(|err| {
-                            tracing::error!("failed to mount cgroup v2: {}", err);
+                            tracing::error!("failed to mount cgroup v1: {}", err);
                             err
                         })?
                     }
@@ -171,10 +171,28 @@ impl Mount {
         tracing::debug!("cgroup mounts: {:?}", host_mounts);
 
         // get process cgroups
+        let ppid = std::os::unix::process::parent_id();
+        let ppid = if ppid == 0 { std::process::id() } else { ppid };
+        let root_cgroups = Process::new(ppid as i32)?.cgroups()?.0;
         let process_cgroups: HashMap<String, String> = Process::myself()?
             .cgroups()?
             .into_iter()
-            .map(|c| (c.controllers.join(","), c.pathname))
+            .map(|c| {
+                let hierarchy = c.hierarchy;
+                // When youki itself is running inside a container, the cgroup path
+                // will include the path of pid-1, which needs to be stripped before
+                // mounting.
+                let root_pathname = root_cgroups
+                    .iter()
+                    .find(|c| c.hierarchy == hierarchy)
+                    .map(|c| c.pathname.as_ref())
+                    .unwrap_or("");
+                let path = c
+                    .pathname
+                    .strip_prefix(root_pathname)
+                    .unwrap_or(&c.pathname);
+                (c.controllers.join(","), path.to_owned())
+            })
             .collect();
         tracing::debug!("Process cgroups: {:?}", process_cgroups);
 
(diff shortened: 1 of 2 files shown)


crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -171,10 +171,26 @@ impl Mount {
         tracing::debug!("cgroup mounts: {:?}", host_mounts);
 
         // get process cgroups
+        let pid1_cgroups = Process::new(1)?.cgroups()?.0;
```
utam0k, line 174: If users don't separate the PID namespace, what happens?

jprendes: Good point!
It fails in the same way as without the fix :-(

jprendes: When not unsharing PID namespace, using the parent PID does the trick :-)
I just pushed a fix

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -171,10 +171,28 @@ impl Mount {
         tracing::debug!("cgroup mounts: {:?}", host_mounts);
 
         // get process cgroups
+        let ppid = std::os::unix::process::parent_id();
+        let ppid = if ppid == 0 { std::process::id() } else { ppid };
```
utam0k, line 175:
```suggestion
        // The non-zero ppid means that the PID Namespace is not separated.
        let ppid = if ppid == 0 { std::process::id() } else { ppid };
```

utam0k (approved):
It looks good to me. Let's continue to add a test on this PR if you want.

jprendes:
> Let's continue to add a test on this PR if you want.

Yes, that's a good plan :-)

YJDoc2:
Hey, should we go ahead and merge this fix as it is? We can add the tests in a separate PR if you want :+1:

utam0k:
> Hey, should we go ahead and merge this fix as it is? We can add the tests in a separate PR if you want 👍

I'd like to include this PR in the next patch release, so please send out another PR 🙏

jprendes:
> please send out another PR

Will do. The tests will need the musl builds PR to be merged. I'll work on that first.
Sorry for the delay, I have limited bandwidth as I'm currently on parental leave.

PR #2536, opened 2023-11-18
(feat) add support for `musl` using `cross-rs`

jprendes:
This PR adds support for building and testing against `musl` targets.

A summary of changes:
* Add musl targets to `Cross.toml` and the corresponding `Dockerfile.musl`
* Install dependencies in `Dockerfile.gnu` so that we can run tests using cross.
* Add a `cargo.sh` script that wraps `cargo` calls, using either `cargo` or `cross` based on the target.
* Fix `mock.rs` so that it builds with `musl` targets
* Use the `cargo.sh` script in the `features_test.sh` and `justfile`
* Remove now redundant `musl_test.sh`.
* Update CI to install `cross`, use `cargo.sh`, and force using `cross` for all targets (including host target).
* Update CI to relese `musl` binaries as well as `gnu` binaries.

Example of a release generated from this PR: https://github.com/jprendes/youki/releases/tag/v0.4.1

Note: This PR builds on top of #2541. I will rebase once that is merged.


utam0k:
Sorry, I won't be able to find the time this week, so I'll look at it in a few days 🙏

scripts/cargo.sh
```diff
@@ -0,0 +1,66 @@
...
+# `cross-<target>` to avoid libc conflicts in host binaries (e.g.,
+# build scripts, proc macros). The computed value of the target
+# directory can be obtained with `cargo.sh --print-target-dir`.
+#
+# Lastly, when using `cross` this scrips sets some configuration
+# to allow running `youki` tests inside the `cross`` container.
+# Please check the comments in this scrips to learm more about that.
+#
+# Limitations:
+#  * You **must** use the `CARGO_BUILD_TARGET` environment variable
```
adrianalin, line 26: Isn't this overwritten by
https://github.com/containers/youki/pull/2536/files#diff-3d64996115d0ac3b74e36f57ce6b00d6b4982614c41b1ed39eab8ef43b2e9a2bR38 ?

scripts/cargo.sh
```diff
@@ -0,0 +1,66 @@
...
+#    to specify the build target instead of using the `--target` CLI
+#    argument, or a configuration file like `.cargo/config.toml`.
+#  * If **must** use the `CARGO_TARGET_DIR` environment variable to
+#    specify the target directory instead of using the `--target-dir`
+#    CLI argument, or a configuration file like `.cargo/config.toml`.
+#
+
+ROOT=$(cargo metadata --format-version=1 | jq -r .workspace_root)
```
adrianalin, line 34: No need to add new dependency (jq), you can use 
```suggestion
ROOT=$(git rev-parse --show-toplevel)
```

scripts/cargo.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
adrianalin, line 1:
```suggestion
#!/usr/bin/env bash
set -euo pipefail
```

scripts/cargo.sh
jprendes: Thanks, good catch :-)

jprendes:
Reabsed.
@utam0k @adrianalin @cuisongliu PTAL

An example release generated by this PR is:
https://github.com/jprendes/youki/releases/tag/v0.3.2

YJDoc2:
Self-reminder : need to update  merge CI requirements to correct CI names after this is merged.
(The three pending CI seems to be because of renaming of the workflows, so should be ok.)

utam0k (approved):
May I ask a favor of the documentation for developers in another PR?
https://containers.github.io/youki/developer/introduction.html

utam0k:
Thanks!! @jprendes

utam0k:
> Self-reminder : need to update merge CI requirements to correct CI names after this is merged. (The three pending CI seems to be because of renaming of the workflows, so should be ok.)

Done ✅

YJDoc2:
Thanks @utam0k 
Is this ready to merge? The CI checks which are updated in this are keeping other PRs from passing...
I can revert the CI check requirements temp until this is merged, or as we don't have anything urgent, just wait out for this.
Let me know.

.github/workflows/main.yml
```diff
@@ -35,21 +35,26 @@ jobs:
...
+      - name: Install just
+        uses: taiki-e/install-action@just
+      - name: Install cross-rs
+        run: cargo install cross --git https://github.com/cross-rs/cross
+      - name: Setup target
+        run: |
+          echo "CARGO=cross" >> ${GITHUB_ENV}
+          echo "TARGET=${{ matrix.arch }}-unknown-linux-${{ matrix.libc }}" >> ${GITHUB_ENV}
       - name: Check formatting and lints
         run: just lint
```
YJDoc2, line 59: If I'm right, then I don't think we need matrix for lint checking, the `just lint` target should checks for all targets and features. Would that be affected by arch or libc type?

Cross.toml
```diff
@@ -1,8 +1,14 @@
 [build]
-default-target = "x86_64-unknown-linux-gnu"
+default-target = "x86_64-unknown-linux-musl"
```
YJDoc2, line 2: I think we should keep the default target to gnu, rather than musl unless we have some strong reason not to...

scripts/build.sh
YJDoc2, line 1: If I'm correct, this would still use the normal cargo for building unless we specify cross, right? I'd prefer not needing cross or other related deps for local development unless needed in development.

YJDoc2 (approved):
LGTM :+1:

I have some nitpicks / questions, but nothing that would block merging, so as @utam0k has already approved, I'll wait sometime and merge this.

@jprendes thanks a lot for this, this is really amazing work! Take your time for replying to the comments :+1:

scripts/build.sh
jprendes: Thats correct. It defaults to cargo when building for the host triple, and to cross when building for another triple.

Cross can be forced with the env var `CARGO=cross` or with the `-x` CLI argument.

.github/workflows/main.yml
jprendes: The arch could make a difference with clippy when conditional compilation is used.

This also applies for dependencies using conditional compilation. An example of this is in `mock.rs` in this PR, where libc doesn't expose `__rlimit_resource_t` when building for musl.

Cross.toml
jprendes: This only impacts when cross is called without specifying a target. The cargo.sh script always specifies a target, so it doesn't use the default.

I'll change it to gnu for consistency.

utam0k:
Ops sorry I'll merge it

adrianalin:
Sorry, I could not find where documentation was updated, @jprendes please provide link. Thank you!

jprendes:
> Sorry, I could not find where documentation was updated (related to musl build)

I still need to create a PR. I'll do that as soon as I have some bandwidth again.

PR #2544, opened 2023-11-23
Support namespace path or mount_label or rootfs_propagation empty in json

Vanient:
If we use youki as crate, config.json maybe created by upper container engine not youki self, namespace path, apparmor_profile, mount_label or rootfs_propagation may be set to Some("") not None in config.json, but it should be treated same as None.

### DIFF
--- a/crates/libcontainer/src/container/init_builder.rs
+++ b/crates/libcontainer/src/container/init_builder.rs
@@ -164,14 +164,16 @@ impl InitContainerBuilder {
 
         if let Some(process) = spec.process() {
             if let Some(profile) = process.apparmor_profile() {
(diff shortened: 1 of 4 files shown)


YJDoc2:
Hey, thanks for opening this PR!
Is there any upper container engine where you have observed this case? Or does this occur when we do some specific settings only?
Also, I think a better place to handle this / at least the spec part of this would be in the oci_spec library itself, rather than in youki, wdyt?

Vanient:
> Hey, thanks for opening this PR! Is there any upper container engine where you have observed this case? Or does this occur when we do some specific settings only? Also, I think a better place to handle this / at least the spec part of this would be in the oci_spec library itself, rather than in youki, wdyt?

Yes, you are right i found it should not show in json if String is "", close it.

PR #2548, opened 2023-11-29
Set up userns in a straightforward way

utam0k:
It is just a functionalization. Easier to test.


utam0k:
Thanks for your review 🙇

PR #2583, opened 2023-12-21
New Releases needs approval from the maintainer

utam0k:
We have already introduced an automatic release workflow. But it leads to accidents. Now we merge https://github.com/containers/youki/pull/2582 and will start our release workflow.

I'd like to prevent it by reviewing the deployment flow.
https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -60,6 +60,8 @@ jobs:
 
   release:
     name: Create Draft Release
+    environment:
+      name: release
     runs-on: ubuntu-20.04
     permissions:
       contents: write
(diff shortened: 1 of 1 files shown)


YJDoc2:
Hey, so this will not run the release CI, however the tag PR would  still run right? Also, are we doing 0.3.2 release immediately?

utam0k:
> Also, are we doing 0.3.2 release immediately?

I don't have a plan.

> Hey, so this will not run the release CI, however the tag PR would still run right?

Yes, but this PR prevents the release workflow without approval. Frankly, it is not so much a question of an accidental tag in itself being created.

YJDoc2 (approved):

PR #2621, opened 2024-01-09
[doc] Update doc with `cross-rs` and `musl` builds

jprendes:
This PR updates the mdbook documentation to take into account the changes introduced in #2536

### DIFF
--- a/docs/src/user/basic_setup.md
+++ b/docs/src/user/basic_setup.md
@@ -6,16 +6,35 @@ Youki currently only supports Linux Platform, and to use it on other platform yo
 
 Also note that Youki currently only supports and expects systemd as init system, and would not work on other systems. There is currently work on-going to put systemd dependent features behind a feature flag, but till then you will need a systemd enabled system to work with Youki.
 
-## Requirements
+## Build Requirements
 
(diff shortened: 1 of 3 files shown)


jprendes:
PTAL, thanks!
@utam0k @YJDoc2 @adrianalin

docs/src/user/basic_setup.md
```diff
@@ -6,16 +6,35 @@ Youki currently only supports Linux Platform, and to use it on other platform yo
...
+```console
+$ just youki-dev # or youki-release
+```
 
-### Debian, Ubuntu and related distributions
+Install the build depdencies using your distribution's package manger
 
+#### Debian, Ubuntu and related distributions
 ```console
 $ sudo apt-get install    \
```
jprendes, line 39: I'm not sure that we need all these dependencies.
IIUC we don't need neither of `libsystemd-dev`, `libelf-dev`, `libclang-dev` or `libssl-dev`.
Are they needed by any of the tests?

utam0k: Good point. I often forget to update the dependencies. A good way to check is to delete them from there and attempt our CI. When all CI are passed, it means they aren't necessary.
https://github.com/containers/youki/blob/656ccc3f69329c2ed10129e046d0474e6c3dee13/justfile#L153-L160

utam0k (approved):
Good

PR #2629, opened 2024-01-17
Fix the missed naming changes in integration test validation CI

YJDoc2:
The name changes were missed for the validate-integration-test CI, so it started failing. This fixes that.


YJDoc2:
There are no core code changes, and blocking a couple of PRs, so skipping branch protection for this. @containers/artwork-maintainers , please take a look later, and let me know if any changes are needed, will open another PR for that.

PR #2638, opened 2024-01-18
Roll up various minor and major version dep upgrade

YJDoc2:
Rolls up minor version dep updates from : #2632 , #2633 , #2635
Also updates wasmtime major version from v10->v16 from #2627 and #2636


utam0k (approved):
Thanks!

PR #2645, opened 2024-01-19
Add docker-in-docker e2e test

jprendes:
This PR adds tests to lock the behavior introduced by #2532.
It tests that youki can be used in docker-in-docker with both cgroups v1 (through Ubuntu 20.04) and v2 (through Ubuntu 22.04).

### DIFF
--- a/tests/dind/run.sh
+++ b/tests/dind/run.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+set -e
+
+ROOT=$(git rev-parse --show-toplevel)
+
+docker run --privileged -dq \
+  --name youki-test-dind \
+  -v $ROOT/youki:/usr/bin/youki \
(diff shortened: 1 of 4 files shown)


tests/dind/run.sh
```diff
@@ -0,0 +1,19 @@
...
+docker run --privileged -dq \
+  --name youki-test-dind \
+  -v $ROOT/youki:/usr/bin/youki \
+  -v $ROOT/tests/dind/daemon.json:/etc/docker/daemon.json \
+  docker:dind > /dev/null
+
+trap "docker rm -f youki-test-dind > /dev/null" EXIT
+
+# give dockerd time to start
+sleep 2
```
jprendes, line 16: This timeout might be to short

utam0k (approved):
Awesome!

PR #2647, opened 2024-01-21
Re enable skipped e2e tests

### LINKED ISSUE #2616: Decide which runc to use for validating the rust oci test validation
Even though PRs for features such as domain name and sched_entity are merged in rnuc's main branch, no release till now contains the support for them. That is causing causing block for merging #1544 and #2495 , as the rust oci tests fail on runc's released version. When tested on runc compiled from main branch's  latest commit, the tests are passing, as the supporting code is present.

We can take two approaches for solving this :
1. In the CI, clone runc and build from main instead of using the released versions. That way we will get all latest features supported by runc, but then we would be validating our tests on latest runc instead of the released versions, also issues with potential regressions that might get in runc itself. We can pin by runc commit.
2. Do a more complex scheme of tests, where we either ignore the un-supported tests or validate that  the tests are failing on runc, and consider that failure as success for runc. That way we can track what features we support not yet supported by runc etc. This would require a bit complex cfg based or feature based building and some code/macros to appropriately validate tests. Also would make our tests more tailored to youki and runc instead of generic oci adhering runtime.

I would like to take one approach or the other soon, rather than waiting for runc to release the features, as both PR have been pending for long time only due to the runc validation failing.

cc @utam0k

YJDoc2:
closes #2616 

There are some tests which are (for now) only supported by youki, and not by the latest release of runc. Instead of not running those tests, we instead check an env var to see what runtime we are running, and skip those for runc.

### DIFF
--- a/tests/contest/contest/src/tests/scheduler/scheduler_policy.rs
+++ b/tests/contest/contest/src/tests/scheduler/scheduler_policy.rs
@@ -4,7 +4,7 @@ use oci_spec::runtime::{
 };
 use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
 
-use crate::utils::test_inside_container;
+use crate::utils::{is_runtime_runc, test_inside_container};
 
 fn create_spec(policy: LinuxSchedulerPolicy, execute_test: &str) -> Result<Spec> {
     let sc = SchedulerBuilder::default()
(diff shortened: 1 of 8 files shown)


tests/contest/contest/src/tests/scheduler/scheduler_policy.rs
```diff
@@ -46,11 +46,23 @@ fn scheduler_policy_batch_test() -> TestResult {
...
-    let policy_rr_test = Test::new("policy_batch", Box::new(scheduler_policy_batch_test));
-
-    scheduler_policy_group.add(vec![Box::new(policy_fifo_test)]);
-    scheduler_policy_group.add(vec![Box::new(policy_rr_test)]);
+    let policy_fifo_test = ConditionalTest::new(
+        "policy_other",
+        Box::new(|| match std::env::var("RUNTIME_KIND") {
+            Err(_) => true,
+            Ok(s) => s != "runc",
+        }),
```
utam0k, line 54: How about turn into a function and putting others together?

utam0k: others: https://github.com/containers/youki/pull/2647/files#diff-54c4c6b1c0869e602e1d1c3f81dd7162b5ad8e237a6fd7dc477416e27624233cR30-R33

YJDoc2: Done, please take a look!

utam0k (approved):

Issue #2144, opened 2023-07-06
cargo test with `musl` hangs occasionally


yihuaf:
We observed in the field that unit tests with `musl` hangs from time to time. This happens fairly often but not all the time. Several CI runs now shows this behavior and I am able to reproduce this locally with:

```console
[eng@stardust youki]$ cargo +nightly test fork -Zbuild-std --target $(uname -m)-unknown-linux-musl --package libcontainer --no-default-features -F v2
    Finished test [unoptimized + debuginfo] target(s) in 0.12s
     Running unittests src/lib.rs (target/x86_64-unknown-linux-musl/debug/deps/libcontainer-0019f012789e0a63)

running 3 tests
test process::fork::test::test_container_clone_sibling ... ok
test process::fork::test::test_container_fork ... ok
test process::fork::test::test_container_err_fork has been running for over 60 seconds
```

The test that hangs is a simple test where we fork/clone3 a with a function that returns error right away. The test itself runs correctly in the normal unit test with glibc. This seems to only happen in musl.

I was able to reproduce this case locally that happens about once in a while. I see 3 process related to cargo test at the time of the hang:

```
[eng@stardust youki]$ ps aux | grep libcontainer
eng      1325755  0.7  0.1 119340 44988 pts/5    S+   16:42   0:00 /home/eng/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo test fork -Zbuild-std --target x86_64-unknown-linux-musl --package libcontainer --no-default-features -F v2
eng      1325757  0.0  0.0  14016  6400 pts/5    Sl+  16:42   0:00 /home/eng/private/youki/target/x86_64-unknown-linux-musl/debug/deps/libcontainer-0019f012789e0a63 fork
eng      1325762  0.0  0.0  18128  1532 pts/5    S+   16:42   0:00 /home/eng/private/youki/target/x86_64-unknown-linux-musl/debug/deps/libcontainer-0019f012789e0a63 fork
```

Specifically, 1325757 seems to be the actual unit test process. The process correctly is calling `waitpid`:

```
[eng@stardust youki]$ strace -f -p 1325757
strace: Process 1325757 attached with 2 threads
[pid 1325757] restart_syscall(<... resuming interrupted read ...> <unfinished ...>
[pid 1325759] wait4(1325762, ^Cstrace: Process 1325757 detached
```

The 1325762 is the forked process and interestingly it is stuck in a futex:

```
[...]

PR #2685, opened 2024-02-15
Set '--test-threads' option to 1 in unit tests

YJDoc2:
Another try at https://github.com/containers/youki/pull/2615 and #2144 
 
done with ref https://github.com/containers/youki/pull/2615#issuecomment-1882697445

Instead of splitting the tests by configs, tried setting the number of test threads to 1, so the tests are always run one at a time. There is no considerable change in time, the basic check CI (whole) takes about 5-6 minutes now instead of 3 minutes previously. I ran it 10 times, and it didn't fail in any attempt : https://github.com/YJDoc2/youki/actions/workflows/basic.yml?query=branch%3Atests%2Ffix-for-2144-v2

This is a much simpler solution for the issue, and as mentioned in original problem, the root cause comes from tests running in multiple threads, and should not occur in actual use ; so doing this seems a better option than code changes, if this works.

### DIFF
--- a/justfile
+++ b/justfile
@@ -39,11 +39,11 @@ test-basic: test-unit test-doc
 
 # run cargo unit tests
 test-unit:
-    {{ cwd }}/scripts/cargo.sh test --lib --bins --all --all-targets --all-features --no-fail-fast
+    {{ cwd }}/scripts/cargo.sh test --lib --bins --all --all-targets --all-features --no-fail-fast -- --test-threads=1
 
 # run cargo doc tests
 test-doc:
-    {{ cwd }}/scripts/cargo.sh test --doc
+    {{ cwd }}/scripts/cargo.sh test --doc -- --test-threads=1
 
 # run permutated feature compilation tests
 test-features:


jprendes (approved):
LGTM.

IIUC, the process will still have 2 threads, one driving the testing framework and another running the actual test. But the thread running the framework will be sleepping the whole time the test thread is running, so it should be fine.

YJDoc2:
> IIUC, the process will still have 2 threads, one driving the testing framework and another running the actual test. But the thread running the framework will be sleepping the whole time the test thread is running, so it should be fine.

Yes, I think that is correct. But if we are to use cargo test, or anything else, I don't  think that can be avoided, so limiting the thread to 1 was the best option I could think of :) 

Another approach was as utam0k suggested, to separate the hanging tests with some kind of feature, and run them single threaded separately, but I found that some of them already had a `serial` attribute to them, which ideally should have run them in a single thread. Thus,  went for this appraoch. If nothing else, I'm hoping single threading this would at least reduce the failure rate to a degree which can be considered rare enough to be ignored. AFAIK, the increased time due to running them serially is not that much more (considering time taken by all CI jobs, specially containerd tests) so would be ok, compared against having to re run CI due to test hangs.

utam0k (approved):
The code is a bit changed. However, this change has been based on a lot of your research. And it solves a critical problem that bothers Youki's maintainer. I appreciate you 😍

PR #2686, opened 2024-02-15
Add test for cgroups_relative_memory

omprakaash:
Adds an integration test for cgroups_relative_memory. The only difference from the non-relative test seems to be the cgroup_path of the specified spec while testing.

### DIFF
--- a/crates/libcgroups/src/v1/util.rs
+++ b/crates/libcgroups/src/v1/util.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::HashMap,
+    fs, io,
+    path::{Path, PathBuf},
+};
 
 use procfs::{process::Process, ProcError};
 
@@ -78,4 +82,25 @@ pub fn get_subsystem_mount_point(subsystem: &ControllerType) -> Result<PathBuf, 
         .ok_or(V1MountPointError::NotFound {
             subsystem: *subsystem,
         })
+}
+
+pub fn get_subsystem_path(pid: i32, subsystem: &str) -> Result<PathBuf, io::Error> {
+    let contents = fs::read_to_string(Path::new(&format!("/proc/{}/cgroup", pid)))
+        .unwrap_or_else(|_| panic!("failed to read /proc/{}/cgroup", pid));
+    for line in contents.lines() {
+        let parts: Vec<&str> = line.splitn(3, ':').collect();
+        if parts.len() < 3 {
+            continue;
+        }
+        let subparts: Vec<&str> = parts[1].split(',').collect();
+        for subpart in subparts {
+            if subpart == subsystem {
+                return Ok(PathBuf::from(parts[2].to_string()));
+            }
+        }
+    }
+    Err(io::Error::new(
+        io::ErrorKind::Other,
+        format!("subsystem {} not found", subsystem),
+    ))
 }
--- a/tests/contest/contest/src/tests/cgroups/memory.rs
+++ b/tests/contest/contest/src/tests/cgroups/memory.rs
@@ -1,12 +1,12 @@
 use std::path::Path;
 
+use crate::utils::{linux_resource_memory::validate_linux_resource_memory, test_outside_container};
 use anyhow::{Context, Result};
+use libcgroups::common::{get_cgroup_setup, CgroupSetup};
 use oci_spec::runtime::{
     LinuxBuilder, LinuxMemoryBuilder, LinuxResourcesBuilder, Spec, SpecBuilder,
 };
 use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
-
-use crate::utils::{test_outside_container, test_utils::check_container_created};
 
 const CGROUP_MEMORY_LIMIT: &str = "/sys/fs/cgroup/memory/memory.limit_in_bytes";
 const CGROUP_MEMORY_SWAPPINESS: &str = "/sys/fs/cgroup/memory/memory.swappiness";
@@ -50,8 +50,8 @@ fn test_memory_cgroups() -> TestResult {
     ];
 
     for spec in cases.into_iter() {
-        let test_result = test_outside_container(spec, &|data| {
-            test_result!(check_container_created(&data));
+        let test_result = test_outside_container(spec.clone(), &|data| {
+            test_result!(validate_linux_resource_memory(&spec, data));
 
             TestResult::Passed
         });
@@ -64,7 +64,10 @@ fn test_memory_cgroups() -> TestResult {
 }
 
 fn can_run() -> bool {
-    Path::new(CGROUP_MEMORY_LIMIT).exists() && Path::new(CGROUP_MEMORY_SWAPPINESS).exists()
+    let cgroup_setup = get_cgroup_setup();
+    matches!(cgroup_setup, Ok(CgroupSetup::Legacy))
+        && Path::new(CGROUP_MEMORY_LIMIT).exists()
+        && Path::new(CGROUP_MEMORY_SWAPPINESS).exists()
 }
 
 pub fn get_test_group() -> TestGroup {
--- a/tests/contest/contest/src/tests/cgroups/relative_memory.rs
+++ b/tests/contest/contest/src/tests/cgroups/relative_memory.rs
@@ -1,0 +1,67 @@
+use std::path::Path;
+
+use anyhow::{Context, Result};
+use oci_spec::runtime::{
+    LinuxBuilder, LinuxMemoryBuilder, LinuxResourcesBuilder, Spec, SpecBuilder,
+};
+use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
+
+use crate::utils::{test_outside_container, test_utils::check_container_created};
+
+const CGROUP_MEMORY_LIMIT: &str = "/sys/fs/cgroup/memory/memory.limit_in_bytes";
+const CGROUP_MEMORY_SWAPPINESS: &str = "/sys/fs/cgroup/memory/memory.swappiness";
+
+fn create_spec(cgroup_name: &str, limit: i64, swappiness: u64) -> Result<Spec> {
+    let spec = SpecBuilder::default()
+        .linux(
+            LinuxBuilder::default()
+                .cgroups_path(Path::new("/testdir/runtime-test/container").join(cgroup_name))
+                .resources(
+                    LinuxResourcesBuilder::default()
+                        .memory(
+                            LinuxMemoryBuilder::default()
+                                .limit(limit)
+                                .swappiness(swappiness)
+                                .build()
+                                .context("failed to build memory spec")?,
+                        )
+                        .build()
+                        .context("failed to build resource spec")?,
+                )
+                .build()
+                .context("failed to build linux spec")?,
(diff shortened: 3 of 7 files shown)


tsturzl (approved):

lengrongfu:
LGTM

YJDoc2:
Hey, as @omprakaash had mentioned in discord we seem to be skipping a check for memory cgroups validation, as done [here](https://github.com/opencontainers/runtime-tools/blob/408c51e934dc558e90e13f63a808f1bd6b155fd2/validation/linux_cgroups_memory/linux_cgroups_memory.go#L45) in runtime-tools. @tsturzl do you have any idea, why we might be intentionally skipping this, or we might have just missed it originally?

If so, @omprakaash may I ask you to add that in both this and the other test? (You can do it in separate PR, we can merge this one before it)

tsturzl:
@YJDoc2 Hmm, if you compare any of the integration tests to runtime-tools it seems like we're missing a similar step for almost every cgroups integration test. I think many people assumed there was something in `check_container_created` that was checking the container against the spec somehow. It's been 3 years, so I can't recall my thinking from back then. It looks like all current cgroups tests need some step to validate the spec against the actual cgroups values. I was largely just copying what other cgroups integration tests had done at the time, but it looks like we're missing the important validation step currently.

A simple solution would be to just read the respective cgroups files and comparing them. However, there is a lot of opportunity here to create a read/write abstraction for cgroups files. Right now we just have a constant for each file name and use a `write_cgroup_file` utility function to write to a given path. If reading (or writing) cgroups file is going to become more useful outside the context of libcgroups crate then maybe we want a generalized abstraction to handle read/write? This would clean up the implementation and unit tests, as well as provide obvious utility to these integration tests in checking cgroups values without having to implement a lot of boilerplate. It's something I thought about introducing to make it easier to more easily swap out the kernel IO API being used which is something I've experimented with in the past. I don't know if that effort is too large to block fixing the issues with the current cgroups integration tests, but it's something I'd be interested in working on if no one else is immediately interested.

tsturzl:
@omprakaash that's probably a better immediate solution. If we can just implement something similar to what runtime tools is doing.

YJDoc2:
I think I have some changes to request in this, I'll try to get to a full review soon.

tests/contest/contest/src/tests/cgroups/memory.rs
```diff
@@ -5,8 +5,9 @@ use oci_spec::runtime::{
     LinuxBuilder, LinuxMemoryBuilder, LinuxResourcesBuilder, Spec, SpecBuilder,
 };
 use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
+use utils::linux_resource_memory::validate_linux_resource_memory;
 
-use crate::utils::{test_outside_container, test_utils::check_container_created};
+use crate::utils::{self, test_outside_container};
```
YJDoc2, line 10: nit: can we combine this import with the utils import above?

tests/contest/contest/src/tests/cgroups/relative_memory.rs
```diff
@@ -0,0 +1,67 @@
...
+use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
+
+const CGROUP_MEMORY_LIMIT: &str = "/sys/fs/cgroup/memory/memory.limit_in_bytes";
+const CGROUP_MEMORY_SWAPPINESS: &str = "/sys/fs/cgroup/memory/memory.swappiness";
+
+fn create_spec(cgroup_name: &str, limit: i64, swappiness: u64) -> Result<Spec> {
+    let spec = SpecBuilder::default()
+        .linux(
+            LinuxBuilder::default()
+                .cgroups_path(Path::new("/testdir/runtime-test/container").join(cgroup_name))
```
YJDoc2, line 18: nit: extract this path at top as a const

tests/contest/contest/src/tests/cgroups/relative_memory.rs
```diff
@@ -0,0 +1,67 @@
...
+    test_outside_container(spec.clone(), &|data| {
+        test_result!(validate_linux_resource_memory(&spec, data));
+
+        TestResult::Passed
+    })
+}
+
+fn can_run() -> bool {
+    Path::new(CGROUP_MEMORY_LIMIT).exists() && Path::new(CGROUP_MEMORY_SWAPPINESS).exists()
+}
```
YJDoc2, line 54: If this is cgroups v1 specific test, we should also check that cgroups is v1 here

tests/contest/contest/src/utils/linux_resource_memory.rs
```diff
@@ -0,0 +1,36 @@
...
+pub(crate) fn validate_linux_resource_memory(spec: &Spec, data: ContainerData) -> Result<()> {
+    let expected_memory = spec
+        .linux()
+        .as_ref()
+        .unwrap()
+        .resources()
+        .as_ref()
+        .unwrap()
+        .memory()
+        .as_ref();
```
YJDoc2, line 15: I think we can do without the `as_ref` calls here, as spec is already a immutable ref? Please check once.

tests/contest/contest/src/utils/linux_resource_memory.rs
```diff
@@ -0,0 +1,36 @@
...
+    let expected_memory = match expected_memory {
+        Some(m) => m,
+        None => bail!("expected memory to be set, but it was not"),
+    };
+
+    let memory = get_memory_data(data.state.unwrap().pid.unwrap());
+
+    if memory.is_err() {
+        bail!("failed to get memory data: {:?}", memory.err().unwrap());
+    }
```
YJDoc2, line 25:
```suggestion
    if let Err(e) = memory {
        bail!("failed to get memory data: {:?}", e);
    }
```

tests/contest/contest/src/utils/linux_resource_memory.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    if expected_memory.limit().unwrap() != memory.as_ref().unwrap().limit().unwrap() {
+        bail!("expected memory {:?}, got {:?}", expected_memory, memory);
+    }
+
+    if expected_memory.swappiness().unwrap() != memory.as_ref().unwrap().swappiness().unwrap() {
+        bail!("expected memory {:?}, got {:?}", expected_memory, memory);
+    }
```
YJDoc2, line 33: For both of these, can we extract the corresponding params in vars, and compare those? Additionally we should also report specifically what param mismatch was there, as potentially (although it shouldn't) some other param in memory data can be diff, making it confusing as to exactly why this failed. So in error message it should be something like `expected memory limit {expected_limit} but got {actual_limit} instead` .

YJDoc2 (changes requested):
Hey, Apologies it took me long for the review. There are some changed needed, so I have added comments, please take a look.

Is this test supposed to be cgroups  v1 specific? If not we should not use the v1 functions explicitly, and use a generic version that can work on both : v1 and v2. 

Also, for the functions added in libcgroup : if we are using them in only tests, and they might not be as useful otherwise, I'd prefer to move them in the test crate itself rather then libcgroup.

omprakaash:
Thank you for the review ! Will make the required changes.

crates/libcgroups/src/v1/util.rs
```diff
@@ -41,6 +46,71 @@ pub fn list_supported_mount_points() -> Result<HashMap<ControllerType, PathBuf>,
...
+                }
+                "memory.oom_control" => {
+                    let oom_control = value.split_whitespace().collect::<Vec<&str>>();
+                    let oom_control = oom_control
+                        .get(1)
+                        .ok_or("Failed to get oom_control")?
+                        .parse::<u64>()?;
+                    memory_data = memory_data.disable_oom_killer(oom_control == 1);
+                }
+                _ => {}
```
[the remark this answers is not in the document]
omprakaash: Got it. Will change it

tests/contest/contest/src/tests/cgroups/memory.rs
omprakaash: Yep !

crates/libcgroups/src/v1/util.rs
```diff
@@ -41,6 +46,71 @@ pub fn list_supported_mount_points() -> Result<HashMap<ControllerType, PathBuf>,
     Ok(mount_paths)
 }
 
+pub fn get_memory_data(pid: i32) -> Result<LinuxMemory, Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
omprakaash: I do not see any other functions that do this. Would it be a better idea to just move this into the testing crate.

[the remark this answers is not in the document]
YJDoc2: Yep. While it is not a bad idea to have this sort of function, I'd prefer not to add this to a core crate in a PR related to tests. If you are interested, we can open another issue and discuss the API for this ; but for this case, I'll request you to move the functions into the tests itself.

YJDoc2:
@omprakaash ping! There are still some review changes pending, so pinging you

tests/contest/contest/src/utils/linux_resource_memory.rs
omprakaash: Oh ,the linux() function call returns a &Option<Linux>, so as_ref() was needed to make it a Option<&Linux>

crates/libcgroups/src/v1/util.rs
```diff
@@ -79,3 +149,24 @@ pub fn get_subsystem_mount_point(subsystem: &ControllerType) -> Result<PathBuf,
...
+            if subpart == subsystem {
+                return Ok(PathBuf::from(parts[2].to_string()));
+            }
+        }
+    }
+    Err(io::Error::new(
+        io::ErrorKind::Other,
+        format!("subsystem {} not found", subsystem),
+    ))
+}
```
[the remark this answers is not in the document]
omprakaash: Yes there is, but it is tied to V1Manager and is for the current process only.

crates/libcgroups/src/v1/util.rs
```diff
@@ -41,6 +46,71 @@ pub fn list_supported_mount_points() -> Result<HashMap<ControllerType, PathBuf>,
...
+    let mut memory_data = LinuxMemoryBuilder::default();
+    let cgroup_memory_files = vec![
+        "memory.limit_in_bytes",
+        "memory.soft_limit_in_bytes",
+        "memory.memsw.limit_in_bytes",
+        "memory.kmem.limit_in_bytes",
+        "memory.kmem.tcp.limit_in_bytes",
+        "memory.swappiness",
+        "memory.oom_control",
+    ];
```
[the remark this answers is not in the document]
omprakaash: Hmm, I don't think so.

omprakaash:
Sorry for the delay, finally got a chance to work on this.

YJDoc2:
@omprakaash Ping!
You'll also need to rebase on latest main.

saku3:
Thank you for your contribution.

As mentioned in #3285, cgroup v1 is no longer maintained. Unless there is a specific reason to continue working on cgroup v1, new cgroup-related test work should focus on cgroup v2.

If you would like to work on cgroup v2 tests, please see #3605

For this reason, I’m going to close this PR.

PR #2699, opened 2024-02-20
Add missed test-threads=1 to coverage CI

YJDoc2:
In #2685 I missed adding the test threads option to the coverage CI, due to which it still occasionally times out and fails. This adds that.

Also added in feature test script for uniformity, even though those are not failing


YJDoc2:
@containers/youki-maintainers going ahead and merging this as 
1. the test-threads = 1 concept is already reviewed and approved in other PR
2. This only changes CI file and a script, both are validated by passing their corresponding CI

Issue #2695, opened 2024-02-19
Convert unwraps to `?` and improve logging


YJDoc2:
Right now, after a certain point in the init process, the logs do not get forwarded to systemd/journalctl as we change the root mount. After that point we don't send the error back via the main sender to the main process either, so we don't get any error logs. Thus, if the init process fails due to any reason, we only get `Error waiting for init ready, pipe broken` as the log, which makes it hard to debug.

Also, there are several places where we have called `unwrap` on results, due to which the process panics, and as a result the errors are not propagated correctly. I think a better way here is to let the error propagate up by using `?` , so that users of our library (in our case youki) can handle them as they want, which is a better practice for libraries.

YJDoc2:
@utam0k @jprendes May I ask you to take a look at above, and let me know if you have any suggestions, comments or concerns in this? I will work on this accordingly. Thanks!

utam0k:
Good idea. Let's move on to the next step. How about establishing the procedure that produces this error message for starters?

utam0k:
https://github.com/containers/youki/pull/2705#issuecomment-1961344115

yihuaf:
Is there anything still left to do for this issue?

utam0k:
From my view, nothing. @YJDoc2  WDYT?

YJDoc2:
No, we can close this one, and be careful of unwraps in the code changes we do :+1:

PR #2705, opened 2024-02-23
Improve error reporting and logging

YJDoc2:
ref : #2695 

This converts a lot of `.unwrap` calls to propagated errors, and also improves general error reporting that is seen by users of youki.

I'll suggest to review both of the commits separately,  as they focus on two different aspects of this :

- 7f8422f2ef1babbe4b65d86152ade37f74f605bd deals with the issue that when init process fails to exec, it does not report back to main process, thus the error gets reported as `channel broken` which is not at all useful. This now sends the error back via the channel. This also changes the `InvalidArg` error that is given by validation of executable to `ArgValdationError` . I feel the latter is better suited, and does not collide with another error of the same name that we have. This might be a breaking change for runwasi If they are match-ing on the enum values.

- 62431aa0cb8da9814029785784ba39bb305de58f : This converts various `.unwrap` s we have into error propagation so that users of the libraries can decide how they want to deal with it. I have left some unwraps where we first check that it cannot fail, and then call unwrap, as in those places, converting it to more idiomatic ways is either not possible to too complex code. I have also left out unwraps in all the tests, as it does not make sense to convert them. While the error propagation behavior is changed, otherwise the behavior should not change.

I also have some other comments, which I'm leaving as review comments below.

Thanks :)

### DIFF
--- a/crates/libcgroups/src/v2/io.rs
+++ b/crates/libcgroups/src/v2/io.rs
@@ -127,10 +127,12 @@ impl Io {
     fn apply(root_path: &Path, blkio: &LinuxBlockIo) -> Result<(), V2IoControllerError> {
         if let Some(weight_device) = blkio.weight_device() {
             for wd in weight_device {
-                common::write_cgroup_file(
-                    root_path.join(CGROUP_BFQ_IO_WEIGHT),
-                    format!("{}:{} {}", wd.major(), wd.minor(), wd.weight().unwrap()),
-                )?;
+                if let Some(weight) = wd.weight() {
+                    common::write_cgroup_file(
+                        root_path.join(CGROUP_BFQ_IO_WEIGHT),
+                        format!("{}:{} {}", wd.major(), wd.minor(), weight),
+                    )?;
+                }
             }
         }
         if let Some(leaf_weight) = blkio.leaf_weight() {
--- a/crates/libcontainer/src/container/container_checkpoint.rs
+++ b/crates/libcontainer/src/container/container_checkpoint.rs
@@ -13,6 +13,12 @@ use std::os::unix::io::AsRawFd;
 const CRIU_CHECKPOINT_LOG_FILE: &str = "dump.log";
 const DESCRIPTORS_JSON: &str = "descriptors.json";
 
+#[derive(thiserror::Error, Debug)]
+pub enum CheckpointError {
+    #[error("criu error: {0}")]
+    CriuError(String),
+}
+
 impl Container {
     pub fn checkpoint(&mut self, opts: &CheckpointOptions) -> Result<(), LibcontainerError> {
         self.refresh_status()?;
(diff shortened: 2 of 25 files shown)


crates/libcgroups/src/v2/io.rs
```diff
@@ -127,10 +127,12 @@ impl Io {
...
-                common::write_cgroup_file(
-                    root_path.join(CGROUP_BFQ_IO_WEIGHT),
-                    format!("{}:{} {}", wd.major(), wd.minor(), wd.weight().unwrap()),
-                )?;
+                if let Some(weight) = wd.weight() {
+                    common::write_cgroup_file(
+                        root_path.join(CGROUP_BFQ_IO_WEIGHT),
+                        format!("{}:{} {}", wd.major(), wd.minor(), weight),
+                    )?;
+                }
```
YJDoc2, line 135: I have changed this based upon https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/fs/blkio.go#L29-L33 from runc ; and if the weight is `None` the existing code would just panic. However, I'm not sure if this is the correct approach, or the `None` wight should be defaulted to 0 or reported as error.

crates/libcontainer/src/container/container_checkpoint.rs
YJDoc2, line 1: Is this too much? For now we don't support checkpoint at all, so I feel this might be overkill. We can also potentially do these changes when we fix the support for checkpointing.

crates/libcontainer/src/user_ns.rs
```diff
@@ -78,6 +78,8 @@ pub enum UserNamespaceError {
     UnknownUnprivilegedUsernsClone(u8),
     #[error(transparent)]
     IDMapping(#[from] MappingError),
+    #[error(transparent)]
+    OtherIO(#[from] std::io::Error),
```
YJDoc2, line 82: Should the `OtherIO` be some other higher level error, as this is duplicated in another error enum below.

YJDoc2 (commented):
1. I'm not sure is converting the `is_true_root` , `is_in_new_userns` and `rootless_required` this way is a good idea or not. The only error case here is failing to read `/proc/self/uid_map`  , which I'm not sure how much possible it is, and it bloats the signatures and forces to use IO error everywhere it is called. I'd be happy to convert it back to original if someone else also feels the same way
2. While doing this I have found that right now youki does not work with `podman exec` at all. I have found the rc, and have a fix in another branch ; but some unit tests are failing. If that works, then the failing podman tests goes down from 161 to 99 :tada: This PR will help me diagnose further errors, so opening this first and waiting for merge before that one.

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -122,19 +124,19 @@ pub fn container_intermediate_process(
...
                         let buf = format!("{e}");
                         if let Err(err) = write(exec_notify_fd, buf.as_bytes()) {
                             tracing::error!(?err, "failed to write to exec notify fd");
-                            return -1;
                         }
                         if let Err(err) = close(exec_notify_fd) {
                             tracing::error!(?err, "failed to close exec notify fd");
-                            return -1;
                         }
                     }
```
YJDoc2, line 139: I moved the `return -1` from individual arms, as we should still try to close the socket even if we fail to write it, and we are returning -1 at the block end

utam0k:
Overall, it looks good to me, but I have one question. How about using `unwraped_used` in cargo clippy to detect using `unwrap()` in CI.
https://rust-lang.github.io/rust-clippy/master/index.html#/unwrap_used

utam0k (approved):
Awesome improvement!

YJDoc2:
I'll wait maybe a day  or two more and see if anyone has any comments, if not will merge.

YJDoc2:
Going ahead and merging this, as no further comments were added . Thanks :)

PR #2707, opened 2024-02-26
Fix integration test validation CI, make io_priority test conditional

YJDoc2:
After the name change of integration test to contest, this CI file was missed, and as a result we didn't validate the tests even when there was a change in them.

I think implementation of io_priority in runc is still pending on https://github.com/opencontainers/runc/pull/3783 , so made them condition with runc check.

### DIFF
--- a/.github/workflows/integration_tests_validation.yaml
+++ b/.github/workflows/integration_tests_validation.yaml
@@ -22,9 +22,7 @@ jobs:
         with:
           files: |
             .github/workflows/integration_tests_validation.yaml
-            tests/runtimetest/**
-            tests/test_framework/**
-            tests/integration_test/**
+            tests/contest/**
           files_ignore: |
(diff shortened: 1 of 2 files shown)


lengrongfu:
when this pr https://github.com/opencontainers/runc/pull/3783 merge after, we should delete `Box::new(|| !is_runtime_runc()),` this line of code?

YJDoc2:
> when this pr https://github.com/opencontainers/runc/pull/3783 merge after, we should delete Box::new(|| !is_runtime_runc()), this line of code?

Not just PR merge, this also needs to be released by runc in one of their releases. For example, right now they have domainname support on their main, but not in any released versions. Once they release any version with the PR code, we can update runc version in our CI and change the rust test code to previous.

lengrongfu (approved):

PR #2708, opened 2024-02-26
Add linux_devices test

omprakaash:
Adds the linux_devices integration test.


YJDoc2:
Hey, due to a bug in our CI, currently the tests are not validated on runc. May I ask you to wait for https://github.com/containers/youki/pull/2707 to get merged and rebase on it, so we know that the added tests also work with runc? Thanks :)

YJDoc2:
It is merged now, So can you rebase and push?

omprakaash:
Thank you. Should be rebased now

YJDoc2 (commented):
Hey I have a general question regarding this. I think you have copied over the impl from the original go test, which is ok ; but can we do better?

My point here is that if we are only creating three specific devices, then in the test we should only validate those three for the expected values. In the original validation function, we also consider other devices such as tty etc, but I feel we shouldn't...

Can you check if in the original go code, the validation fn is being used for any other test, and if not, can we trim it down here to just check the three devices we control?

Thanks!

omprakaash:
Hello, yes this is a straightforward port from the test in oci-runtime tools and there is another test `validateDefaultDevices` that seems to use this function as well (in oci-runtime tools).

YJDoc2:
> Hello, yes this is a straightforward port from the test in oci-runtime tools and there is another test [validateDefaultDevices](https://github.com/opencontainers/runtime-tools/blob/408c51e934dc558e90e13f63a808f1bd6b155fd2/cmd/runtimetest/main.go#L880) that seems to use this function as well (in oci-runtime tools).

Ok, so do you think we should separate both of these, and have smaller, maybe more modular functions that do tests specific to their test cases? Or this is the best solution, and breaking up function would be worse code than this? This does not have to be done in this PR, just want to know your thoughts.

YJDoc2:
@omprakaash ping!

omprakaash:
Hey sorry for the delay. Will get back to this tomorrow.

omprakaash:
Hey I believe that this is fine, since these are used only for testing and the other test when added, could just call this function. Having separate functions for each type seems unnecessary to me since a lot of the logic seems to overlap for multiple device types, but happy to change this if you strongly feel otherwise.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -381,6 +384,114 @@ pub fn validate_scheduler_policy(spec: &Spec) {
...
+    }
+}
+
+fn validate_device(device: &LinuxDevice, description: &str) {
+    let file_data = match fs::metadata(device.path()) {
+        Ok(data) => data,
+        Err(e) => {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                return;
+            }
```
YJDoc2, line 409: I believe if the device is not present, that needs to be reported as error

tests/contest/runtimetest/src/tests.rs
```diff
@@ -381,6 +384,114 @@ pub fn validate_scheduler_policy(spec: &Spec) {
...
+                minor
+            );
+        }
+    }
+
+    let actual_permissions = file_data.permissions().mode() & 0o777;
+    let expected_permissions = device.file_mode();
+    if expected_permissions.is_none() {
+        return;
+    }
```
YJDoc2, line 463: We should probably get the actual permissions after checking if expected permissions are set

YJDoc2 (commented):
some minor nits

YJDoc2:
>Hey I believe that this is fine, since these are used only for testing and the other test when added, could just call this function. Having separate functions for each type seems unnecessary to me since a lot of the logic seems to overlap for multiple device types, but happy to change this if you strongly feel otherwise.

Fair enough. I don't have any better idea either, so if in future this proves to be an issue, we can fix ti then. Let's keep it as is now :+1:

tests/contest/contest/src/tests/devices/devices_test.rs
```diff
@@ -0,0 +1,71 @@
...
+        .context("failed to create device 1")?;
+
+    let device3 = LinuxDeviceBuilder::default()
+        .path("/dev/test3")
+        .typ(LinuxDeviceType::P)
+        .major(8)
+        .minor(666)
+        .file_mode(432u32)
+        .build()
+        .context("failed to create device 1")?;
```
YJDoc2, line 38: nit: Error messages need to be changed for each device :)

tests/contest/runtimetest/src/tests.rs
```diff
@@ -381,6 +384,119 @@ pub fn validate_scheduler_policy(spec: &Spec) {
...
+
+    if let Some(expected_gid) = device.gid() {
+        if file_data.st_gid() != expected_gid {
+            eprintln!(
+                "error due to device gid want {}, got {}",
+                expected_gid,
+                file_data.st_gid()
+            );
+        }
+    }
```
YJDoc2, line 497: Hey, for these, can we 
- move the perm check in the else clause of is_none check
- not return from uid and gid checks, so both happen irrespective of whether permissions and uid are set or not?

Would this be possible/sensible?

Apart from this, lgtm

omprakaash: Gotcha, I think both are good suggestions.

YJDoc2:
Hey I mad a minor change where we don't return immediately from check function if permissions are not configured, it is similar to how uid/gid validation is done in our implementation.

YJDoc2 (approved):
lgtm!

Thanks @omprakaash for your contribution and following up and doing the requested changes :)

PR #2720, opened 2024-03-04
Fix cgroups determination in exec implementation

YJDoc2:
This fixes the way we determine cgroup name as well as the way we set the cgroup in case of exec. Right now on the main, following would fail even though they should succeed

```
podman run -d --runtime /path/to/youki busybox sleep 10m
podman exec -it --runtime /path/to/youki SHA-HERE /bin/sh
```
This is because 
1. we copy over the namespaces from "main" container to tenant, but not the cgroups
2. In case of rootless, we directly use container id as cgroup name, but that fails when trying to decode the cgroup path

Thus, we fix it here by also copying over the cgroup config from init to tenant, and if cgroup path is not present using `:youki:id` instead .

I have also added a test for exec in the rootless tests we have.

~NOT READY FOR MERGE YET.~

### DIFF
--- a/crates/libcgroups/src/systemd/manager.rs
+++ b/crates/libcgroups/src/systemd/manager.rs
@@ -435,8 +435,11 @@ mod tests {
     use anyhow::{Context, Result};
 
     use super::*;
-    use crate::systemd::dbus_native::{
-        client::SystemdClient, serialize::Variant, utils::SystemdClientError,
+    use crate::{
+        common::DEFAULT_CGROUP_ROOT,
+        systemd::dbus_native::{
+            client::SystemdClient, serialize::Variant, utils::SystemdClientError,
+        },
     };
 
     struct TestSystemdClient {}
@@ -532,5 +535,38 @@ mod tests {
         );
 
         Ok(())
+    }
+    #[test]
+    fn test_task_addition() {
+        let manager = Manager::new(
+            DEFAULT_CGROUP_ROOT.into(),
+            ":youki:test".into(),
+            "youki_test_container".into(),
+            false,
+        )
+        .unwrap();
+        fs::create_dir_all(&manager.full_path).unwrap();
+        let mut p1 = std::process::Command::new("sleep")
+            .arg("1s")
+            .spawn()
+            .unwrap();
+        let p1_id = nix::unistd::Pid::from_raw(p1.id() as i32);
+        let mut p2 = std::process::Command::new("sleep")
+            .arg("1s")
+            .spawn()
+            .unwrap();
+        let p2_id = nix::unistd::Pid::from_raw(p2.id() as i32);
+        manager.add_task(p1_id).unwrap();
+        manager.add_task(p2_id).unwrap();
+        let all_pids = manager.get_all_pids().unwrap();
+        assert!(all_pids.contains(&p1_id));
+        assert!(all_pids.contains(&p2_id));
+        // wait till both processes are finished so we can cleanup the cgroup
+        let _ = p1.wait();
+        let _ = p2.wait();
+        manager.remove().unwrap();
+        // the remove call above should remove the dir, we just do this again
+        // for contingency, and thus ignore the result
+        let _ = fs::remove_dir(&manager.full_path);
     }
 }
--- a/crates/libcgroups/src/systemd/manager.rs
+++ b/crates/libcgroups/src/systemd/manager.rs
@@ -21,7 +21,7 @@ use super::{
 use crate::{
     common::{
         self, AnyCgroupManager, CgroupManager, ControllerOpt, FreezerState, JoinSafelyError,
-        PathBufExt, WrapIoResult, WrappedIoError, CGROUP_PROCS,
+        PathBufExt, WrapIoResult, WrappedIoError,
     },
     systemd::{dbus_native::serialize::Variant, unified::Unified},
     v2::manager::V2ManagerError,
@@ -355,7 +355,8 @@ impl CgroupManager for Manager {
         }
         if self.client.transient_unit_exists(&self.unit_name) {
             tracing::debug!("Transient unit {:?} already exists", self.unit_name);
-            common::write_cgroup_file(self.full_path.join(CGROUP_PROCS), pid)?;
+            self.client
+                .add_process_to_unit(&self.unit_name, "", pid.as_raw() as u32)?;
             return Ok(());
         }
 
@@ -482,6 +483,15 @@ mod tests {
         fn control_cgroup_root(&self) -> Result<PathBuf, SystemdClientError> {
             Ok(PathBuf::from("/"))
         }
+
+        fn add_process_to_unit(
+            &self,
+            _unit_name: &str,
+            _subcgroup: &str,
+            _pid: u32,
+        ) -> Result<(), SystemdClientError> {
+            Ok(())
+        }
     }
 
     #[test]
@@ -545,7 +555,6 @@ mod tests {
             false,
         )
         .unwrap();
-        fs::create_dir_all(&manager.full_path).unwrap();
         let mut p1 = std::process::Command::new("sleep")
             .arg("1s")
             .spawn()
--- a/crates/libcgroups/src/systemd/manager.rs
+++ b/crates/libcgroups/src/systemd/manager.rs
@@ -21,7 +21,7 @@ use super::{
 use crate::{
     common::{
         self, AnyCgroupManager, CgroupManager, ControllerOpt, FreezerState, JoinSafelyError,
-        PathBufExt, WrapIoResult, WrappedIoError,
+        PathBufExt, WrapIoResult, WrappedIoError, CGROUP_PROCS,
     },
     systemd::{dbus_native::serialize::Variant, unified::Unified},
     v2::manager::V2ManagerError,
@@ -355,6 +355,7 @@ impl CgroupManager for Manager {
         }
         if self.client.transient_unit_exists(&self.unit_name) {
             tracing::debug!("Transient unit {:?} already exists", self.unit_name);
+            common::write_cgroup_file(self.full_path.join(CGROUP_PROCS), pid)?;
             return Ok(());
         }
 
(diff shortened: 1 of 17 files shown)


YJDoc2:
With this PR, the number of failed podman tests goes down to 105 from 157 : https://github.com/YJDoc2/youki/actions/runs/8138706191 :tada:

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -353,6 +353,10 @@ impl CgroupManager for Manager {
         if pid.as_raw() == -1 {
             return Ok(());
         }
+        if self.client.transient_unit_exists(&self.unit_name) {
+            tracing::debug!("Transient unit {:?} already exists", self.unit_name);
+            return Ok(());
+        }
```
YJDoc2, line 359: I'm not sure if this  is correct. For exec, the unit will exist (as the container is running) , so we should probably add the pid of the new exec process into it? If so then in case of tenant container, we need a call to set unit props with pid modified? @utam0k may I ask you to take a look at this?

utam0k: In terms of OCI Runtime spec, such a behavior isn't defined. I'm not an expert on systemd, but anyway, we have to put a tenant pid into a cgroup which is the same cgroup of running container pid.

lengrongfu:
LGTM

lengrongfu (approved):

crates/libcgroups/src/systemd/manager.rs
YJDoc2: Hey, fixed this. Now the add_task will either create a new cgroup with the provided pid, or if cgroup already exists, it will add the pid into that. I have verified it using systemd cgroup impl, and changed for cgroups-c2 , but not sure how to verify that, as even docker uses systemd cgroups.

YJDoc2:
@lengrongfu @utam0k May I ask one of you to take a look at the  newer changes I have pushed?

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -327,8 +327,16 @@ impl TenantContainerBuilder {
...
 
+        let linux = spec.linux().as_ref().unwrap();
+        let linux = if linux.cgroups_path().is_some() {
+            LinuxBuilder::default()
+                .namespaces(ns)
+                .cgroups_path(linux.cgroups_path().as_ref().unwrap().clone())
+                .build()?
+        } else {
+            LinuxBuilder::default().namespaces(ns).build()?
+        };
```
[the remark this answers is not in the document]
lengrongfu:
```rust
        let linux_spec = spec.linux().as_ref().unwrap();
        let linux = if linux_spec.cgroups_path().is_some() {
            LinuxBuilder::default()
                .namespaces(ns)
                .cgroups_path(linux_spec.cgroups_path().as_ref().unwrap().clone())
                .build()?
        } else {
            LinuxBuilder::default().namespaces(ns).build()?
        };
```
Maybe it would be better if we rename this var name.

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -353,6 +353,11 @@ impl CgroupManager for Manager {
         if pid.as_raw() == -1 {
             return Ok(());
         }
+        if self.client.transient_unit_exists(&self.unit_name) {
```
utam0k, line 356: Can we add a unit tet for this change?

utam0k (changes requested):
I noticed this leads to breaking changes for existing users due to changing a default cgroup path. In my opinion, affecting rootless users can be allowed, but breaking the change for regular users(non-rootless) is more important. Can we avoid affecting regular users?

crates/libcontainer/src/container/tenant_builder.rs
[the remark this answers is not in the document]
YJDoc2: yep, will do that as well :+1:

YJDoc2:
> I noticed this leads to breaking changes for existing users due to changing a default cgroup path. In my opinion, affecting rootless users can be allowed, but breaking the change for regular users(non-rootless) is more important. Can we avoid affecting regular users?

Hey, while this is breaking , I'm not sure which part specifically are you talking about with respect to default cgroup path. If it is this :  https://github.com/containers/youki/pull/2720/files#diff-36cd5e5a1222b5c819958e4045a242cbff00ae22a8a1699229f5c6220dafdb6bR150-R153

Then, this only changes the when the cgroup path was not set, and I had observed this behavior when creating a tenant container via exec. For the normal container, the path is correctly set and would not change in this. Also the previous behavior of using just the `container_id` as the path is somewhat wrong, as in a valid cgroup path check, we expect it to have two `:` indicating the parent and prefix [here](https://github.com/containers/youki/blob/main/crates/libcgroups/src/systemd/manager.rs#L77-L109) in the systemd manager, which using just the `container_id` is wrong (I have observed that docker and podman with root still use the systemd manager) .

If you meant something else, please correct me. Thanks!

crates/libcgroups/src/systemd/manager.rs
YJDoc2: ok, willl do :+1:

utam0k:
> Hey, while this is breaking , I'm not sure which part specifically are you talking about with respect to default cgroup path. If it is this : https://github.com/containers/youki/pull/2720/files#diff-36cd5e5a1222b5c819958e4045a242cbff00ae22a8a1699229f5c6220dafdb6bR150-R153

Yes

> Then, this only changes the when the cgroup path was not set, and I had observed this behavior when creating a tenant container via exec. For the normal container, the path is correctly set and would not change in this. 

I wonder what happens in the following case:
1. a user runs a container(Container A) with no-set cgroup path with the latest youki version(`v0.3.2`)
2. a user updates the release version, including this PR
3. a user tries `exec` to Container A

In this above case, Every 1 and 3 processes find out different cgroup paths, right?

YJDoc2:
Hey,

>I wonder what happens in the following case:
    a user runs a container(Container A) with no-set cgroup path with the latest youki version(v0.3.2)
    a user updates the release version, including this PR
    a user tries exec to Container A
In this above case, Every 1 and 3 processes find out different cgroup paths, right?

Note : below youki 3.2 = youki on tag v0.3.2 ; branch youki = youki compiled from this branch

So I tried this and when I launch container A (with sudo  pdoman) with youki 3.2 , then try to exec into it with youki 3.2, it gives error of
```
Error: OCI runtime error: thread 'main' panicked at crates/libcontainer/src/process/container_intermediate_process.rs:49:82:
called `Result::unwrap()` on an `Err` value: Systemd(CgroupsPath(MalformedPath("348136f0a314677e10ac6b3d09b2e7430a70ee6b2b01a23239f0bb4f18cfd150")))
exec failed : failed to receive. "waiting for intermediate process". BrokenChannel
```
which is what I mentioned about the default cgroup path being incorrect in my previous comment.

However, if I try to launch container A (with sudo podman) youki 3.2 and exec into it using branch youki , it works due to the cgroups path being set correctly in this branch. Also, I verified that there were two processes in the cgroup of container A when I tried to exec into it using branch youki.

Similarly using branch youki for container A and exec works too. Trying to exec into container launched with branch youki using youki 3.2 also fails due to malformed path.

In short, the exec was broken for both rootful and rootless on v0.3.2 , so I think changing the cgroup path determination wouldn't cause any issues when using containers launched with youki 0.3.2 with branch youki.

Does this address your concern? Do you want me to try any other permutations?

YJDoc2:
Also, I need to check this in detail, but I suspect that changes in the cgroup path determination don't get applied at all due to the changes in setting the cgroups  path [here](https://github.com/containers/youki/pull/2720/files#diff-f1b7c5673621e51900ef703849bf21a2b78474d14349f133a6c82e137a13fb92R331-R339) for exec containers. 

Previously we would not set that path, thus the path determination finding `None` and using `container_id` as path, which lead to the failure. However, after we set the cgroup path, it will always take that as the cgroup path for exec/tenant container. I need to see in which case does the cgroup path can be `None` (I think it can be case if someone is using youki directly withou docker/podman or they are doing some manual cgroup setup) .

utam0k:
> I think it can be case if someone is using youki directly withou docker/podman or they are doing some manual cgroup setup

Yes, I imagined this situation(Sorry for my lack of words 🙏)
Is there any case at all where podman, contained, and CRI-O is not setting the cgroup path? I think it is possible to write a Migration Guide and get away with it.

YJDoc2:
@utam0k I have added a test for systemd cgroup manager regarding the changes here. The CI is failing as the dbus pipe is not set up correctly, so I'll fix that up ; but can you take a look at the test I have added, and check if that approach is ok, or any changes are needed - commit : https://github.com/containers/youki/pull/2720/commits/4a952119fa0d1bf585d0202a4f1402b031075c59 ? Thanks :)

utam0k:
The commit in itself looks acceptable to me. I wonder whether this PR should be written as a breaking change or not.

YJDoc2:
Thanks, I'll work on fixing the CI then.

>The commit in itself looks acceptable to me. 

I'm not sure, while some of the internals have changed, the overall effect is making exec work which was not working previously, thus I added the bug label. I'll think more on this :+1:

> I wonder whether this PR should be written as a breaking change or not.

utam0k:
LGTM except for the CI failed

YJDoc2:
@utam0k please take a look. I have changed the approach of adding the task - before I wrote directly to the cgroup proc file, which could be incorrect / have permission errors ; Now I use the systemd dbus api to add the process via dbus call. I have also fixed the tests, although for the task addition test, I needed to add some more mounts and one hack script so that it can run properly in the cross container.

See these three commits for these new changes : https://github.com/containers/youki/pull/2720/commits/75c851b78d4dab8eb54e3cb868bcb1b271c98a7a https://github.com/containers/youki/pull/2720/commits/3fe7d04ae4fa173ccdf2ab342ab56ec09cfa4e4c https://github.com/containers/youki/pull/2720/commits/df5ebb29ef85b8cd8106227667ea244e38552aae

Thanks :)

utam0k (approved):
Depending on DBus is better than the previous approach, +1

utam0k:
Thanks a lot, @YJDoc2

PR #2725, opened 2024-03-11
Limt dependabot updates to only direct dependencies

utam0k:
The `indirect` type creates PR too much.


YJDoc2 (approved):
I'd prefer if we allow security updates for indirect (but not feature updates) , but for now I'm approving and merging this.

Issue #2723, opened 2024-03-06
Update nix to 0.28.0


YJDoc2:
ref : https://github.com/containers/youki/pull/2712

Nix update has some breaking changes , specifically in terms of deprecated fns , so need to update those to use non-deprecated things instead.

omprakaash:
Hey can I take up this issue ?

YJDoc2:
> Hey can I take up this issue ?

Yep, thanks! Assigned :+1:

zahash:
i think a few types like `ContainerType::TenantContainer { exec_notify_fd: RawFd }` 
must be changed to `ContainerType::TenantContainer { exec_notify_fd: OwnedFd }`

because in a few places like `TenantContainerBuilder::build` we get a `OwnedFd` but soon convert it to a `RawFd`.

```rust
        let (read_end: OwnedFd, write_end: OwnedFd) =
            pipe2(OFlag::O_CLOEXEC).map_err(LibcontainerError::OtherSyscall)?;

        let mut builder_impl = ContainerBuilderImpl {
            container_type: ContainerType::TenantContainer {
                exec_notify_fd: write_end.as_raw_fd(),
            },
            ...
            
```

PR #2728, opened 2024-03-16
Update nix to 0.28.0

omprakaash:
Updates nix to 0.28.0. Changed the deprecated functions to new recommended ones. Another noticeable change was the write() ~and close()~ function~s~ in nix now take in `OwnedFd` instead of `RawFd`. Included changes for them as well.

### DIFF
--- a/crates/libcontainer/src/container/tenant_builder.rs
+++ b/crates/libcontainer/src/container/tenant_builder.rs
@@ -3,6 +3,7 @@ use std::convert::TryFrom;
 use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::io::BufReader;
+use std::os::fd::AsRawFd;
 use std::os::unix::prelude::RawFd;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
@@ -10,7 +11,7 @@ use std::str::FromStr;
 
 use caps::Capability;
 use nix::fcntl::OFlag;
-use nix::unistd::{close, pipe2, read, Pid};
+use nix::unistd::{pipe2, read, Pid};
 use oci_spec::runtime::{
     Capabilities as SpecCapabilities, Capability as SpecCapability, LinuxBuilder,
     LinuxCapabilities, LinuxCapabilitiesBuilder, LinuxNamespace, LinuxNamespaceBuilder,
@@ -126,7 +127,7 @@ impl TenantContainerBuilder {
 
         let mut builder_impl = ContainerBuilderImpl {
             container_type: ContainerType::TenantContainer {
-                exec_notify_fd: write_end,
+                exec_notify_fd: write_end.as_raw_fd(),
             },
             syscall: self.base.syscall,
             container_id: self.base.container_id,
@@ -148,13 +149,19 @@ impl TenantContainerBuilder {
         let mut notify_socket = NotifySocket::new(notify_path);
         notify_socket.notify_container_start()?;
 
-        close(write_end).map_err(LibcontainerError::OtherSyscall)?;
+        // Explicitly close the write end of the pipe here to notify the
+        // `read_end` that the init process is able to move forward. Closing one
+        // end of the pipe will immediately signal the other end of the pipe,
+        // which we use in the init thread as a form of barrier.  `drop` is used
+        // here becuase `OwnedFd` supports it, so we don't have to use `close`
+        // here with `RawFd`.
+        drop(write_end);
 
         let mut err_str_buf = Vec::new();
 
         loop {
             let mut buf = [0; 3];
-            match read(read_end, &mut buf).map_err(LibcontainerError::OtherSyscall)? {
+            match read(read_end.as_raw_fd(), &mut buf).map_err(LibcontainerError::OtherSyscall)? {
                 0 => {
                     if err_str_buf.is_empty() {
                         return Ok(pid);
--- a/crates/libcontainer/src/process/container_intermediate_process.rs
+++ b/crates/libcontainer/src/process/container_intermediate_process.rs
@@ -1,3 +1,5 @@
+use std::os::fd::FromRawFd;
+
 use libcgroups::common::CgroupManager;
 use nix::unistd::{close, write, Gid, Pid, Uid};
 use oci_spec::runtime::{LinuxNamespace, LinuxNamespaceType, LinuxResources};
@@ -130,12 +132,15 @@ pub fn container_intermediate_process(
                     }
                     if let ContainerType::TenantContainer { exec_notify_fd } = args.container_type {
                         let buf = format!("{e}");
-                        if let Err(err) = write(exec_notify_fd, buf.as_bytes()) {
+                        let exec_notify_fd =
+                            unsafe { std::os::fd::OwnedFd::from_raw_fd(exec_notify_fd) };
+                        if let Err(err) = write(&exec_notify_fd, buf.as_bytes()) {
                             tracing::error!(?err, "failed to write to exec notify fd");
                         }
-                        if let Err(err) = close(exec_notify_fd) {
-                            tracing::error!(?err, "failed to close exec notify fd");
-                        }
+
+                        // After sending the error through the exec_notify_fd,
+                        // we need to explicitly close the pipe.
+                        drop(exec_notify_fd);
                     }
                     -1
                 }
@@ -206,7 +211,7 @@ fn setup_userns(
     prctl::set_dumpable(true).map_err(|e| {
         IntermediateProcessError::Other(format!(
             "error in setting dumpable to true : {}",
-            nix::errno::from_i32(e)
+            nix::errno::Errno::from_raw(e)
         ))
     })?;
     sender.identifier_mapping_request().map_err(|err| {
@@ -220,7 +225,7 @@ fn setup_userns(
     prctl::set_dumpable(false).map_err(|e| {
         IntermediateProcessError::Other(format!(
             "error in setting dumplable to false : {}",
-            nix::errno::from_i32(e)
+            nix::errno::Errno::from_raw(e)
         ))
     })?;
     Ok(())
--- a/crates/libcontainer/src/process/fork.rs
+++ b/crates/libcontainer/src/process/fork.rs
@@ -1,5 +1,4 @@
 use std::ffi::c_int;
-use std::fs::File;
 use std::num::NonZeroUsize;
 
 use libc::SIGCHLD;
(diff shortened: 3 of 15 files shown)


utam0k:
>  Another noticeable change was the write() and close() functions in nix now take in OwnedFd instead of RawFd. Included changes for them as well.

Where can I find this change? I couldn't find the change you mentioned.
https://github.com/nix-rust/nix/blob/master/CHANGELOG.md#changed

omprakaash:
Yeah, I could not find that in the change-log too. The change was done in this PR: https://github.com/nix-rust/nix/pull/2100.

crates/libcontainer/src/process/fork.rs
```diff
@@ -166,13 +166,11 @@ fn clone(cb: CloneCb, flags: u64, exit_signal: Option<u64>) -> Result<Pid, Clone
...
         // `::<File>` doesn't have any meaning because we won't use it.
-        mman::mmap::<File>(
+        mman::mmap_anonymous(
             None,
             NonZeroUsize::new(default_stack_size).ok_or(CloneError::ZeroStackSize)?,
             mman::ProtFlags::PROT_READ | mman::ProtFlags::PROT_WRITE,
             mman::MapFlags::MAP_PRIVATE | mman::MapFlags::MAP_ANONYMOUS | mman::MapFlags::MAP_STACK,
-            None,
-            0,
         )
```
YJDoc2, line 174: Need to check if this change is ok (should be, just have to confirm), but in that case, we should also remove the comment above, as not applicable anymore.

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -130,10 +132,11 @@ pub fn container_intermediate_process(
                     }
                     if let ContainerType::TenantContainer { exec_notify_fd } = args.container_type {
                         let buf = format!("{e}");
-                        if let Err(err) = write(exec_notify_fd, buf.as_bytes()) {
+                        let exec_notify_fd = unsafe { std::fs::File::from_raw_fd(exec_notify_fd) };
```
zahash, line 135: why not use `unsafe { std::os::fd::OwnedFd::from_raw_fd( exec_notify_fd ) }`
i think `RawFd` is more semantically related to `OwnedFd` instead of `File`

crates/libcontainer/src/process/fork.rs
```diff
@@ -166,13 +166,11 @@ fn clone(cb: CloneCb, flags: u64, exit_signal: Option<u64>) -> Result<Pid, Clone
     let child_stack = unsafe {
         // Since nix = "0.27.1", `mmap()` requires a generic type `F: AsFd`.
         // `::<File>` doesn't have any meaning because we won't use it.
-        mman::mmap::<File>(
+        mman::mmap_anonymous(
             None,
             NonZeroUsize::new(default_stack_size).ok_or(CloneError::ZeroStackSize)?,
             mman::ProtFlags::PROT_READ | mman::ProtFlags::PROT_WRITE,
             mman::MapFlags::MAP_PRIVATE | mman::MapFlags::MAP_ANONYMOUS | mman::MapFlags::MAP_STACK,
```
anti-entropy123, line 173: It seems we don't need `mman::MapFlags::MAP_ANONYMOUS` anymore.

zahash: yes. i have also confirmed that `mmap_anonymous` has this line

```rust
pub unsafe fn mmap_anonymous(
    addr: Option<NonZeroUsize>,
    length: NonZeroUsize,
    prot: ProtFlags,
    flags: MapFlags,
) -> Result<NonNull<c_void>> {
    // ...

    let flags = MapFlags::MAP_ANONYMOUS | flags;

    // ...
}
```

crates/libcontainer/src/process/container_intermediate_process.rs
omprakaash: Oh Yes ! That is much better. Will change it.

crates/libcontainer/src/process/fork.rs
omprakaash: Yep that is redundant now. Will remove it. Thanks !

utam0k:
@omprakaash It seems CI failed with lint. May I ask you to check it?

omprakaash:
Oh sorry about that, should have fixed the linting issues now.

utam0k:
@omprakaash Thanks!

utam0k:
@zahash @anti-entropy123 @YJDoc2 May I ask you to leave this PR to you?

zahash (approved):
looks good

crates/libcontainer/src/process/container_intermediate_process.rs
YJDoc2: Marking as resolved, as changed accordingly

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -148,13 +149,13 @@ impl TenantContainerBuilder {
         let mut notify_socket = NotifySocket::new(notify_path);
         notify_socket.notify_container_start()?;
 
-        close(write_end).map_err(LibcontainerError::OtherSyscall)?;
+        close(write_end.as_raw_fd()).map_err(LibcontainerError::OtherSyscall)?;
```
anti-entropy123, line 152: Since write_end is `OwnedFd`, why you do not close it by `drop(write_end)`? 

```rust
#[stable(feature = "io_safety", since = "1.63.0")]
impl Drop for OwnedFd {
    #[inline]
    fn drop(&mut self) {
        unsafe {
            ...
            let _ = libc::close(self.fd);
            ...
```
https://doc.rust-lang.org/nightly/src/std/os/fd/owned.rs.html#174-186

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -130,10 +132,12 @@ pub fn container_intermediate_process(
...
+                        let exec_notify_fd =
+                            unsafe { std::os::fd::OwnedFd::from_raw_fd(exec_notify_fd) };
+                        if let Err(err) = write(&exec_notify_fd, buf.as_bytes()) {
                             tracing::error!(?err, "failed to write to exec notify fd");
                         }
-                        if let Err(err) = close(exec_notify_fd) {
+                        if let Err(err) = close(exec_notify_fd.as_raw_fd()) {
                             tracing::error!(?err, "failed to close exec notify fd");
                         }
                     }
```
YJDoc2, line 143: Previously, as everything was `rawFd`, we would manually close it in the last close call. However, now we are converting it to the `ownedFd`, using it for `write` and then doing another manual `close` call on it. As per docs, `OwnedFd` will attempt to close the underlying file description on `Drop`, so here wouldn't that cause to get called `close` twice on same fd? I think that is not error-ing, as the impl of drop [as seen here](https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#174-186) says (essentially) `we ignore errors so to not accidentally close another valid file desc` . Do we calling close explicitly here can run into the same problem?

Should we instead convert to `ownedFd` for the write call, let its drop take care of closing the fd, and remove the close call altogether?

crates/libcontainer/src/container/tenant_builder.rs
YJDoc2, line 152: Same question as the other comment, `write_end` as an `OwnedFd` will call close on drop ; so should we not call close here? If not, a better option to force the closing would be manually calling `drop()` on it at this point.

crates/libcontainer/src/process/fork.rs
```diff
@@ -187,7 +185,7 @@ fn clone(cb: CloneCb, flags: u64, exit_signal: Option<u64>) -> Result<Pid, Clone
 
     // Since the child stack for clone grows downward, we need to pass in
     // the top of the stack address.
-    let child_stack_top = unsafe { child_stack.add(default_stack_size) };
+    let child_stack_top = unsafe { child_stack.as_ptr().add(default_stack_size) };
```
YJDoc2, line 188: Why is `as_ptr` needs to be added, when the return type of (original) `mmap` and (new) `mmap_anonymous` is same of `Result<NonNull<c_void>>` ?

YJDoc2 (changes requested):
While other changes are ok, I have some points of concern here regarding the mix of `OwnedFd` and `RawFd` . Main point of concern for this PR, is that in several places, we are converting a RawFd to OwnedFd, and then calling close on it manually when the OwnedFd gets closed in its drop. I feel this should be addressed here.

crates/libcontainer/src/container/tenant_builder.rs
YJDoc2: You beat me to it :joy: https://github.com/containers/youki/pull/2728#discussion_r1547168975

crates/libcontainer/src/process/fork.rs
YJDoc2: Hey, this still needs to be addressed, the function change is valid, so comment should be updated/removed.

crates/libcontainer/src/process/fork.rs
anti-entropy123: > Why is `as_ptr` needs to be added, when the return type of (original) `mmap` and (new) `mmap_anonymous` is same of `Result<NonNull<c_void>>` ?

Perhaps because with nix version "0.27.1", the mmap function returns a `Result<*mut c_void>`.

YJDoc2: oh yes, I was comparing both from 0.28.0, my bad. Thanks!

crates/libcontainer/src/process/container_intermediate_process.rs
omprakaash: Ah yes good point, I missed this. Will look into this. I think letting the drop automatically close the fd would be good, since OwnedFd has that feature built in.

crates/libcontainer/src/container/tenant_builder.rs
omprakaash: Yes, since drop is automatically called at end of scope, the close() seems to be redundant.

YJDoc2:
Hey, thanks for updating. Two notes :
- there's conflict in lock file, you can recreate it or copy it over from master and run build to fix it
- Can you add explicit `drop` call where there was close before, and comment the reasoning, maybe linking this PR there? That way in future we know for sure that we have taken care of the fds and the reason for calling drop instead of close

YJDoc2:
Hey @omprakaash , it seems there are still some more conflicts here. In case you are busy, is it ok if I resolve them and push? because the conflicts are in cargo* files, not code, it would be pretty straight forward to resolve. Apart from that, the code is lgtm.

omprakaash:
Hey sorry for the late reply. I did not have the chance to look at this and missed your comment. Would be great if you could that. Thank you !

yihuaf:
I will take care of this.

yihuaf:
Merge conflict is resolved. I will look over the review comments over the weekend and address them.

yihuaf:
@YJDoc2 This is ready for review one last time. I have addressed the merge conflicts and went over the reviews. One note, the closing of the pipes (The fds that were converted from `RawFd` to `OwnedFd`) is actually required. We use the pipes as both a way to communicate between processes and as a barrier for synchronization. The finer control to explicitly close is required, instead of letting the scope takes care of the close through `Drop` trait. By closing one end of the pipe, we immediately signal to the other end of the pipe, so that any blocking calls on the pipe will immediately return. The PR already included the change, but I changed the comments to reflect this information. The previous comment implies that the `drop` was there only to maintain the legacy code, but the explicit close actual serves a purpose in this case, not just because we have to manually close due to the use of `RawFd`.

Let me know if there is anything else you would like to make change.

YJDoc2 (approved):
Code looks good, but unfortunately there's another conflict due to latest release. Approving so that after resolving and @utam0k 's approval, this can get merged. Thank you @omprakaash and @yihuaf for taking care of this :pray:

utam0k (approved):
Thanks!

yihuaf:
close #2723

PR #2729, opened 2024-03-16
Init a seccomp project

utam0k:
This is an experimental crate. I'll implement a seccomp crate step-by-step. 
When it is mature, I'll move it under the `crate/` and use it instead of libseccomp.

ref: https://github.com/containers/youki/issues/2724


YJDoc2:
I'll try to take a look at this later today :+1:

saschagrunert (approved):

experiment/seccomp/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "seccomp"
```
YJDoc2, line 2: There is already a crate on crates.io named seccomp ; maybe we should rename this to something else if we are considering publishing this eventually, maybe seccomp-rs ?

experiment/seccomp/src/instruction/arch.rs
```diff
@@ -0,0 +1,20 @@
+use crate::instruction::Instruction;
+use crate::instruction::*;
+
+pub enum Arch {
+    X86,
+}
+
+impl Arch {
+    pub fn gen_validate(&self) -> Vec<Instruction> {
```
YJDoc2, line 9: For this, can we make this as an interface? For example, something like we have done in cgroups or syscall, so we have a function that will return dyn Arch, and we call method on it.

Alternatively, if Arch itself does not need to hold any data, its just an enum, can we instead have an empty struct, and have static methods (no &self) , and inside those methods, we either match over the arch at runtime, or do conditional compilation based on cfg. That way adding support for new arch will be easier.

experiment/seccomp/src/instruction/consts.rs
```diff
@@ -0,0 +1,63 @@
...
+// struct seccomp_data {
+//     int nr;
+//     __u32 arch;
+//     __u64 instruction_pointer;
+//     __u64 args[6];
+// };
+// ```
+pub const SECCOMP_DATA_ARCH_OFFSET: u8 = 4;
+pub const SECCOMP_DATA_ARGS_OFFSET: u8 = 16;
+pub const SECCOMP_DATA_ARG_SIZE: u8 = 8;
```
YJDoc2, line 61: While I think int is always 4 byte wide on all platforms, should we do a size_of::<libc::c_int> or something so that we correctly get size of the int?

experiment/seccomp/src/instruction/inst.rs
```diff
@@ -0,0 +1,69 @@
...
+            multiuse_field,
+        }
+    }
+
+    pub fn jump(
+        code: c_ushort,
+        multiuse_field: c_uint,
+        jump_true: c_uchar,
+        jump_false: c_uchar,
+    ) -> Self {
```
YJDoc2, line 34: Can we keep order of params same as new (so multiuse_field is last in both)?

experiment/seccomp/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+        SockType::Stream,
+        None,
+        SockFlag::empty(),
+    )?;
+
+    let _ = prctl::set_no_new_privileges(true);
+
+    let mut bpf_prog = Arch::X86.gen_validate();
+    bpf_prog.append(&mut vec![
+        // A: Check if syscall is getpid
```
YJDoc2, line 71:
```suggestion
        // A: Check if syscall is getcwd
```
?

YJDoc2 (approved):
Hey, great as a starting point! I have left some comments, which we might want to consider, but we can consider them in later commits, this is good to go, so approving.

Also, eventually when we have enough of lib ready, we should consider portin https://github.com/seccomp/libseccomp/tree/main/tests tests over here, so we can validate our impl. I think in their CI they compile and link these tests against branch build, which in case of rust is similar to unit tests, so we should be able to use those examples as part of crate/top level unit tests.

YJDoc2:
@utam0k Should I go ahead and merge this?

experiment/seccomp/Cargo.toml
utam0k: Yes, I know. I gonna change the name when it is published.

utam0k:
> @utam0k Should I go ahead and merge this?

Thanks for your review. I'll fix them on this PR

experiment/seccomp/src/instruction/consts.rs
utam0k: I'm thinking we can use `offset_of!` . It would be better in this use case.  WDYT?
https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#offset_of

YJDoc2: Yep, that would be much better and stable :+1:

experiment/seccomp/src/instruction/arch.rs
utam0k: I prefer later 👍
3266094a

experiment/seccomp/src/instruction/consts.rs
utam0k: Okay, I'll handle it in another PR.

utam0k:
@YJDoc2 Do you want to take another look?

experiment/seccomp/src/instruction/consts.rs
YJDoc2: yep :+1:

YJDoc2 (approved):
lgtm :+1:

Can merge once the CI is done :+1:

utam0k:
@YJDoc2 Thanks for your quick review 😍

PR #2763, opened 2024-04-17
seccomp: Use offset_of!

utam0k:
`offset_of` is better because it allows us not to consider the difference between arches.
https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#offset_of

### DIFF
--- a/experiment/seccomp/src/instruction/consts.rs
+++ b/experiment/seccomp/src/instruction/consts.rs
@@ -71,7 +71,7 @@ pub const fn seccomp_data_arch_offset() -> u8 {
     offset_of!(SeccompData, arch) as u8
 }
 
-pub const fn seccomp_data_arg_size_offset() -> u8 {
+pub const fn seccomp_data_arg_size() -> u8 {
     8
 }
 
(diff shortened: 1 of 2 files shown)


experiment/seccomp/src/instruction/consts.rs
```diff
@@ -56,8 +58,51 @@ pub const AUDIT_ARCH_AARCH64: u32 = 183 | 0x8000_0000 | 0x4000_0000;
...
+    args: [u64; 6],
+}
+
+pub const fn seccomp_data_arch_offset() -> u8 {
+    offset_of!(SeccompData, arch) as u8
+}
+
+pub const fn seccomp_data_arg_size_offset() -> u8 {
+    8
+}
```
YJDoc2, line 76: Does this need the `offset` in name?
```suggestion
pub const fn seccomp_data_arg_size() -> u8 {
    8
}
```

YJDoc2 (commented):
One minor nitpick, otherwise lgtm.
Also, I think we should label all seccomp package related prs with `experimental`, and not use other labels for these, so the release notes will not mix youki changes with these.

PR #2775, opened 2024-04-29
seccomp: Add a case for checking arguments

utam0k:
As the title says.

Part of https://github.com/containers/youki/issues/2724

Next, I'd like to make it abstract(e.g., SeccompRule) based on this PR.

### DIFF
--- a/experiment/seccomp/Cargo.lock
+++ b/experiment/seccomp/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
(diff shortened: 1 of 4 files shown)


YJDoc2 (approved):
RSLGTM
I don't understand all of the changes here, but  in general looks good, and experimental changes, so approving.

Also, I hope the async is only for the demo/example, I'd prefer the actual library code to not have any async in it if possible.

utam0k:
> Also, I hope the async is only for the demo/example, I'd prefer the actual library code to not have any async in it if possible.

Of course, I intend to use async/await in main.rs for seccomp notify.

PR #2780, opened 2024-05-03
Remove unnecessary chdir

utam0k:
Fix: https://github.com/containers/youki/issues/2772

These `chdir` are unexpected for `libcontainer`'s users.

### DIFF
--- a/Vagrantfile.containerd2youki
+++ b/Vagrantfile.containerd2youki
@@ -44,8 +44,7 @@ Vagrant.configure("2") do |config|
         ./script/setup/install-cni
         ./script/setup/install-critools
         rm -rf /bin/runc /sbin/runc /usr/sbin/runc /usr/bin/runc
-
-        cp /vagrant/youki/youki /usr/bin/runc
+        ln -s /vagrant/youki/youki /usr/bin/runc
       SHELL
     end
 
--- a/crates/libcontainer/src/container/builder_impl.rs
+++ b/crates/libcontainer/src/container/builder_impl.rs
@@ -1,3 +1,9 @@
+use std::{fs, io::Write, os::unix::prelude::RawFd, path::PathBuf, rc::Rc};
+
+use libcgroups::common::CgroupManager;
+use nix::unistd::Pid;
+use oci_spec::runtime::Spec;
+
 use super::{Container, ContainerStatus};
 use crate::{
     error::{LibcontainerError, MissingSpecError},
@@ -13,10 +19,6 @@ use crate::{
     utils,
     workload::Executor,
 };
-use libcgroups::common::CgroupManager;
-use nix::unistd::Pid;
-use oci_spec::runtime::Spec;
-use std::{fs, io::Write, os::unix::prelude::RawFd, path::PathBuf, rc::Rc};
 
 pub(super) struct ContainerBuilderImpl {
     /// Flag indicating if an init or a tenant container should be created
@@ -82,7 +84,11 @@ impl ContainerBuilderImpl {
 
         if matches!(self.container_type, ContainerType::InitContainer) {
             if let Some(hooks) = self.spec.hooks() {
-                hooks::run_hooks(hooks.create_runtime().as_ref(), self.container.as_ref())?
+                hooks::run_hooks(
+                    hooks.create_runtime().as_ref(),
+                    self.container.as_ref(),
+                    None,
+                )?
             }
         }
 
--- a/crates/libcontainer/src/container/container_start.rs
+++ b/crates/libcontainer/src/container/container_start.rs
@@ -48,16 +48,6 @@ impl Container {
             err
         })?;
         if let Some(hooks) = config.hooks.as_ref() {
-            let original_dir = env::current_dir().map_err(|err| {
-                tracing::error!("failed to get current directory: {}", err);
-                LibcontainerError::Other(format!("failed to get current directory: {}", err))
-            })?;
-
-            unistd::chdir(self.root.as_os_str()).map_err(|err| {
-                tracing::error!("failed to change directory to container root: {}", err);
-                LibcontainerError::OtherSyscall(err)
-            })?;
-
             // While prestart is marked as deprecated in the OCI spec, the docker and integration test still
             // uses it.
             #[allow(deprecated)]
@@ -68,11 +58,6 @@ impl Container {
                 let _ = self.kill(signal::Signal::SIGKILL, true);
 
                 err
-            })?;
-
-            unistd::chdir(original_dir.as_path()).map_err(|err| {
-                tracing::error!("failed to change directory to container root: {}", err);
-                LibcontainerError::OtherSyscall(err)
             })?;
         }
 
@@ -88,9 +73,24 @@ impl Container {
         // Run post start hooks. It runs after the container process is started.
         // It is called in the runtime namespace.
         if let Some(hooks) = config.hooks.as_ref() {
+            let original_dir = env::current_dir().map_err(|err| {
+                tracing::error!("failed to get current directory: {}", err);
+                LibcontainerError::Other(format!("failed to get current directory: {}", err))
+            })?;
+
+            unistd::chdir(self.root.as_os_str()).map_err(|err| {
+                tracing::error!("failed to change directory to container root: {}", err);
+                LibcontainerError::OtherSyscall(err)
+            })?;
+
             hooks::run_hooks(hooks.poststart().as_ref(), Some(self)).map_err(|err| {
                 tracing::error!("failed to run post start hooks: {}", err);
                 err
+            })?;
+
+            unistd::chdir(original_dir.as_path()).map_err(|err| {
+                tracing::error!("failed to change directory to container root: {}", err);
+                LibcontainerError::OtherSyscall(err)
             })?;
         }
 
--- a/crates/libcontainer/src/container/container_start.rs
+++ b/crates/libcontainer/src/container/container_start.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 
 use super::{Container, ContainerStatus};
-use nix::{sys::signal, unistd};
+use nix::sys::signal;
 
 impl Container {
     /// Starts a previously created container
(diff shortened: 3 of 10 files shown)


utam0k:
I've checked the relevant tests that came from containerd were passed.
https://github.com/containers/youki/pull/2777#issuecomment-2091658540

Vagrantfile.containerd2youki
```diff
@@ -44,8 +44,7 @@ Vagrant.configure("2") do |config|
         ./script/setup/install-cni
         ./script/setup/install-critools
         rm -rf /bin/runc /sbin/runc /usr/sbin/runc /usr/bin/runc
-
-        cp /vagrant/youki/youki /usr/bin/runc
+        ln -s /vagrant/youki/youki /usr/bin/runc
```
utam0k, line 47: It'd be more beneficial because we don't need to copy again when we rebuild.

utam0k:
I'll check CI.

utam0k:
@Furisto Can you take a look at this PR?

Mossaka (approved):

crates/libcontainer/src/container/container_start.rs
```diff
@@ -59,11 +59,6 @@ impl Container {
             })?;
         }
 
-        unistd::chdir(self.root.as_os_str()).map_err(|err| {
-            tracing::error!("failed to change directory to container root: {}", err);
-            LibcontainerError::OtherSyscall(err)
-        })?;
-
```
YJDoc2, line 66: Would there be any issues with the hook behavior by running this? I checked the spec once, and I don't think it requires to run the hooks in same dir as the container, only in the same namespace, but just to confirm.

utam0k: Good catch! You are right, thanks.
444146eb

YJDoc2: Hey, I think we need to re-change the cwd back to the original one after the `run_hooks` call, so that the dir is correct? Otherwise now the dir will change when hooks are there, and won't when they're not, it'll be inconsistent behavior.

Also, When I originally commented , I was talking about the hooks invocation code below this, [line 77->85.](https://github.com/containers/youki/pull/2780/files#diff-e21c19897fbff6f13bb303b180c4879d40d96f5a30a65497eb4f605fc3fb8341L78-L83)

Given that the existing invocation of hooks above  [line 52->60](https://github.com/containers/youki/pull/2780/files#diff-e21c19897fbff6f13bb303b180c4879d40d96f5a30a65497eb4f605fc3fb8341R56-R63) didn't have chdir before, I don't think it is needed now?

Also my concern while raising this point was that I'm not sure if hooks even need the chdir, as the spec says `in the same namespace` which does include the mount namespace, but does not say anything about the cwd if I understand it correctly.

utam0k: 524ae7a8

You are probably right, but if we delete it, we must be responsible for the migration guide or something. So I don't prefer introducing such breaking change in this PR.
If we do, we should create another PR.

> Also my concern while raising this point was that I'm not sure if hooks even need the chdir, as the spec says in the same namespace which does include the mount namespace, but does not say anything about the cwd if I understand it correctly.

YJDoc2: Hey, I agree that we should not have breaking changes here.
But I think I'm still not able to convey my concern properly - The changes you have done are around the pre-start hooks, which are marked deprecated. I don't think this hook invocation was running in the root dir before, so for these hooks, these changes are breaking. On the other hand, there is the invocation of post-start hook _after_ the currently removed chdir which should be added between two chdir calls. Let me know if you still have questions, maybe I can just make a patch PR to you fork to make it better understandable.

YJDoc2:
@Mossaka can you confirm that this branch's youki build resolved the original perf issues you were facing? So that once this PR is checked from the our side, we can merge and make a release.

crates/libcontainer/src/container/container_start.rs
YJDoc2: I opened https://github.com/utam0k/youki/pull/528 on your fork.

utam0k: Ops, you are right. I misunderstood the point you pointed out. Is it enough?
8de223377e21d5b1e9f1ce3c5d56e6d22f60e601

crates/libcontainer/src/container/container_start.rs
```diff
@@ -88,10 +73,25 @@ impl Container {
         // Run post start hooks. It runs after the container process is started.
         // It is called in the runtime namespace.
         if let Some(hooks) = config.hooks.as_ref() {
+            let original_dir = env::current_dir().map_err(|err| {
+                tracing::error!("failed to get current directory: {}", err);
+                LibcontainerError::Other(format!("failed to get current directory: {}", err))
+            })?;
+
+            unistd::chdir(self.root.as_os_str()).map_err(|err| {
```
jprendes, line 81: Instead of chaging directory here and then reverting it, could we set `current_dir` in the new process instead?

Here: https://github.com/containers/youki/blob/c6fffb2c6a59b5fc2efc85189d485d322c5454e8/crates/libcontainer/src/hooks.rs#L39

```diff
-let mut hook_command = process::Command::new(hook.path());
+let mut hook_command = process::Command::new(hook.path())
+    .current_dir(root_dir);
```

Mossaka:
> @Mossaka can you confirm that this branch's youki build resolved the original perf issues you were facing? So that once this PR is checked from the our side, we can merge and make a release.

Just confirmed, it solved the original perf issue we are facing

Mossaka (approved):

crates/libcontainer/src/tty.rs
```diff
@@ -165,84 +165,72 @@ fn connect_stdio(stdin: &RawFd, stdout: &RawFd, stderr: &RawFd) -> Result<()> {
...
-    fn setup() -> Result<(tempfile::TempDir, PathBuf, PathBuf)> {
-        let testdir = tempfile::tempdir()?;
-        let rundir_path = Path::join(testdir.path(), "run");
-        fs::create_dir(&rundir_path)?;
-        let socket_path = Path::new(&rundir_path).join("socket");
-        let _ = File::create(&socket_path);
-        env::set_current_dir(&testdir)?;
-        Ok((testdir, rundir_path, socket_path))
-    }
-
```
YJDoc2, line 186: Hey, what is the reason behind changing this logic? I'm not saying changing it is wrong, just not sure.

crates/libcontainer/src/hooks.rs
```diff
@@ -27,16 +28,35 @@ pub enum HookError {
...
+            } else {
+                cmd.current_dir(
+                    env::current_dir()
+                        .map_err(|err| {
+                            tracing::error!("failed to get current directory: {}", err);
+                            HookError::OtherIO(err)
+                        })?
+                        .as_path(),
+                )
+            };
```
YJDoc2, line 59: I don't think the else section is needed, as by default cmd is run in the calling process's directory: ref - third example in the docs [here](https://doc.rust-lang.org/std/process/struct.Command.html)

YJDoc2 (commented):
Hey, can you take a look at two comments I have? Apart from them, it is lgtm

crates/libcontainer/src/tty.rs
utam0k: Just refactoring. I don't intend to change the core logic of these tests.

crates/libcontainer/src/hooks.rs
utam0k: Initially, I want to ensure cmd works fine even if the execution of `chdir` after this line.  But I thought about it again, we can notice the cover in the unite test, so I'll turn it off.

utam0k: 73cffe381028c034686d042a6e11de9649a59a3b

crates/libcontainer/src/hooks.rs
```diff
@@ -27,16 +27,28 @@ pub enum HookError {
...
+) -> Result<()> {
     let state = &(container.ok_or(HookError::MissingContainerState)?.state);
 
     if let Some(hooks) = hooks {
         for hook in hooks {
-            let mut hook_command = process::Command::new(hook.path());
+            let mut hook_command = &mut process::Command::new(hook.path());
+
+            if let Some(cwd) = cwd {
+                // This is error. please fix it
```
jprendes, line 48: What do you mean with this comment?

crates/libcontainer/src/hooks.rs
```diff
@@ -27,16 +27,28 @@ pub enum HookError {
...
+                hook_command = hook_command.current_dir(cwd);
```
jprendes, line 49:
```suggestion
                hook_command.current_dir(cwd);
```

crates/libcontainer/src/hooks.rs
utam0k: 🤦‍♂

YJDoc2 (approved):
LGTM :+1:
Not merging it in case you want to do something before release, or anyone else has a review comment

Thanks @utam0k !

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -1,3 +1,9 @@
+use std::{fs, io::Write, os::unix::prelude::RawFd, path::PathBuf, rc::Rc};
+
+use libcgroups::common::CgroupManager;
+use nix::unistd::Pid;
+use oci_spec::runtime::Spec;
+
```
jprendes, line 6: We should have a `rustfmt.toml` to standardize this.
I personally like the one that @Mossaka put together for runwasi.
https://github.com/containerd/runwasi/blob/main/rustfmt.toml

crates/libcontainer/src/hooks.rs
```diff
@@ -27,16 +27,27 @@ pub enum HookError {
     MissingContainerState,
     #[error("failed to write container state to stdin")]
     WriteContainerState(#[source] std::io::Error),
+    #[error("io error")]
+    OtherIO(#[source] std::io::Error),
```
jprendes, line 31: I don't see where this is used.

jprendes (approved):
Thanks @utam0k !
I left some minor comments.
Otherwise, LGTM.

crates/libcontainer/src/container/builder_impl.rs
utam0k: I didn't know that. It is what I want! Thanks for sharing.

crates/libcontainer/src/hooks.rs
utam0k: https://github.com/containers/youki/pull/2780/commits/4a4975954e4c8ee1b88c176789225eef37f7a48a

PR #2787, opened 2024-05-15
Add `rustfmt.toml` to standardize formatting

jprendes:
This PR adds a `rustfmt.toml` which configures rustfmt.
In particular, it specifies formatting for imports grouping:
```toml
group_imports = "StdExternalCrate" # create three groups for std, external and local crates
imports_granularity = "Module" # Merge imports from the same module
```

These options are still unstable, so they need to be run with `cargo +nightly fmt ...`.

For reviewing it's easier to review it by commit instead of as a whole.
* The first commit adds `rustfmt.toml` and updates the `cargo.sh` script to use nightly for `cargo fmt`.
* The second commit is the result of running `just format`.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -44,7 +44,9 @@ jobs:
       - name: Setup Rust toolchain and cache
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          components: rustfmt, clippy
+          components: clippy
+      - name: Install nightly rustfmt
+        run: rustup toolchain install nightly --component rustfmt --profile minimal --no-self-update
       - name: typos-action
         uses: crate-ci/typos@v1.14.12
       - name: Install just
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,6 @@
+# Ignore commits in the blame view
+# see:
+#   https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
+
+# Bulk apply formatting
+13a3fd516e9904d5cd7ef20ead76dd70cd172fe7
--- a/crates/libcgroups/examples/bpf.rs
+++ b/crates/libcgroups/examples/bpf.rs
@@ -2,19 +2,15 @@ use anyhow::Result;
 
 #[cfg(feature = "cgroupsv2_devices")]
 mod bpf {
-    use anyhow::{bail, Result};
-    use clap::Arg;
-    use clap::Command;
-    use oci_spec::runtime::LinuxDeviceCgroup;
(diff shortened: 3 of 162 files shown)


.github/workflows/basic.yml
```diff
@@ -44,7 +44,9 @@ jobs:
       - name: Setup Rust toolchain and cache
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          components: rustfmt, clippy
+          components: clippy
+      - name: Install nightly rustfmt
+        run: rustup toolchain install nightly --component rustfmt --profile minimal --no-self-update
```
YJDoc2, line 49: nit: can we add the nightly directly in the setup rust step? If not can we have a comment here mentioning we use this for formatting?

YJDoc2:
Hey can we add a rev-ignore file [like this](https://github.com/blender/blender/blob/main/.git-blame-ignore-revs) , and add the second commit to it? That way the blame would not consider the formatting changes. Corresponding git doc : https://github.com/git/git/blob/ae3f36dea16e51041c56ba9ed6b38380c8421816/Documentation/blame-options.txt#L113-L125

Not sure how well the blame would pick up the changes , but given that blender uses it for its formatting commits, it should be good-enough for us to start with this PR. Otherwise the setup commit looks, good, ci passed, so assuming that works :+1:

Thanks!

.github/workflows/basic.yml
jprendes: > can we add the nightly directly in the setup rust step?

That would result in using nightly `clippy`, which is not my intention here.
I'm open to suggestions :-)

jprendes:
The ignore file seems to be working:
* with ignore: https://github.com/jprendes/youki/blame/rustfmt/crates/libcontainer/src/hooks.rs
* without ignore: https://github.com/jprendes/youki/blame/rustfmt~/crates/libcontainer/src/hooks.rs

Bypassing the ignore file is done appending a `~` to the SHA/branch in the URL.

.github/workflows/basic.yml
YJDoc2: No, I can't think of anything better. We can keep this as is :+1:

utam0k:
> Hey can we add a rev-ignore file [like this](https://github.com/blender/blender/blob/main/.git-blame-ignore-revs) , and add the second commit to it? That way the blame would not consider the formatting changes. Corresponding git doc : https://github.com/git/git/blob/ae3f36dea16e51041c56ba9ed6b38380c8421816/Documentation/blame-options.txt#L113-L125
> 
> Not sure how well the blame would pick up the changes , but given that blender uses it for its formatting commits, it should be good-enough for us to start with this PR. Otherwise the setup commit looks, good, ci passed, so assuming that works 👍
> 
> Thanks!

I didn't know that. Good idea.

PR #2799, opened 2024-05-25
Add dependencies that support the loongarch64 architecture

yzewei:
Add dependencies that support the loongarch64 architecture
https://github.com/rust-lang/libc/pull/2765

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -533,7 +533,7 @@ dependencies = [
  "flate2",
  "libcgroups",
  "libcontainer",
- "nix 0.28.0",
+ "nix",
  "num_cpus",
  "oci-spec",
  "once_cell",
@@ -1916,7 +1916,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "800bccc21b216764f96334a241a072b842121843bf679f5a03b0c2c21cb339ed"
 dependencies = [
  "cc",
- "nix 0.28.0",
+ "nix",
  "pkg-config",
 ]
 
@@ -1938,7 +1938,7 @@ dependencies = [
  "libbpf-sys",
  "libc",
  "mockall",
- "nix 0.28.0",
+ "nix",
  "oci-spec",
  "procfs",
  "quickcheck",
@@ -1965,7 +1965,7 @@ dependencies = [
  "libcgroups",
  "libseccomp",
  "nc",
- "nix 0.28.0",
+ "nix",
  "oci-spec",
  "once_cell",
  "prctl",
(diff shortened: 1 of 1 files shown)


utam0k (changes requested):
Thanks for your first contribution! Let me know why this change leads to support the loongarch64 arch?

yzewei:
> Thanks for your first contribution! Let me know why this change leads to support the loongarch64 arch?

Thanks for your reply!
After modifying the libc version, this project can be compiled on loongarch. This is because the latest version of libc introduces support for loongarch. For details, see https://github.com/rust-lang/libc/pull/2765.

utam0k:
@yzewei I understood, but we don't have the CI platform to test the loongarch64 arch. It means, unfortunately, it's hard to maintain it for us. Do you have any good ideas?

yzewei:
> @yzewei I understood, but we don't have the CI platform to test the loongarch64 arch. It means, unfortunately, it's hard to maintain it for us. Do you have any good ideas?

@utam0k 
The cross-rs project already supports loongarch64. I think it can support loongarch64 like aarch64. What do you think? If it's ok, I will initiate the pr of adding loongarch64 corresponding to the ci test.

utam0k:
@yzewei Thanks for your response. Who will maintain the loongarch64 arch? In other words, is it worth the maintenance? Would you use it? Sorry for the spicy question, but we consider it because our bandwidth is very limited.

yzewei:
> @yzewei Thanks for your response. Who will maintain the loongarch64 arch? In other words, is it worth the maintenance? Would you use it? Sorry for the spicy question, but we consider it because our bandwidth is very limited.

@utam0k It doesn't matter, the reason why I submitted pr is because the alpine community is about to support loongacrh. As an open-source contributor to the loongacrh architecture, I found on aports that youki has not updated libc, resulting in the failure of the package compilation under loongarch. If you add loongarch support, I believe that with the active number of people in the alpine community, there will be a lot of people using it

utam0k:
> It doesn't matter,

Why so? Who will continue to maintain the loongarch64 arch in youki? Do you want to contribute youki to supporting it?

> If you add loongarch support, I believe that with the active number of people in the alpine community, there will be a lot of people using it

Do you run containers on Alpine?

Also, we need to review our dependencies. For example https://github.com/opencontainers/runtime-spec/pull/1187

utam0k:
ref: https://github.com/opencontainers/runc/pull/3765#issuecomment-1464543237

yihuaf:
> Who will maintain the loongarch64 arch?

I share the same concern. We have limited bandwidth to support an extra architecture at the moment. 

With that being said, I am unclear how dropping the explicit `nix` version would help with the latest `libc`. Since we don't have the hardware with the given architecture, we won't be able to meaningfully test this. If you can provide us with a bit more details, including how to cross compile to verify, it would be much more helpful for the review.

Issue #2718, opened 2024-03-03
Implement a crate like opencontainers/selinux


Gekko0114:
# background
In this PR https://github.com/containers/youki/pull/2688, it was found that the implementation of linux mount label is different between runc and youki.

> I have some question regarding the MountLabel implementation itself : As per the oci spec, the MountLabel field marks the selinux label to be used for that mount. Thus in runc https://github.com/opencontainers/runc/blob/d5e4c33001d74176222fe8f48a323f3e8ad89999/libcontainer/rootfs_linux.go#L536 , they use the selinux package and set the label using that. However in youki, we are passing that from https://github.com/containers/youki/blob/main/crates/libcontainer/src/rootfs/rootfs.rs#L90 to the syscall implementation, which finally reaches and then goes https://github.com/containers/youki/blob/main/crates/libcontainer/src/syscall/linux.rs#L471 . Where it finally calls nix::mount. I'm not sure how that is supposed to correspond to selinux. Can you check if there is some issue with out implementation , or am I doing some wrong code follow?

Youki should follow runc's implementation.
Therefore, we will implement the crate like opencontainers/selinux in this issue.

Gekko0114:
@utam0k 
I created a issue here, please correct me if I misunderstand.

YJDoc2:
Hey @Gekko0114 , do you need any help with this?

Gekko0114:
Thanks, but I don't have enough time to work on these days..
I plan to work on it when I have time. Sorry for inconvenience.
I thought it is not so urgent, but  if it is urgent, please let me know.

YJDoc2:
> Thanks, but I don't have enough time to work on these days..

I completely understand, no worries!

>I plan to work on it when I have time. Sorry for inconvenience.
I thought it is not so urgent, but if it is urgent, please let me know.

This is not exactly urgent, but given that this is an incorrect implementation , I would prefer to have it fixed sooner than later. I was just wondering if you are still planning to work on this or not, so pinged you. Take your time :purple_heart:

Gekko0114:
Hi @YJDoc2, @utam0k 

Sorry for not doing this task for a while. I will resume it.
Seems that https://github.com/opencontainers/selinux has several files. So I would like to take this task into several PRs, like these. Do you have any suggestions about this?

- go-selinux/label/
- go-selinux/selinux.go
- selinux_stub.go
- pkg/pwalk/
and so on...

YJDoc2:
[...]

PR #2800, opened 2024-05-26
Init a selinux project

Gekko0114:
This is an experimental crate. I'll implement a selinux crate step-by-step.
When it is mature, I'll move it under the 'crate/' and use it.
ref: #2718


Gekko0114:
@YJDoc2 @utam0k 
As the initial PR, what function should I implement at least? Or is any small PR fine ?

utam0k:
I'm not familiar with this area. So, It would be nice if you could make the PR small enough so that one function could run on e2e.

Gekko0114:
Sure, thanks!

Gekko0114:
Hi @utam0k , @YJDoc2 

Though this PR is small, I would like to complete this before becoming too complicated.
Could you review it?

utam0k:
> Though this PR is small, I would like to complete this before becoming too complicated.
Could you review it?
 
This PR is the perfect size for me to review! Thanks. But sorry, I'll revisit it next weekend 🙏

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,302 @@
...
+// set_file_label sets the SELinux label for this path, following symlinks, or returns an error.
+pub fn set_file_label(fpath: &str, label: &str) -> Result<(), std::io::Error> {
+    if fpath.is_empty() {
+        return Err(std::io::Error::new(io::ErrorKind::InvalidInput, ERR_EMPTY_PATH));
+    }
+
+    loop {
+        match set_xattr(fpath, XATTR_NAME_SELINUX, label.as_bytes(), 0) {
+            Ok(_) => break,
+// TODO            Err(Errno::EINTR) => continue,
```
YJDoc2, line 37: Is there specific reason for keeping this TODO for a next PR?
Same for other todos below

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,302 @@
...
+        return Err(std::io::Error::new(io::ErrorKind::InvalidInput, ERR_EMPTY_PATH));
+    }
+    let label = get_xattr(fpath, XATTR_NAME_SELINUX);
+    match label {
+        Ok(mut v) => {            
+            if (v.len() > 0) && (v.chars().nth(v.len() - 1) == Some('\x00')) {
+                v = (&v[0..v.len() - 1]).to_string();
+            }
+            return Ok(v);
+        },
```
YJDoc2, line 80: as per the current definition, `get_xattr` already returns a `String` . So can we instead do this checking there (when it will be implemented in next PR) , and instead just return the label from here?

Also, instead of `len()` check, more idiomatic way would be `!v.is_empty()` , and for the  last element, we can do `v.chars().last()` . I also think there would be some other way to convert null terminated string into rust String; although the way will depend on how we are getting the original string in `get_xattr` implementation.

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+    Ok(())
+}
+
+pub fn read_con_fd(file: &mut File) -> Result<String, std::io::Error> {
+    let mut data = String::new();
+    file.read_to_string(&mut data)?;
+
+    let trimmed_data = data.trim_end_matches(char::from(0));
```
YJDoc2, line 233: Is this necessary? If I'm correct, then `read_to_string` returns a Rust String, which should not be null terminated in first place. Can the file content itself have a null byte in it which we are trying to remove here?

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,302 @@
...
+            HAVE_THREAD_SELF = path.is_dir();
+        }
+    });
+
+    unsafe {
+        if HAVE_THREAD_SELF {
+            return format!("{}/{}", THREAD_SELF_PREFIX, attr);
+        }
+    }
+
```
YJDoc2, line 261: Can you explain the use of this function? I am not able to understand it correctly :sweat_smile:

1. What is the reason for using Once and call_once here? Assuming youki as the primary use-case, is there still a chance of having a multi-thread race? If so can be use AtomicBool instead and remove dependency of Once? (Suggesting as underlying data is simple bool type). We might  be able to do an compare-and-exchange operation instead of the call_once here.
2. Do we need the `unsafe` block if we are only reading the `HAVE_THREAD_SELF` value? I thought unsafe was needed only if we were writing to the variable.

YJDoc2 (commented):
Hey, some comments, some nitpicks, overall the implementation seems to be going in the right direction :+1:

Can we add doc comments for all the `pub` functions (which are implemented here), so it will be easier to review what the function is supposed to do + documentation

minor nitpicks : 
- instead of `panic!(...)` I feel `unimplemented!()` would be a better fit. Even though both  behave the same, I feel unimplemented is semantically better. I am not asking to change the current use, but maybe for future PRs.
- Can you check if clippy is run here and does not give any errors? I have had issues in past when running clippy in CI with projects excluded via cargo.toml.

experiment/selinux/src/selinux.rs
Gekko0114: Thanks for your comment!
Some functions like this depend on xattr library, which is not implemented yet.
Therefore, I keep this TODO. Will fix it after implementing xattr.
(I will add this explanation in the code as well)

experiment/selinux/src/selinux.rs
Gekko0114: > So can we instead do this checking there (when it will be implemented in next PR) , and instead just return the label from here?

Make sense, thanks. `get_xattr` is used in `file_label` and `lfile_label`. These two files handle label same. So we can move these logic to `get_attr`.

experiment/selinux/src/selinux.rs
Gekko0114: I added test cases in `test_read_con_fd`.
Seems that  trimming is necessary if we want to remove null bytes on the end of a file.
Does it make sense?
Sorry if I misunderstood the context or something

experiment/selinux/src/selinux.rs
Gekko0114: Thanks, I've updated the code using atomicBool instead.

I implemented using Once because go-selinux implemented like that.
I guess that this is to avoid conflicts between multi-thread like goroutine.
I thought it is possible that similar issue happens on Youki as well.
https://github.com/opencontainers/selinux/blob/9dee859cbffb974e779f65efaef9f5ef6e1d260b/go-selinux/selinux_linux.go#L451-L467

I am not familiar with selinux nor youki, please let me know if I misunderstood.

YJDoc2: Hey, thanks. I am also not that familiar with selinux-go, so will take a look and confirm once.

experiment/selinux/src/selinux.rs
YJDoc2: Hey, I think this is ok for now, we can re-visit this in later PRs if needed. I was just wondering that if the `read_to_string` fn itself will remove the trailing `0` s, ie. would `test_read_con_fd` pass even if we remove this trim call.

experiment/selinux/README.md
```diff
@@ -0,0 +1,9 @@
+This is an experimental project in order to create selinux library in Rust.  
+Ref: https://github.com/containers/youki/issues/2718.  
+Reimplementation of (selinux)[https://github.com/opencontainers/selinux] in Rust.  
+Also selinux depends on xattr, but nix doesn't cover xattr function. 
+Therefore, this PR will implement xattr in Rust.
+
+```console
+$ cargo run
```
utam0k, line 8: I couldn't run it because there isn't a `main` function.

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,433 @@
...
+use std::io::{self, Read};
+use std::os::fd::{AsFd, AsRawFd};
+use std::sync::atomic::{AtomicBool, Ordering};
+
+const XATTR_NAME_SELINUX: &str = "security.selinux";
+const ERR_EMPTY_PATH: &str = "empty path";
+static HAVE_THREAD_SELF: AtomicBool = AtomicBool::new(false);
+static INIT_DONE: AtomicBool = AtomicBool::new(false);
+
+// function compatible with setDisabled in go-selinux repo.
```
utam0k, line 16: I don't think we have to keep it compatible with the `go-selinux` repo. Is there any specific reason to keep it?

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,433 @@
...
+use std::path::PathBuf;
+use std::fs::File;
+use std::io::{self, Read};
+use std::os::fd::{AsFd, AsRawFd};
+use std::sync::atomic::{AtomicBool, Ordering};
+
+const XATTR_NAME_SELINUX: &str = "security.selinux";
+const ERR_EMPTY_PATH: &str = "empty path";
+static HAVE_THREAD_SELF: AtomicBool = AtomicBool::new(false);
+static INIT_DONE: AtomicBool = AtomicBool::new(false);
```
utam0k, line 14: Can a structure or something have them to prevent from using `static`.

experiment/selinux/src/selinux.rs
utam0k: In my view, prioritizing Rust-friendliness over maintaining compatibility with `go-selinux` is more important.

utam0k (changes requested):
Thanks, @Gekko0114! This is a good first step for us. I left some comments.

experiment/selinux/src/selinux.rs
Gekko0114: > In my view, prioritizing Rust-friendliness over maintaining compatibility with go-selinux is more important.

Yeah, agree. I couldn't write Rust-friendliness because I am not so familiar with Rust yet.
Please let me know if you feel weird in terms of Rust. Thank you so much!

experiment/selinux/README.md
Gekko0114: Thanks, removed this line.

experiment/selinux/src/selinux.rs
utam0k: > Please let me know if you feel weird in terms of Rust. Thank you so much!

Sure. First of all, we can take using `struct` instead of global variables into consideration. WDYT?

Gekko0114: Sure, as you commented in another thread, I will use struct.
I will fix it after reading experiment/seccomp and other codes.

If we use `struct` here and call this `struct` more than once, I thought maybe the selinux struct is initialized more than once. But the performance caused by this is tiny, so I will use struct here.

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,433 @@
...
+            }
+        }
+    }
+    Ok(())
+}
+
+// function compatible with lSetFileLabel in go-selinux repo.
+// lset_file_label sets the SELinux label for this path, not following symlinks,
+// or returns an error.
+pub fn lset_file_label(fpath: &str, label: &str) -> Result<(), std::io::Error> {
```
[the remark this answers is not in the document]
Gekko0114: Sure, I introduced `thiserror`. Please let me know if I don't use `thiserror` correctly.

experiment/selinux/src/selinux.rs
Gekko0114: Thanks. Changed the code using struct.

Gekko0114:
Hi @utam0k, @YJDoc2,

Thanks to your comments, I've fixed this PR.
However, I think there are some parts which is not rust-friendly.
I would appreciate it if you could let me know.

experiment/selinux/src/selinux.rs
```diff
@@ -0,0 +1,469 @@
...
+    #[error("Failed to call read_con_fd for SELinux: {0}")]
+    ReadConFd(String),
+    #[error("Failed to call read_con for SELinux: {0}")]
+    ReadCon(String),    
+}
+
+pub struct SELinux {
+    have_thread_self: AtomicBool,
+    init_done: AtomicBool,
+}
```
utam0k, line 35: 👍 👍 👍

utam0k (approved):
It looks good to me as a first PR for a Selinxu project. In the next PR, it'd be great if there was an excellent example to run with the `selinux` crate.

Gekko0114:
@utam0k 
Thanks! Since I don't have the write access, I can't merge this PR.
Could you merge this PR if it is ok?

utam0k:
Thanks, @Gekko0114 !

PR #2802, opened 2024-05-27
Export max_usage in cgroups v2 mode

HeRaNO:
Kernel 5.19 introduces `memory.peak` and kernel 6.5 introduces `memory.swap.peak` to expose the memory and swap usage watermark. This change uses these values to fill `max_usage`.


crates/libcgroups/src/v2/memory.rs
```diff
@@ -77,6 +77,8 @@ impl Memory {
             stats::parse_single_value(&cgroup_path.join(format!("{}.{}", file_prefix, "current")))?;
         let limit =
             stats::parse_single_value(&cgroup_path.join(format!("{}.{}", file_prefix, "max")))?;
+        let max_usage =
```
utam0k, line 80: Please add the error handling if this file doesn't exist, i.e. the users' kernel version doesn't meet it.

HeRaNO: done, is it ok?

utam0k: Thanks, but won't this hide the actual error?

HeRaNO: The function `parse_single_value` will return two kinds of error: `WrappedIoError` and `InvalidData`. `WrappedIoError` is what we want to mute. And the file `memory.peak` (also `memory.swap.peak`) is a read-only file provided by kernel, so `InvalidData` error is rare. So here I treat them both as the peak file is missing.

Actually, I found that in cgroups v1, `memsw` part is not using error handling. But mem+swap accounting is not enabled by default (should add `swapaccount=1` to grub), so the situation is similar to here. I'm not quite sure if we need more detailed error handling, but I think it's enough for us.

Please feel free to comment and add suggestions, thanks.

utam0k: Okay, I see, but please cover the test case for the missing file.

utam0k: Others, looks good +1

HeRaNO: OK, I add a test `test_get_memory_data_with_peak` and kept the origin test for the missing file case. PTAL, thx @utam0k

utam0k (approved):
Thanks, @HeRaNO

PR #2805, opened 2024-05-31
Fix word order in README sentence justifying Rust usage

andrewimeson:
I believe the intent is to say that it's also tricky in Rust, not that it's "too tricky" (i.e. _impossible_) in Rust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ youki is named after the Japanese word 'youki', which means 'a container'. In Ja
 
 Here is why we are writing a new container runtime in Rust.
 
-- Rust is one of the best languages to implement the oci-runtime spec. Many very nice container tools are currently written in Go. However, the container runtime requires the use of system calls, which requires a bit of special handling when implemented in Go. This is too tricky (e.g. _namespaces(7)_, _fork(2)_); with Rust, it's not that tricky. And, unlike in C, Rust provides the benefit of memory safety. While Rust is not yet a major player in the container field, it has the potential to contribute a lot: something this project attempts to exemplify.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ youki is named after the Japanese word 'youki', which means 'a container'. In Ja
 
 Here is why we are writing a new container runtime in Rust.
 
-- Rust is one of the best languages to implement the oci-runtime spec. Many very nice container tools are currently written in Go. However, the container runtime requires the use of system calls, which requires a bit of special handling when implemented in Go. This is too tricky (e.g. _namespaces(7)_, _fork(2)_); with Rust, it's not that tricky. And, unlike in C, Rust provides the benefit of memory safety. While Rust is not yet a major player in the container field, it has the potential to contribute a lot: something this project attempts to exemplify.
+- Rust is one of the best languages to implement the oci-runtime spec. Many very nice container tools are currently written in Go. However, the container runtime requires the use of system calls, which requires a bit of special handling when implemented in Go. This too is tricky (e.g. _namespaces(7)_, _fork(2)_); with Rust, it's not that tricky. And, unlike in C, Rust provides the benefit of memory safety. While Rust is not yet a major player in the container field, it has the potential to contribute a lot: something this project attempts to exemplify.
```
YJDoc2, line 38:
```suggestion
- Rust is one of the best languages to implement the oci-runtime spec. Many very nice container tools are currently written in Go. However, the container runtime requires the use of system calls, which requires a bit of special handling when implemented in Go. This tricky (e.g. _namespaces(7)_, _fork(2)_); with Rust too, but it's not that tricky. And, unlike in C, Rust provides the benefit of memory safety. While Rust is not yet a major player in the container field, it has the potential to contribute a lot: something this project attempts to exemplify.
```

How about this?

utam0k (approved):
Thanks, @andrewimeson 😍

YJDoc2:
As this is a readme changes, it did not run the rust check CI, and hence the CI is pending. I'll go ahead and merge this bypassing the protection rules, as just a readme text change, no code change.

PR #2809, opened 2024-06-05
initial progress on supporting OwnedFd

zahash:
related to issue https://github.com/containers/youki/issues/2417

### DIFF
--- a/crates/libcontainer/src/channel.rs
+++ b/crates/libcontainer/src/channel.rs
@@ -16,13 +16,11 @@ pub enum ChannelError {
     #[error("channel connection broken")]
     BrokenChannel,
 }
-#[derive(Clone)]
 pub struct Receiver<T> {
     receiver: RawFd,
     phantom: PhantomData<T>,
 }
 
-#[derive(Clone)]
 pub struct Sender<T> {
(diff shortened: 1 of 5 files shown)


utam0k (changes requested):
@zahash Please elaborate a bit more 🙏

zahash:
@utam0k Sorry i think i linked the wrong issue. I changed it to the correct one just now.

also, a lot of explanation was done in the issue comments

utam0k:
@zahash Thanks for your update, but it's not enough information to start by reviewing it. Apparently, the title of the PR is wrong.

YJDoc2:
@zahash ping!

zahash:
@utam0k @YJDoc2 

i have changed the title.
the idea is, `OwnedFd` cannot be cloned. so before switching to `OwnedFd`, we have to make sure no cloning is done.
the next step would be to use `OwnedFd` inside sender and reciever instead of `RawFd`

utam0k:
💯 

> the idea is, OwnedFd cannot be cloned. so before switching to OwnedFd, we have to make sure no cloning is done.

utam0k (approved):

PR #2813, opened 2024-06-12
move macro define youki_version to use before

lengrongfu:
Fixes: https://github.com/containers/youki/issues/2812

### DIFF
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -23,6 +23,23 @@ struct YoukiExtendOpts {
     pub log_level: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


lengrongfu:
kind/cleanup

lengrongfu:
@YJDoc2 Can you help me review this PR? thanks ~

YJDoc2 (approved):
Ideally this should be fixed in rust rover, as this seems to work in other places like vs code and nvim lsp. Anyways, not a big change, only definition is moved around in same file, so acceptable.

Also apologies for the delay, I saw failing CI and thought it was an issue on the code side, when actually it was PR label problem.

PR #2817, opened 2024-06-15
Use HashMap for envs in container_init_process

musaprg:
The environment variables `envs` for the container processes are obtained in `Vec<String>` and manipulated based on the various circumstances in `container_init_process`[^manipulate-envs-1]. They're eventually converted into the HashMap, so the logic to manipulate environment variables should be performed for the parsed HashMap instead of `Vec<String>`. It would reduce the conditional logic based on the string match[^conditional-logic-code].

There might be a downside to applying HashMap for manipulation, for example, inserting a new value could be slow. AFAIU, there are a few insertions in the current `container_process_init` function, so I don't think it would put the additional latency for the initialization process.

[^manipulate-envs-1]: https://github.com/containers/youki/blob/6f8b39c28a69d9f608b773e4e18e1e08a81e7bd4/crates/libcontainer/src/process/container_init_process.rs#L484-L489
[^conditional-logic-code]: https://github.com/containers/youki/blob/6f8b39c28a69d9f608b773e4e18e1e08a81e7bd4/crates/libcontainer/src/process/container_init_process.rs#L563-L568

### DIFF
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -279,7 +279,8 @@ pub fn container_init_process(
     let spec = &args.spec;
     let linux = spec.linux().as_ref().ok_or(MissingSpecError::Linux)?;
     let proc = spec.process().as_ref().ok_or(MissingSpecError::Process)?;
-    let mut envs: Vec<String> = proc.env().as_ref().unwrap_or(&vec![]).clone();
+    let mut envs: HashMap<String, String> =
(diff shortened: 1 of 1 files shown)


yihuaf:
> There might be a downside to applying HashMap for manipulation, for example, inserting a new value could be slow. AFAIU, there are a few insertions in the current container_process_init function, so I don't think it would put the additional latency for the initialization process.

The improvements looks to be worth it. I would not be too concerned with the performance of insert with Hashmap vs. Array. For this specific usecase.

yihuaf (approved):

PR #2818, opened 2024-06-17
Rename to improve readability

utam0k:
At the moment, recursive options are the only way to do this. Wouldn't it be easier to understand if the names were more specific?


crates/libcontainer/src/syscall/linux.rs
```diff
@@ -38,87 +38,65 @@ const MOUNT_ATTR_NODIRATIME: u64 = 0x00000080;
 const MOUNT_ATTR_NOSYMFOLLOW: u64 = 0x00200000;
 
 /// Constants used by mount_setattr(2).
-pub enum MountAttrOption {
+pub enum MountRecursive {
```
yihuaf, line 41: I like the changes to drop the `MountAttr` prefix to the enum fields. I am not sure about `MountAttrOption` -> `MountRecursive`. These are defined in the syscalls interface as `MOUNT_ATTR`, so should we keep the name `MountAttrOption` to avoid confusion?

yihuaf: @utam0k Do you remember why we did not implement the rest of the mount attr options in this `enum` and choose to match string directly here? https://github.com/containers/youki/blob/main/crates/libcontainer/src/rootfs/utils.rs#L131

yihuaf: Nvm, after reading through the code again, I know understand. You are correct. These options only corresponds to recursive mount options and gets passed through to `rec_attr` in `MountAttr` and then requires the set of `AT_RECURSIVE` flag when applied. I agree the `MountAttrOption` is confusing and rename to something having recursive in the name is the right thing to do.

yihuaf (approved):
Nit: change to `MountRecursiveAttr`? Optional :)

YJDoc2 (approved):
LGTM :+1: This change makes so much sense , thank you :pray: 

Given that this changes public enum names and error name, I'm adding breaking change label too this. If you think that is not needed, feel free to remove.

utam0k:
> Nit: change to `MountRecursiveAttr`? Optional :)

Thanks for your suggestion. But I think `Rdonly` and others are obviously MountAttr.

PR #2820, opened 2024-06-18
Add new `setup_envs` method for the `Executor` trait

### LINKED ISSUE #2815: Options to pass env vars from OCI Spec to WASI Context
Wasm containers are different from Linux containers in many ways, one of them being their handling of environment variables. A Wasm runtime, by default, does not grant any environment variables to Wasm modules unless explicitly passed to the WASI Context (e.g., [using wasmtime's API](https://docs.rs/wasmtime-wasi/21.0.1/wasmtime_wasi/struct.WasiCtxBuilder.html#method.envs)). This contrasts with Linux containers, which can utilize all the environment variables available to them.

Given this difference, it seems unnecessary for `libcontainer` to modify the main process's environment variables before fork-exec container processes (as seen [here](https://github.com/containers/youki/blob/main/crates/libcontainer/src/process/container_init_process.rs#L570-L574)) in Wasm use cases. Moreover, mutating environment variables before calling `Executor::exec` complicates the delineation between the Containerd shim and Wasm containers (see more in [this issue](https://github.com/containerd/runwasi/issues/619) from `runwasi`). The runwasi shim sets up the Wasm runtime (e.g., Spin, Wasmtime, WasmEdge, etc.) within the container namespace, but this setup is not part of the user's Wasm container code. Runwasi introduces a third layer, the Wasm runtime, between the traditional shim and Linux container layers.

We propose that container environment variables, such as those from the PodSpec in Kubernetes, should be directly passed to the WASI Context instead of the shim process, which is still configuring the Wasm runtime.

I am not sure how exactly we can achieve this, but I have shared some ideas in [this issue](https://github.com/containerd/runwasi/issues/619#issuecomment-2163456418). I welcome your thoughts on this matter.

musaprg:
fixes #2815 

This PR introduces a new `setup_envs` method for the `Executor` trait. It allows any implementors of the `Executor` trait to control how environment variables specified in the OCI spec are handled. The default implementation mirrors the behavior of the one implemented in the `container_init_process`, which involves removing existing environment variables and setting the ones specified in the OCI spec.

## TODO

- [x] Make consensus about the interface
- [x] Consider defining a dedicated error instead of `ExecutorError`
- [x] Add missing test cases

### DIFF
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -589,7 +589,7 @@ pub fn container_init_process(
     }
 
     args.executor.validate(spec)?;
-    args.executor.set_envs(envs)?;
+    args.executor.setup_envs(envs)?;
 
     // Notify main process that the init process is ready to execute the
     // payload.  Note, because we are already inside the pid namespace, the pid
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -73,6 +73,8 @@ pub enum InitProcessError {
     Workload(#[from] workload::ExecutorError),
     #[error(transparent)]
     WorkloadValidation(#[from] workload::ExecutorValidationError),
+    #[error(transparent)]
+    WorkloadSetEnvs(#[from] workload::ExecutorSetEnvsError),
     #[error("invalid io priority class: {0}")]
     IoPriorityClass(String),
     #[error("call exec sched_setattr error: {0}")]
--- a/crates/libcontainer/src/process/container_init_process.rs
+++ b/crates/libcontainer/src/process/container_init_process.rs
@@ -558,18 +558,6 @@ pub fn container_init_process(
         err
     })?;
 
-    // add HOME into envs if not exists
-    if !envs.contains_key("HOME") {
-        if let Some(dir_home) = utils::get_user_home(proc.user().uid()) {
-            envs.insert("HOME".to_owned(), dir_home.to_string_lossy().to_string());
-        }
-    }
-
-    // Reset the process env based on oci spec.
-    env::vars().for_each(|(key, _value)| env::remove_var(key));
-    envs.iter()
-        .for_each(|(key, value)| env::set_var(key, value));
-
     // Initialize seccomp profile right before we are ready to execute the
     // payload so as few syscalls will happen between here and payload exec. The
     // notify socket will still need network related syscalls.
@@ -591,7 +579,15 @@ pub fn container_init_process(
         tracing::warn!("seccomp not available, unable to set seccomp privileges!")
     }
 
+    // add HOME into envs if not exists
+    if !envs.contains_key("HOME") {
+        if let Some(dir_home) = utils::get_user_home(proc.user().uid()) {
+            envs.insert("HOME".to_owned(), dir_home.to_string_lossy().to_string());
+        }
+    }
+
     args.executor.validate(spec)?;
+    args.executor.set_envs(envs)?;
 
     // Notify main process that the init process is ready to execute the
     // payload.  Note, because we are already inside the pid namespace, the pid
--- a/crates/libcontainer/src/workload/default.rs
+++ b/crates/libcontainer/src/workload/default.rs
@@ -138,6 +138,11 @@ fn is_executable(path: &Path) -> std::result::Result<bool, std::io::Error> {
 
 #[cfg(test)]
 mod tests {
+    use std::collections::HashMap;
+    use std::env;
+
+    use serial_test::serial;
+
     use super::*;
 
     #[test]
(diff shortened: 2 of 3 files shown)


utam0k:
cc: @Mossaka

crates/libcontainer/src/workload/mod.rs
```diff
@@ -1,4 +1,5 @@
 use oci_spec::runtime::Spec;
+use std::{collections::HashMap, env};
```
jprendes, line 2: Formatting issue:
```suggestion
use std::collections::HashMap;
use std::env;
```

You can catch formatting issues locally by running `just lint`.

jprendes (commented):
LGTM
Thanks @musaprg for working on this!
Now we just needs to get a green CI run :-)

crates/libcontainer/src/workload/mod.rs
musaprg: fixed in 6471f71

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -591,7 +581,15 @@ pub fn container_init_process(
...
 
+    // add HOME into envs if not exists
+    if !envs.contains_key("HOME") {
+        if let Some(dir_home) = utils::get_user_home(proc.user().uid()) {
+            envs.insert("HOME".to_owned(), dir_home.to_string_lossy().to_string());
+        }
+    }
+
     args.executor.validate(spec)?;
+    args.executor.set_envs(envs)?;
```
Mossaka, line 592: Is the idea for the wasm shim to override this method by giving an empty implementation and decides what to do with the env vars inside of the `Executor::execute()` function?

Mossaka (commented):
LGMT, one question to clarify

crates/libcontainer/src/workload/default.rs
```diff
@@ -177,4 +182,35 @@ mod tests {
         assert!(!is_executable(&non_executable_path).unwrap());
         assert!(!is_executable(directory_path).unwrap());
     }
+
+    #[test]
+    #[serial]
+    fn test_executor_set_envs() {
```
utam0k, line 188: 👍

crates/libcontainer/src/process/container_init_process.rs
```diff
@@ -73,6 +73,8 @@ pub enum InitProcessError {
     Workload(#[from] workload::ExecutorError),
     #[error(transparent)]
     WorkloadValidation(#[from] workload::ExecutorValidationError),
+    #[error(transparent)]
+    WorkloadSetEnvs(#[from] workload::ExecutorSetEnvsError),
```
utam0k, line 77: :+1:

crates/libcontainer/src/workload/mod.rs
```diff
@@ -60,6 +71,16 @@ pub trait Executor: CloneBoxExecutor {
     /// namespace and cgroups, and pivot_root into the rootfs. But this step
     /// runs before waiting for the container start signal.
     fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError>;
+
+    /// Set environment variables for the container process to be executed.
+    fn set_envs(&self, envs: HashMap<String, String>) -> Result<(), ExecutorSetEnvsError> {
```
utam0k, line 76: I don't think `set` is a proper word for this method because we need to delete and add the envs. How about `update` or `setup`?

crates/libcontainer/src/workload/mod.rs
```diff
@@ -60,6 +71,16 @@ pub trait Executor: CloneBoxExecutor {
     /// namespace and cgroups, and pivot_root into the rootfs. But this step
     /// runs before waiting for the container start signal.
     fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError>;
+
+    /// Set environment variables for the container process to be executed.
```
utam0k, line 75: It'd be better to explain the situation this method is called. For example, this method is called the init process and hasn't cleared the host's envs yet.

crates/libcontainer/src/process/container_init_process.rs
musaprg: > Is the idea for the wasm shim to override this method by giving an empty implementation and decides what to do with the env vars inside of the Executor::execute() function?

Yes, that's my intention.

musaprg: Another option was moving the logic of setting envs under the `Executor::execute()` default implementation, which would affect any existing `Executor` implementations that assume the original behavior regarding environment variables. IMO defining another dedicated method would be good as suggested in https://github.com/containers/youki/issues/2815#issuecomment-2169088272.

crates/libcontainer/src/workload/mod.rs
musaprg: `setup_envs` sounds good to me. Let me rename the method and related type names.

musaprg:
@Mossaka @utam0k @jprendes Hi. Thanks for the reviews! I've updated my PR based on your comments. Would you recheck again?

crates/libcontainer/src/workload/mod.rs
musaprg: I've added a detailed description for the `setup_envs` in b20b8fa.

Mossaka (approved):
lgtm!

yihuaf (commented):
LGTM

utam0k:
@musaprg Congrats! 🎉 Thanks for your contribution.

PR #2825, opened 2024-06-22
selinux: write xattr related codes.

Gekko0114:
This is an experimental crate. I'll implement a selinux crate step-by-step.
In this PR, I implemented functions related to xattr.

ref: https://github.com/containers/youki/issues/2718 https://github.com/containers/youki/pull/2800


experiment/selinux/src/xattrs/xattr.rs
```diff
@@ -0,0 +1,105 @@
...
+        let attr_name = "user.test_attr";
+        let attr_value = "system_u:object_r:some_label_t";
+        let file_name = "../test_file";
+        let file_path = Path::new(file_name);
+        let symlink_path = Path::new("test_symlink");
+        create_temp_file(b"test", file_name);
+        symlink(&file_path, &symlink_path).expect("Failed to create symlink");
+
+        lset_xattr(&symlink_path, attr_name, attr_value.as_bytes()).expect("Failed to lset xattr");
+        let actual = lget_xattr(&symlink_path, attr_name).expect("Failed to lget xattr");
```
Gekko0114, line 100: @utam0k @YJDoc2 
I wrote the test, but it failed on my codespace.
Do you know how to enable labeling lset_xattr on this environment ?

```
---- xattrs::xattr::tests::test_lset_xattr_and_lget_xattr stdout ----
thread 'xattrs::xattr::tests::test_lset_xattr_and_lget_xattr' panicked at src/xattrs/xattr.rs:99:69:
Failed to lset xattr: LSetXattr("Operation not permitted (os error 1)")
```

Gekko0114:
Hi @utam0k, @YJDoc2,
I've created this small PR adding the functions related to xattr. Could you review it?

experiment/selinux/src/xattrs/xattr.rs
YJDoc2: Hey, so after digging through https://www.man7.org/linux/man-pages/man7/xattr.7.html , https://www.man7.org/linux/man-pages/man2/listxattr.2.html, https://www.man7.org/linux/man-pages/man1/setfattr.1.html etc. I think the reason behind this is that `user.*` attributes are not supported for symlinks. Unfortunately the rest (system,security and trusted) all require CAP_SYS_ADMIN i.e. root perms to be set, which means we cannot really test this in a straightforward way as far as I can think. Hope this helps.

I also saw the code in rustix itself, and even they don't seem to be testing actual get/set for symlinks, only some error conditions.

Gekko0114: I see, thank you so much for your investigation.
Though I will investigate a bit more on whether there is a way running a test for this, I think we can skip lset_xattr test here.

Gekko0114: Added "ignore"

utam0k:
I'll visit this PR in a couple of days. Please give me some time.

experiment/selinux/src/xattrs/xattr.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::xattrs::*;
+    use std::fs::{self, File};
+    use std::path::Path;
+    use std::io::Write;
+    use std::os::unix::fs::symlink;
+
+    fn create_temp_file(content: &[u8], file_name: &str) {
```
utam0k, line 66: How about using `tempfile` crate?
https://github.com/containers/youki/blob/11ca60dc8883968f1e69cfe34e6803e909391bc7/crates/libcontainer/Cargo.toml#L62

experiment/selinux/src/xattrs/xattr.rs
```diff
@@ -0,0 +1,106 @@
...
+
+// function similar with getattr in go-selinux repo.
+// get_xattr returns the value of an extended attribute attr set for path.
+pub fn get_xattr(fpath: &Path, attr: &str) -> Result<String, XattrError> {
+    // SELinux label is not so big, so we allocate 1024 bytes for the buffer.
+    let mut buf = vec![0u8; 1024];
+    let size = rfs::getxattr(fpath, attr, &mut buf)
+        .map_err(|e| XattrError::GetXattr(e.to_string()))?;
+    let mut value = String::from_utf8_lossy(&buf[..size]).into_owned();
+    if (!value.is_empty()) && (value.ends_with('\x00')) {
```
utam0k, line 38: Shouldn't the \x00 be checked and the buffer size adjusted if it isn't termination.

experiment/selinux/src/xattrs/xattr.rs
Gekko0114: Thanks, fixed it

experiment/selinux/src/xattrs/xattr.rs
Gekko0114: Fixed it. Is this implementation same as your intention?

Gekko0114:
Hi @utam0k, @YJDoc2,
Thanks for your comments and fixed them. Could you take a look again when you have time?

experiment/selinux/src/xattrs/xattr.rs
```diff
@@ -0,0 +1,119 @@
...
+        let temp_file = NamedTempFile::new().expect("Failed to create temp file");
+        let file_path = temp_file.path();
+
+        set_xattr(&file_path, attr_name, attr_value.as_bytes()).expect("Failed to set xattr");
+        let actual = get_xattr(&file_path, attr_name).expect("Failed to get xattr");
+        assert_eq!(actual, attr_value);
+    }
+
+    #[test]
+    #[ignore]
```
YJDoc2, line 104: Hey, if we are ignoring this test, should we add this in the first place? If our original discussion on this was correct, then this will never pass unless run with root permissions, and in this case this does not actually test anything at all when running as non-root.

YJDoc2 (commented):
Hey, a couple of questions, and the CI is failing due to unrelated issues, can you rebase on current main so CI will pass? Thanks!

YJDoc2:
> ⚠️ Please install the 'codecov app svg image' to ensure uploads and comments are reliably processed by Codecov.

@utam0k should we remove the codecov bot for now? We are not enforcing the coverage value in CI, and also not maintaining a required threshold for it.

experiment/selinux/src/selinux.rs
```diff
@@ -63,19 +63,23 @@ impl SELinux {
...
+    #[allow(unused_variables)]
     pub fn class_index(class: &str) -> Result<i64, String> {
         unimplemented!("not implemented yet")
     }
 
     // function similar with setFileLabel in go-selinux repo.
     // set_file_label sets the SELinux label for this path, following symlinks, or returns an error.
+    #[allow(unused_variables)]
+    #[allow(unreachable_patterns)]
+    #[allow(non_snake_case)]
```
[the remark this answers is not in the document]
Gekko0114: These functions are not implemented yet, therefore I've added these macros. But I don't mind whether these should be ignored or not for now.
Do you think these are not necessary?

experiment/selinux/src/xattrs/xattr.rs
Gekko0114: Sure. If we can't run the test with root permissions, I will remove this test.

YJDoc2: Yes, I think that would be preferable, as in our CI and most places we run tests as non-root

experiment/selinux/src/selinux.rs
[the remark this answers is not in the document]
YJDoc2: Unless clippy is complaining about these, I don't think they are needed here.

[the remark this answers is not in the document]
Gekko0114: Sure, removed them. Thanks!

experiment/selinux/src/xattrs/xattr.rs
Gekko0114: Removed them, thanks!

utam0k:
> @utam0k should we remove the codecov bot for now? We are not enforcing the coverage value in CI, and also not maintaining a required threshold for it.

+1 for now. But we'd like to improve our code coverage in the future. It encourages us to add the unit tests.

experiment/selinux/src/xattrs/xattr.rs
utam0k: Yes, is this 💯

experiment/selinux/src/xattrs/xattr.rs
```diff
@@ -0,0 +1,119 @@
...
+            Err(e) => return Err(XattrError::GetXattr(e.to_string())),
+        }
+    }
+}
+
+// function similar with lgetxattr in go-selinux repo.
+// lget_xattr returns the value of an extended attribute attr set for path.
+pub fn lget_xattr(fpath: &Path, attr: &str) -> Result<String, XattrError> {
+    // SELinux label is not so big, so we allocate 1024 bytes for the buffer.
+    let mut buf_size = 1024;
```
utam0k, line 79: It'd be defined as `const`.

experiment/selinux/src/xattrs/xattr.rs
```diff
@@ -0,0 +1,119 @@
...
+                let mut value = String::from_utf8_lossy(&buf[..size]).into_owned();
+                if value.ends_with('\x00') {
+                    value.pop();
+                }
+                return Ok(value);
+            }
+            Err(e) => return Err(XattrError::LGetXattr(e.to_string())),
+        }
+    }
+}
```
utam0k, line 99: What do you think about implementing it as a trait for `Path` or `File`?
Like that:

```rust
pub trait PathBufExt {
    lget_xattr(&self, attr: &str) -> Result<String, XattrError>
...
}

impl PathBufExt for Path {
	fn lget_xattr(&self, attr: &str) -> Result<String, XattrError> {
		...
	}
	...
}
```
https://github.com/containers/youki/blob/a2390311b44f9beacd7b801bea0078daf8bbc1b4/crates/libcontainer/src/utils.rs#L35-L42

utam0k: WDYT? @YJDoc2 @Gekko0114

Gekko0114: Yeah, sure. I think this change will make the code easier to handle.

Gekko0114: Fixed it

Gekko0114:
Hi @utam0k, @YJDoc2 
Thanks for your review comments and I've fixed it. Could you take a look again when you have time?

utam0k (approved):
Awesome!

PR #2829, opened 2024-06-23
Add issue templates

YJDoc2:
This adds Bug report and Feature request issue templates.


utam0k (changes requested):
May I ask you to use issue forms instead of `*.md`?
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
...
+A clear and concise description of what you expected to happen.
+
+**System Info**
+Add information about your system. If Youki is compiling and running on your system, please add o/p of `youki info` .
+
+**Setup Info**
+If you are experiencing issues in compiling Youki, or related deps, please add details regarding your setup, dependencies etc. Also if you are running Youki via something else (Docker, podman, k8s) add its info here.
+
+**Additional context**
+Add any other context about the problem here.
```
utam0k, line 30: Is it really necessary to have all this information to create an issue? I feel that it'd be a waste to miss the opportunity to have an issue created at this stage of the phase. But, it might be useful to have labels and so on.

YJDoc2: > I feel that it'd be a waste to miss the opportunity to have an issue created at this stage of the phase

I'm not sure what you mean here :sweat_smile: 

As for the details, I think we can also add an empty issue template, which will not have any fields, and behave similar to how current 'new issue' works. I agree that having to add this much info can be troublesome, but I feel having these details would be good for understanding and debugging the issues.

YJDoc2:
> May I ask you to use issue forms instead of *.md?

Sure, I had not used them as they are in beta stage, and can potentially change, but after the above comment is discussed, I'll change this for forms.

.github/ISSUE_TEMPLATE/bug_report.md
utam0k: Sorry, let me tell you more details. 

> but I feel having these details would be good for understanding and debugging the issues.

Yes, this helps **us** understand the issue. However, how about an issue's author?  Perhaps filling in some fields is tedious, and they may stop making issues. I'm worried about it.
But, if we use the issue form, we can mark require or not require. We use it in order to solve this problem.

YJDoc2: Hey, yep I get it now. That is a valid concern. I'll update the  PR to use forms instead.

YJDoc2:
@utam0k I have changed to forms now. I have made only a couple of fields as required, others are optional. Please take a look.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: true # allow people to not use forms, just directly create issues
```
utam0k, line 1: I didn't know this option 👀

utam0k (approved):
Perfect! Thanks

utam0k:
@YJDoc2 LGTM except for DCO
https://github.com/containers/youki/pull/2829/checks?check_run_id=27320996017

PR #2838, opened 2024-06-30
Fix/dbus call issue

YJDoc2:
This fixes the issues with dbus method call not receiving reply. The RC was that the current implementation stopped after receiving one message, and did not continue.  Thus in some cases the signal type message would be returned, and we would discard it as its not the method reply type. As we did not look for further messages, it seemed that there was no reply from dbus. Now if the sent message was of type method call, we keep looping until we get either a method reply or method error type of message. Ideally the correct way would be to keep collecting messages in a stateful buffer and return when we get a message corresponding to the serial number we sent, but it can be tricky to implement that correctly, and at least for now we do not need that.

This should still be considered WIP, I needed review so opened this, but I need to add some related shell scripts I used for future needs. 

Apart from that this also has fix for when intermediate process would error and leave a broken socket, without any error. Now it properly attempts to send the error message over the socket so that main process can correctly identify and report it.

### DIFF
--- a/crates/libcgroups/src/systemd/dbus_native/dbus.rs
+++ b/crates/libcgroups/src/systemd/dbus_native/dbus.rs
@@ -4,6 +4,7 @@ use std::os::fd::AsRawFd;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU32, Ordering};
 
+use nix::errno::Errno;
 use nix::sys::socket;
 
 use super::client::SystemdClient;
@@ -228,9 +229,10 @@ impl DbusConnection {
             .filter(|m| m.preamble.mtype == MessageType::MethodReturn)
             .collect();
 
-        let res = res.first().ok_or(DbusError::MethodCallErr(
-            "expected method call to have reply, found no reply message".into(),
-        ))?;
+        let res = res.first().ok_or(DbusError::AuthenticationErr(format!(
+            "expected Hello call to have reply, found no reply message, got {:?} instead",
+            res
+        )))?;
         let mut ctr = 0;
         let id = String::deserialize(&res.body, &mut ctr)?;
         self.id = Some(id);
@@ -247,13 +249,18 @@ impl DbusConnection {
             let mut reply: [u8; REPLY_BUF_SIZE] = [0_u8; REPLY_BUF_SIZE];
             let mut reply_buffer = [IoSliceMut::new(&mut reply[0..])];
 
-            let reply_rcvd = socket::recvmsg::<()>(
+            let reply_res = socket::recvmsg::<()>(
                 self.socket,
                 &mut reply_buffer,
                 None,
                 socket::MsgFlags::empty(),
-            )?;
+            );
 
+            let reply_rcvd = match reply_res {
+                Ok(msg) => msg,
+                Err(Errno::EAGAIN) => continue,
+                Err(e) => return Err(e.into()),
+            };
             let received_byte_count = reply_rcvd.bytes;
 
             ret.extend_from_slice(&reply[0..received_byte_count]);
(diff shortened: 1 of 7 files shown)


crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -296,20 +303,47 @@ impl DbusConnection {
...
+                .iter()
+                .filter(|m| {
+                    m.preamble.mtype == MessageType::MethodReturn
+                        || m.preamble.mtype == MessageType::Error
+                })
+                .count();
+
+            if return_message_count > 0 {
+                break;
+            }
```
YJDoc2, line 346: I'd like suggestion for other ways to implement this. Ideally I feel that `send_message` function should not have to deal with the specific message type, but as we collect the response here, we need to account for it. Other way I can think is if we separate the send message and get response functions, so that the higher level calling functions can choose to loop over the get response.  But that means we would have to do two function calls at each place we are doing a send message call. Would that be a better way?

YJDoc2:
As per the https://github.com/containers/youki/issues/2826#issuecomment-2204425249 , this PR likely solves the issue, so requesting a review of the code now (not urgent).

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -247,13 +249,18 @@ impl DbusConnection {
...
-            let reply_rcvd = socket::recvmsg::<()>(
+            let reply_res = socket::recvmsg::<()>(
                 self.socket,
                 &mut reply_buffer,
                 None,
                 socket::MsgFlags::empty(),
-            )?;
+            );
 
+            let reply_rcvd = match reply_res {
```
Furisto, line 259: :+1:

crates/libcgroups/src/systemd/dbus_native/dbus.rs
Furisto: > But that means we would have to do two function calls at each place we are doing a send message call. Would that be a better way?

I do not think it is. The complexity of should be contained in `send_message.` The caller should not have to think about it.

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -296,20 +303,47 @@ impl DbusConnection {
...
-            // we reset the buf, because I couldn't figure out how the adjust_counter function
-            // should should be changed to work correctly with non-zero start counter, and this solved that issue
-            buf = &buf[ctr..];
-            ret.push(msg);
+
+        // it is possible that while receiving messages, we get some extra/previous message
+        // for method calls, we need to have an error or method return type message, so
+        // we keep looping until we get either of these. see https://github.com/containers/youki/issues/2826
+        // for more detailed analysis.
+        loop {
```
Furisto, line 312: I am not sure I understand why this loop is necessary. Shouldn't `receive_complete_response` have read everything from the socket the first time?

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -296,20 +303,47 @@ impl DbusConnection {
...
-
-        while !buf.is_empty() {
-            let mut ctr = 0;
-            let msg = Message::deserialize(&buf[ctr..], &mut ctr)?;
-            // we reset the buf, because I couldn't figure out how the adjust_counter function
-            // should should be changed to work correctly with non-zero start counter, and this solved that issue
-            buf = &buf[ctr..];
-            ret.push(msg);
+
+        // it is possible that while receiving messages, we get some extra/previous message
```
Furisto, line 308: Do we know why we get these extra messages?

crates/libcgroups/src/systemd/dbus_native/dbus.rs
YJDoc2: The way `receive_complete_response` is implemented, it loops internally to read from socket into a buffer, and if read_bytes < buffer_size, it returns the buffer. This is because dbus uses same socket for communications, and the read will block if there is no further data to read. Hence we return once the received data is less than buffer. 
Unfortunately, the dbus messages can come in multiple bursts, so we can have two messages that get delivered together, or they can have few milliseconds between them, and thus one read call might only give single message < size of buffer. We need to loop here to keep consuming messages until we find one that we are looking for, which in this case is the method_return or error type.

crates/libcgroups/src/systemd/dbus_native/dbus.rs
YJDoc2: As of now the only extra messages that are observed are of type signal, which basically broadcast some event to all connected clients. The thing is we used to get these before as well, but usually these would get consumed in buffers of some other method calls or such, and get ignored. In few cases that would not happen, which lead to the dbus error in question. To solve this, now we keep consuming all messages until we get method return or error message (in case of method call message).

Furisto (approved):

Furisto:
Thanks @YJDoc2 !

PR #2845, opened 2024-07-07
Install nightly for running linter inside devcontainer

musaprg:
Nightly release is required for running formatter and linter as [`rustfmt.toml` uses unstable features](https://github.com/containers/youki/blob/aaa1590b564ad102832edd948ea8dbe462e24914/rustfmt.toml#L2C1-L2C86), however devcontainer only installs stable release. This PR adds a line to the devcontainer's Dockerfile to install nightly along with stable release.


YJDoc2 (approved):
Thanks for your contribution!

PR #2850, opened 2024-07-14
selinux: implemented remaining selinux functions

Gekko0114:
This is an experimental create. This PR is for selinux crate written in Rust.
In this PR, most of remaining selinux functions are implemented.

ref: #2718 #2800 #2825


Gekko0114:
@utam0k @YJDoc2 
Could you review this PR? I think most of remaining selinux functions are covered in this PR.
Though there are several remaining functions, I am not sure all of them are necessary.

experiment/selinux/src/selinux.rs
```diff
@@ -44,27 +87,196 @@ impl SELinux {
...
+                Err(Errno::EAGAIN) | Err(Errno::EINTR) => continue,
+                Err(_) => return false,
+            }
+        }
+    }
+
+    // function similar with findSELinuxfsMount in go-selinux repo.
+    // find_selinux_fs_mount returns a next selinuxfs mount point found,
+    // if there is one, or an empty string in case of EOF or error.
+    fn find_selinux_fs_mount(s: &str) -> String {
```
utam0k, line 192: Is the type of the return value `PathBuf`?
```suggestion
    fn find_selinux_fs_mount(s: &str) ->Option<PathBuf> {
```

experiment/selinux/src/selinux.rs
```diff
@@ -141,8 +353,8 @@ impl SELinux {
     // function similar with setFSCreateLabel in go-selinux repo.
     // set_fscreate_label sets the default label the kernel which the kernel is using
     // for file system objects.
-    pub fn set_fscreate_label(&self, label: &str) -> Result<(), SELinuxError> {
-        return Self::write_con(self.attr_path("fscreate").as_path(), label);
+    pub fn set_fscreate_label(&mut self, label: &str) -> Result<usize, SELinuxError> {
```
utam0k, line 356: Can't we make labels more user-friendly by making them their own structure instead of `&str`?

utam0k: And, I think `experiment/selinux/src/selinux.rs` is too big a file, so It'd be better to separate the functions related to `label` to reduce `experiment/selinux/src/selinux.rs`

utam0k (changes requested):
I'll still review this PR because the changes are significant. But I've done the first my review.

experiment/selinux/src/selinux.rs
Gekko0114: > And, I think experiment/selinux/src/selinux.rs is too big a file, so It'd be better to separate the functions related to label to reduce experiment/selinux/src/selinux.rs

Agree. I divided selinux.rs into two files, sharing same impl `SELinux`. Is this what you meant ?

> Can't we make labels more user-friendly by making them their own structure instead of &str?

Sounds very reasonable, but I am not sure how I should define new structure for labels.
Could you explain a little bit more about your idea on this?
You mean like this?

```
struct SELinuxLabel { aaa }
impl SELinuxLabel {fn ...}
```

If so, I don't think it is a good idea because a label should be passed as an arg for this struct and impl.
Therefore I would like to know your idea before starting the implementation :)

Gekko0114:
Hi @utam0k 

Thank you so much for your review. I've fixed or replied to your comments.
Could you take a look when you have time?

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +95,301 @@ impl SELinux {
...
+            },
+            None => -1,
+        };
+
+        match mode {
+            1 => SELinuxMode::ENFORCING,
+            0 => SELinuxMode::PERMISSIVE,
+            -1 => SELinuxMode::DISABLED,
+            _ => SELinuxMode::DISABLED,
+        }
```
utam0k, line 326: We can use From or Into instead of that.
https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +95,301 @@ impl SELinux {
...
+        if !self.policy_root_init_done.load(Ordering::SeqCst) {
+            let policy_root_path = Self::read_config(SELINUX_TYPE_TAG).unwrap_or_default();
+            self.policy_root = Some(PathBuf::from(policy_root_path));
+            self.policy_root_init_done.store(true, Ordering::SeqCst);
+        }
+        self.policy_root.as_ref()
+    }
+
+    // This function reads SELinux config file and returns the value with a specified key.
+    fn read_config(target: &str) -> Result<String, SELinuxError> {
```
utam0k, line 126: What is `target`? Can we use a type instead of  `&str`?

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +95,301 @@ impl SELinux {
...
+        self.selinuxfs.as_ref()
     }
 
-    // function similar with containerLabels in go-selinux repo.
-    // container_labels returns an allocated processLabel and fileLabel to be used for
-    // container labeling by the calling process.
-    pub fn container_labels() {
-        unimplemented!("not implemented yet")
+    // classIndex returns the int index for an object class in the loaded policy,
+    // or -1 and an error.
```
utam0k, line 263: I guess `-1` never returns, right?

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +95,301 @@ impl SELinux {
...
+    // If class is "file" or "dir", then return the corresponding index for selinux.
+    pub fn class_index(&mut self, class: &str) -> Result<i64, SELinuxError> {
```
utam0k, line 265: Can `class` be represented by enums or something?

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +95,301 @@ impl SELinux {
...
 
-    // function similar with fileLabel in go-selinux repo.
-    // fileLabel returns the SELinux label for this path, following symlinks,
-    // or returns an error.
-    pub fn file_label(fpath: &Path) -> Result<String, SELinuxError> {
-        if !fpath.exists() {
-            return Err(SELinuxError::FileLabel(ERR_EMPTY_PATH.to_string()));
+    // find_selinux_fs_mount returns a next selinuxfs mount point found,
+    // if there is one, or an empty string in case of EOF or error.
+    fn find_selinux_fs_mount(s: &str) -> Option<PathBuf> {
```
utam0k, line 200: What is `s`?

experiment/selinux/src/selinux_label.rs
```diff
@@ -0,0 +1,305 @@
...
+
+const XATTR_NAME_SELINUX: &str = "security.selinux";
+const KEY_LABEL_PATH: &str = "/proc/self/attr/keycreate";
+
+// This impl is for methods related to labels in SELinux struct.
+impl SELinux {
+    // set_file_label sets the SELinux label for this path, following symlinks, or returns an error.
+    pub fn set_file_label<P: AsRef<Path> + PathXattr>(
+        fpath: P,
+        label: &str,
```
utam0k, line 18: Can't you use a struct to represent `label`? e.g., 
```rust
struct SELinuxLabel {
    user: String,
    role: String,
    type_: String,
    level: Option<String>,
}
```

experiment/selinux/src/selinux.rs
Gekko0114: Thanks, fixed it

experiment/selinux/src/selinux.rs
Gekko0114: You're right, fixed it

experiment/selinux/src/selinux.rs
Gekko0114: The function's name and the argument's name are vague, so I renamed these.

In go-selinux repo, `s` stands for scanner. But it is not the correct name for this PR.
Therefore I renamed it.

experiment/selinux/src/selinux.rs
Gekko0114: `target` means a specified key in selinux config file.
In this cargo, read_config is used finding values of `SELINUX` and `SELINUXTYPE`.
However it is possible that there are other keys used in selinux config file, therefore I am not using a type instead of `&str`.

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +115,290 @@ impl SELinux {
...
+            self.policy_root_init_done.store(true, Ordering::SeqCst);
+        }
+        self.policy_root.as_ref()
+    }
+
+    // This function reads SELinux config file and returns the value with a specified key.
+    fn read_config(key: &str) -> Result<String, SELinuxError> {
+        let config_path = Path::new(SELINUX_DIR).join(SELINUX_CONFIG);
+        match File::open(config_path) {
+            Ok(file) => {
```
YJDoc2, line 149: For this -
1. can we rename this to `get_config_key` or something similar?
2. would it be possible to load the config at init time and store it in memory, so we don't have to open and read each time we want to get a key? Or is this behavior intentional as we always want the most updated key-value and config file can continuously change?

experiment/selinux/src/selinux.rs
Gekko0114: If the number of class is not so big, I think this is a good idea.

However, seems that there are more than 100 classes.
So I don't think it is a good idea implementing class as enums or something.

I confirmed the number of classes on amazon linux environment using the command `seinfo -c`.
There are 133 classes.

experiment/selinux/src/selinux.rs
Gekko0114: Sure, I will rename it.
Since the selinux config file is only read at boot time, your proposal is feasible.

experiment/selinux/src/selinux_label.rs
Gekko0114: Thanks, I've updated the implementation. Could you review it again?

Gekko0114:
Hi @utam0k , @YJDoc2 
Thanks for your review. I've updated my PR. Could you take a look?

utam0k:
I'll review it when I have more time to review. But the structure itself seemed very readable 😍

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +122,292 @@ impl SELinux {
...
-    pub fn set_key_label(label: &str) -> Result<(), SELinuxError> {
-        match Self::write_con(Path::new("/proc/self/attr/keycreate"), label) {
-            Ok(v) => Ok(v),
-            // TODO: This line will be fixed after implementing write_con.
-            Err(e) => Err(e),
+    // find_selinux_fs finds the SELinux filesystem mount point.
+    fn find_selinux_fs() -> Option<PathBuf> {
+        // fast path: check the default mount first
+        let selinux_fs_mount_path = PathBuf::from(SELINUX_FS_MOUNT);
+        if Self::verify_selinux_fs_mount(selinux_fs_mount_path.clone()) {
```
utam0k, line 249:
```suggestion
        if Self::verify_selinux_fs_mount(&selinux_fs_mount_path) {
```

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +122,292 @@ impl SELinux {
...
+                Ok(stat) => {
+                    // verify if the file is readonly or not
+                    if !stat.flags().contains(statvfs::FsFlags::ST_RDONLY) {
+                        return false;
+                    }
+                    // verify if the file is SELinux filesystem
+                    return stat.filesystem_type() == statfs::SELINUX_MAGIC;
                 }
+                // check again if there is an issue while calling statfs
+                Err(Errno::EAGAIN) | Err(Errno::EINTR) => continue,
```
utam0k, line 222: Why do we repeat this case? I'm afraid of the infinity loop.

experiment/selinux/src/selinux.rs
```diff
@@ -44,268 +122,292 @@ impl SELinux {
...
+            return None;
+        }
 
-    // function similar with containerLabels in go-selinux repo.
-    // container_labels returns an allocated processLabel and fileLabel to be used for
-    // container labeling by the calling process.
-    pub fn container_labels() {
-        unimplemented!("not implemented yet")
+        // continue reading until finding mount point
+        loop {
```
utam0k, line 260: Please tell me the case where we need the loop.

utam0k (approved):
~~The structure itself looks good to me +1. May I ask you to add `main.rs` to show an example of how to use this crate?~~ I was mistaken.

utam0k (changes requested):
The structure itself looks good to me +1. May I ask you to add main.rs to show an example of how to use this crate?

experiment/selinux/src/selinux.rs
Gekko0114: I had thought loop is needed because go-selinux has it as well.
But I understood that go-seliux is not infinite loop and my implementation is wrong. I will fix it.

https://github.com/opencontainers/selinux/blob/bb1ec25c889f29e35dffec33346af3c7d434dbc1/go-selinux/selinux_linux.go#L173-L182

experiment/selinux/src/selinux.rs
Gekko0114: @utam0k 
I referred to go-selinux's implementation.
I understood that we need infinity loop in case of temporary issues like interruptions.
Do you think I don't need the loop here?

https://github.com/opencontainers/selinux/blob/bb1ec25c889f29e35dffec33346af3c7d434dbc1/go-selinux/selinux_linux.go#L135

utam0k: Okay, I understood. Let's keep it. Thanks for your detailed explanation.

utam0k:
Is this PR ready for re-reviewing?

Gekko0114:
@utam0k 
Yes, I've added `main.rs`.
Please take a look when you have time :)

utam0k:
> @utam0k Yes, I've added `main.rs`. Please take a look when you have time :)

Awesome! Does `cargo run` need root permission, right? 
```
root ➜ /workspaces/youki/experiment/selinux (selinux) $ ls -Z test_file.txt 
unconfined_u:object_r:public_content_t:s1 test_file.txt
```

Gekko0114:
> Does cargo run need root permission, right?

Ah I hadn't tried this as a root on workspace, but seems you're right (I've tried it just now).

Seems that whether main.rs can be executed successfully depends on development environment.
I wrote this code on my ec2 amazon linux because selinux is enabled.
However, on my ec2 linux env, "set_file_label" is failed because of amazon linux's env (though I used root user).

YJDoc2:
Hey @Gekko0114 , do you think adding a vagrant file (or making changes in existing ones) to enable selinux in the vagrant box would be a good/easy idea? That way we can have a common dev env to write and test these things. As you said depending on how selinux is setup on machine, the main file can succeed or fail, I think it is a good idea to have a standard env to test in. wdyt?

Gekko0114:
@YJDoc2 @utam0k 
Agree, I think it is good preparing vagrant file for selinux.
Though I am not very familiar with vagrant, I can work on it.
I will take that work after completing this PR. wdyt?

utam0k:
> I will take that work after completing this PR. wdyt?

+1

utam0k (approved):
Thanks!

PR #2854, opened 2024-07-20
Bump oci-spec

keisku:
Enable to use `ExecCPUAffinity` which was introduced https://github.com/containers/oci-spec-rs/pull/174 and https://github.com/containers/oci-spec-rs/pull/178 to solve https://github.com/containers/youki/issues/2795.


utam0k (approved):
Thanks!

PR #2898, opened 2024-08-28
Integration test: cgroup v1 relative-cpus tests

posutsai:
Hi, I have implemented the test_relative_cpus function to contribute to #361  by referencing [`linux_cgroups_relative_cpus.go`](https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_cgroups_relative_cpus/linux_cgroups_relative_cpus.go) and [`cpu/v2.rs`](https://github.com/containers/youki/blob/main/tests/contest/contest/src/tests/cgroups/cpu/v2.rs). However, I encountered a few issues:

The current implementation closely mirrors the logic in v2.rs. Since there isn't a clear distinction between v1 and v2 logic in linux_cgroups_relative_cpus.go, would it be better to extract the common logic into a shared module?

I am also unsure about the correct way to run the tests. The just test-contest command on the main branch fails on Ubuntu 20.04. Could you provide guidance on resolving this issue?

Thank you for your help. I have learned a lot from this process.

### DIFF
--- a/tests/contest/contest/src/tests/cgroups/cpu/mod.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/mod.rs
@@ -41,11 +41,16 @@ fn create_cpu_spec(
     builder.build().context("failed to build cpu spec")
 }
 
-fn create_spec(cgroup_name: &str, case: LinuxCpu) -> Result<Spec> {
+fn create_spec(cgroup_name: &str, case: LinuxCpu, is_relative: bool) -> Result<Spec> {
+    let cgroups_path = if is_relative {
+        Path::new("testdir/runtime-test").join(cgroup_name)
+    } else {
+        Path::new("/runtime-test").join(cgroup_name)
+    };
     let spec = SpecBuilder::default()
         .linux(
             LinuxBuilder::default()
-                .cgroups_path(Path::new("/runtime-test").join(cgroup_name))
+                .cgroups_path(cgroups_path)
                 .resources(
                     LinuxResourcesBuilder::default()
                         .cpu(case)
--- a/tests/contest/contest/src/tests/cgroups/cpu/mod.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/mod.rs
@@ -41,16 +41,11 @@ fn create_cpu_spec(
     builder.build().context("failed to build cpu spec")
 }
 
-fn create_spec(cgroup_name: &str, case: LinuxCpu, is_relative: bool) -> Result<Spec> {
-    let cgroups_path = if is_relative {
-        Path::new("testdir/runtime-test").join(cgroup_name)
-    } else {
-        Path::new("/runtime-test").join(cgroup_name)
-    };
+fn create_spec_with_cgroup_path(cgroup_path: &str, case: LinuxCpu) -> Result<Spec> {
     let spec = SpecBuilder::default()
         .linux(
             LinuxBuilder::default()
-                .cgroups_path(cgroups_path)
+                .cgroups_path(Path::new(cgroup_path))
                 .resources(
                     LinuxResourcesBuilder::default()
                         .cpu(case)
@@ -64,6 +59,10 @@ fn create_spec(cgroup_name: &str, case: LinuxCpu, is_relative: bool) -> Result<S
         .context("failed to build spec")?;
 
     Ok(spec)
+}
+
+fn create_spec(cgroup_name: &str, case: LinuxCpu) -> Result<Spec> {
+    create_spec_with_cgroup_path(&format!("/runtime-test/{}", cgroup_name), case)
 }
 
 fn create_empty_spec(cgroup_name: &str) -> Result<Spec> {
--- a/tests/contest/contest/src/tests/cgroups/cpu/v1.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/v1.rs
@@ -2,14 +2,14 @@ use std::fs;
 use std::path::Path;
 use std::string::ToString;
 
-use anyhow::Result;
-use libcgroups::common;
-use num_cpus;
-use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
-
 use super::{create_cpu_spec, create_empty_spec, create_spec};
 use crate::utils::test_outside_container;
-use crate::utils::test_utils::{check_container_created, CGROUP_ROOT};
+use crate::utils::test_utils::check_container_created;
+use anyhow::Result;
+use libcgroups::common;
+use libcgroups::v1::{util, ControllerType};
+use num_cpus;
+use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
 
 const CPU_CGROUP_PREFIX: &str = "/sys/fs/cgroup/cpu,cpuacct";
 const DEFAULT_REALTIME_PERIOD: u64 = 1000000;
@@ -224,12 +224,12 @@ fn test_cpu_cgroups() -> TestResult {
 
 fn check_cgroup_subsystem(
     cgroup_name: &str,
-    subsystem: &str,
+    subsystem: &ControllerType,
     filename: &str,
     expected: &dyn ToString,
 ) -> Result<()> {
-    let cgroup_path = Path::new(CGROUP_ROOT)
-        .join(subsystem)
+    let mount_point = util::get_subsystem_mount_point(subsystem)?;
+    let cgroup_path = mount_point
         .join("runtime-test")
         .join(cgroup_name)
         .join(filename);
@@ -258,31 +258,31 @@ fn test_relative_cpus() -> TestResult {
         let cgroup_name = "test_relative_cpus";
         test_result!(check_cgroup_subsystem(
             cgroup_name,
-            "cpu,cpuacct",
+            &ControllerType::CpuAcct,
             "cpu.shares",
             &case.shares().unwrap(),
         ));
         test_result!(check_cgroup_subsystem(
             cgroup_name,
-            "cpu,cpuacct",
+            &ControllerType::CpuAcct,
             "cpu.cfs_period_us",
             &case.period().unwrap(),
         ));
         test_result!(check_cgroup_subsystem(
             cgroup_name,
-            "cpu,cpuacct",
+            &ControllerType::CpuAcct,
             "cpu.cfs_quota_us",
             &case.quota().unwrap(),
         ));
         test_result!(check_cgroup_subsystem(
             cgroup_name,
-            "cpuset",
+            &ControllerType::CpuSet,
             "cpuset.cpus",
             &case.cpus().to_owned().unwrap(),
         ));
         test_result!(check_cgroup_subsystem(
             cgroup_name,
-            "cpuset",
+            &ControllerType::CpuSet,
             "cpuset.mems",
             &case.mems().to_owned().unwrap()
         ));
--- a/tests/contest/contest/src/tests/cgroups/cpu/v1.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/v1.rs
@@ -238,12 +238,13 @@ fn check_cgroup_subsystem(
     debug!("reading value from {:?}", cgroup_path);
     let content = fs::read_to_string(&cgroup_path)
         .with_context(|| format!("failed to read {cgroup_path:?}"))?;
+    let trimmed = content.trim();
     if let Some(v) = expected.downcast_ref::<u64>() {
-        assert_eq!(&content.parse::<u64>()?, v);
+        assert_eq!(&trimmed.parse::<u64>()?, v);
     } else if let Some(v) = expected.downcast_ref::<i64>() {
-        assert_eq!(&content.parse::<i64>()?, v);
+        assert_eq!(&trimmed.parse::<i64>()?, v);
     } else if let Some(v) = expected.downcast_ref::<String>() {
-        assert_eq!(&content, v);
+        assert_eq!(&trimmed, v);
     }
     Ok(())
 }
--- a/tests/contest/contest/src/tests/cgroups/cpu/v1.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/v1.rs
@@ -2,11 +2,10 @@ use std::fs;
 use std::path::Path;
 use std::string::ToString;
 
-use anyhow::{Context, Result};
+use anyhow::Result;
 use libcgroups::common;
 use num_cpus;
 use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
-use tracing::debug;
 
 use super::{create_cpu_spec, create_empty_spec, create_spec};
 use crate::utils::test_outside_container;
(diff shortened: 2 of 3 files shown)


utam0k:
@YJDoc2 May I ask you to review this PR?

posutsai:
Sorry, I would like to make some refinement. @YJDoc2 , Could you please review later?

YJDoc2:
@posutsai , sure. May I request you to mark this as draft and ping me when this is ready to review?
Also instead of having multiple functions or dealing with any, I can suggest two ways -
1. make function generic with type parameter constrained by PartialEq,Eq , that way you can use `==` operator. Might need some more bounds
2. just take everything as string. So instead of giving 1000 as number we give it as a string for expected value. Not the best way, but I think good enough  for test result evaluation.

posutsai:
Thank you. Also I would like to know should I do the check in `test_cpu_cgroups` instead of declaring another function.

posutsai:
Hi, @YJDoc2 I believe the review can resume. I use `ToString` input type to handle different and check equality between `string`. Thank you.

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
```diff
@@ -219,6 +223,81 @@ fn test_cpu_cgroups() -> TestResult {
...
+                .context("no period value in cpu spec"))
+        ));
+        test_result!(check_cgroup_subsystem(
+            "test_relative_cpus",
+            "cpuset",
+            "cpuset.mems",
+            &test_result!(case
+                .mems()
+                .to_owned()
+                .context("no period value in cpu spec")),
```
YJDoc2, line 295: We should be able to unwrap here right? We are making the spec ourselves, so we know these values will always exist?

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
```diff
@@ -1,12 +1,16 @@
+use std::fs;
 use std::path::Path;
+use std::string::ToString;
 
+use anyhow::{Context, Result};
```
YJDoc2, line 5: Do we need to add this here? The existing tests used normal `Result` so should keep new one the same if possible.

YJDoc2 (changes requested):
Some changes, overall ok :+1:
Thanks :)

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
posutsai: I've remove `Context`. However, I believe it's necessary to use `Result` if we would like to check with `test_result` macro. Since there is already lots of functions like `create_cpu_spec` returning `anyhow::Result`, I guess it's okay to import `Result` in v1.rs.

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
posutsai: I've fixed it. Btw, just for curiosity. Is `unwrap` considered a bad practice due to panic? And we are okay to put here because of following reasons.

1. We control the output value
2. The test is okay to panic

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
```diff
@@ -1,12 +1,16 @@
...
 use libcgroups::common;
 use num_cpus;
 use test_framework::{test_result, ConditionalTest, TestGroup, TestResult};
+use tracing::debug;
```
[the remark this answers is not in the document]
posutsai: I believe the debug message is not necessary and remove the line. Thank you.

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
```diff
@@ -219,6 +223,81 @@ fn test_cpu_cgroups() -> TestResult {
...
+        "0",
+        get_realtime_period(),
+        get_realtime_runtime(),
+    ));
+    let spec = test_result!(create_spec("test_relative_cpus", case.clone()));
+
+    test_outside_container(spec, &|data| {
+        test_result!(check_container_created(&data));
+        test_result!(check_cgroup_subsystem(
+            "test_relative_cpus",
```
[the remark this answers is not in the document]
posutsai: Not exactly sure what the suggestion mean. But I guess you are saying to unpack the structure like the following snippet.

```rust
let LinuxCpu {
  shares,
  quota,
  ..
} = case;
```
I've tried. But these members are defined private. I believe use the corresponding method would be better. Please correct me if I have something misunderstanding. Thank you.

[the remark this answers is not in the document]
YJDoc2: Apologies, I meant extract the "test_relative_cpus" string into a variable as that is repeated

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
YJDoc2: ok, works :+1: I was only concerned as the tests were working without the imports before, and didn't wanted to add once which are not strictly needed.

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
YJDoc2: In this particular case it is primarily because we are controlling the output value. As we are building the `case` , we know that those fields will always be present, so we can be sure that this wont panic. While generally it is ok for a test to panic, which indicates a test failure anyways, it is preferable to avoid panicing in tests if possible.

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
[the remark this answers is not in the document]
posutsai: I've declared a `cgroup_name` variable and refined the corresponding function call. Thank you.

YJDoc2:
Hey, so I should have asked this a long time back, but even I'm catching this  right now - What you have implemented is still using the absolute cgroup path right? This does not have anything corresponding to https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_cgroups_relative_cpus/linux_cgroups_relative_cpus.go#L24 which sets the relative cgroup path.

posutsai:
Hi, sorry for late reply. I've modified the code to use `get_subsystem_mount_point` and take relative path as input. Not sure if I do it the right way.

YJDoc2:
> Hi, sorry for late reply.

No worries! I'll try to take a look at this today or so.

YJDoc2:
Hey, sorry it took me a while to get to this, got busy with other stuff. So in your recent changes, you are correctly taking it using the function however you are not setting the cgroup path as relative in the spec. For reference I'd suggest to take a look at https://github.com/containers/youki/pull/2686/files# , and see how it is done there, specifically https://github.com/containers/youki/pull/2686/files#diff-682de1209b41c713baac9258d0206875be877e6453cba0834357f52050d0520eR18-R21 

I'll also suggest to take a look at discussion there, it seems the original author might not be continuing with the PR, so if you're fine with it, you might be able to take that one up after this (if you want).

YJDoc2:
@posutsai , ping!

YJDoc2:
Hey @posutsai , let me know if you have any issues, or cannot work this anymore. It is ok if you are busy with something else, just let me know so we can proceed accordingly. Thanks :)

posutsai:
Hi, @YJDoc2. Sorry for reply lately. I've checked your reference url. I would like to know if the key point is to specify `cgroups` path as following.
```rust
    let spec = SpecBuilder::default()
        .linux(
            LinuxBuilder::default()
                // key here
                .cgroups_path(Path::new(RELATIVE_CGROUPS_PATH).join(cgroup_name))
                .resources(...)
                .build()
                .context("failed to build linux spec")?,
        )
        .build()
        .context("failed to build spec")?;
```
Thus, in our case we should avoid using the utility `create_cpu_spec` since there is no `cgroups_path` argument to pass in.

YJDoc2:
>  I would like to know if the key point is to specify cgroups path as following.

Yes, we should be specifying the cgroups path similar to that.

> Thus, in our case we should avoid using the utility create_cpu_spec since there is no cgroups_path argument to pass in.

Yes, we can either create a new fn to take a path param, or modify the original one to optinally take a param (Option<Path> or something like that) , and default to current behavior if no path is passed (i.e. None is passed)

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
```diff
@@ -219,6 +223,74 @@ fn test_cpu_cgroups() -> TestResult {
...
+        1024,
+        100000,
+        50000,
+        None,
+        "0-1",
+        "0",
+        get_realtime_period(),
+        get_realtime_runtime(),
+    ));
+    let spec = test_result!(create_spec("test_relative_cpus", case.clone()));
```
posutsai, line 255: Please correct me if I am wrong. After checking the [definition](https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/tests/cgroups/cpu/mod.rs#L44) of the `create_spec` function, seems like I already specify the `cgroups_path` by calling it.
```rust
fn create_spec(cgroup_name: &str, case: LinuxCpu) -> Result<Spec> {
    let spec = SpecBuilder::default()
        .linux(
            LinuxBuilder::default()
                .cgroups_path(Path::new("/runtime-test").join(cgroup_name))
                .resources(
                    LinuxResourcesBuilder::default()
                        .cpu(case)
                        .build()
                        .context("failed to build resource spec")?,
                )
                .build()
                .context("failed to build linux spec")?,
        )
        .build()
        .context("failed to build spec")?;

    Ok(spec)
}
```
If that's the case, I guess I have no need to modify `create_cpu_spec` interface. Am I right?

utam0k: Is your intention to edit `Spec` after calling `create_spec`? I prefer passing a cgroup path to `create_cpu_spec`.

posutsai: Thank you for reply. I totally miss the point. I plan to modify `create_spec` and add an optional argument, since the original code already calls the builder there. Is it better than modify `create_cpu_spec`?

utam0k: Yes, it'd be better because `create_cpu_spec` should only be editing related CPU field,s not the cgroup path.

posutsai: I've add a `is_relative` argument in `create_spec`. But I have no idea why the CI fail.

utam0k: 📝 
```
5 / 5 : test_relative_cpus : not ok
```

tests/contest/contest/src/tests/cgroups/cpu/mod.rs
```diff
@@ -41,11 +41,16 @@ fn create_cpu_spec(
     builder.build().context("failed to build cpu spec")
 }
 
-fn create_spec(cgroup_name: &str, case: LinuxCpu) -> Result<Spec> {
+fn create_spec(cgroup_name: &str, case: LinuxCpu, is_relative: bool) -> Result<Spec> {
+    let cgroups_path = if is_relative {
+        Path::new("testdir/runtime-test").join(cgroup_name)
+    } else {
+        Path::new("/runtime-test").join(cgroup_name)
+    };
```
utam0k, line 49: Why don't we pass the cgroup path itself?

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
posutsai: I've tried to run a container with following `config.json` which specified specific relative `cgroupsPath` and everything works fine.
```json
{
	"ociVersion": "1.2.0",
	"linux": {
		"resources": {
			"devices": [
				{
					"allow": false,
					"access": "rwm"
				}
			]
		},
		"cgroupsPath": "testdir/runtime-test",
		"namespaces": [
			{
				"type": "pid"
			},
			{
				"type": "network"
			},
			{
				"type": "ipc"
			},
			{
				"type": "uts"
			},
			{
				"type": "mount"
			}
		],
		"maskedPaths": [
			"/proc/acpi",
			"/proc/asound",
			"/proc/kcore",
			"/proc/keys",
			"/proc/latency_stats",
			"/proc/timer_list",
			"/proc/timer_stats",
			"/proc/sched_debug",
			"/sys/firmware",
			"/proc/scsi"
		],
		"readonlyPaths": [
			"/proc/bus",
			"/proc/fs",
			"/proc/irq",
			"/proc/sys",
			"/proc/sysrq-trigger"
		]
	}
}
```
Also, my `runc` is version `1.2.4`

tests/contest/contest/src/tests/cgroups/cpu/mod.rs
utam0k: This flag increases the number of arguments in this function. The two reasons you pointed out are not enough to increase it.

utam0k: How about defining the functions `create_spec()` and `create_spec_with_cgroup_path`?

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
utam0k: `testdir/runtime-test` has not been created in advance, right? Is that okay?

Also, there isn't enough error message, so why not add an error message in the `?` part?

tests/contest/contest/src/tests/cgroups/cpu/mod.rs
posutsai: okay, I'll change it and call `create_spec_with_cgroup_path` in `create_spec`. Thank you.

tests/contest/contest/src/tests/cgroups/cpu/v1.rs
posutsai: Does the issue comes from the [user.slice](https://github.com/opencontainers/runc/blob/a7d76457f4b630449d04a8caa81c0a7f2052afd3/docs/systemd.md#systemd-unit-name-and-placement) part? Cause my actual cgroup path is `/sys/fs/cgroup/cpu,cpuacct/user.slice/testdir/runtime-test/`

utam0k: > Does the issue come from the [user.slice](https://github.com/opencontainers/runc/blob/a7d76457f4b630449d04a8caa81c0a7f2052afd3/docs/systemd.md#systemd-unit-name-and-placement) part?

I'm not sure, but there is a possibility.
OCI Runtime said that so, `/sys/fs/cgroup/cpu,cpuacct/user.slice/$relative-path` should be okay to pass.
> In the case of a relative path (not starting with /), the runtime MAY interpret the path relative to a runtime-determined location in the cgroups hierarchy.

https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#cgroups-path

utam0k:
> Hi, @utam0k I would like to know if solving this CI issue is included in the scope of this PR? If so, could you give me some hint or direction to fix it. For example, which file to inspect. If not, is there any possible way to leave some expected failure mark in CI?

I think it is within scope so far. Would you like me to take over?

posutsai:
yes, please. I have no idea how to fix the issue. Thank you.

PR #2915, opened 2024-09-16
Fixed ENAMETOOLONG error in setup_console_socket

morganllewellynjones:
In reference to the error found here: https://github.com/containers/youki/issues/2910

sun_path has a max length of 108 bytes: https://man7.org/linux/man-pages/man7/unix.7.html

And including the absolute path name can make the resulting address of the socket path too long for sun_path.

Quick solution, move to the container_dir first, that way the sun_path is always just as long as the name "console-socket" itself, and doesn't break depending on the length of the runtime filepath.

I also wrapped this update in code that saves the previous directory and moves back to it afterwords. I personally doubt this is necessary and may simply be wasted clock cycles. Since we tend to use absolute path names in general, both to navigate and to access specific links. Nevertheless, I didn't want to make that call. If repositioning the directory afterwards is not necessary that could certainly be cleaned out. Cargo test passes in either case.

### DIFF
--- a/crates/libcontainer/src/tty.rs
+++ b/crates/libcontainer/src/tty.rs
@@ -1,5 +1,6 @@
 //! tty (teletype) for user-system interaction
 
+use std::env;
 use std::io::IoSlice;
 use std::os::unix::fs::symlink;
 use std::os::unix::io::AsRawFd;
@@ -75,7 +76,13 @@ pub fn setup_console_socket(
     console_socket_path: &Path,
     socket_name: &str,
 ) -> Result<RawFd> {
-    let linked = container_dir.join(socket_name);
+    // Move into the container directory to avoid sun family conflicts with long socket path names.
+
+    let prev_dir = env::current_dir().unwrap();
+    let _ = env::set_current_dir(container_dir);
+
+    let linked = PathBuf::from(socket_name);
+
     symlink(console_socket_path, &linked).map_err(|err| TTYError::Symlink {
         source: err,
         linked: linked.to_path_buf().into(),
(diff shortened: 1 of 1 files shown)


YJDoc2:
Hey, thanks for helping us with this! Unfortunately I'm afraid we'll need some time to get to reviewing this PR, as there are already several others in the backlog, and the maintainers are a bit busy at the moment. I'll ping you if any changes are needed when I get to checking this, but as the diff is small, this should be ok in general.

Also, is https://github.com/containers/youki/pull/2914 duplicate of this? Can you close that? Thanks!

morganllewellynjones:
No worries! The duplicate has been closed.

crates/libcontainer/src/tty.rs
```diff
@@ -75,7 +76,13 @@ pub fn setup_console_socket(
     console_socket_path: &Path,
     socket_name: &str,
 ) -> Result<RawFd> {
-    let linked = container_dir.join(socket_name);
+    // Move into the container directory to avoid sun family conflicts with long socket path names.
```
utam0k, line 79:
```suggestion
    // Move into the container directory to avoid sun family conflicts with long socket path names.
    // ref: https://github.com/containers/youki/issues/2910
```

utam0k (approved):
Thanks for your first contribution. Your detailed explanation and concise explanation helped me to review this PR.
LGTM with nits.

YJDoc2:
Hey, the DCO is failing probably due to directly committing the code from the suggestion. Given that the original suggestion was by utam0k and you have committed it directly from github, I don't think you have any problem with signing DCO. I will sign the last commit myself and force push in sometime unless you want to do it yourself or have any problems.

YJDoc2:
Manually setting DCO to pass and merging.

PR #2924, opened 2024-09-22
Add seccomp into feature flags of youki to be compiled in

musaprg:
Currently, the youki [disables the `libcontainer/libseccomp` feature by setting `default-feature = false`](https://github.com/containers/youki/blob/0b92791222322fa804bae50feed49983d47005ec/crates/youki/Cargo.toml#L33). If libcontainer has a feature flag for seccomp, youki should also have the feature flag so that we can use it by enabling it through the build option.

Since I'm new to this project, I don't have enough background and status of seccomp support in youki. Please let me know if there's any reason why this feature is explicitly disabled in youki, for example, it's still an experimental feature.


yihuaf:
In short, not all platform supports libseccomp. There are some context in these issues and MR. Please read them and I can help answer further questions :)

- https://github.com/containers/youki/issues/382
- https://github.com/containers/youki/pull/1484

utam0k:
May I ask you to check CI?

musaprg:
Hi. I rechecked and fixed my changes to the test cases. Could you check it again? Thanks.

utam0k (approved):
Make sense

PR #2936, opened 2024-09-28
selinux: fix xattr and remove anyhow

Gekko0114:
Fixed several points based on the feedback I received.
https://github.com/containers/youki/issues/2718

1. This PR updates the logic of xattr.
If a file already has an xattr label, the `REPLACE` method is used.
If a file doesn't have an xattr label, the `CREATE` method is used.

2. This PR removes anyhow from selinux.

### DIFF
--- a/experiment/selinux/src/selinux.rs
+++ b/experiment/selinux/src/selinux.rs
@@ -91,8 +91,6 @@ pub enum SELinuxError {
     GetConfigKey(String),
     #[error("Invalid format for SELinux label: {0}")]
     InvalidSELinuxLabel(String),
-    #[error("Failed to run: {0}")]
-    Run(String),
 }
 
 pub struct SELinux {
--- a/experiment/selinux/src/selinux.rs
+++ b/experiment/selinux/src/selinux.rs
@@ -91,6 +91,8 @@ pub enum SELinuxError {
     GetConfigKey(String),
     #[error("Invalid format for SELinux label: {0}")]
     InvalidSELinuxLabel(String),
+    #[error("Failed to run: {0}")]
+    Run(String),
 }
 
 pub struct SELinux {
@@ -489,7 +491,7 @@ mod tests {
 
     fn create_temp_file(content: &[u8], file_name: &str) {
         let path = Path::new(file_name);
-        let mut file = File::create(&path).expect("Failed to create file");
+        let mut file = File::create(path).expect("Failed to create file");
         file.write_all(content).expect("Failed to write to file");
         file.sync_all().expect("Failed to sync file");
     }
@@ -570,7 +572,7 @@ mod tests {
             let expected = PathBuf::from(expected_array[i]);
             match SELinux::check_line_include_selinux_fs_mount_point(input) {
                 Some(output) => assert_eq!(expected, output),
-                None => assert_eq!(succeeded_array[i], false),
+                None => assert!(!succeeded_array[i]),
             }
         }
     }
--- a/experiment/selinux/src/tools/xattr.rs
+++ b/experiment/selinux/src/tools/xattr.rs
@@ -34,15 +34,27 @@ where
     // set_xattr sets extended attributes on a file specified by its path.
     fn set_xattr(&self, attr: &str, data: &[u8]) -> Result<(), XattrError> {
         let path = self.as_ref();
-        match rfs::setxattr(path, attr, data, rfs::XattrFlags::REPLACE) {
-            Ok(_) => Ok(()),
-            Err(e) => {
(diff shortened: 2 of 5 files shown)


Gekko0114:
Hi @YJDoc2 @utam0k 
I've fixed selinux lib based on feedback from @YJDoc2 https://github.com/containers/youki/issues/2718#issuecomment-2368063122
Could you review it?

experiment/selinux/src/tools/xattr.rs
```diff
@@ -34,31 +34,55 @@ where
...
+            Err(_) => match rfs::lsetxattr(path, attr, data, rfs::XattrFlags::CREATE) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    let errno = e.raw_os_error();
+                    if errno == libc::EINTR {
+                        return Err(XattrError::EINTR(errno));
+                    }
+                    Err(XattrError::LSetXattr(e.to_string()))
+                }
+            },
```
YJDoc2, line 85: Given that rest of the code is same, can we change this match to -
```rust
let op = match path.lget_xattr(attr) {
    Ok(_) => rfs::XattrFlags::REPLACE,
    Err(_) => rfs::XattrFlags::CREATE
};
// code to call lsetxattr which uses `op` ...
```

experiment/selinux/src/tools/xattr.rs
```diff
@@ -34,31 +34,55 @@ where
...
+            Err(_) => match rfs::setxattr(path, attr, data, rfs::XattrFlags::CREATE) {
+                Ok(_) => Ok(()),
+                Err(e) => {
+                    let errno = e.raw_os_error();
+                    if errno == libc::EINTR {
+                        return Err(XattrError::EINTR(errno));
+                    }
+                    Err(XattrError::SetXattr(e.to_string()))
+                }
+            },
```
YJDoc2, line 57: Same comment as the code below.

experiment/selinux/src/selinux.rs
```diff
@@ -91,6 +91,8 @@ pub enum SELinuxError {
     GetConfigKey(String),
     #[error("Invalid format for SELinux label: {0}")]
     InvalidSELinuxLabel(String),
+    #[error("Failed to run: {0}")]
+    Run(String),
```
YJDoc2, line 95: Hey one more question here - If we are introducing `Run` error just for the `main.rs` part, and not using it anywhere else, is it really worth adding an error type for it? My reasoning is `main.rs` is not going to be actually used for anything, I believe it is  just for testing and verifying the code, so we can either ignore the error there or unwrap if it is a fatal error. Correct me if I'm wrong.

YJDoc2 (commented):
Hey overall good, a couple of suggestions in the code, with two questions here -
1. Please run clippy on local and check that all lints are passing, currently that is not getting verified in CI.
2. I saw that for `EINTR`, we are returning it as-is to caller. If my understanding is correct `EINTR` indicates that syscall was interrupted and there might not be any actual error. In which case should we retry? in a loop? I think we had this discussion before somewhere, what was decided from that?

Apart from this, lgtm :+1:

experiment/selinux/src/selinux.rs
Gekko0114: SGTM, removed `Run` error

Gekko0114:
Thanks @YJDoc2,

> Please run clippy on local and check that all lints are passing, currently that is not getting verified in CI.

Sure, I ran clippy. BTW, why not introduing clippy in CI ?

> I saw that for EINTR, we are returning it as-is to caller. If my understanding is correct EINTR indicates that syscall was interrupted and there might not be any actual error. In which case should we retry? in a loop? I think we had this discussion before somewhere, what was decided from that?

Though we didn't discuss it in great depth, we talked about this topic.
https://github.com/containers/youki/pull/2850#discussion_r1702401564
In the xattr function, a loop is not used, but a loop is used in the label function that calls xattr.
https://github.com/Gekko0114/youki/blob/main/experiment/selinux/src/selinux_label.rs#L75
Does this answer your question?

YJDoc2:
> why not introduing clippy in CI ?

Clippy is in CI, but we have intentionally excluded experimental crates from cargo workspace because of which clippy does not run on those crates.

>  Does this answer your question?

Yep, that's exactly what I was looking for. In this case, ok we are handling it correctly.

YJDoc2 (approved):
Thanks!

YJDoc2:
I re-labeled this to experimental, as we have still kept the crate in experimental. Once we move this and start using it in youki, that PR should be marked as feature.

PR #2950, opened 2024-10-09
linux_masked_paths integration test

nayuta-ai:
## Summary
Create linux_masked_paths integration test.

## Related Issue
#361 

## Refer
- https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_masked_paths/linux_masked_paths.go
- https://github.com/containers/youki/tree/main/tests/contest/contest/src/tests/readonly_paths

### DIFF
--- a/tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
+++ b/tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
@@ -1,0 +1,256 @@
+use std::path::PathBuf;
+
+use anyhow::{anyhow, bail};
+use nix::sys::stat::SFlag;
+use oci_spec::runtime::{LinuxBuilder, ProcessBuilder, Spec, SpecBuilder};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::utils::test_inside_container;
+
+fn get_spec(masked_paths: Vec<String>) -> Spec {
+    SpecBuilder::default()
+        .linux(
+            LinuxBuilder::default()
+                .masked_paths(masked_paths)
+                .build()
+                .expect("could not build"),
+        )
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["runtimetest".to_string(), "masked_paths".to_string()])
+                .build()
+                .unwrap(),
+        )
+        .build()
+        .unwrap()
+}
+
+fn check_masked_paths() -> TestResult {
+    let masked_dir = "masked-dir";
+    let masked_subdir = "masked-subdir";
+    let masked_file = "masked-file";
+
+    let masked_dir_top = PathBuf::from(masked_dir);
+    let masked_file_top = PathBuf::from(masked_file);
+
+    let masked_dir_sub = masked_dir_top.join(masked_subdir);
+    let masked_file_sub = masked_dir_top.join(masked_file);
+    let masked_file_sub_sub = masked_dir_sub.join(masked_file);
+
+    let root = PathBuf::from("/");
+
+    let masked_paths = vec![
+        root.join(&masked_dir_top).to_string_lossy().to_string(),
+        root.join(&masked_file_top).to_string_lossy().to_string(),
+        root.join(&masked_dir_sub).to_string_lossy().to_string(),
+        root.join(&masked_file_sub).to_string_lossy().to_string(),
+        root.join(&masked_file_sub_sub)
+            .to_string_lossy()
+            .to_string(),
+    ];
+
+    let spec = get_spec(masked_paths);
+
+    test_inside_container(spec, &|bundle_path| {
+        use std::{fs, io};
+        let test_dir = bundle_path.join(&masked_dir_sub);
+
+        match fs::create_dir_all(&test_dir) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
+
+        match fs::File::create(test_dir.join("tmp")) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
+
+        let test_sub_sub_file = bundle_path.join(&masked_file_sub_sub);
+        match fs::File::create(test_sub_sub_file) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
+
+        let test_sub_file = bundle_path.join(&masked_file_sub);
+        match fs::File::create(test_sub_file) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
+
+        let test_file = bundle_path.join(masked_file);
+        match fs::File::create(test_file) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
+
+        Ok(())
+    })
+}
+
+fn check_masked_rel_paths() -> TestResult {
+    // Deliberately set a relative path to be masked,
+    // and expect an error
+    let masked_rel_path = "masked_rel_path";
+    let masked_paths = vec![masked_rel_path.to_string()];
+    let spec = get_spec(masked_paths);
+
+    test_inside_container(spec, &|bundle_path| {
+        use std::{fs, io};
+        let test_file = bundle_path.join(masked_rel_path);
+        match fs::metadata(&test_file) {
+            io::Result::Ok(md) => {
+                bail!(
+                    "reading path {:?} should have given error, found {:?} instead",
+                    test_file,
+                    md
+                )
+            }
+            io::Result::Err(e) => {
+                let err = e.kind();
+                if let io::ErrorKind::NotFound = err {
+                    Ok(())
+                } else {
+                    bail!("expected not found error, got {:?}", err);
+                }
+            }
+        }
+    })
+}
+
+fn check_masked_symlinks() -> TestResult {
+    // Deliberately create a masked symlink that points an invalid file,
+    // and expect an error.
+    let root = PathBuf::from("/");
+    let masked_symlink = "masked_symlink";
+    let masked_paths = vec![root.join(masked_symlink).to_string_lossy().to_string()];
+    let spec = get_spec(masked_paths);
+
+    let res = test_inside_container(spec, &|bundle_path| {
+        use std::{fs, io};
+        let test_file = bundle_path.join(masked_symlink);
+        // ln -s .. /masked-symlink ; readlink -f /masked-symlink; ls -L /masked-symlink
+        match std::os::unix::fs::symlink("../masked_symlink", &test_file) {
+            io::Result::Ok(_) => { /* This is expected */ }
+            io::Result::Err(e) => {
+                bail!("error in creating symlink, to {:?} {:?}", test_file, e);
+            }
+        }
+
+        let r_path = match fs::read_link(&test_file) {
+            io::Result::Ok(p) => p,
+            io::Result::Err(e) => {
+                bail!("error in reading symlink at {:?} : {:?}", test_file, e);
+            }
+        };
+
+        match fs::metadata(r_path) {
+            io::Result::Ok(md) => {
+                bail!(
+                    "reading path {:?} should have given error, found {:?} instead",
+                    test_file,
+                    md
+                )
+            }
+            io::Result::Err(e) => {
+                let err = e.kind();
+                if let io::ErrorKind::NotFound = err {
+                    Ok(())
+                } else {
+                    bail!("expected not found error, got {:?}", err);
+                }
+            }
+        }
+    });
+
+    if let TestResult::Passed = res {
+        TestResult::Failed(anyhow!(
+            "expected error in container creation with invalid symlink, found no error"
+        ))
+    } else {
+        TestResult::Passed
+    }
+}
+
+fn test_node(mode: u32) -> TestResult {
+    let root = PathBuf::from("/");
+    let masked_device = "masked_device";
+    let masked_paths = vec![root.join(masked_device).to_string_lossy().to_string()];
+    let spec = get_spec(masked_paths);
+
+    test_inside_container(spec, &|bundle_path| {
+        use std::os::unix::fs::OpenOptionsExt;
+        use std::{fs, io};
+        let test_file = bundle_path.join(masked_device);
+
+        let mut opts = fs::OpenOptions::new();
+        opts.mode(mode);
+        opts.create(true);
+        if let io::Result::Err(e) = fs::OpenOptions::new()
+            .mode(mode)
+            .create(true)
+            .write(true)
+            .open(&test_file)
+        {
+            bail!(
+                "could not create file {:?} with mode {:?} : {:?}",
+                test_file,
+                mode ^ 0o666,
+                e
+            );
+        }
+
+        match fs::metadata(&test_file) {
+            io::Result::Ok(_) => Ok(()),
+            io::Result::Err(e) => {
+                let err = e.kind();
+                if let io::ErrorKind::NotFound = err {
+                    bail!("error in creating device node, {:?}", e)
+                } else {
+                    Ok(())
+                }
+            }
+        }
+    })
+}
+
+fn check_masked_device_nodes() -> TestResult {
+    let modes = [
+        SFlag::S_IFBLK.bits() | 0o666,
+        SFlag::S_IFCHR.bits() | 0o666,
+        SFlag::S_IFIFO.bits() | 0o666,
+    ];
+    for mode in modes {
+        let res = test_node(mode);
+        if let TestResult::Failed(_) = res {
+            return res;
+        }
+        std::thread::sleep(std::time::Duration::from_millis(1000));
+    }
+    TestResult::Passed
+}
+
+pub fn get_linux_masked_paths_tests() -> TestGroup {
+    let mut tg = TestGroup::new("masked_paths");
+    let masked_paths_test = Test::new("masked_paths", Box::new(check_masked_paths));
+    let masked_rel_paths_test = Test::new("masked_rel_paths", Box::new(check_masked_rel_paths));
+    let masked_symlinks_test = Test::new("masked_symlinks", Box::new(check_masked_symlinks));
+    let masked_device_nodes_test =
+        Test::new("masked_device_nodes", Box::new(check_masked_device_nodes));
+    tg.add(vec![
+        Box::new(masked_paths_test),
+        Box::new(masked_rel_paths_test),
+        Box::new(masked_symlinks_test),
+        Box::new(masked_device_nodes_test),
+    ]);
+    tg
+}
--- a/tests/contest/runtimetest/src/main.rs
+++ b/tests/contest/runtimetest/src/main.rs
@@ -33,6 +33,7 @@ fn main() {
         "hello_world" => tests::hello_world(&spec),
         ////////// ANCHOR_END: example_runtimetest_main
         "readonly_paths" => tests::validate_readonly_paths(&spec),
+        "masked_paths" => tests::validate_masked_paths(&spec),
         "set_host_name" => tests::validate_hostname(&spec),
         "mounts_recursive" => tests::validate_mounts_recursive(&spec),
         "domainname_test" => tests::validate_domainname(&spec),
--- a/tests/contest/runtimetest/src/tests.rs
+++ b/tests/contest/runtimetest/src/tests.rs
@@ -545,3 +545,34 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
         eprintln!("error ioprio_get expected priority {expected_priority:?}, got {priority}")
     }
 }
+
+pub fn validate_masked_paths(spec: &Spec) {
+    let linux = spec.linux().as_ref().unwrap();
+    let masked_paths = match linux.masked_paths() {
+        Some(p) => p,
+        None => {
+            eprintln!("in readonly paths, expected some readonly paths to be set, found none");
+            return;
+        }
+    };
+
+    if masked_paths.is_empty() {
+        return;
+    }
+
+    // TODO when https://github.com/rust-lang/rust/issues/86442 stabilizes,
+    // change manual matching of i32 to e.kind() and match statement
+    for path in masked_paths {
+        if let std::io::Result::Err(e) = test_read_access(path) {
+            let errno = Errno::from_raw(e.raw_os_error().unwrap());
+            if errno == Errno::ENOENT {
+                /* This is expected */
+            } else {
+                eprintln!("in masked paths, error in testing read access for path {path} : {e:?}");
+                return;
+            }
+        } else {
+            /* Expected */
+        }
+    }
+}
--- a/tests/contest/contest/src/main.rs
+++ b/tests/contest/contest/src/main.rs
@@ -17,6 +17,7 @@ use crate::tests::hostname::get_hostname_test;
 use crate::tests::intel_rdt::get_intel_rdt_test;
 use crate::tests::io_priority::get_io_priority_test;
 use crate::tests::lifecycle::{ContainerCreate, ContainerLifecycle};
+use crate::tests::linux_masked_paths::get_linux_masked_paths_tests;
 use crate::tests::linux_ns_itype::get_ns_itype_tests;
 use crate::tests::mounts_recursive::get_mounts_recursive_test;
 use crate::tests::pidfile::get_pidfile_test;
@@ -113,6 +114,7 @@ fn main() -> Result<()> {
     let scheduler = get_scheduler_test();
     let io_priority_test = get_io_priority_test();
     let devices = get_devices_test();
+    let masked_paths = get_linux_masked_paths_tests();
 
     tm.add_test_group(Box::new(cl));
     tm.add_test_group(Box::new(cc));
@@ -136,6 +138,7 @@ fn main() -> Result<()> {
     tm.add_test_group(Box::new(sysctl));
     tm.add_test_group(Box::new(scheduler));
     tm.add_test_group(Box::new(devices));
+    tm.add_test_group(Box::new(masked_paths));
 
     tm.add_test_group(Box::new(io_priority_test));
     tm.add_cleanup(Box::new(cgroups::cleanup_v1));
--- a/tests/contest/contest/src/tests/linux_masked_paths/mod.rs
+++ b/tests/contest/contest/src/tests/linux_masked_paths/mod.rs
@@ -1,0 +1,3 @@
+mod masked_paths;
+
+pub use masked_paths::get_linux_masked_paths_tests;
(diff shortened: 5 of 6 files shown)


tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,256 @@
...
+    test_inside_container(spec, &|bundle_path| {
+        use std::{fs, io};
+        let test_dir = bundle_path.join(&masked_dir_sub);
+
+        match fs::create_dir_all(&test_dir) {
+            io::Result::Ok(_) => { /*This is expected*/ }
+            io::Result::Err(e) => {
+                bail!(e)
+            }
+        }
```
utam0k, line 63:
```suggestion
		if let Err(e) = fs::create_dir_all(&test_dir) {
			bail!(e)
        }
```

nayuta-ai:
@utam0k Thank you for the review! Please review my code again.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,3 +545,34 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+        Some(p) => p,
+        None => {
+            eprintln!("in readonly paths, expected some readonly paths to be set, found none");
+            return;
+        }
+    };
+
+    if masked_paths.is_empty() {
+        return;
+    }
```
YJDoc2, line 561: Should we error here also like above?

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,3 +545,34 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+            let errno = Errno::from_raw(e.raw_os_error().unwrap());
+            if errno == Errno::ENOENT {
+                /* This is expected */
+            } else {
+                eprintln!("in masked paths, error in testing read access for path {path} : {e:?}");
+                return;
+            }
+        } else {
+            /* Expected */
+        }
```
YJDoc2, line 576: Is this expected? For masked path we are expecting the path to NOT be readable inside the container, right? So if this is `Ok`, then that should be error, right?

YJDoc2 (changes requested):
I still have to check the main test implementation, have a couple of comments, and clippy CI is failing, so please fix that as well.

tests/contest/runtimetest/src/tests.rs
nayuta-ai: In the [Go implementation](https://github.com/opencontainers/runtime-tools/blob/master/cmd/runtimetest/main.go#L904-L907), it is skipped, so it doesn't seem necessary to return an error here. However, since entering this process would be problematic for the test cases we are considering, I will add an error.
[62e5e1a](https://github.com/containers/youki/pull/2950/commits/62e5e1a1a7bd04ea1a06275661f0325ea75da37f)

tests/contest/runtimetest/src/tests.rs
nayuta-ai: Sorry, you're right. I fixed it. [62e5e1a](https://github.com/containers/youki/pull/2950/commits/62e5e1a1a7bd04ea1a06275661f0325ea75da37f)

nayuta-ai: I added this line but encountered the following error, so I'm currently investigating the cause. As soon as I finish, I'll request a review.
```
# Start group masked_paths
1 / 4 : masked_device_nodes : not ok
	container stderr was not empty, found : in masked paths, error in testing read access for path /masked_device: 8630

2 / 4 : masked_paths : not ok
	container stderr was not empty, found : in masked paths, error in testing read access for path /masked-file: 8630

3 / 4 : masked_rel_paths : ok
4 / 4 : masked_symlinks : ok
# End group masked_paths
```

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,3 +545,36 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+            if errno == Errno::ENOENT {
+                /* This is expected */
+            } else {
+                eprintln!("in masked paths, error in testing read access for path {path} : {e:?}");
+                return;
+            }
+        } else {
+            eprintln!("in masked paths, error in testing read access for path {path}");
+            return;
+        }
```
YJDoc2, line 578:
```suggestion
        } else {
            eprintln!("in masked paths, expected path {path} to be masked, but was found readable");
            return;
        }
```

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,3 +545,36 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+    // change manual matching of i32 to e.kind() and match statement
+    for path in masked_paths {
+        if let std::io::Result::Err(e) = test_read_access(path) {
+            let errno = Errno::from_raw(e.raw_os_error().unwrap());
+            if errno == Errno::ENOENT {
+                /* This is expected */
+            } else {
+                eprintln!("in masked paths, error in testing read access for path {path} : {e:?}");
+                return;
+            }
```
YJDoc2, line 574:
```suggestion
            } else {
                eprintln!("in masked paths, error in testing read access for path {path} : {errno:?}");
                return;
            }
```

I think errno would be better readability than raw error code.

nayuta-ai:
@utam0k @YJDoc2 
I have completed the revisions. Please check!

YJDoc2:
>  I have completed the revisions. Please check!

Hey, thanks a lot, I'll try to get to this today or so.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,251 @@
...
+            io::Result::Err(e) => {
+                let err = e.kind();
+                if let io::ErrorKind::NotFound = err {
+                    Ok(())
+                } else {
+                    bail!("expected not found error, got {:?}", err);
+                }
+            }
+        }
+    })
```
YJDoc2, line 122: Why is this check here? we should be checking this in the runtime test and here we should be creating the path, right? Or did I misunderstood anything?

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        let r_path = match fs::read_link(&test_file) {
+            io::Result::Ok(p) => p,
+            io::Result::Err(e) => {
+                bail!("error in reading symlink at {:?} : {:?}", test_file, e);
+            }
+        };
+
+        match fs::metadata(r_path) {
```
YJDoc2, line 151: same as one before.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,251 @@
...
+    if let TestResult::Passed = res {
+        TestResult::Failed(anyhow!(
+            "expected error in container creation with invalid symlink, found no error"
+        ))
+    } else {
+        TestResult::Passed
+    }
+}
+
+fn test_node(mode: u32) -> TestResult {
```
YJDoc2, line 179:
```suggestion
fn test_mode(mode: u32) -> TestResult {
```
typo?

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,251 @@
...
+        SFlag::S_IFBLK.bits() | 0o666,
+        SFlag::S_IFCHR.bits() | 0o666,
+        SFlag::S_IFIFO.bits() | 0o666,
+    ];
+    for mode in modes {
+        let res = test_node(mode);
+        if let TestResult::Failed(_) = res {
+            return res;
+        }
+        std::thread::sleep(std::time::Duration::from_millis(1000));
```
YJDoc2, line 232: Are we sleeping to make sure the container exists? or some other reason? Please add a comment in code.

tests/contest/runtimetest/src/utils.rs
```diff
@@ -1,22 +1,52 @@
...
+                        Ok(_) => Ok(true),   // If the entry is Ok, then it's readable
+                        Err(_) => Ok(false), // If the entry is Err, then it's not readable
+                    }
+                }
+                None => Ok(false), // If there are no entries, then it's not readable
+            }
+        }
+        Err(_) => {
+            Ok(false) // If there's an error, then it's not readable
+        }
```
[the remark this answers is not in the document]
nayuta-ai: Yes, it is possible. Basically, it is the case that there is no file in the path, but there are also temporary errors, such as network issues.

tests/contest/runtimetest/src/utils.rs
```diff
@@ -1,22 +1,52 @@
...
+    // Create a buffer with a capacity of 1 byte
+    let mut buffer = [0u8; 1];
+    match file.read(&mut buffer) {
+        Ok(_) => {
+            let mut contents = String::new();
+            file.read_to_string(&mut contents)?;
+            if contents.is_empty() {
+                // empty file
+                return Ok(false);
+            }
```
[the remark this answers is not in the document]
nayuta-ai: In a masked path, references to file names are possible, but the contents of the file cannot be accessed. Therefore, in the read access test, we check whether the file is empty. According to what was mentioned, the test for masked paths can determine if the file is empty by checking if the number of bytes read with read is 0, which indicates EOF.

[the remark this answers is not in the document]
nayuta-ai: I think it would be good to refactor it as shown below, so I would like to modify this implementation approach and add comments. Is this method sufficient?
```suggestion
fn test_file_read_access(path: &str) -> Result<bool, std::io::Error> {
    let mut file = OpenOptions::new().create(false).read(true).open(path)?;

    // Create a buffer with a capacity of 1 byte
    let mut buffer = [0u8; 1];
    match file.read(&mut buffer) {
        Ok(0) => Ok(false),
        Ok(_) => Ok(true),
        Err(e) => Err(e),
    }
}
```

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: When setting a masked path, it is impossible to determine whether it is a relative path, so I added this to verify whether it meets the test requirements at the time of creation. This implementation is also adopted in runtime-tools, so verification is necessary. However, I overlooked a few points regarding file creation, so I will make corrections.

tests/contest/runtimetest/src/utils.rs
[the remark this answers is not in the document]
YJDoc2: ok. How frequently have you run into these other errors if at all? If the frequency of them is small, then this is ok. If that is the case, please reply and mark this comment as resolved, as no change is needed.

tests/contest/runtimetest/src/utils.rs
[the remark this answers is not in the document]
YJDoc2: ok, so going in original test, [this comment](https://github.com/opencontainers/runtime-tools/blob/master/cmd/runtimetest/main.go#L464-L468) explains that if we get eof (or in our case a 0 bytes read) that indicates blocked read, _because_ the tests use files with some content. Please add a similar comment.

Also in your refactored approach, you need to account for EOF error like the current code has. Other than that the refactored code has much better readability.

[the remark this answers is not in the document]
YJDoc2: Also see [this](https://github.com/opencontainers/runtime-tools/blob/master/validation/linux_masked_paths/linux_masked_paths.go#L49-L51)

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
YJDoc2: ok, so I see that you have carried over the implementation from go tests, but I'm still having a bit of difficulty understanding this -
1. The go code seems to checks  that if error is not null and error is not found, then return error, else return null. I.e. if the error is not found, then the prepare function will return error, other wise not. Here the behavior seems to be opposite.
2. Also, this function/closure we pass, both here and in go code, will be executed before runtime has done anything. This is supposed to prepare the rootfs for tests running inside the container. Given that, what exactly is this checking? This will be running on plain bundle fs that is created for each test, It is ofcourse not going to find any of the files. 

What I'm thinking here is that even the go tests have slightly incorrect implementation (which has happened before). Can you confirm going through comments in go code and seeing what the test is supposed to check, and what it is actually validating?

YJDoc2:
@nayuta-ai , I have replied to comments, also the PR has conflicts with main, Please make the suggested changes and resolve conflicts. Thanks.

tests/contest/runtimetest/src/utils.rs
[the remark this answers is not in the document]
nayuta-ai: Apologies for the additional comment. I revised this to return an error and expect only the ENOENT case, which I believe addresses the concern.
[48739f7](https://github.com/youki-dev/youki/pull/2950/commits/48739f7323eddca02fc9b0443f91d2b4cb553da7)

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: There are cases where the Go implementation may be incorrect. I'll check it out.

tests/contest/runtimetest/src/utils.rs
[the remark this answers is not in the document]
nayuta-ai: [8ec531f](https://github.com/youki-dev/youki/pull/2950/commits/8ec531f6465408df0883f4074e2fa37132cf255d)

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: It is unnecessary for this test, so I have deleted it.
[8ec531f](https://github.com/youki-dev/youki/pull/2950/commits/8ec531f6465408df0883f4074e2fa37132cf255d)

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: I posted a question in the Issue because I couldn’t fully grasp the purpose of this test.（https://github.com/opencontainers/runtime-tools/issues/780 ）

This is my own opinion, but I believe the test might be trying to illustrate a discrepancy where the path added to `masked_path` is specified from the host system’s perspective, resulting in an unintended relative path from `/` inside the container that doesn’t match the expected path.

As for addressing this test, we could:
1. Add the above comment
2. Wait for a response on the Issue before deciding
3. Delete the test
4. Refactor the test

If we choose to improve the test, I think the following approach could work:
- Create a file inside the container we want to mask (`masked-relpath`).
- Specify the relative path from the host system (`./bundle/rootfs/masked-relpath`) as the path to mask.
- After creating the container, verify that the file is masked by checking if it exists at the expected path without the `./bundle/rootfs` prefix.

Do you think this would be a good approach?

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: In this case, we are creating a symbolic link to a non-existent path and verifying that the symlink cannot be read as expected. If the relative path is not set correctly within the container's root filesystem, resolving the link will be problematic, leading to an error. This test aims to detect that scenario.

The validation for `fs::metadata` will be similar to the following comment.
https://github.com/youki-dev/youki/pull/2950#discussion_r1824747287

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: As stated in the [runtime-spec](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#masked-paths), setting a relative path is not allowed. Therefore, I believe the correct responsibility of contest is to produce an error when a relative_path is set during testing. Based on this, I implemented the following.
d1fa11b

Regarding `runtimetest`, I made it verify that paths passed to the namespace are not relative path to ensure the behavior complies with the [runtime-spec](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#masked-paths).

One concern is whether to add an implementation to validate relative_path in the spec's validation. This would also affect the test behavior, so I am leaving this comment for discussion.
https://github.com/youki-dev/youki/blob/main/crates/libcontainer/src/container/init_builder.rs#L152

BTW, in `runc`, validation to ensure `masked_path` is an absolute path is not performed, so it might not be necessary.

nayuta-ai:
@YJDoc2 
Thank you for the review! Please review my code again.

YJDoc2:
Hey, can I ask you to rebase/merge latest main? There is a small breaking change in `test_inside_container` and the functions `test_read/write_access` are changed slightly as well. I'll do a review after that. Thanks!

nayuta-ai:
@YJDoc2 
I've rebased it. Please review!

utam0k:
@nayuta-ai I'd like to review this PR. Would you mind fixing the conflict before I go on reviewing?

nayuta-ai:
@utam0k 
I'm sorry for not noticing this late. I've resolved the conflicts, so please review it.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,225 @@
...
+                .unwrap(),
+        )
+        .build()
+        .unwrap()
+}
+
+fn check_masked_paths() -> TestResult {
+    let masked_dir = "masked-dir";
+    let masked_subdir = "masked-subdir";
+    let masked_file = "masked-file";
```
utam0k, line 32: Please make them the const values.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,225 @@
...
+
+use anyhow::{anyhow, bail};
+use nix::sys::stat::SFlag;
+use oci_spec::runtime::{LinuxBuilder, ProcessBuilder, Spec, SpecBuilder};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::utils::test_inside_container;
+use crate::utils::test_utils::CreateOptions;
+
+fn get_spec(masked_paths: Vec<String>) -> Spec {
```
utam0k, line 11:
```suggestion
fn get_spec(masked_paths: Vec<PathBuf>) -> Spec {
    let paths: Vec<String> = masked_paths
        .iter()
        .map(|p| p.to_string_lossy().to_string())
        .collect();
```

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,225 @@
...
+    let spec = get_spec(masked_paths);
+
+    test_inside_container(spec, &CreateOptions::default(), &|bundle_path| {
+        use std::os::unix::fs::OpenOptionsExt;
+        use std::{fs, io};
+        let test_file = bundle_path.join(masked_device);
+
+        let mut opts = fs::OpenOptions::new();
+        opts.mode(mode);
+        opts.create(true);
```
utam0k, line 167: I think `opts` isn't used, right?

tests/contest/runtimetest/src/main.rs
```diff
@@ -33,6 +33,7 @@ fn main() {
         "hello_world" => tests::hello_world(&spec),
         ////////// ANCHOR_END: example_runtimetest_main
         "readonly_paths" => tests::validate_readonly_paths(&spec),
+        "masked_paths" => tests::validate_masked_paths(&spec),
```
utam0k, line 36: BTW, I realized `validate_masked_paths` should always be called, even if the subcommand isn't passed.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -820,3 +820,46 @@ pub fn validate_fd_control(_spec: &Spec) {
...
+        None => {
+            eprintln!("in masked paths, expected some masked paths to be set, found none");
+            return;
+        }
+    };
+
+    if masked_paths.is_empty() {
+        eprintln!("in masked paths, expected some masked paths to be set, found none");
+        return;
+    }
```
utam0k, line 837: We don't need it.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
utam0k: Others are also the same.

tests/contest/runtimetest/src/utils.rs
```diff
@@ -1,22 +1,46 @@
...
 
 use nix::sys::stat::{stat, SFlag};
 
-fn test_file_read_access(path: &str) -> Result<(), std::io::Error> {
-    let _ = std::fs::OpenOptions::new()
-        .create(false)
-        .read(true)
-        .open(path)?;
-    Ok(())
+fn test_file_read_access(path: &str) -> Result<bool, std::io::Error> {
```
utam0k, line 10: May I ask you to make the `bool` meaning clear?

nayuta-ai: There was an implementation mistake. The following code seems unnecessary:

```rust
Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => Ok(false),
```

I have addressed this in d120803. Is everything clear now? If additional docstrings are needed, please let me know.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: You are correct. It seems that `opts` is not being used in the code. Thank you for pointing that out!

tests/contest/runtimetest/src/main.rs
nayuta-ai: It does seem necessary to call it every time. As for how to fix it, I believe the approach would be as follows, but I'm not entirely sure about the correct way to handle it, so I have a question.  

The idea is to remove the processing from the `subcommand` call and instead add an implementation that explicitly calls `subcommand`. However, this would likely result in the removal of the `masked_paths` test group. I'm unclear on where the test cases that were originally part of the `masked_paths` test group should be moved.

utam0k: You don't need to take care of it in this PR ;)

utam0k: If you are interested in it, please do it in another PR.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
```diff
@@ -0,0 +1,227 @@
...
+        SFlag::S_IFCHR.bits() | 0o666,
+        SFlag::S_IFIFO.bits() | 0o666,
+    ];
+    for mode in modes {
+        let res = test_mode(mode);
+        if let TestResult::Failed(_) = res {
+            return res;
+        }
+    }
+    TestResult::Passed
```
utam0k, line 210:
```suggestion
    [
        SFlag::S_IFBLK.bits() | 0o666,
        SFlag::S_IFCHR.bits() | 0o666,
        SFlag::S_IFIFO.bits() | 0o666,
    ]
    .iter()
    .map(|mode| test_mode(*mode))
    .find(|res| {
        if let TestResult::Failed(_) = res {
            true
        } else {
            false
        }
    })
    .unwrap_or(TestResult::Passed)
```

Ideally, we should return the failure result of all modes.

tests/contest/runtimetest/src/utils.rs
utam0k: Please add a comment about the return value or use a type alias.

tests/contest/contest/src/tests/linux_masked_paths/masked_paths.rs
nayuta-ai: I fixed it in commit [5d79a48](https://github.com/youki-dev/youki/pull/2950/commits/5d79a486a3b67a977f1c789723424910bbd0b6ee).

tests/contest/runtimetest/src/utils.rs
nayuta-ai: I fixed it in commit [b4e3511](https://github.com/youki-dev/youki/pull/2950/commits/b4e3511b883426fe60675d68203552d69048fdee).

tests/contest/runtimetest/src/tests.rs
```diff
@@ -820,3 +820,41 @@ pub fn validate_fd_control(_spec: &Spec) {
...
+    let linux = spec.linux().as_ref().unwrap();
+    let masked_paths = match linux.masked_paths() {
+        Some(p) => p,
+        None => {
+            eprintln!("in masked paths, expected some masked paths to be set, found none");
+            return;
+        }
+    };
+
+    for path_str in masked_paths {
```
utam0k, line 834: How about  handling `path_str` as Path, not &str?
```suggestion
    for path in masked_paths.iter().map(Path::new) {
```

Following the above change, we'll amend some functions, like:
```
pub fn test_read_access<P: AsRef<Path>>(path: P) -> Result<Readable, std::io::Error> {
```

nayuta-ai: Most of the function arguments in `tests/contest/runtimetest/src/utils.rs`, including `test_read_access`, are handled as `&str`. Should they still be modified? If so, should the change be applied to all of them?

utam0k: You don't need to take care of other functions in this PR. I'm glad to address them in another PR if possible 🙏

nayuta-ai:
@utam0k 
I have tested this multiple times locally, and the tests pass on GNU/Linux. Would it be possible to rerun the CI?

YJDoc2:
I reran the CI, and it passed. I'll try to take a look at it, as I missed some discussion in between, and if all seems good, I'll merge it, otherwise utam0k might get to it before I do.

utam0k (approved):
LGTM from my eyes.

utam0k (approved):
LGTM from my eyes.

utam0k (approved):
LGTM from my eyes.

tests/contest/runtimetest/src/utils.rs
```diff
@@ -1,22 +1,48 @@
...
 use nix::sys::stat::{stat, SFlag};
 
-fn test_file_read_access(path: &str) -> Result<(), std::io::Error> {
-    let _ = std::fs::OpenOptions::new()
-        .create(false)
-        .read(true)
-        .open(path)?;
-    Ok(())
+// It means the file or directory is readable
+type Readable = bool;
```
YJDoc2, line 11: Why did we newtype this? Not opposed to do so, but because we are not implementing any new methods on this, and the actual values we return are just bool values, is there any special advantage for making a newtype here instead of just using plain bool?

YJDoc2 (approved):
LGTM to me as well.
Have a basic question, but non-blocking, so will go ahead and merge.

tests/contest/runtimetest/src/utils.rs
nayuta-ai: Sorry for the late reply.We were discussing the following link:https://github.com/youki-dev/youki/pull/2950#discussion_r1932074342
The background of this change is that the bool value was unclear in its meaning, so we addressed it in this way.

PR #2923, opened 2024-09-21
fix(libcontainer) no_pivot args is not used

xujihui1985:
fix the problem that no_pivot args is not used when create container when we prepare rootfs with chroot, we should move_mount the rootfs before chroot, otherwise we will not able to use the new rootfs when exec into the container

PR #2688, opened 2024-02-18
linux_mount_label integration test

Gekko0114:
Created linux_mount_label integration test.

Referred this linux_mount_label test.
https://github.com/opencontainers/runtime-tools/tree/master/validation/linux_mount_label

#361

PR #2953, opened 2024-10-13
create mount_rootfs method

Gekko0114:
The implementation of mounting differs between Youki and runc.
https://github.com/containers/youki/issues/2718

Therefore, we want to modify Youki to adopt runc's mounting logic.
As the first step, I've created a new method `mount_rootfs`.
I will implement the mounting logic step by step in this method.

### DIFF
--- a/crates/libcontainer/src/rootfs/rootfs.rs
+++ b/crates/libcontainer/src/rootfs/rootfs.rs
@@ -31,17 +31,14 @@ impl RootFS {
         }
     }
 
-    pub fn prepare_rootfs(
+    pub fn mount_to_rootfs(
         &self,
+        linux: &Linux,
         spec: &Spec,
         rootfs: &Path,
-        bind_devices: bool,
         cgroup_ns: bool,
     ) -> Result<()> {
-        tracing::debug!(?rootfs, "prepare rootfs");
         let mut flags = MsFlags::MS_REC;
-        let linux = spec.linux().as_ref().ok_or(MissingSpecError::Linux)?;
-
         match linux.rootfs_propagation().as_deref() {
             Some("shared") => flags |= MsFlags::MS_SHARED,
             Some("private") => flags |= MsFlags::MS_PRIVATE,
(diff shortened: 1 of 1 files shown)


Gekko0114:
Hi @utam0k, @YJDoc2,

I've submitted a small PR as the first step towards integrating SELinux into Youki's mounting logic.
Could you take a look?
Also I'd like to know which commands should be run to verify that the code changes are safe.

YJDoc2:
@Gekko0114 , Thanks for implementing this, but may I ask you to wait for/coordinate with #2923 as that also changes the mount impl.

Gekko0114:
Sure, then I will wait this PR will be merged.

YJDoc2:
@Gekko0114 the other PR is merged, you can rebase on main and continue your work here.

Gekko0114:
Hi @YJDoc2,
Could you review it since I've rebased this PR ?
I would like to merge this PR first, and after that I will create other PRs using SELinux like runc.

YJDoc2:
Hey, it looks ok overall, but can you give a bit more reasoning behind this change? I mean, I understand that this is done to make the logic same in youki and runc, but why is that needed, and what this split of function will achieve (eventually) ?

Gekko0114:
Hi @YJDoc2 

In Runc, processing differs for each device of the mount point. https://github.com/opencontainers/runc/blob/d5e4c33001d74176222fe8f48a323f3e8ad89999/libcontainer/rootfs_linux.go#L485
However, Youki doesn't currently handle this like Runc. As a result, Youki fails the OCI tests. https://github.com/youki-dev/youki/pull/2688

Therefore, I plan to address this issue. First I created a mount_to_rootfs function, and then I will add processing for each device of the mount point.

YJDoc2:
ok, Ideally we should try to do it our own way in a single function if possible, but I see that the individual mount fn in runc is quite complex, so maybe we should break our function in two as well. In the following PRs, please check if we can do it differently or more rust-idiomatically. Merging this for now.

YJDoc2 (approved):

PR #2956, opened 2024-10-18
Add unittest to expertiment seccomp programs

sat0ken:
I've added the unit test code to the PR below.

https://github.com/containers/youki/pull/2946


utam0k (approved):
Thanks!

PR #2976, opened 2024-11-04
Add test root readonly

sat0ken:
This implements the root_readonly_true validation in https://github.com/youki-dev/youki/issues/361


tests/contest/runtimetest/src/tests.rs
```diff
@@ -546,6 +546,19 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+        let errno = Errno::from_raw(e.raw_os_error().unwrap());
+        if errno == Errno::ENOENT {
+            /* This is expected */
+        } else {
+            eprintln!("in readonly paths, error in testing read access for / : {e:?}");
+        }
+    } else {
+        /* Expected */
+    }
+}
```
YJDoc2, line 560: I think this logic is wrong - we need to check if write access is false, not read-access is giving any error or not. Also if root is readonly , why would the read access give any error at all ? It should succeed, even ENOENT is not expected.

tests/contest/runtimetest/src/main.rs
```diff
@@ -44,7 +44,9 @@ fn main() {
         "io_priority_class_be" => tests::test_io_priority_class(&spec, IoprioClassBe),
         "io_priority_class_idle" => tests::test_io_priority_class(&spec, IoprioClassIdle),
         "devices" => tests::validate_devices(&spec),
+        "root_readonly" => tests::test_validate_root_readonly(),
         "no_pivot" => tests::validate_rootfs(),
+
```
YJDoc2, line 49: Remove this extra blank

YJDoc2 (changes requested):
Hey, I think we need to break this in two tests - one with readonly true and other with false. We will have a corresponding test in the `runtime` test, which will check that read operation is successful (maybe create a file and validate the contents) and the the write access will be tested according to config's value.

This is small deviation from the go test, but I think this would be a better test.

sat0ken:
@YJDoc2 HI, I've fixed the code according to the comments. Could you please check the code?

tests/contest/runtimetest/src/main.rs
YJDoc2: @sat0ken This is still to be done

tests/contest/contest/src/tests/root_readonly_true/root_readonly_tests.rs
```diff
@@ -0,0 +1,36 @@
...
+        .context("failed to build spec")?;
+
+    Ok(spec)
+}
+
+fn root_readonly_test() -> TestResult {
+    let spec_true = test_result!(create_spec(true));
+    let spec_false = test_result!(create_spec(false));
+    test_inside_container(spec_true, &|_| Ok(()));
+    test_inside_container(spec_false, &|_| Ok(()))
```
YJDoc2, line 26: I think we should split this into two tests, maybe `root_readonly_true_test` and `root_readonly_false_test`. Current impl is correct, but I feel it is logically better to have two tests to check two behaviors rather than one test checking both.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -546,6 +546,27 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+    let root = spec.root().as_ref().unwrap();
+    if root.readonly().unwrap() {
+        if let Err(e) = test_dir_write_access("/") {
+            let errno = Errno::from_raw(e.raw_os_error().unwrap());
+            if errno == Errno::EROFS {
+                /* This is expected */
+            } else {
+                eprintln!("readonly root filesystem, error in testing write access for path /");
+            }
+        }
```
YJDoc2, line 559: This is ok, but also check for read access is `Ok` and not err _if possible_. For the `!= EROFS` case, also add the actual `errno` in the `eprintl` for debugging

tests/contest/runtimetest/src/tests.rs
```diff
@@ -546,6 +546,27 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+    } else if let Err(e) = test_dir_write_access("/") {
+        let errno = Errno::from_raw(e.raw_os_error().unwrap());
+        if errno == Errno::EROFS {
+            eprintln!("readt only root filesystem is false but write access for path / is err");
+        } else {
+            /* This is expected */
+        }
```
YJDoc2, line 566: I think for `readonly = false`, write access should not give any errors at all. So if it not ok, that should be considered error irrespective of errno.

YJDoc2 (changes requested):
Hey, overall ok, but some changes needed.

tests/contest/runtimetest/src/main.rs
sat0ken: Sorry, I fixed.
https://github.com/youki-dev/youki/pull/2976/commits/a49ca55f3b66d513820022b21b44b20371188ed7

tests/contest/contest/src/tests/root_readonly_true/root_readonly_tests.rs
sat0ken: @YJDoc2 
I split into tow test. Is this correct?
https://github.com/youki-dev/youki/pull/2976/commits/81d6128c54e1e8cf9cf92b182c6298be25fa6010

tests/contest/runtimetest/src/tests.rs
sat0ken: I fixed this.
https://github.com/youki-dev/youki/pull/2976/commits/c3c62fbf865aeaedfb6fdce1b33aba31301988ca

tests/contest/runtimetest/src/tests.rs
sat0ken: I fixed this.
https://github.com/youki-dev/youki/pull/2976/commits/c3c62fbf865aeaedfb6fdce1b33aba31301988ca

sat0ken:
@YJDoc2 Thank you for review and comment! I fixed code, Could you please check the code?

sat0ken:
@YJDoc2 Ping!

tests/contest/runtimetest/src/tests.rs
```diff
@@ -550,6 +552,44 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+            let errno = Errno::from_raw(e.raw_os_error().unwrap());
+            if errno == Errno::EROFS {
+                /* This is expected */
+            } else {
+                eprintln!(
+                    "readonly root filesystem, error in testing read access for path /, error: {}",
+                    errno
+                );
+            }
+        }
```
YJDoc2, line 579: Hey, for readonly=true, read_access should not give any error at all. So any error should be considered test fail. I think you might have incorrectly copy-pasted from the write access case. Please update.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -550,6 +552,44 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+        if e.raw_os_error().is_some() {
+            let errno = Errno::from_raw(e.raw_os_error().unwrap());
+            eprintln!(
+                "readt only root filesystem is false but write access for path / is err, error: {}",
+                errno
+            );
+        } else {
+            /* This is expected */
+        }
+    }
```
YJDoc2, line 590: Why we are checking `raw_os_error().is_some()` ? If this was error, any error while writing, it should be considered as a test fail, right? Also can we have a read_access check in this branch as well? Basically :

if readonly=true{
   check write_access == error;
   check read_access == ok;

}else{
   check write_access == ok;
   check read_access == ok;
}

YJDoc2 (changes requested):
some minor changes needed.

sat0ken:
@YJDoc2 Thank you for review and comments. I've fixed the code. Could you please check the code?

tests/contest/runtimetest/src/tests.rs
```diff
@@ -550,6 +552,51 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+        if let Err(e) = test_dir_read_access("/") {
+            if let Some(errno_code) = e.raw_os_error() {
+                let errno = Errno::from_raw(errno_code);
+                eprintln!(
+                    "readonly root filesystem, but error in testing read access for path /, error: {}",
+                    errno
+                );
+            }
+        }
+    } else {
```
YJDoc2, line 578:
```suggestion
        if let Err(e) = test_dir_read_access("/") {
            eprintln!(
                "readonly root filesystem, but error in testing read access for path /, error: {}",
                e
            );
    } else {
```
As I had mentioned in my previous comment,  we don't need specific check for raw_os_error. As long as this is error, the test is failing. Similar for the  two cases below this. Also I think here, you are converting the error to raw os error, and then using errorno to convert it back into the original error.

YJDoc2 (changes requested):
One final change.

tests/contest/runtimetest/src/tests.rs
sat0ken: Sorry for any inconvenience caused by my misunderstanding.
I fixed this. Is this ok?

https://github.com/youki-dev/youki/pull/2976/commits/736127756158af9dd9328f8bb77cb08975f78858

YJDoc2: Yep, this is what I wanted :+1:

YJDoc2:
@sat0ken , can you merge/rebase the latest main? the `test_inside_container` function has changed, so this will error if merged in current state. Apart from that this is good to go, so once that is done I'll merge.

Thanks for all the efforts, and apologies for so many back-and-forths :)

sat0ken:
@YJDoc2 

> can you merge/rebase the latest main?
Sure ! I worked it. Is this ok?

https://github.com/youki-dev/youki/pull/2976/commits/397a8c5532f083d5dacf40a8a1c3b29fcfd62231

YJDoc2 (approved):
LGTM!
Thanks for your efforts, and apologies for so many back-and-forth :)

PR #2996, opened 2024-11-17
add kill test

YamasouA:
part of #361 
I implement kill test case in [runtime-tools](https://github.com/opencontainers/runtime-tools) to youki.

### DIFF
--- a/tests/contest/contest/src/tests/kill/kill_test.rs
+++ b/tests/contest/contest/src/tests/kill/kill_test.rs
@@ -1,0 +1,163 @@
+use std::time::Duration;
+
+use anyhow::{anyhow, Context, Result};
+use oci_spec::runtime::{ProcessBuilder, Spec, SpecBuilder};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::tests::lifecycle::ContainerLifecycle;
+
+fn create_spec(args: &[&str]) -> Result<Spec> {
+    let args_vec: Vec<String> = args.iter().map(|&a| a.into()).collect();
+    let spec = SpecBuilder::default()
+        .process(
+            ProcessBuilder::default()
+                .args(args_vec)
+                .build()
+                .context("failed to build process spec")?,
+        )
+        .build()
+        .context("failed to build spec")?;
+    Ok(spec)
+}
+
+fn failed_and_delete(text: String, container: ContainerLifecycle) -> TestResult {
+    let delete_result = container.delete();
+    match delete_result {
+        TestResult::Passed => TestResult::Failed(anyhow!(text)),
+        TestResult::Failed(err) => TestResult::Failed(anyhow!(
+            "{}; also container deletion failed: {:?}",
+            text,
+            err
+        )),
+        _ => TestResult::Failed(anyhow!("{}; unexpected delete result", text)),
+    }
+}
+
+fn merge_test_results(kill_result: TestResult, delete_result: TestResult) -> TestResult {
+    match (kill_result, delete_result) {
+        (TestResult::Failed(err), _) => TestResult::Failed(err),
+        (TestResult::Passed, TestResult::Failed(err)) => {
+            TestResult::Failed(anyhow!("Delete failed: {:?}", err))
+        }
+        (TestResult::Passed, TestResult::Passed) => TestResult::Passed,
+        _ => TestResult::Failed(anyhow!("Unexpected result")),
+    }
+}
+
+fn kill_with_empty_id_test() -> TestResult {
+    let mut container = ContainerLifecycle::new();
+
+    // kill with empty id
+    container.set_id("");
+    match container.kill() {
+        TestResult::Failed(_) => TestResult::Passed,
+        TestResult::Passed => TestResult::Failed(anyhow!(
+            "Expected killing container with empty id to fail, but was successful"
+        )),
+        _ => TestResult::Failed(anyhow!(
+            "Unexpected killing container with empty id test result"
+        )),
+    }
+}
+
+fn kill_non_existed_container() -> TestResult {
+    let container = ContainerLifecycle::new();
+
+    // kill for non existed container
+    match container.kill() {
+        TestResult::Failed(_) => TestResult::Passed,
+        TestResult::Passed => TestResult::Failed(anyhow!(
+            "Expected killing non existed container to fail, but was successful"
+        )),
+        _ => TestResult::Failed(anyhow!(
+            "Unexpected killing non existed container test result"
+        )),
+    }
+}
+
+fn kill_created_container_test() -> TestResult {
+    let container = ContainerLifecycle::new();
+
+    // kill created container
+    match container.create() {
+        TestResult::Passed => {}
+        _ => return failed_and_delete("Failed to create container".to_string(), container),
+    }
+    let kill_result = container.kill();
+    let delete_result = container.delete();
+    merge_test_results(kill_result, delete_result)
+}
+
+fn kill_stopped_container_test() -> TestResult {
+    let container = ContainerLifecycle::new();
+    let spec = create_spec(&["true"]).unwrap();
+
+    // kill stopped container
+    match container.create_with_spec(spec) {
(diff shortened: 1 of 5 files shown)


YJDoc2:
@YamasouA  please sign your commits. The instructions can be found at https://github.com/youki-dev/youki/pull/2996/checks?check_run_id=33094881373

YamasouA:
@YJDoc2 
Thank you for your review!
I pass DCO.

tests/contest/contest/src/tests/kill/kill_test.rs
```diff
@@ -0,0 +1,111 @@
...
+    TestResult::Passed
+}
+
+pub fn get_kill_test() -> TestGroup {
+    let mut test_group = TestGroup::new("kill_container");
+    let kill_test = ConditionalTest::new(
+        "test_kill_container",
+        Box::new(|| true),
+        Box::new(run_kill_test_cases),
+    );
```
YJDoc2, line 108: You can use normal test instead of ConditionalTest here, that will not require a conditional function.

YJDoc2:
Hey @YamasouA , thanks for the PR. I have taken a basic look, and want to request a change -
The go tests usually put all test cases in a single function, but here we try to separate each test case as a different test, so it is easier to understand and maintain. see the code in https://github.com/youki-dev/youki/pull/2976 , how it has created one test function for readonly=true and another for readonly=false.

Can I request you to change this code to split each test condition into its own testcase? So it will be one function for testing kill without id, one for kill non-existing container and so on.

YamasouA:
@YJDoc2 
I fix your comment! 
please review again!!

tests/contest/contest/src/tests/kill/kill_test.rs
```diff
@@ -0,0 +1,131 @@
...
+        TestResult::Passed => {}
+        _ => return TestResult::Failed(anyhow!("Failed to create container")),
+    }
+    let result = match container.kill() {
+        TestResult::Passed => TestResult::Passed,
+        TestResult::Failed(_) => {
+            TestResult::Failed(anyhow!("Expected success but got failure"))
+        }
+        _ => TestResult::Failed(anyhow!("Unexpected test result")),
+    };
```
YJDoc2, line 47: I think in both these cases you can use `test_result!` macro to directly return the error without having to do manual match. Same for the cases below.

YJDoc2 (changes requested):
overall looks ok, two things I'd request you to fix, Then I'll do the final review.

YamasouA:
@YJDoc2 
Thank you for your review.
I fix your comment and fix lint error.

tests/contest/contest/src/tests/kill/kill_test.rs
YamasouA: container.kill() return TestResult. type.
So I think we don't have to use test_result macro.
Is it true?

YJDoc2 (commented):
Hey, so few things need change -
1. I have left a couple of comments, please take a look.
2. In the original go tests, it uses special config for stopped container and running container :
    - It uses a config with command `true` which would immediately exit, and thus we get a stopped container
    - It uses a config with command `sleep 30` and waits 10 seconds so to make sure container is running and then attempts to kill the container for running container
In our current lifecycle impl, we do not have support for passing custom command, we always use sleep 10. So to get the correct behavior we can either modify the lifecycle impl (maybe with a new fn) to allow customizing the command, or for stopped container test, we sleep 15 seconds, then kill and for running test we sleep ~3 seconds and then kill.

Can you check what would be better and update the test accordingly? Thanks :)

YamasouA:
@YJDoc2 
I wrote some util function to support this test like your advice, and update test.

YamasouA:
@YJDoc2 
I understand your situation!
No problem.
Thanks for the reply!

utam0k:
@YamasouA I've checked your PR this weekend.

tests/contest/contest/src/tests/kill/kill_test.rs
```diff
@@ -0,0 +1,146 @@
+use std::time::Duration;
+
+use anyhow::{anyhow, Context, Result};
+use oci_spec::runtime::{ProcessBuilder, Spec, SpecBuilder};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::tests::lifecycle::ContainerLifecycle;
+
+fn create_spec(args: &[&str]) -> Result<Spec> {
+    let args_vec: Vec<String> = args.iter().map(|&a| a.to_string()).collect();
```
utam0k, line 10:
```suggestion
    let args_vec: Vec<String> = args.iter().map(|&a| a.into()).collect();
```

tests/contest/contest/src/tests/kill/kill_test.rs
```diff
@@ -0,0 +1,146 @@
...
+    let result = match container.kill() {
+        TestResult::Failed(_) => TestResult::Passed,
+        TestResult::Passed => TestResult::Failed(anyhow!(
+            "Expected killing non existed container to fail, but was successful"
+        )),
+        _ => TestResult::Failed(anyhow!(
+            "Unexpected killing non existed container test result"
+        )),
+    };
+    container.delete();
```
utam0k, line 55: I guess we should delete the container even if the test failed to clean up.

utam0k: Also, please take care of the return value of `delete()`

utam0k (changes requested):

YamasouA:
@utam0k 
Thank you for your review!
I fix your review point.
I take a mistake to rebase and commit log is dirty.
Should I create new PR?

utam0k:
> Should I create new PR?

No needed. How about re-checking the branch from the latest main branch and applying the diff? Of course, I don't mind using force-push.
https://patch-diff.githubusercontent.com/raw/youki-dev/youki/pull/2996.diff

YamasouA:
@utam0k 
Super thank you!!
I didn't know apply patch command.
It's very useful knowledge!!

YamasouA:
@utam0k 
The review points were corrected in this commit. [78df79a](https://github.com/youki-dev/youki/pull/2996/commits/78df79a594a475728eb6945351dcbb7a99a65785)
Sorry for not separating the commits.
Could you please review it again?
And how to rerun tests?

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
```diff
@@ -86,6 +97,31 @@ impl ContainerLifecycle {
             &self.container_id,
         )
     }
+
+    pub fn waiting_for_status(
```
utam0k, line 101: This function is duplicated with https://github.com/youki-dev/youki/pull/3082/files#diff-7c81e28b8064c3c2c13f68c038ca10661530915ffd2159b939e1dba7684b40e8

utam0k: The PR has been merged.

YJDoc2:
@YamasouA , are you still working on this? If it is not possible for you , that is ok, just let us know.
If you are still working on this, there are couple of comments and conflicts that needs to be fixed , may I ask you to take a look at them? Thanks :)

YJDoc2:
Hey, thanks for the update :)
I think you need to either rebase or merge with main to fix the failing CIs, can you also take a look at that? Thanks :)

YamasouA:
@YJDoc2 
Sorry, I overlooked a few corrections.

YJDoc2 (approved):
Hey, LGTM! Thanks for the PR. One last thing - Can I ask you to add comments to the test functions and explanation for what is the expected behavior for those tests? That will  help when we work on these tests in future. Not a blocking though, so if you can't get to it in few days, I'll merge this as-is.

Thanks again :)

YamasouA:
@YJDoc2 
I add some comment.
Please check!!

YJDoc2:
@utam0k I think your change request has been satisfied, so I'll go ahead and merge.

PR #3018, opened 2024-12-10
Fix : fix gid duplication, additional_gids etc not getting used in exec

YJDoc2:
This PR does following fixes:

1. Fix gid duplication in additional_gids : if we provide duplicate gid, runc simply ignore it, whereas we used to retain duplicates and create two distinct groups with same gid. Now we filter the duplicates and only set unique gids
2. Fix additional_gid, user, group not getting set in exec : exec command accepts these as args, but we did not use them in tenant container. Now we use them as well.
3. Inject `PATH` from original container in case not provided to exec : exec allows supplying env vars, but by default it does not set any. This causes issue when running exec directly via youki, as it would error with `PATH is not set` . Given that we are exec-ing into an existing container, the expectation would be the same env would be applied, or atleast the path would be taken from it. I guess any higher runtime (docker/podman) automatically sets envs from original container, but I have injected path if missing.

I want to also add tests for these, but wanted to get a review on the changes before that.

### DIFF
--- a/crates/libcontainer/src/container/tenant_builder.rs
+++ b/crates/libcontainer/src/container/tenant_builder.rs
@@ -14,7 +14,7 @@ use nix::unistd::{pipe2, read, Pid};
 use oci_spec::runtime::{
     Capabilities as SpecCapabilities, Capability as SpecCapability, LinuxBuilder,
     LinuxCapabilities, LinuxCapabilitiesBuilder, LinuxNamespace, LinuxNamespaceBuilder,
-    LinuxNamespaceType, LinuxSchedulerPolicy, Process, ProcessBuilder, Spec,
+    LinuxNamespaceType, LinuxSchedulerPolicy, Process, ProcessBuilder, Spec, UserBuilder,
 };
 use procfs::process::Namespace;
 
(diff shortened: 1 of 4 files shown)


crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -397,8 +451,28 @@ impl TenantContainerBuilder {
...
+        // It is not possible in normal flow that path is None. The original container
+        // creation would have failed if path was absent. However we use Option
+        // just as a caution, and if neither exec cmd not original spec has PATH,
+        // the container creation will fail later which is ok
+        if let Some(p) = path {
+            if !env_exists {
+                env.push(p);
+            }
+        }
+        env
```
YJDoc2, line 475: 1. If there is a more semantic way, let me know, these multiple checks are not particularly idiomatic.
2. If there are no envs at all (self.env is empty) should we simply clone from the original container? I feel that is what the user would expect.

YJDoc2: @YJDoc2 rename `env_exists` to `path_exists`

YJDoc2:
There are a couple of other fixes needed in exec impl as well, so probably split this PR into another one. duplicate gid is pretty small in scope, so not mixing it with exec changes.

PR #2900, opened 2024-09-01
selinux: create Vagrantfile for SELinux

Gekko0114:
In this PR, created Vagrantfile for an SELinux environment.

ref: https://github.com/containers/youki/pull/2850

PR #3021, opened 2024-12-14
Refactor the selinux crate

utam0k:


utam0k:
@Gekko0114 I'm trying to remove `inti_*` in the selinux crate. WDYT?

Gekko0114:
What do you mean by `inti_*` in the selinux crate ?
I couldn't find any functions or values with the prefix `inti_`.

utam0k:
> What do you mean by `inti_*` in the selinux crate ? I couldn't find any functions or values with the prefix `inti_`.

Ops, like that one `policy_root_init_done`

Gekko0114:
I see. I am not familiar with Rust, therefore it is possible that there should be better implementation rather than using atomicBool.
Thanks, I will wait for your PR :)

utam0k:
I've pushed my changes. Overall, I have already finished it, but I still need to edit it to run it on VM.

utam0k:
@Gekko0114 I couldn't completely run `main.rs` on VM based on Vagrantfile, even if it's in the main branch. Did you pass it?
```
[vagrant@centos8 youki]$ sudo ./target/debug/selinux
current enforce mode is: permissive
SELinux is enabled
SELinux label of current process is: unconfined_u:unconfined_r:unconfined_t:s0-s0
Error: SetFileLabel("Failed to set_xattr: Operation not supported (os error 95)")
```

Gekko0114:
Hi @utam0k 
I noticed this message now, so sorry for late reply. I will have a look.

Gekko0114:
Hi @utam0k 

Though I am using libvirt instead of virtualbox because of my PC environment, I could run the code.
```
selinux is enabled
default enforce mode is: enforcing
current enforce mode is: enforcing
SELinux label of current process is: unconfined_u:unconfined_r:unconfined_t:s0-s0
file label is system_u:object_r:public_content_t:s0
```


Based on previous comments from YJDoc2, it appears this error is likely related to directory mounting within Vagrant, rather than an issue with the code itself. YJDoc2 found that copying the binary to /tmp resolved the issue, indicating that the problem arises from how Vagrant mounts directories.
https://github.com/youki-dev/youki/pull/2900#issuecomment-2363855380

> Hmm, the issue was that with the directory mounting. When I copied the binary into /tmp it worked correctly as expected and sets the label. This is probably issue with mounting dirs into vagrant, not not with our code, so can ignore. Going ahead and merging this PR. Thanks a lot!

Therefore, while the error is occurring, it seems to be an environmental issue rather than code issue.
What do you think?

PR #3034, opened 2024-12-31
Use selinux.format_mount_label when formatting mount label

Gekko0114:
Previously, the mount label was formatted without using the SELinux library.
However, since the SELinux library was implemented in Rust, we can use it for formatting the mount label.
Therefore, in this PR, SELinux.format_mount_label is used.

This PR is related to the issue #2718

### DIFF
--- a/crates/libcontainer/Cargo.toml
+++ b/crates/libcontainer/Cargo.toml
@@ -51,6 +51,7 @@ thiserror = "2.0.9"
 tracing = { version = "0.1.41", features = ["attributes"] }
 safe-path = "0.1.0"
 nc = "0.9.5"
+selinux = { path = "../../experiment/selinux" }
 
 [dev-dependencies]
 oci-spec = { version = "~0.7.1", features = ["proptests", "runtime"] }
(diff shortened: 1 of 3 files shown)


Gekko0114:
Hi @utam0k @YJDoc2 
I used experiment/selinux crate, but I am not sure this implementation is correct.
Could you review it when you have time?
If this implementation is correct, then I will create other PRs related to SELinux.

YJDoc2:
Hey @Gekko0114 , couple of things :
- please rebase on latest main, that will solve some of the CI failures
- I don't think we should add the selinux crate dep from `experiment/` . If we are done with major work of the implementation of that create, we can move it form `experiment/` to the `crates/` and use that. My reasoning is that once we merge this PR , we will face issues when releasing as crates io needs direct deps not path deps.
- What I can suggest is that we can have one commit which moves the selinux to `crates/` and then changes in this PR. This can be in same PR or individual PRs.
- Also I'll suggest waiting till we have made a release, as otherwise this will break the release, we need to check if we need make any changes in our release flow for accommodating new crates.

Apart from that I'll take a look on the actual changes again. Thanks for the PR!

Gekko0114:
Thanks @YJDoc2 for your suggestion!

Hi @utam0k,
@YJDoc2 suggested that the SELinux crate should be moved under the `crates/` directory first.
However, you are working on a PR refactoring the SELinux crate https://github.com/youki-dev/youki/pull/3021.
Should I wait to move SELinux crate to `crates/` until you've completed your PR first?

utam0k:
@Gekko0114 Yes, please wait for it. I'm working on it, but I need more time.

PR #3036, opened 2025-01-01
Don't specify the versionFile for tagpr

utam0k:
It looks like we don't need to specify because the release versions should be synced with git tags.
https://github.com/Songmu/tagpr#tagprversionfile

### DIFF
--- a/.tagpr
+++ b/.tagpr
@@ -1,7 +1,7 @@
 [tagpr]
     vPrefix = true
     releaseBranch = main
-    versionFile = .tagpr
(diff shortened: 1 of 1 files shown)


YJDoc2 (approved):
lgtm. I think we should also remove the `.tagpr` file from repo once we have verified this works in next release.

utam0k:
> I think we should also remove the .tagpr file from repo once we have verified this works in next release.

No, we don't do that because `.tagpr` is the configure file for tagpr.

PR #3064, opened 2025-01-31
ci: Publish packages after the release flow

utam0k:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [x] CI/CD related changes

## Testing

N/A

## Related Issues

None

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -90,9 +90,7 @@ jobs:
 
   publish:
     name: Publish Packages
-    needs: build
-    environment:
-      name: release
+    needs: release
     runs-on: ubuntu-22.04
     if: ${{ github.repository == 'containers/youki' }}
     env:


.github/workflows/release.yaml
```diff
@@ -90,9 +90,7 @@ jobs:
 
   publish:
     name: Publish Packages
-    needs: build
-    environment:
-      name: release
+    needs: release
```
YJDoc2, line 93: We should still keep the env as release right? That way the maintainers can approve the env and only the publish the package. Or are we removing that functionality, and going directly to release if the test and builds pass?

utam0k: I didn't think that was necessary since we'd already have to approve the environment on the release job. What do you think?

YJDoc2: oh, yes. I missed that in the diff. Apologies, this makes sense then 👍

YJDoc2 (approved):

PR #3085, opened 2025-02-25
update rust version to 1.85.0

YJDoc2:
## Description
This updates rust version to latest stable 1.85.0. This also aligns with the new rust editions, and some deps like wasmtime require us to update our rust version to 1.82+ anyways.

The code changes are simply fixes suggested by clippy.

NOTE: it seems cross-rs has not yet caught up with latest rust version, so it is failing to compile. Will need to wait for them, or use pre-compiled binaries to avoid this.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes
- [X] Other (please describe): Rust version update

## Testing

- [ ] Added new unit tests
- [X] Ran existing test suite

## Related Issues
N/A

## Additional Context
Even though  the fix suggestions by clippy should be correct, we should have another look for confirmation and sanity checks.

### DIFF
--- a/tests/contest/contest/src/utils/test_utils.rs
+++ b/tests/contest/contest/src/utils/test_utils.rs
@@ -222,7 +222,12 @@ pub fn test_inside_container(
         .wait_with_output()
     {
         Ok(c) => c,
-        Err(e) => return TestResult::Failed(anyhow!("container start failed : {:?}", e)),
+        Err(e) => {
+            // given that start has failed, we can be pretty sure that create has either failed
+            // or completed already, so we wait on it so it does not become a zombie process
+            let _ = create_process.wait_with_output();
+            return TestResult::Failed(anyhow!("container start failed : {:?}", e));
+        }
     };
 
     let create_output = create_process
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -95,7 +95,7 @@ jobs:
       - name: Install just
         uses: taiki-e/install-action@just
       - name: Install cross-rs
-        run: cargo install cross --git https://github.com/cross-rs/cross
+        run: RUSTFLAGS="" cargo install cross --git https://github.com/cross-rs/cross
       - name: Setup target
         run: |
           echo "CARGO=cross" >> ${GITHUB_ENV}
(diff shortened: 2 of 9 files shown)


tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -222,7 +222,12 @@ pub fn test_inside_container(
         .wait_with_output()
     {
         Ok(c) => c,
-        Err(e) => return TestResult::Failed(anyhow!("container start failed : {:?}", e)),
+        Err(e) => {
+            // given that start has failed, we can be pretty sure that create has either failed
+            // or completed already, so we wait on it so it does not become a zombie process
+            let _ = create_process.wait_with_output();
+            return TestResult::Failed(anyhow!("container start failed : {:?}", e));
```
YJDoc2, line 229: I would like someone to take another look at this, the clippy error was that in the early return codepath here, we do not wait on the create process, hence it can become a zombie process. Hence I added a wait here with assumption that if the `start_container` has failed, we can be fairly certain that the original create has also errored or has already exited.

However, would like someone else to confirm the logic.

utam0k: Why don't we create a different PR for only this part? This could be an important change.

YJDoc2: Hey, ok, I'll create a  separate PR just for this change, and once that is resolved I can rebase this PR.

YJDoc2:
Hey, should I also update the rust edition to 2024 in this PR, so just rust version update for now, and maybe update the edition later?

yihuaf:
> Hey, should I also update the rust edition to 2024 in this PR, so just rust version update for now, and maybe update the edition later?

Without knowing how much changes to upgrade to 2024 editions, we should do it in a different PR.

tests/contest/contest/src/utils/test_utils.rs
YJDoc2: I have opened https://github.com/youki-dev/youki/pull/3130 to extract just this change, once that is merged, I can remove it from here if needed.

Issue #3087, opened 2025-02-25
[Bug]: disable_oom_killer in cgroup_v1


xujihui1985:
### Bug Description

https://github.com/youki-dev/youki/blob/a8b8fe2423282cfaface461b252165dcc0a3a641/crates/libcgroups/src/v1/memory.rs#L115

The logic here is incorrect. The intended behavior is to enable the OOM killer when disable_oom_killer is not set, which is the common case. However, since 0 means enable and 1 means disable, the if condition is currently reversed.

This is a really critical issue, as it will cause processes in the container to hang.

### Steps to Reproduce

1. create container
2. check memory.oom_control

xujihui1985:
@YJDoc2  Hi, it’s been a while since the last patch—hope you’re doing well!

I’ve found a serious issue in cgroup v1 management. The logic for setting the OOM killer is incorrect. The intended behavior is to enable the OOM killer when disable_oom_killer is not set, which is the common case. However, since 0 enables it and 1 disables it, the current if condition is reversed. This can cause processes in the container to hang.

Let me know what you think!

utam0k:
Fixed by https://github.com/youki-dev/youki/pull/3090

PR #3090, opened 2025-02-26
fix(libcgroup): fix disable_oom_killer in cgroup v1

xujihui1985:
The intended behavior is to enable the OOM killer when disable_oom_killer is not set, which is the common case. However, since 0 means enable and 1 means disable, the if condition is currently reversed.

This is a really critical issue, as it will cause processes in the container to hang.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #

### DIFF
--- a/crates/libcgroups/src/v1/memory.rs
+++ b/crates/libcgroups/src/v1/memory.rs
@@ -204,10 +204,7 @@ impl Memory {
         Ok(memory_data)
     }
 
-    fn set_oom_control(
-        cgroup_root: &Path,
-        disable_oom_killer: bool,
-    ) -> Result<(), WrappedIoError> {
+    fn set_oom_control(cgroup_root: &Path, disable_oom_killer: bool) -> Result<(), WrappedIoError> {
         if disable_oom_killer {
             common::write_cgroup_file(cgroup_root.join(CGROUP_MEMORY_OOM_CONTROL), 1)
         } else {
--- a/crates/libcgroups/src/v1/memory.rs
+++ b/crates/libcgroups/src/v1/memory.rs
@@ -111,7 +111,7 @@ impl Controller for Memory {
                 )?;
             }
 
-            Self::set_disable_oom_killer(cgroup_root, controller_opt.disable_oom_killer)?;
+            Self::set_oom_control(cgroup_root, controller_opt.disable_oom_killer)?;
 
             if let Some(swappiness) = memory.swappiness() {
                 if swappiness <= 100 {
@@ -204,7 +204,7 @@ impl Memory {
         Ok(memory_data)
     }
 
-    fn set_disable_oom_killer(
+    fn set_oom_control(
         cgroup_root: &Path,
         disable_oom_killer: bool,
     ) -> Result<(), WrappedIoError> {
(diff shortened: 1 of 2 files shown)


xujihui1985:
@YJDoc2 fix issue https://github.com/youki-dev/youki/issues/3087

crates/libcgroups/src/v1/memory.rs
```diff
@@ -208,6 +204,17 @@ impl Memory {
         Ok(memory_data)
     }
 
+    fn set_disable_oom_killer(
```
utam0k, line 207: I felt that the name of this function disabled the OOM killer. But it's actually determined by looking at the option. Could you please change the name of this function, or why don't you check the option on the caller side?

xujihui1985: Yeah, I changed the function name from **set_disable_oom_killer** to **set_oom_control** to make it more meaningful. The reason I created a separate function is to make it easier to test its correctness

xujihui1985: @utam0k  I think this is a very serious bug that causes unexpected behavior in the container. The process will not be killed by the OOM killer, leading to all memory allocations hanging. As a result, you’ll notice a large number of tasks in the “D” state, and you won’t be able to exec into the container since every memory allocation will be stuck.

utam0k: > @utam0k I think this is a very serious bug that causes unexpected behavior in the container. The process will not be killed by the OOM killer, leading to all memory allocations hanging. As a result, you’ll notice a large number of tasks in the “D” state, and you won’t be able to exec into the container since every memory allocation will be stuck.

I understood the situation. Do you use you in your production? If yes, I'd rush to publish the new release.

utam0k (approved):

utam0k:
@xujihui1985 Thanks for your contribution!

crates/libcgroups/src/v1/memory.rs
xujihui1985: thanks, yes, we are using in production, but no rush, I have forked the repo to fix the issue for our production environment.

utam0k: @xujihui1985 
I see! Anyway, thanks for your contribution. If you give me the production use case, I'd really appreciate it and give me the motivation.

utam0k: > Please release the new version carefully, but I think it’s necessary to highlight this issue in the release notes.

I really agree with you.

> I’m really shocked that no one else has encountered this issue so far—maybe most users have already migrated to cgroup v2. :-)

And support for Ubuntu 20.04(the default is cgroup v1) on github actions is ending :)

xujihui1985: > @xujihui1985 
> 
> I see! Anyway, thanks for your contribution. If you give me the production use case, I'd really appreciate it and give me the motivation.

We are using Youki to sandbox the user code execution environment in an IDE production, but I can’t disclose its name at the moment. Once the product matures, we may reveal more technical details. We chose Youki for its low memory footprint and well-maintained source code and test cases.

utam0k: > We are using Youki to sandbox the user code execution environment in an IDE production, but I can’t disclose its name at the moment. Once the product matures, we may reveal more technical details. We chose Youki for its low memory footprint and well-maintained source code and test cases.

Awesome!! I am very happy to hear that. I wish you the best of luck with your product. Please let us know if you have any more problems.

Issue #2917, opened 2024-09-18
[Bug]: Fix Wasm executors to exit after run completion


YJDoc2:
### Bug Description

In youki we expect that `executor.exec` will not return if the exec is successful. This behavior is consistent for tradition containers, where we do an exec syscall and replace process image, thus never returning from exec call. However for wasm workloads our `exec` implementation does return, and causes crash as it reaches the `unreachable!` statement. We can either allow the executor to return, or just do a `process.exit` with exit code from wasm code in our wasm executors to conform it to traditional containers.

### Steps to Reproduce

Follow https://containers.github.io/youki/user/webassembly.html to create a wasm container and run it. It will run, but in the end it will error saying -
```
Printing args
1
2
3
Printing envs
("PATH", "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
("TERM", "xterm")
("container", "podman")
("HOSTNAME", "b569c0459007")
thread 'main' panicked at crates/libcontainer/src/process/container_init_process.rs:644:5:
internal error: entered unreachable code: the executor should not return if it is successful.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Expectation

Youki should not crash in this way. The exit should be graceful, i.e. with a proper exit code (non-zero in case of wasm module returning non-zero exit)

### System and Setup Info

N/A , should be reproducible on all systems

### Additional Context

FYI, if a new contributor is interested in taking this , ping me (@YJDoc2 ) for help before starting!

kiokuless:
@YJDoc2 I'm interested in this issue, but I won't be able to work on it until the weekends in October. If that’s okay, I would love to take it on.

YJDoc2:
Hey, sure I will assign this to you. 

Context for this is - for normal binaries we use `exec` syscall so that process image is replcaed by that binary. For wasm, we use wasm runtimes which execute the wasm file. Unlike binary here we do not replcae the process image so after the runtime is done executing, it returns and causes the error. So in the main wasm executor implementation we want to take the exit code returned by those runtime and basically do a `process.exit` with that exit code.

kiokuless:
@YJDoc2 Thanks for sharing contexts! My new PC has arrived, I just started using a Linux desktop machine from today. So I can finally tackle this issue. I need an additional bit of time, but I will do this for sure.

YJDoc2:
[...]

PR #3099, opened 2025-03-06
fix: Gracefully terminate processes after successful execution of Wasm executors

z63d:
## Description

This PR ensures that processes are gracefully terminated after the successful execution of a Wasm executors.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes https://github.com/youki-dev/youki/issues/2917

### DIFF
--- a/crates/youki/src/workload/wasmedge.rs
+++ b/crates/youki/src/workload/wasmedge.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use libcontainer::oci_spec::runtime::Spec;
 use libcontainer::workload::{Executor, ExecutorError, ExecutorValidationError};
+use wasmedge_sdk::error::{CoreError, CoreExecutionError, WasmEdgeError};
 use wasmedge_sdk::wasi::WasiModule;
 use wasmedge_sdk::{params, Module, Store, Vm};
 
@@ -47,9 +48,15 @@ impl Executor for WasmedgeExecutor {
         vm.register_module(Some("main"), module).unwrap();
 
         vm.run_func(Some("main"), "_start", params!())
-            .map_err(|err| ExecutorError::Execution(err))?;
+            .map_err(|err| match *err {
+                // This case indicates that the wasm code panicked.
+                WasmEdgeError::Core(CoreError::Execution(CoreExecutionError::Unreachable)) => {
+                    std::process::exit(101)
+                }
+                _ => ExecutorError::Execution(err),
+            })?;
 
-        Ok(())
+        std::process::exit(0)
     }
 
     fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {
--- a/crates/youki/src/workload/wasmer.rs
+++ b/crates/youki/src/workload/wasmer.rs
@@ -1,7 +1,9 @@
+use std::error::Error;
+
 use libcontainer::oci_spec::runtime::Spec;
 use libcontainer::workload::{Executor, ExecutorError, ExecutorValidationError, EMPTY};
 use wasmer::{Instance, Module, Store};
-use wasmer_wasix::WasiEnv;
+use wasmer_wasix::{WasiEnv, WasiError};
 
 const EXECUTOR_NAME: &str = "wasmer";
 
@@ -71,13 +73,21 @@ impl Executor for WasmerExecutor {
                 "could not retrieve wasm module main function: {err}"
             ))
         })?;
-        start
-            .call(&mut store, &[])
-            .map_err(|err| ExecutorError::Execution(err.into()))?;
+
+        start.call(&mut store, &[]).map_err(|err| {
+            if let Some(WasiError::Exit(exit_code)) = err
+                .source()
+                .and_then(|err_source| err_source.downcast_ref::<WasiError>())
+            {
+                std::process::exit(exit_code.raw())
+            } else {
+                ExecutorError::Execution(err.into())
+            }
+        })?;
 
         wasi_env.cleanup(&mut store, None);
 
-        Ok(())
+        std::process::exit(0)
     }
 
     fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {
--- a/crates/youki/src/workload/wasmtime.rs
+++ b/crates/youki/src/workload/wasmtime.rs
@@ -1,6 +1,7 @@
 use libcontainer::oci_spec::runtime::Spec;
 use libcontainer::workload::{Executor, ExecutorError, ExecutorValidationError, EMPTY};
 use wasi_common::sync::{add_to_linker, WasiCtxBuilder};
+use wasi_common::I32Exit;
 use wasmtime::{Engine, Linker, Module, Store};
 
 const EXECUTOR_NAME: &str = "wasmtime";
@@ -85,9 +86,14 @@ impl Executor for WasmtimeExecutor {
             ExecutorError::Other("could not retrieve wasm module main function".into())
         })?;
 
-        start
-            .call(&mut store, &[], &mut [])
-            .map_err(|err| ExecutorError::Execution(err.into()))
+        start.call(&mut store, &[], &mut []).map_err(|err| {
+            match err.downcast_ref::<I32Exit>() {
+                Some(exit) => std::process::exit(exit.0),
+                None => ExecutorError::Execution(err.into()),
+            }
+        })?;
+
+        std::process::exit(0)
     }
 
     fn validate(&self, spec: &Spec) -> Result<(), ExecutorValidationError> {


YJDoc2:
Hey @z63d , this is a bit different that what I was thinking. I was thinking that instead of propagating the result back and branch according to it, the executors for was themselves should exit appropriately.
So the current behavior for default (non-wasm) workload is that we do and exec syscall, which basically replaces the youki process by the new command. In wasm however, the executor simply returns when the wasm execution is done. I was thinking that we capture the result there and process it there only. Now there can be three cases here -
1. the wasm exited with status 0. This will be probably reported as Ok(()). In this case we can simply process::exit(0) from the executor itself.
2. the wasm exited with non-zero status. We need to check what is the return value in this case, but it might get reported as error. We (youki) needs to process::exit with that status code.
3. the wasm execution failed due to some other error, and the program was not executed. In this case, we need to propagate the error back.

My though for moving this logic into executor itself was that this way the behavior of all executors are same, all run the process and never return. (like the default).

wdyt?

z63d:
@YJDoc2 
Thanks for the advice!

> the wasm exited with non-zero status. We need to check what is the return value in this case, but it might get reported as error. We (youki) needs to process::exit with that status code.

How do I get the exit code?

YJDoc2:
Hey @z63d , so I checked the doc of run_func [here](https://wasmedge.github.io/wasmedge-rust-sdk/wasmedge_sdk/struct.Vm.html#method.run_func) . So it runs Err if there is any error when executing the function, and if the execution is successful it returns Ok with `WasmValue` .  We can simply return the error in the same way we are handling it right now, and for Ok we can do this -
The wasm value is enum [here](https://wasmedge.github.io/wasmedge-rust-sdk/wasmedge_sys/types/struct.WasmValue.html) and the actual type can be found using `ty` function on it. If the type is `I32`, we should exit with that value, and in any other case we should exit with `0` as the result is Ok. Similar things should be done for other runtimes as well.

wdyt? Any issues or caveats you can see with this?

z63d:
I checked the return value of the function in the Ok case at each runtime.
But there doesn't seem to be any information...

WasmEdge
https://github.com/youki-dev/youki/blob/a91e3470aff8be4a46bea2af16d30b9d1e647e8e/crates/youki/src/workload/wasmedge.rs#L49-L50
```
[]
```

Wasmer
https://github.com/youki-dev/youki/blob/a91e3470aff8be4a46bea2af16d30b9d1e647e8e/crates/youki/src/workload/wasmer.rs#L74-L76
```
[]
```

Wasmtime
https://github.com/youki-dev/youki/blob/a91e3470aff8be4a46bea2af16d30b9d1e647e8e/crates/youki/src/workload/wasmtime.rs#L88-L90
```
Ok(())
```

YJDoc2:
Hey how did you check this? I can try the same on my side and check again. I guess we need to return something from the wasm code in order to get the value here. Need to check more.

z63d:
@YJDoc2 
I checked using `tracing::debug!` or `panic!`.
The reason I used panic! was because Wasmer and Wasmtime display a huge amount of debug logs, and for some reason I had an issue where the logs I had inserted weren't displayed.
If you have any good methods or advice on debugging youki, please let me know.

I used this as reference.
https://youki-dev.github.io/youki/user/webassembly.html?search=#build-youki-with-wasm-wasmedge-wasm-wasmer-or-wasm-wasmtime-feature-flag-enabled

WasmEdge
```
let result = vm
    .run_func(Some("main"), "_start", params!())
    .map_err(|err| ExecutorError::Execution(err))?;
tracing::debug!("result: {:?}", result);
```

```
$ ./scripts/build.sh -o . -f wasm-wasmtime

$ sudo podman --runtime /home/kaita_nakamura/youki/youki run localhost/wasm-module 1 2 3
DEBUG youki: started by user 0 with ArgsOs { inner: ["/home/kaita_nakamura/youki/youki", "start", "f1f8366fd0e38e2cfac4b19893e1145a21bf7ed644757d0adc4d02ae1c571054"] }
DEBUG libcontainer::notify_socket: notify container start
DEBUG libcontainer::notify_socket: notify finished
DEBUG libcontainer::container::container: Save container status: Container { state: State { oci_version: "v1.0.2", id: "f1f8366fd0e38e2cfac4b19893e1145a21bf7ed644757d0adc4d02ae1c571054", status: Running, pid: Some(30226), bundle: "/var/lib/containers/storage/overlay-containers/f1f8366fd0e38e2cfac4b19893e1145a21bf7ed644757d0adc4d02ae1c571054/userdata", annotations: Some({"io.podman.annotations.autoremove": "FALSE", "io.kubernetes.cri-o.TTY": "false", "io.podman.annotations.privileged": "FALSE", "io.kubernetes.cri-o.Created": "2025-03-24T11:49:21.528539426Z", "io.container.manager": "libpod", "io.podman.annotations.init": "FALSE", "io.podman.annotations.publish-all": "FALSE", "module.wasm.image/variant": "compat", "org.opencontainers.image.stopSignal": "15"}), created: Some(2025-03-24T11:49:21.899044680Z), creator: Some(0), use_systemd: true, clean_up_intel_rdt_subdirectory: Some(false) }, root: "/run/youki/f1f8366fd0e38e2cfac4b19893e1145a21bf7ed644757d0adc4d02ae1c571054" } in "/run/youki/f1f8366fd0e38e2cfac4b19893e1145a21bf7ed644757d0adc4d02ae1c571054"
DEBUG libcontainer::notify_socket: received: start container
DEBUG youki::workload::wasmedge: executing workload with wasmedge handler
Printing args
1
2
3
Printing envs
("PATH", "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
("TERM", "xterm")
("container", "podman")
("HOSTNAME", "f1f8366fd0e3")
DEBUG youki::workload::wasmedge: result: []
thread 'main' panicked at crates/libcontainer/src/process/container_init_process.rs:719:5:
internal error: entered unreachable code: the executor should not return if it is successful.
fatal runtime error: IO Safety violation: owned file descriptor already closed
```

Wasmer
```
let result = start
    .call(&mut store, &[])
    .map_err(|err| ExecutorError::Execution(err.into()))?;
panic!("result: {:?}", result);
```

```
$ ./scripts/build.sh -o . -r -f wasm-wasmer

$ sudo podman --runtime /home/kaita_nakamura/youki/youki run --env WASMER_DIR=/home/kaita_nakamura/.wasmer/bin/wasmer localhost/wasm-module 1 2 3
Printing args
1
2
3
Printing envs
("PATH", "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
("TERM", "xterm")
("container", "podman")
("WASMER_DIR", "/home/kaita_nakamura/.wasmer/bin/wasmer")
("HOSTNAME", "a659346cfeb8")
thread 'main' panicked at crates/youki/src/workload/wasmer.rs:79:9:
result: []
```

Wasmtime

```
let result = start
    .call(&mut store, &[], &mut [])
    .map_err(|err| ExecutorError::Execution(err.into()));
panic!("result: {:?}", result);
```

```
$ ./scripts/build.sh -o . -r -f wasm-wasmtime

$ sudo podman --runtime /home/kaita_nakamura/youki/youki run localhost/wasm-module 1 2 3
Printing args
1
2
3
Printing envs
("PATH", "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
("TERM", "xterm")
("container", "podman")
("HOSTNAME", "ac2815855de0")
thread 'main' panicked at crates/youki/src/workload/wasmtime.rs:91:9:
result: Ok(())
```

YJDoc2:
Hey @z63d Thanks for checking. I think you used the example program from our documentations which is like
```
fn main() {
    println!("Printing args");
    for arg in std::env::args().skip(1) {
        println!("{}", arg);
    }

    println!("Printing envs");
    for envs in std::env::vars() {
        println!("{:?}", envs);
    }  
}
```

What I meant in my previous comment (https://github.com/youki-dev/youki/pull/3099#issuecomment-2709412360) , is that in this case, we are not returning anything so we should simply exit with 0. However if we update the main function in the rust code example above, to return some value ; say `return 0` or `return 5` ; in that case we should check what the executors return. I think in this particular cases they will simply return that as INT32 value or equivalent. In all other case with Ok(...), we should exit 0 ; in case of error we are propagating it anyway. (you can also check what happens if the wasm itself errors by maybe adding panic! in the code example)

May I ask you to check this?

z63d:
@YJDoc2 
I fixed in the case of Wasmer and Wasmtime.
However, in the case of WasmEdge, I cannot capture the exit code. It always returns `[]`. No errors will occur.
What should I do?

I tried it with the following source code.

```rust
use std::process::ExitCode;

fn main() -> ExitCode {
    ExitCode::from(5)
}
```

z63d:
> Instead we should return the exit code as Ok - change the current signature of method .

This change affects [runwasi](https://github.com/containerd/runwasi).
https://github.com/containerd/runwasi/blob/a73e5bcea5a19f88981b97daaa1c4b7d1832b8c8/crates/containerd-shim-wasm/src/sys/unix/container/executor.rs#L57

I spoke with @utam0k and I think if we are going to make this change it would be best to let the runwasi members know in advance.
What do you think @YJDoc2 ?

z63d:
@YJDoc2 
Has there been any progress?

YJDoc2:
Hey @jprendes can you comment from runwasi's side how much problematic it can be if we changed the function signature?

short version is - Currently our wasm executors return after executing the wasm code , which is problematic because the liibcontainer expects that executors will never return. Runwasi handles this correctly by calling process::exit after executing the wasm function. We want to do the same, and want to try to propagate the return code (if any) returned by wasm code. However to do so would require change in the executor trait function signature to return i32 instead of current () . Would this breakage have any major implications from runwasi's side, or would you be fine with us changing the signature and making a major release?

jprendes:
It would be fine from runwasi side.
That's not a big change, we would just need to adjust the signature of our executor when we bump libcontainer.
I tried to get the context, but why can't the wasm executors un youki do the same process::exit as in runwasi? or is it just to improve the libcontainer API?

YJDoc2:
> I tried to get the context, but why can't the wasm executors un youki do the same process::exit as in runwasi? or is it just to improve the libcontainer API?

Ok, so my original reasoning behind this was that this way the process::exit would be called in a single place, and the wasm will simply return the value. However after your response and more thinking, maybe it would be better to avoid breakage and just call process::exit from the wasm executors . The issue is also from `!` as return type (never-return) is not stabilized so we cannot use it as the `exec` 's signature.

@z63d may I ask you that instead of returning value from wasm executor, just call process::exit similar to how runwasi is doing in the exec function itself? Thanks :)

z63d:
@YJDoc2 
I have modified the source code.
Does this match your intent?

YJDoc2:
Hey @z63d , sorry, what I meant was instead of returning the exit code and changing the function signature in the wasm executor's exec fn, we exit from there. So instead of propagating the exit code, just exit from there immediately.

YJDoc2:
Hey, ok let me check this and get back to you.

YJDoc2:
ok, @z63d so for this, what we should do is
- if result is Ok, exit with status code 0. There is no way for us to get exist code from run_func or otherwise as far as I checked.
- instead of directly calling `.map_err(|err| ExecutorError::Execution(err))?` on the error, we match to check if the error is specifically `Err(Core(Execution(Unreachable)))` , and in that case exit with code 101 (corresponding to rust's panic).
- In other cases, we propagate the error like before.

wdyt?

z63d:
@YJDoc2

z63d:
fixed.

crates/youki/src/workload/wasmedge.rs
```diff
@@ -47,7 +48,13 @@ impl Executor for WasmedgeExecutor {
         vm.register_module(Some("main"), module).unwrap();
 
         vm.run_func(Some("main"), "_start", params!())
-            .map_err(|err| ExecutorError::Execution(err))?;
+            .map_err(|err| match *err {
+                // This case indicates that the WasmEdge panicked.
```
YJDoc2, line 52:
```suggestion
                // This case indicates that the wasm code panicked.
```

YJDoc2 (commented):
Hey one last change - Let the exec function itself `process::exit` with 0 if the wasm code has successfully executed. This keeps with the expectation that the `exec` function of executors never returns if user's command starts successfully. We can remove the `process::exit` from executor.rs in that case.

Otherwise LGTM, thanks for your help :)

z63d:
@YJDoc2 
Please review.

YJDoc2 (approved):
LGTM!
Thanks for the PR :)

PR #3097, opened 2025-03-05
chore(ci): Upgrade GitHub Actions workflows for `ubuntu-24.04`

sou1118:
## Description

This pull request involves updating the GitHub Actions workflows to use a newer version of the Ubuntu runner and making some additional changes to the workflows. Below are the most important changes:

### Ubuntu Runner Version Updates:
[...]

PR #3102, opened 2025-03-07
chore(ci): add cgroup v1 compatibility for tests on ubuntu-24.04

### LINKED ISSUE #3095: [FEATURE]: Upgrade GitHub Actions workflows from Ubuntu 20.04 to 24.04
## Background
[GitHub has announced the upcoming removal of Ubuntu 20.04 support in GitHub Actions.](https://github.blog/changelog/2025-02-12-notice-of-upcoming-deprecations-and-breaking-changes-for-github-actions/) They will be implementing brownouts (temporary failures) for workflows using ubuntu-20.04 during specific time windows in March 2025, before completely removing support. This affects our CI/CD pipeline as we currently use Ubuntu 20.04 in our GitHub Actions workflows.

## Feature Request
I'm proposing that we update all our GitHub Actions workflow files to use Ubuntu 24.04 instead of Ubuntu 20.04.

## Is the request related to some problem running youki?
This is not directly related to running youki, but rather ensuring our CI/CD pipeline continues to function without interruption. If we don't update, our GitHub Actions workflows will start failing during the brownout periods and eventually stop working altogether.

## Proposed Solution
We should update all occurrences of `runs-on: ubuntu-20.04` to `runs-on: ubuntu-24.04` in our workflow YAML files. This change will require additional modifications since our tests specifically need cgroup-v1, while Ubuntu 24.04 uses cgroup-v2 by default.

For the cgroup-v1 tests, we'll need to implement one of these approaches:
1. Configure the Ubuntu 24.04 runner to use cgroup-v1 (may require kernel boot parameters)
2. Use a container or VM within the workflow that has cgroup-v1 enabled
3. Create a custom solution that allows testing with cgroup-v1 on newer Ubuntu versions

## Considerations
- Ubuntu 24.04 uses cgroup-v2 by default, while our tests specifically require cgroup-v1
- According to GitHub issue discussions, configuring cgroup-v1 on newer Ubuntu versions may require kernel boot parameters, which isn't straightforward on GitHub-hosted runners
- We may need to create a virtualized testing environment within our workflows
- We should explore whether Docker-in-Docker or similar approaches could help maintain a cgroup-v1 testing environment
- Ubuntu 24.04 includes newer package versions which may affect our build environment

## Additional Context
According to GitHub's announcement on February 12, 2025, the brownouts will occur on:
- March 4, 2025: 14:00 - 22:00 UTC
- March 11, 2025: 13:00 - 21:00 UTC
- March 18, 2025: 13:00 - 21:00 UTC
- March 25, 2025: 13:00 - 21:00 UTC

I recommend we complete this migration before the first brownout on March 4th to avoid any disruptions.

If no one else is already working on this issue, I'm willing to take on this task and create a pull request with the necessary changes. However, since I don't have permissions to run GitHub Actions workflows directly, I may need assistance from team members with appropriate access rights for testing the changes. I can research possible solutions for maintaining cgroup-v1 tests and prepare the initial changes, but would appreciate collaboration during the implementation and testing phase.

sou1118:
## Description
This PR adds support for testing with cgroups v1 on Ubuntu 24.04 by integrating Lima with AlmaLinux 8. AlmaLinux 8 provides an environment with cgroups v1 which allows us to maintain compatibility testing while upgrading the CI infrastructure to Ubuntu 24.04.

The implementation:
- Uses Lima to create a virtual machine with AlmaLinux 8 when running cgroups v1 tests
- Configures Docker in the guest VM and forwards the Docker socket to the host
- Maintains matrix testing for both cgroups v1 and v2 environments
- Keeps the workflow streamlined with conditional steps based on the cgroup version

This is a follow-up to PR #3097 which upgraded the GitHub Actions workflows to Ubuntu 24.04.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [x] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite in both cgroups v1 (Lima/AlmaLinux) and cgroups v2 (native) environments
- [x] Tested manually by running the workflow locally with Lima

## Related Issues

Fixes #3095

## Additional Context
The approach was inspired by the containerd/nerdctl implementation ([4d76aa96b44df51e975171216d54879794b50605](https://github.com/containerd/nerdctl/commit/4d76aa96b44df51e975171216d54879794b50605)) which successfully uses Lima to provide cgroups v1 compatibility testing on modern Ubuntu versions.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -184,14 +184,11 @@ jobs:
         run: just test-rootless-podman
   
   docker-in-docker:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-24.04
     needs: [youki-build]
-    timeout-minutes: 5
     strategy:
       matrix:
-        # ubuntu 20.04 has cgroups-v1
-        # ubuntu 22.04 has cgroups-v2
-        os: [ "ubuntu-22.04", "ubuntu-20.04" ]
+        cgroup-version: ["v1", "v2"]
     steps:
       - uses: actions/checkout@v4
       - name: Install just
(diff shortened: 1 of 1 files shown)


utam0k:
> The approach was inspired by the containerd/nerdctl implementation ([4d76aa96b44df51e975171216d54879794b50605](https://github.com/containerd/nerdctl/commit/4d76aa96b44df51e975171216d54879794b50605)) which successfully uses Lima to provide cgroups v1 compatibility testing on modern Ubuntu versions.

💯

.github/workflows/e2e.yaml
```diff
@@ -184,25 +184,73 @@ jobs:
...
       matrix:
-        # ubuntu 20.04 has cgroups-v1
-        # ubuntu 22.04 has cgroups-v2
-        os: [ "ubuntu-22.04", "ubuntu-20.04" ]
+        cgroup-version: ["v1", "v2"]
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      # Common step for both versions
```
utam0k, line 196: I don't think this comment needs.

sou1118: I fixed it!

utam0k (approved):
Thanks!

PR #3106, opened 2025-03-10
fix: CPU controller tests for Kernel 6.10 cgroup v2 changes

### LINKED ISSUE #3104: [FEATURE]: Align with cgroup v2 CPU controller kernel documentation implementation
### Background

Recent Linux kernels (including Ubuntu 24.04) now strictly enforce the cgroup v2 CPU controller behavior as documented in the kernel specifications. These changes affect youki's current implementation and tests:

1. `cpu.max` configuration: According to kernel documentation, cpu.max is a "read-write two value file" in format "$MAX $PERIOD" and "If only one number is written, $MAX is updated." In previous implementations, syntax like `- 250000` could be used to set only the period while keeping quota unchanged. Current kernels strictly enforce the documented format, requiring explicit use of `max 250000` to set the period while keeping quota unlimited.

2. `cpu.weight` range validation: The kernel docs specify the weight range is [1, 10000]. Previously, extremely high weight values (e.g., 500000) would be automatically capped at the maximum value (10000). Current kernels strictly enforce this range and reject out-of-range values with "Numerical result out of range" error.

A related kernel patch addressing cgroup v2 CPU controller parameter handling has been merged into Linux Kernel 6.10: https://lore.kernel.org/all/20240424132438.514720-1-serein.chengyu@huawei.com/

These changes cause existing test cases to fail in Ubuntu 24.04 and will likely affect other distributions using recent kernels.

### Feature Request

Extend youki's cgroup v2 support to properly align with kernel documentation and handle both legacy and current implementations:

1. Update the CPU controller's quota/period setting mechanism to use the proper format
2. Add more flexible handling of out-of-range CPU weight values
3. Modify test cases to work with both legacy and newer implementations

When running youki tests on Ubuntu 24.04, the following failures are reported:

```
5 / 11 : test_cpu_period_valid_set : not ok expected cpu quota to be 250000, but was 'max' ... 
9 / 11 : test_cpu_weight_too_high_maximum_set : not ok container creation was not successful. Exit code was Some(1)
```

These failures occur because of the strict enforcement of kernel documentation in cgroup v2 implementations. Direct testing confirms that attempting to set only the period in `cpu.max` results in an "Invalid argument" error, and setting excessively high CPU weight values causes a "Numerical result out of range" error.

### Proposed Solution

1. Update `cpu.max` handling:
   - When modifying only the period, explicitly include the current quota value (or "max")
   - Use the documented format "$MAX $PERIOD" for all operations
   - Extend the `prepare_cpu_max` function to support string values

2. Improve CPU weight handling:
   - Add error handling for out-of-range CPU weight settings
   - Consider container creation failure with out-of-range values as an expected result

3. Make test cases more flexible:
   - Modify `test_cpu_period_valid_set` to verify proper period setting while accepting "max" quota
   - Update `test_cpu_weight_too_high_maximum_set` to pass whether the implementation caps to 10000 or rejects the value

### Considerations

1. Backward compatibility: The solution should work with all kernel versions
2. Future enhancement: May consider kernel version detection to set specific test expectations
3. Error handling: Provide clearer error messages to make debugging easier

### Additional Context

Actual system behavior observed on Ubuntu 24.04:

```
$ cat /sys/fs/cgroup/test_cpu/cpu.max
max 100000

$ echo '- 250000' > /sys/fs/cgroup/test_cpu/cpu.max
bash: line 1: echo: write error: Invalid argument

$ echo 'max 250000' > /sys/fs/cgroup/test_cpu/cpu.max
$ cat /sys/fs/cgroup/test_cpu/cpu.max
max 250000

$ cat /sys/fs/cgroup/test_cpu/cpu.weight
100

$ echo '500000' > /sys/fs/cgroup/test_cpu/cpu.weight
bash: line 1: echo: write error: Numerical result out of range
```

These behaviors align with the kernel documentation for cgroup v2 CPU controller:
- `cpu.max`: "A read-write two value file... If only one number is written, $MAX is updated."
- `cpu.weight`: "For non idle groups (cpu.idle = 0), the weight is in the range [1, 10000]."

This issue is important for ensuring youki works correctly on current and future Linux distributions that properly implement the kernel specifications.

sou1118:
## Description
Fix CPU controller tests to work with Ubuntu 24.04's cgroup v2 implementation, which now more strictly follows the Linux kernel documentation.

Ubuntu 24.04 has changed the behavior of cgroup v2 CPU controller in two significant ways:
1. `cpu.max` format is now strictly enforced as "$MAX $PERIOD" as per kernel docs. When updating only period, you must use "max $PERIOD" format (previously "- $PERIOD" worked)
2. `cpu.weight` range [1, 10000] is now strictly enforced, rejecting out-of-range values with "Numerical result out of range" rather than silently capping them to the maximum

These changes align with the kernel documentation, which states:
- For `cpu.max`: "The maximum bandwidth limit. It's in the following format: $MAX $PERIOD... If only one number is written, $MAX is updated." 
- For `cpu.weight`: "For non idle groups (cpu.idle = 0), the weight is in the range [1, 10000]."

This PR modifies the tests to be compatible with both the historic behavior and the new strict enforcement in Ubuntu 24.04.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing
- [ ] Added new unit tests
- [x] Ran existing test suite
  - Tested on Ubuntu 24.04 with cgroup v2 enabled
  - Confirmed that tests now pass on both Ubuntu 24.04 and older systems

## Related Issues

fixed #3104
Related to opencontainers/runc#4166 (similar cgroup v2 behavior changes)

## Additional Context
A recent kernel patch (https://lore.kernel.org/all/20240424132438.514720-1-serein.chengyu@huawei.com/) related to cgroup v2 CPU controller parameter handling has been merged into Linux Kernel 6.10. This confirms that these behavioral changes are now official part of the mainline kernel.

Ubuntu 24.04 appears to have backported these or similar changes to their kernel (likely 6.8+), resulting in the stricter enforcement of the cgroup v2 documentation specifications.

In the future, we could consider adding kernel version detection to make tests more deterministic (e.g., expecting strict behavior for kernel 6.10+ and the legacy behavior for older versions), but the current approach ensures compatibility across all kernel versions by accepting both behaviors.

This is similar to the issue found in opencontainers/runc where newer kernel implementations of cgroup v2 have subtle behavioral differences from older versions. The fix makes the tests more flexible to accommodate these differences while still validating the core functionality.

This is follow-up PR #3097

### DIFF
--- a/tests/contest/contest/src/tests/cgroups/cpu/v2.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/v2.rs
@@ -108,12 +108,16 @@ fn test_cpu_weight_too_high_maximum_set() -> TestResult {
     let spec = test_result!(create_spec("test_cpu_weight_too_high_maximum_set", cpu));
     // We accept both behaviors: either container creation fails due to
     // out-of-range value, or it succeeds with weight set to maximum (10000)
-    test_outside_container(&spec, &|data| match check_container_created(&data) {
-        Ok(_) => match check_cpu_weight("test_cpu_weight_too_high_maximum_set", 10_000) {
-            Ok(_) => TestResult::Passed,
-            Err(e) => TestResult::Failed(anyhow::anyhow!("Weight check failed: {}", e)),
-        },
-        Err(_) => TestResult::Passed, // Failure is expected on kernels that strictly enforce range
+    test_outside_container(&spec, &|data| {
+        check_container_created(&data).map_or_else(
+            |_e| TestResult::Passed, // Failure is expected on kernels that strictly enforce range
+            |_| {
+                check_cpu_weight("test_cpu_weight_too_high_maximum_set", 10_000).map_or_else(
+                    |e| TestResult::Failed(anyhow::anyhow!("Weight check failed: {}", e)),
+                    |_| TestResult::Passed,
+                )
+            },
+        )
     })
 }
 
--- a/tests/contest/contest/src/tests/cgroups/cpu/v2.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/v2.rs
@@ -178,10 +178,8 @@ fn test_cpu_quota_negative_default_set() -> TestResult {
     })
 }
 
-/// Tests if a valid cpu period (x > 0) is set successfully. Cpu quota needs to
-/// remain unchanged
+/// Tests if a valid cpu period (x > 0) is set successfully
 fn test_cpu_period_valid_set() -> TestResult {
-    let quota = 250_000;
     let expected_period = 250_000;
     let cpu = test_result!(LinuxCpuBuilder::default()
         .period(expected_period)
@@ -189,19 +187,27 @@ fn test_cpu_period_valid_set() -> TestResult {
         .context("build cpu spec"));
 
     let spec = test_result!(create_spec("test_cpu_period_valid_set", cpu));
-    test_result!(prepare_cpu_max(
-        &spec,
-        &quota.to_string(),
-        &expected_period.to_string()
-    ));
+    test_result!(prepare_cpu_max(&spec, "max", &expected_period.to_string()));
 
     test_outside_container(&spec, &|data| {
         test_result!(check_container_created(&data));
-        test_result!(check_cpu_max(
-            "test_cpu_period_valid_set",
-            quota,
-            expected_period
-        ));
+        let data = test_result!(read_cgroup_data("test_cpu_period_valid_set", "cpu.max"));
+        let parts: Vec<&str> = data.split_whitespace().collect();
+        if parts.len() != 2 {
+            return TestResult::Failed(anyhow::anyhow!("Invalid cpu.max format: {}", data));
+        }
+        let period = parts[1].trim();
+        let actual_period = match period.parse::<u64>() {
+            Ok(p) => p,
+            Err(e) => return TestResult::Failed(anyhow::anyhow!("Failed to parse period: {}", e)),
+        };
+        if actual_period != expected_period {
+            return TestResult::Failed(anyhow::anyhow!(
+                "expected cpu period to be {}, but was {}",
+                expected_period,
+                actual_period
+            ));
+        }
         TestResult::Passed
     })
 }
--- a/tests/contest/contest/src/tests/cgroups/cpu/v2.rs
+++ b/tests/contest/contest/src/tests/cgroups/cpu/v2.rs
@@ -95,23 +95,25 @@ fn test_cpu_weight_zero_ignored() -> TestResult {
     })
 }
 
-/// Tests if a cpu weight that is too high (over 10000 after conversion) is set to the maximum value
+/// Tests behavior when setting a cpu weight that is too high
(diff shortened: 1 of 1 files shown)


tests/contest/contest/src/tests/cgroups/cpu/v2.rs
```diff
@@ -95,23 +95,20 @@ fn test_cpu_weight_zero_ignored() -> TestResult {
...
-
     let spec = test_result!(create_spec("test_cpu_weight_too_high_maximum_set", cpu));
-    test_outside_container(&spec, &|data| {
-        test_result!(check_container_created(&data));
-        test_result!(check_cpu_weight(
-            "test_cpu_weight_too_high_maximum_set",
-            converted_cpu_weight
-        ));
-        TestResult::Passed
+    test_outside_container(&spec, &|data| match check_container_created(&data) {
```
utam0k, line 106: What changed?

sou1118: The test was updated to handle two valid behaviors across different version OS.

Before: silently cap weight to 10000 and succeed
After: reject out-of-range values and fail container creation

Since both behaviors are valid according to kernel documentation (which specifies the range as [1, 10000]), the test now passes in either case, making it compatible with both implementations.

I added a comment because it was unclear.

tests/contest/contest/src/tests/cgroups/cpu/v2.rs
```diff
@@ -95,23 +95,25 @@ fn test_cpu_weight_zero_ignored() -> TestResult {
...
-        ));
-        TestResult::Passed
+    // We accept both behaviors: either container creation fails due to
+    // out-of-range value, or it succeeds with weight set to maximum (10000)
+    test_outside_container(&spec, &|data| match check_container_created(&data) {
+        Ok(_) => match check_cpu_weight("test_cpu_weight_too_high_maximum_set", 10_000) {
+            Ok(_) => TestResult::Passed,
+            Err(e) => TestResult::Failed(anyhow::anyhow!("Weight check failed: {}", e)),
+        },
+        Err(_) => TestResult::Passed, // Failure is expected on kernels that strictly enforce range
```
utam0k, line 116:
```suggestion
    test_outside_container(&spec, &|data| {
        check_container_created(&data).map_or_else(
            |_e| TestResult::Passed,  // Failure is expected on kernels that strictly enforce range
            |_| {
                check_cpu_weight("test_cpu_weight_too_high_maximum_set", 10_000).map_or_else(
                    |e| TestResult::Failed(anyhow::anyhow!("Weight check failed: {}", e)),
                    |_| TestResult::Passed,
                )
            },
        )
    })
```

utam0k (approved):
LGTM w/ a nit.

tests/contest/contest/src/tests/cgroups/cpu/v2.rs
utam0k: It'd be better to check the kernel version, whether the `check_container_created` should fail.

Issue #3111, opened 2025-03-15
[Bug/Security]: tj-actions/changed-files compromised version in workflows


sou1118:
### Bug Description

The popular GitHub Action tj-actions/changed-files has been compromised in a supply chain attack. This action is currently used in our repository workflows, specifically in `basic.yml` and `integration_tests_validation.yaml`. According to security reports, the compromised action exfiltrates CI/CD secrets by dumping them from the runner's memory.

### Steps to Reproduce

The issue affects all users of tj-actions/changed-files, including installations pinned to specific version tags, as the attacker has retroactively updated multiple version tags to reference the malicious commit.

In our repository, we're using this action in:
1. `.github/workflows/basic.yml` - using `tj-actions/changed-files@v41`
2. `.github/workflows/integration_tests_validation.yaml` - using `tj-actions/changed-files@v41`

### Expectation

The GitHub Action should function as expected without introducing security vulnerabilities or exfiltrating secrets.

### System and Setup Info

This affects any CI/CD environment running GitHub Actions with the compromised tj-actions/changed-files action.

## Additional Context

According to StepSecurity's report:

> The tj-actions/changed-files GitHub Action, which is currently used in over 23,000 repositories, has been compromised. In this attack, the attackers modified the action's code and retroactively updated multiple version tags to reference the malicious commit. The compromised Action prints CI/CD secrets in GitHub Actions build logs. If the workflow logs are publicly accessible (such as in public repositories), anyone could potentially read these logs and obtain exposed secrets.

> The compromised Action now executes a malicious Python script that dumps CI/CD secrets from the Runner Worker process. Most of the existing Action release tags have been updated to refer to the malicious commit mentioned below. Note: All these tags now point to the same malicious commit hash:0e58ed8671d6b60d0890c21b07f8835ace038e67, indicating the retroactive compromise of multiple versions.

As reported by [roblaszczak on Twitter](https://x.com/roblaszczak/status/1900849264693002386):

> tj-actions/changed-files GitHub action was compromised, and if you are using it, your secrets have been leaked to the attacker. The compromised version has been out for 16 hours already. To make matters worse, removing this action efficiently is pretty hard.
> 
> ALL tags are affected, so if you have pinned it by a tag, you are also affected. What makes it worse is that it affects all your branches. So commiting anithing to any branch that uses this action leaks secrets.

### Recommended action:
1. **Immediate fix**: Replace the action with a secure alternative or pin to a specific commit hash from before the compromise.
[...]

PR #3114, opened 2025-03-18
Add GitHub Actions to Dependabot configuration and pin actions to specific SHAs

sou1118:
## Description
This PR implements two security best practices for our GitHub Actions workflow:
1. Adds GitHub Actions to Dependabot configuration for weekly dependency checks
2. Pins all GitHub Actions to specific SHA commits rather than using tags for improved security and reproducibility

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [x] CI/CD related changes

## Testing
- [ ] Added new unit tests
- [x] Ran existing test suite
  - Verified Dependabot configuration syntax is valid
  - Confirmed GitHub Actions workflow runs successfully with pinned SHAs

## Related Issues

Fixes #

## Additional Context
Following GitHub security best practices, we've updated our workflow to pin GitHub Actions to specific SHAs instead of version tags. This prevents potential supply chain attacks where a tag could be modified to point to malicious code.

The Dependabot configuration now includes weekly checks for GitHub Actions updates to ensure we stay current with the latest security fixes while maintaining the ability to review changes before implementation.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -14,8 +14,9 @@ EOF
 
 RUN <<EOF
 apt-get update
+
+# For building
 apt-get install -y \
-  # For building
   build-essential \
   git \
   libclang-dev \
@@ -23,10 +24,12 @@ apt-get install -y \
   libseccomp-dev \
   libssl-dev \
   libsystemd-dev \
-  pkg-config \
-  # For debugging
-  bpftrace \
-  podman
+  pkg-config
+
+# For debugging
+apt-get install -y \
+  podman \
+  bpftrace
 
 curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/bin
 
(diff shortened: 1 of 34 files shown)


utam0k (changes requested):
This practice is appropriate for business How about for us? I don't think this will solve the essential problem. What do you think?
https://github.com/youki-dev/youki/issues/3111#issuecomment-2732792936

sou1118:
You're absolutely right. After considering the maintenance burden for our OSS project, I believe a targeted approach makes more sense:

1. Pin SHAs only for actions that:
   - Handle secrets or sensitive data
   - Have deployment capabilities
   - Require elevated permissions

2. For other actions, we can continue using version tags with reasonable constraints (e.g., `v4` → `v4.x.y`)

3. We'll also review our workflow permissions to implement the principle of least privilege, ensuring actions only have access to what they absolutely need.

I apologize for submitting this PR without fully considering the practical implications for our project scale.

Issue #2827, opened 2024-06-23
[RFC] Consider Alternate CI for running Podman tests


YJDoc2:
## Background

Currently we have a CI for running podman e2e tests using Youki as the runtime. This is intended to serve as a conformance test to check that Youki can work with podman correctly. We only run tests with sudo, so this does not test rootless behavior. Currently Youki are not passing, details are below.

## Motivation

The tests which are failing currently can be differentiated in 3 categories :
- Tests failing because Youki has different/incorrect impl. For eg, a test is failing because the error message given by youki is not in the format expected by the test. These kind of tests can be fixed, and should be fixed so that youki can be used with podman.
- Tests failing because missing configuration. There are some tests that depend on certain config files such as ssh certs etc. being present. The podman repo injects these via secrets, and we cannot provide these, hence these tests will always fail. Thankfully, most of these tests does not test things related to running containers, so we can safely ignore these.
- Tests failing because of incorrect configuration. Github CI configures the CI VM in such way that we cannot override certain config which are needed for the tests. If you run these tests in the vagrant VM, these tests pass. Also some other issues are : bash and bats version in CI is older, hence some tests do not run correctly at all.

Thus in order to fix 1st and 3rd tests, I am proposing to run podman tests CI in some other CI provider, such as Cirrus CI or Circle CI. These both provide a free tier for public OSS repo with credits and unlike github, provide VM setup that we have complete control over. Note that podman itself uses Cirrus CI. 

I am NOT proposing to move any other CI to these, as that is not needed, and does not make any sense.

## Considerations

Some considerations I have done :

- The average CI runtime for podman tests is 40 min. However, because it uses ubuntu, we need to compile some deps from source such as netavark and aardvark-dns. These take about 1-2 minutes. Also after fixing the bash and bats version, I expect few more minutes to be shaved off.  Thus we need ~35*31 = 1085 minutes per month credits.
- Cirrus CI provides 10,000 cpu-minutes on Linux VM per month (https://cirrus-ci.org/faq/#are-there-any-limits) . Note that the CPU here can be faster, so the tests might actually be a bit faster here, or can be potentially parallelized.
- Circle CI provides 6000 build-minutes per month in free tier, but their Linux VM consumes 10 credits per minute. I'm not sure what is relation between build-minutes and credits, so this may or may not be viable.
[...]

PR #3125, opened 2025-03-24
chore(ci): fix podman-tests ci to use fedora on lima

sou1118:
## Description

This PR fixes a CI workflow for testing the compatibility between Podman and Youki as an OCI runtime. The workflow now uses a Fedora VM via Lima to provide a more stable and appropriate testing environment, ensuring a more accurate assessment of how Youki performs with higher-level container management tools.

This change effect could check it https://github.com/youki-dev/youki/actions/runs/14129458601 .

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [x] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #

## Additional Context

The updated workflow:
1. Sets up a Fedora VM using Lima (rather than the previous environment)
2. Builds Youki from source and installs it as the system's OCI runtime
3. Installs all necessary Podman dependencies from Fedora's package repositories
4. Builds Podman from source with appropriate tags for full feature support
5. Executes Podman's system tests with Youki as the runtime

The move to Fedora provides a more suitable environment for Podman testing, as Fedora is the primary development platform for Podman. This change resolves previous environment issues including PATH variable problems and missing dependencies that were causing test failures.

The test suite now shows significantly improved results, with most tests passing successfully. This confirms that Youki functions properly as a drop-in replacement for other OCI runtimes when used with Podman.

### DIFF
--- a/.github/workflows/podman_tests.yaml
+++ b/.github/workflows/podman_tests.yaml
@@ -2,7 +2,7 @@ name: 🧪 Test for podman
 
 on:
   schedule:
-  - cron: "0 0 * * *"
+    - cron: "0 0 * * *"
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


YJDoc2:
Hey @sou1118 , thanks for doing this! There are a couple of things in this
1. run all of the tests instead of just first one. I'm pretty sure there are several tests later on which used to fail because of missing skopeo or aardvark dep. If we are removing the step to install it, we need to check if that actually works or not
2. If we are not building podman from main, we need to use tests from tag which was used to build the podman. It is possible that tests in main are not passable with podman version that we have installed in runner.

Also when I commented we need to fix podman tests, my scope was larger than just this. See the [issue](https://github.com/youki-dev/youki/issues/2827) for more details, but my main intention was that maybe with lima or something (I'm not much familiar with it) , we could run these tests in fedora instead of ubuntu which solves a lof of our dependencies issues (skopeo and aardvark are packaged for fedora officially and are pretty recent, the bats and bash are also pretty new). Depending on if lima launches a proper vm or just a container we would also be able to solve 3rd point mentioned in the issue.

Can you check that? Thanks :)

sou1118:
Hi @YJDoc2! 
I've fixed the podman-tests CI workflow and it's now able to run. 
Could you help me understand which test cases are critical and need to pass for us to consider this integration successful?

PR #3157, opened 2025-04-30
fix: mount retry and logging

z63d:
## Description

Fixes mount_into_container handling.
Also, since the function name seems to have changed in the https://github.com/youki-dev/youki/pull/378, I adjusted the test function name to match.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

https://github.com/youki-dev/youki/issues/3132

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -2,6 +2,7 @@ use std::fs::{canonicalize, create_dir_all, OpenOptions};
 use std::mem;
 use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 #[cfg(feature = "v1")]
 use std::{borrow::Cow, collections::HashMap};
 
@@ -24,7 +25,16 @@ use super::symlink::SymlinkError;
 use super::utils::{parse_mount, MountOptionConfig};
 use crate::syscall::syscall::create_syscall;
 use crate::syscall::{linux, Syscall, SyscallError};
-use crate::utils::PathBufExt;
+use crate::utils::{retry, PathBufExt};
+
+const MAX_EBUSY_MOUNT_ATTEMPTS: u32 = 3;
+// runc has a retry interval of 100ms. We are following this.
+// https://github.com/opencontainers/runc/blob/v1.3.0/libcontainer/rootfs_linux.go#L1235
+#[cfg(not(test))]
+const MOUNT_RETRY_DELAY_MS: u64 = 100;
+// In tests, there is no need to delay, so set it to 0ms.
+#[cfg(test)]
+const MOUNT_RETRY_DELAY_MS: u64 = 0;
 
 #[derive(Debug, thiserror::Error)]
 pub enum MountError {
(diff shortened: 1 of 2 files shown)


crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -546,29 +546,27 @@ impl Mount {
             PathBuf::from(source)
         };
 
-        if let Err(err) =
-            self.syscall
+        let max_retries: u32 = 3;
+        let mut retries = 0;
+        loop {
```
utam0k, line 551: I don't prefer using the infinity loop. Please use `for` instead of that, at least.

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -546,29 +546,27 @@ impl Mount {
...
-                    tracing::error!("mount of {:?} failed. {}", m.destination(), errno);
+            {
+                Ok(_) => break,
+                Err(err) => {
+                    if let SyscallError::Nix(errno) = err {
+                        if (matches!(errno, Errno::EINVAL) || matches!(errno, Errno::EBUSY))
+                            && retries < max_retries
+                        {
+                            retries += 1;
+                            std::thread::sleep(std::time::Duration::from_millis(100));
```
utam0k, line 563: Is `100ms` reasonable?

utam0k (changes requested):
May I ask you to consider the unit test?

crates/libcontainer/src/rootfs/mount.rs
z63d: runc seems to set it to 100ms when doing the same thing.
https://github.com/opencontainers/runc/blob/8d90e3dba696ac787ee64de4445517ddf1063b04/libcontainer/rootfs_linux.go#L1232
But I don't know if this is reasonable 🤔

utam0k: Could you add a comment regarding where `100ms` came from?

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -546,29 +546,25 @@ impl Mount {
             PathBuf::from(source)
         };
 
-        if let Err(err) =
-            self.syscall
+        let max_attempts: u32 = 3;
```
utam0k: Please Make this a const variable

utam0k (commented):
Use `use` as appropriate.

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -546,29 +548,24 @@ impl Mount {
             PathBuf::from(source)
         };
 
-        if let Err(err) =
-            self.syscall
+        for attempt in 0..MAX_MOUNT_ATTEMPTS {
```
utam0k, line 551: The logic of retry is not of interest here. Why not cut this out for utili and others?

```rust
fn retry<F, T, E, P>(
    mut op: F,
    attempts: usize,
    delay: Duration,
    policy: P,
) -> Result<T, E>
where
    F: FnMut() -> Result<T, E>,
    P: Fn(&E) -> bool,
{
    fn go<F, T, E, P>( // Maybe “for” is easier to understand.
        mut op: F,
        remaining: usize,
        delay: Duration,
        policy: &P,
    ) -> Result<T, E>
    where
        F: FnMut() -> Result<T, E>,
        P: Fn(&E) -> bool,
    {
        op().or_else(|err| {
            if remaining > 1 && policy(&err) {
                std::thread::sleep(delay);
                go(op, remaining - 1, delay, policy)
            } else {
                Err(err) 
            }
        })
    }
    go(&mut op, attempts, delay, &policy)
}
```

YJDoc2:
@z63d CI are failing, please take a look

utam0k:
@z63d It seems the format error occurred.

utam0k (approved):
One nit, otherwise looks good.

crates/libcontainer/src/utils.rs
```diff
@@ -282,6 +283,39 @@ pub fn validate_spec_for_new_user_ns(spec: &Spec) -> Result<(), LibcontainerErro
...
 }
 
+// Generic retry function with delay and policy.
+// Retries the operation `op` up to `attempts` times if it fails.
+// Waits for `delay` duration between retries.
+// Only retries if the error satisfies the `policy` function.
+pub fn retry<F, T, E, P>(
+    mut op: F,
+    attempts: u32,
+    #[cfg_attr(test, allow(unused_variables))] delay: Duration,
```
YJDoc2, line 293: Hey, I think it should be ok to keep the delay in tests as well. with 100 ms delay and 2 attempts after delay, each test would take at most 200 ms ; even if there are 10 tests we might have a delay of 2 sec total, I think that is ok.

z63d: That is certainly true.
I'd like to hear @utam0k's opinion.
https://github.com/youki-dev/youki/pull/3157#discussion_r2070215467

utam0k: My concern is that if this function is used in many places, the waiting time will increase. However, since this is not the case now, it is acceptable to just leave the comment and make it const.

PR #3163, opened 2025-05-06
Add net device feature

### LINKED ISSUE #3136: [FEATURE]: Define Linux Network Devices
### Background

In this issue https://github.com/opencontainers/runtime-spec/issues/1239 and PR https://github.com/opencontainers/runtime-spec/pull/1271 on [runtime-spec](https://github.com/opencontainers/runtime-spec), support for the `netdevices` field in the OCI specification is being added.

### Feature Request

We are upgrading [oci-spec-rs](https://github.com/youki-dev/oci-spec-rs) by supporting the `netdevices` field and adding the feature of moving existing network devices into the container namespace.

_No response_

### Proposed Solution

_No response_

### Considerations

_No response_

### Additional Context

Since it's a recently added field, neither [runc](https://github.com/opencontainers/runc) nor [runtime-tools](https://github.com/opencontainers/runtime-tools) has implemented it yet. A pull request is currently open for `runc`. As for `runtime-tools`, the implementation is still missing, so additional test implementations will be necessary.

nayuta723:
## Description

Add the function to move the network device to a network namespace and the corresponding test.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [ ] Ran existing test suite
  - Set up a new network namespace and device, as shown below.
```
$ sudo ip netns add dummy0
$ sudo ip link add dummy0 typ dummy
$ sudo ip addr add 192.168.1.1/24 dev dummy0
```
  - Set up the configuration using `youki spec` and fix the netns and netDevice configurations as below
```
...
"linux": {
    "netDevices": {
      "dummy0": {}
    },
    ...
    "namespaces": [
      ...
      {
        "type": "network",
        "path": "/var/run/netns/dummy0"
      ...
```
  - Run the container
```
$ sudo ./youki create -b tutorial tutorial_container
```
  - Check the expected results
```
$ sudo ip netns exec dummy0 ip link
...
60: dummy0: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether 1a:95:f0:e7:3d:d8 brd ff:ff:ff:ff:ff:ff
```

## Related Issues

Fixes #3136 

## Additional Context

Please review the placement of the `setup_network_device` function, whether the asynchronous logic is implemented correctly, and whether the test case is sufficient.

### DIFF
--- a/crates/libcontainer/src/network/wrapper.rs
+++ b/crates/libcontainer/src/network/wrapper.rs
@@ -1,16 +1,20 @@
 use netlink_packet_core::NetlinkMessage;
 use netlink_packet_route::RouteNetlinkMessage;
 
-use super::Result;
 use super::client::NetlinkClient;
 use super::fake::FakeNetlinkClient;
 use super::traits::{Client, NetlinkMessageHandler};
+use super::{NetworkError, Result};
 
 /// Enum wrapper for different client types
 /// The `Client` trait contains generic methods, which makes it impossible to use as a trait object.
 /// Therefore, we define `ClientWrapper` as an enum-based dynamic dispatch to handle this.
 pub enum ClientWrapper {
+    /// Real NetlinkClient instance for production use
     Client(NetlinkClient),
+    /// Error state when NetlinkClient initialization failed
+    ErrorState,
+    /// Fake client for testing purposes
     Fake(FakeNetlinkClient),
 }
 
@@ -18,6 +22,7 @@ impl Client for ClientWrapper {
     fn send(&mut self, req: &NetlinkMessage<RouteNetlinkMessage>) -> Result<()> {
         match self {
             ClientWrapper::Client(client) => client.send(req),
+            ClientWrapper::ErrorState => Err(NetworkError::ClientInitializeError),
             ClientWrapper::Fake(client) => client.send(req),
         }
     }
@@ -28,6 +33,7 @@ impl Client for ClientWrapper {
     {
         match self {
             ClientWrapper::Client(client) => client.receive(handler),
+            ClientWrapper::ErrorState => Err(NetworkError::ClientInitializeError),
             ClientWrapper::Fake(client) => client.receive(handler),
         }
     }
@@ -38,6 +44,7 @@ impl Client for ClientWrapper {
     {
         match self {
             ClientWrapper::Client(client) => client.receive_multiple(handler),
+            ClientWrapper::ErrorState => Err(NetworkError::ClientInitializeError),
             ClientWrapper::Fake(client) => client.receive_multiple(handler),
         }
     }
@@ -52,6 +59,7 @@ impl Client for ClientWrapper {
     {
         match self {
             ClientWrapper::Client(client) => client.send_and_receive(req, handler),
+            ClientWrapper::ErrorState => Err(NetworkError::ClientInitializeError),
             ClientWrapper::Fake(client) => client.send_and_receive(req, handler),
         }
     }
@@ -66,6 +74,7 @@ impl Client for ClientWrapper {
     {
         match self {
             ClientWrapper::Client(client) => client.send_and_receive_multiple(req, handler),
+            ClientWrapper::ErrorState => Err(NetworkError::ClientInitializeError),
             ClientWrapper::Fake(client) => client.send_and_receive_multiple(req, handler),
         }
     }
@@ -76,9 +85,12 @@ impl Default for ClientWrapper {
         if cfg!(test) {
             ClientWrapper::Fake(FakeNetlinkClient::new())
         } else {
-            ClientWrapper::Client(
-                NetlinkClient::new().unwrap_or_else(|_| panic!("Failed to create NetlinkClient")),
-            )
+            // If NetlinkClient initialization fails, we store the error state
+            // instead of panicking. The error will be returned when the client is used.
+            match NetlinkClient::new() {
+                Ok(client) => ClientWrapper::Client(client),
+                Err(_) => ClientWrapper::ErrorState,
+            }
         }
     }
 }
--- a/crates/libcontainer/src/process/init/process.rs
+++ b/crates/libcontainer/src/process/init/process.rs
@@ -921,6 +921,10 @@ fn configure_container_network_devices(
     main_sender: &mut channel::MainSender,
     init_receiver: &mut channel::InitReceiver,
 ) -> Result<()> {
+    if net_device.is_empty() {
+        return Ok(());
+    }
+
     main_sender.network_setup_ready()?;
 
     let addrs_map = init_receiver.wait_for_move_network_device()?;
--- a/crates/libcontainer/src/network/mod.rs
+++ b/crates/libcontainer/src/network/mod.rs
@@ -13,6 +13,8 @@ pub enum NetworkError {
     Nix(#[from] nix::Error),
     #[error(transparent)]
     IO(#[from] std::io::Error),
+    #[error("failed to initialize NetlinkClient")]
+    ClientInitializeError,
 }
 
 type Result<T> = std::result::Result<T, NetworkError>;


nayuta-ai:
I have a few questions:
- In the runc tests, some tests fail since the net-devices feature is not yet implemented. Would it be acceptable to add logic on the CI side to skip these tests?
- The newly added unit test includes a process that checks the euid, which causes it to fail when executed as a non-root user. Which approach would be preferable: removing the test for non-root users, skipping the test when not running as root, or modifying the test to avoid checking the euid?
- Regarding the newly added contest, it creates and deletes network namespaces, but the tests seem to fail due to insufficient permissions. Would adding support in the CI environment to grant permission to manipulate network namespaces be acceptable?

Thank you for your attention to these points.

(WIP) In the containerd integration test, I haven’t identified the cause yet, so I’d like to continue investigating it.

nayuta-ai:
@utam0k 
I forgot to mention you. Please take a look when you have a moment!
https://github.com/youki-dev/youki/pull/3163#issuecomment-2858646215

utam0k:
> In the runc tests, some tests fail since the net-devices feature is not yet implemented. Would it be acceptable to add logic on the CI side to skip these tests?

Yes, but please leave the comment.

utam0k:
> Regarding the newly added contest, it creates and deletes network namespaces, but the tests seem to fail due to insufficient permissions. Would adding support in the CI environment to grant permission to manipulate network namespaces be acceptable?

Yes, it is acceptable.

utam0k:
> The newly added unit test includes a process that checks the euid, which causes it to fail when executed as a non-root user. Which approach would be preferable: removing the test for non-root users, skipping the test when not running as root, or modifying the test to avoid checking the euid?

How about using a system call mock?

utam0k (changes requested):

nayuta-ai:
> How about using a system call mock?

I'll proceed with that approach!

BTW, is my understanding correct that this refers to using the `mock` in the `libcontainer/src/syscall` package?
I have a few concerns about using the `syscall` package. Before introducing getter functions like `get_xxid`, we need to implement the `set_id` function on the mock side. Additionally, the existing `MockCalls` implementation in `TestHelperSyscall` is not sufficient, and we will likely need an implementation that allocates the ID values in memory, such as:

```rust
pub struct TestHelperSyscall {
    id: RefCell<MockId>,
    mocks: MockCalls,
}

pub struct MockId {
    uid: Uid,
    gid: Gid,
    euid: Uid,
    egid: Gid,
}
```
Furthermore, if we go with this approach, we would need to refactor many parts of the codebase that currently use `getxxid` directly, and switch them to go through the `syscall` package. Although I plan to support only `rootless_required` in this PR for now, it looks like it would involve quite a large change overall.

I'm currently thinking that building a mock like the above would be the right direction. Does this approach sound reasonable to you?

crates/libcontainer/Cargo.toml
```diff
@@ -46,11 +46,15 @@ libseccomp = { version = "0.3.0", optional = true }
...
 rust-criu = "0.4.0"
+rtnetlink = "0.16.0"
 regex = { version = "1.10.6", default-features = false, features = ["std", "unicode-perl"] }
 thiserror = "2.0.12"
 tracing = { version = "0.1.41", features = ["attributes"] }
 safe-path = "0.1.0"
 nc = "0.9.6"
+netlink-packet-route = "0.22.0"
+futures = "0.3.31"
+tokio = { version = "1", features = ["full"] }
```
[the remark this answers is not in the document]
utam0k: Do we really need to directly use the command?

utam0k:
> BTW, is my understanding correct that this refers to using the mock in the libcontainer/src/syscall package?

Yes.

> Furthermore, if we go with this approach, we would need to refactor many parts of the codebase that currently use getxxid directly, and switch them to go through the syscall package. Although I plan to support only rootless_required in this PR for now, it looks like it would involve quite a large change overall.

How large is it? You may use a different PR.

crates/libcontainer/Cargo.toml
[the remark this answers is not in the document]
nayuta-ai: I understand. I plan to implement it using a syscall based approach instead of using command. Thank you.

nayuta-ai:
> How large is it? You may use a different PR.

I'll address this in a separate PR.

crates/libcontainer/Cargo.toml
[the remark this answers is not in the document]
utam0k: Thanks for your understanding 🙏

nayuta-ai:
I’ve created a separate PR: #3173 .

nayuta723:
I have confirmed that it works locally, and it also succeeded the last time, as shown in the link below. Can we rerun the workflow?
https://github.com/youki-dev/youki/actions/runs/15473318906

I'd appreciate it if you could review it once the job has passed.

crates/libcontainer/src/network.rs
```diff
@@ -0,0 +1,657 @@
...
+            }
+        }
+
+        Ok(message)
+    }
+}
+
+/// dev_change_netns allows to move a device given by name to a network namespace given by nsPath
+/// and optionally change the device name.
+/// The device name will be kept the same if device.Name is the zero value.
```
utam0k: > if device.Name is the zero value.

What if device.Name is None

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -248,4 +251,45 @@ impl ContainerBuilderImpl {
 
         Ok(())
     }
+
+    /// setup_network_device sets up and initializes any defined network interface inside the container.
+    fn setup_network_device(&self, spec: &Spec, init_pid: Pid) -> Result<(), LibcontainerError> {
```
utam0k: &Linux is enough instead of &Spec.

crates/libcontainer/src/network.rs
```diff
@@ -0,0 +1,657 @@
...
+    // This prevents potential conflicts or disruptions on the host network during the transition,
+    // particularly if other host components depend on this specific interface or its properties.
+    link_client.set_down(index)?;
+
+    // Get the existing IP addresses on the interface.
+    let addrs = addr_client.get_by_index(index)?;
+
+    link_client.set_ns_fd(index, &new_name, netns_file.as_raw_fd())?;
+
+    let thread_handle = std::thread::spawn({
```
utam0k: Please let me know why we need to use threads in this process.

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -248,4 +251,45 @@ impl ContainerBuilderImpl {
...
+                .any(|ns| ns.typ() == LinuxNamespaceType::Network)
+            {
+                return Ok(());
+            }
+
+            // get the namespace defined by the config and fall back
+            // to the one created by youki to run the container process.
+            let ns_path = namespaces.iter().find_map(|ns| {
+                if ns.typ() == LinuxNamespaceType::Network {
+                    ns.path().clone()
```
utam0k: Wouldn't a referenece suffice?

crates/libcontainer/src/network.rs
```diff
@@ -0,0 +1,657 @@
...
+
+/// dev_change_netns allows to move a device given by name to a network namespace given by nsPath
+/// and optionally change the device name.
+/// The device name will be kept the same if device.Name is the zero value.
+/// This function ensures that the move and rename operations occur atomically.
+/// It preserves existing interface attributes, including IP addresses.
+pub fn dev_change_net_namespace(
+    name: String,
+    netns_path: String,
+    device: LinuxNetDevice,
```
utam0k:
```suggestion
    device: &LinuxNetDevice,
```

crates/libcontainer/src/network.rs
```diff
@@ -0,0 +1,657 @@
...
+    let mut addr_client = AddressClient::new()?;
+
+    let netns_file = File::open(netns_path)?;
+    let origin_netns_file = File::open("/proc/self/ns/net")?;
+
+    let new_name = device
+        .name()
+        .as_ref()
+        .filter(|d| !d.is_empty())
+        .map_or(name.clone(), |d| d.to_string());
```
utam0k: new_name has to be validated.

nayuta-ai: The new_name has already been validated during the execution of `validate_spec_for_net_devices`.
Also, if the `device.Name` is empty, the device will keep its original name, so this does not cause any issues.

crates/libcontainer/src/network.rs
nayuta723: Addressed in [66f8b6e](https://github.com/youki-dev/youki/pull/3163/commits/66f8b6e1cdb0e32f43f98920412d7e49f01a4c60).

crates/libcontainer/src/network.rs
nayuta723: Addressed in [66f8b6e](https://github.com/youki-dev/youki/pull/3163/commits/66f8b6e1cdb0e32f43f98920412d7e49f01a4c60).

crates/libcontainer/src/network.rs
nayuta723: Addressed in [4a692a1](https://github.com/youki-dev/youki/pull/3163/commits/4a692a18c89c76a1e66a44723745424f6712d8d3).

crates/libcontainer/src/container/builder_impl.rs
nayuta723: Addressed in [4a692a1](https://github.com/youki-dev/youki/pull/3163/commits/4a692a18c89c76a1e66a44723745424f6712d8d3).

crates/libcontainer/src/container/builder_impl.rs
nayuta723: Addressed in [2424422](https://github.com/youki-dev/youki/pull/3163/commits/2424422124ea96e28b4073326581604781e2abac).

crates/libcontainer/src/container/init_builder.rs
```diff
@@ -200,7 +201,10 @@ impl InitContainerBuilder {
             }
         }
 
-        utils::validate_spec_for_new_user_ns(spec)?;
+        let syscall = create_syscall();
+        utils::validate_spec_for_new_user_ns(spec, &*syscall)?;
+        utils::validate_spec_for_net_devices(spec, &*syscall)
+            .map_err(LibcontainerError::NetDevicesError)?;
```
utam0k, line 214: What is the reason for doing this validation here?

crates/libcontainer/src/network.rs
utam0k: @nayuta-ai I can understand if you do this in the init process, but if you do it in main or intermediate, I don't see the problem. Does it need to be a thread? Can't it be a process?

crates/libcontainer/src/network.rs
```diff
@@ -0,0 +1,672 @@
...
+
+    let link = link_client.get_by_name(&name)?;
+
+    let index = link.header.index;
+
+    // Set the interface link state to DOWN before modifying attributes like namespace or name.
+    // This prevents potential conflicts or disruptions on the host network during the transition,
+    // particularly if other host components depend on this specific interface or its properties.
+    link_client.set_down(index)?;
```
utam0k: If it fails under this line, does it leave the interface down?

crates/libcontainer/src/network.rs
utam0k, line 1: Test coverage for this file is too low. And, I feel it's too fat.

crates/libcontainer/src/network.rs
nayuta723: Yes, the interface remains in the down state. This behavior is based on an implementation aligned with how `runc` handles it.
ref: https://github.com/opencontainers/runc/pull/4538

To avoid leaving the interface down in cases where the operations before moving the interface into the network namespace (specifically `addr_client.get_by_index` and `set_ns_fd`) fail, it seems reasonable to add logic that brings the interface back up when such failures occur.

crates/libcontainer/src/container/init_builder.rs
nayuta723: The `validate_spec_for_net_devices` function performs validation at this stage because we want to ensure the spec is valid before proceeding with container creation.
We intentionally avoid performing this check during container creation itself (e.g., within `dev_change_net_namespace`), as it is more appropriate to catch invalid configurations early in the lifecycle.
Is there any concern about this approach?

crates/libcontainer/src/network.rs
nayuta723: I'm sorry about that. I'll fix that.
I’d like to ask a few questions about this.
- I'm debating whether to create mocks using the existing `Syscall` trait framework or to define a new `NetworkOps` trait specifically for mocking.
  - Using the `Syscall` trait would require mocking at the socket level (e.g., `send`/`recv`), but since these operate on raw byte values, writing meaningful tests becomes difficult. On the other hand, mocking `send`, `receive`, and `receive_multiple` in `NetlinkClient` makes testing much easier, but doing so would go beyond the intended responsibilities of the `Syscall` trait.
  - This leads to the idea of introducing a new trait like `NetworkOps`, which could handle mocking for network components beyond just `Link` and `Address`. However, this also raises concerns about fragmentation and the proliferation of traits and mock components, including overlap with the existing `Syscall` trait.
- I'm considering refactoring the logic in `network.rs` into a separate `network` module, such as the `device` module. Would this be an appropriate approach?
  - ref: 96ef3be0f3735113e21b5ac41f93259a9d67e104

crates/libcontainer/src/network.rs
nayuta723: Given our previous discussion, it seems safer to perform the setns operations from a separate process rather than a thread, as it doesn't strictly need to be in the main process. Furthermore, since these operations don't require the main process itself, it would be beneficial to port them to the init process. Thank you for the confirmation.

Therefore, my proposed approach, which I believe is a suitable solution to the earlier questions, is as follows:
```mermaid
sequenceDiagram
    participant Main Process
    participant Intermediate Process
    participant Init Process

    Intermediate Process->>Main Process: Send "ready for intermediate process" and init PID
    Init Process->>Main Process: Send seccomp notify FD
    Main Process->>Init Process: Send "seccomp notify done"

    Note over Main Process: Setup network device and move<br>interface to target netns

    Main Process->>Init Process: Send "ready to setup network device"
    Note over Init Process: Setup network device for init process
    Init Process->>Main Process: Send "ready"
```

Seccomp is not directly related to the current implementation; it was included to indicate that the network device setup takes place after seccomp.

If the above understanding is correct, I think it should be feasible, so I'd like to try making the changes.

crates/libcontainer/src/container/init_builder.rs
utam0k: > The validate_spec_for_net_devices function performs validation at this stage because we want to ensure the spec is valid before proceeding with container creation.
We intentionally avoid performing this check during container creation itself (e.g., within dev_change_net_namespace), as it is more appropriate to catch invalid configurations early in the lifecycle.
Is there any concern about this approach?

Ops, don't worry this comment. It's my misunderstanding.

crates/libcontainer/src/network.rs
utam0k: @aojea 👋 Hi, we are implementing nevdev feature. I have a question about it. What if a failure occurs after the link goes down? Should we recover or ignore it?
Probably, we should define this behavior  on OCI Runtime Spec.

crates/libcontainer/src/network.rs
utam0k: > this also raises concerns about fragmentation and the proliferation of traits and mock components, including overlap with the existing Syscall trait.

I don't think so. I prefer NetoworkOps trait.

> I'm considering refactoring the logic in network.rs into a separate network module, such as the device module. Would this be an appropriate approach?

Yes, I think so too.

crates/libcontainer/src/network.rs
aojea: that is a good question, the main problem with dealing with interfaces is that they impact the entire namespace,so bringing an interface up that has changed or modified any network parameters can be dangerous as it impacts the entire host networking in this case

 My opinion in anything networking related is that failing close/safe is always the best option, so leaving it down you ensure that there is no risk associated with the operation despite it failed ...  if something in the host was already depending on this interface being up it will have the same impact that the operation succeeded than not succeed but leaving it down

crates/libcontainer/src/network.rs
nayuta723: Thanks, I'll work on that.

crates/libcontainer/src/network.rs
nayuta723: @utam0k 
Please confirm if the above policy is acceptable. If so, there's one difficult implementation point: how to pass `Vec<AddressMessage>` to the init process.

The challenge is that `AddressMessage` itself is used for Netlink responses and is not serializable, so it cannot be directly added to the `Message` Enum.

I've considered two workarounds:
* Keep the data as `bytes` internally within the `addr_client.get_by_index` function, without parsing it into `Vec<AddressMessage>`.
* Create a new struct, like `SerializableAddress`, and convert the data into a structure that can be described in the `Message` Enum.
  * The current undecided policy is described in [this Draft PR](https://github.com/nayuta-ai/youki/pull/3). Please refer to it for our discussion. If it seems fine, I plan to move it to this PR.
  * The relevant commit for the discussion is 6a055ec8b16bb0c4f81f20fe7b611bda96712cb3.

Personally, I think creating `SerializableAddress` and performing type-safe operations for each process is the better approach, but I'd like your input on how to implement this.

nayuta723: The advantage of handling address and link activation in the main process seems to be that network interface operations likely require root privileges. For network interface operations within a container, the `CAP_NET_ADMIN` capability would probably be necessary.

nayuta723: > https://github.com/youki-dev/youki/pull/3163#discussion_r2173738611

I'd like to provide a consolidated summary, as the previous discussions have become somewhat convoluted.

Initially, we explored whether it's feasible to complete the entire setup within a "main process" or "intermediate process" without relying on a dedicated "init process" within the new namespaces. We've concluded that this approach presents significant challenges:

* **Problem with Network Device Setup alongside UID/GID Mapping (in intermediate process):**
    While it's possible to perform network namespace-related operations within an intermediate process, a subsequent `mount` namespace operation (specifically, setting up a new root filesystem or remounting `/proc` and `/sys`) would render the previously configured network devices **invisible** from within the container.（My understanding might be incorrect.）This is because the file system view (provided by the mount namespace) would no longer reflect the host's `/proc` or `/sys` where network information is typically exposed, unless `/proc` and `/sys` are carefully re-mounted *after* the network setup *within the new mount namespace*.

* **Problem with Network Operations in an Init Process:**
    Although it's technically possible to configure network devices between the main process and an "init process" (the first process within the new PID namespace), this approach would necessitate that the "init process" itself possesses the necessary network-related capabilities (e.g., `CAP_NET_ADMIN`) *within its own namespace context*. This can complicate privilege management and might require elevated privileges for the init process, which we aim to avoid where possible.

Given these challenges, we believe that spawning a **kernel thread from the main process** to handle network device setup is the most suitable approach. This strategy allows for the necessary low-level network operations to be performed while maintaining separation and avoiding the complexities associated with mount namespace visibility and init process capabilities.

aojea: is what you are describing the reason of kubernetes having to use the `pause` container? some ref in https://github.com/containerd/containerd/issues/10505

nayuta723: My apologies for the confusion. I was referring to which process is most appropriate for executing `dev_change_net_namespace`.
I didn't intend to explain the reason why the `pause` container is necessary.

crates/libcontainer/src/network.rs
nayuta723: Apologies for the delayed response. I've already performed the validation here: https://github.com/youki-dev/youki/blob/60768cd6ab03cedeb7090b40d47b625f63f6c2f0/crates/libcontainer/src/utils.rs#L361-L372. Do I need to perform the validation again?

nayuta723:
@utam0k 
Please review this at your convenience. I'd particularly appreciate your confirmation on [this point](https://github.com/youki-dev/youki/pull/3163#discussion_r2159869857).

crates/libcontainer/src/network.rs
utam0k: @nayuta-ai Let me clarify my understanding. In other words, `dev_change_net_namespace` requires these condition, right?
- a.After changing mount ns
- b. the process has `CAP_NET_ADMIN`

I didn't understand why creating a thread would grant privilege.

nayuta723: I'm correcting my previous comment, as there was a misunderstanding.
The prerequisite for executing dev_change_net_namespace is the presence of the `CAP_NET_ADMIN` capability.

> I didn't understand why creating a thread would grant privilege.

The current implementation runs the thread within the `main process`, which has root privileges. Since the thread only joins the network namespace internally, it inherits the main process's privileges, allowing it to perform operations with those elevated permissions.

nayuta723: @utam0k 
To meet these requirements, we've identified three potential approaches:

* Continue with the current thread-based implementation.
* Move the thread's operations to an **intermediate process** (this is feasible because network namespaces are inherited by child processes).
* Perform the operations currently handled by the thread within the **init process**, specifically before capabilities are set (somewhere in `https://github.com/youki-dev/youki/blob/main/crates/libcontainer/src/process/init/process.rs#L71-L299`).

Personally, I'm leaning towards implementing the operations within the init process. My reasoning is that threads can be **unsafe**, and this approach helps prevent over-centralizing logic in an intermediate process. What are your thoughts on this? If you agree, I'll proceed with implementing this strategy.

utam0k: Why are the threads unsafe?

Do you prefer this option?
> Perform the operations currently handled by the thread within the init process, specifically before capabilities are set (somewhere in https://github.com/youki-dev/youki/blob/main/crates/libcontainer/src/process/init/process.rs#L71-L299).

nayuta723: > Why are the threads unsafe?

Factors like a file descriptor being closed prematurely can introduce thread safety issues, requiring careful handling (e.g., ensuring file descriptors are not invalidated during thread execution).

> Do you prefer this option?

Yes, I prefer this one because of its safe handling.

FYI
```
Thread Approach
- Pros:
  - Shared Memory for Data: Threads operate within the same memory space as the main process, allowing for easy sharing of variables (like address lists).
  - Simple Lifecycle Management: The processing is consolidated within the main process, leading to a simpler and more centralized lifecycle management.
- Cons:
  - Overhead from Thread Creation: There's an overhead associated with creating and tearing down threads, which scales with the number of devices being processed.
  - File Descriptor Safety Concerns: Factors like a file descriptor being closed prematurely can introduce thread safety issues, requiring careful handling (e.g., ensuring file descriptors are not invalidated during thread execution).
  - Potential Deviation from `youki`'s Design: If `youki` primarily uses processes rather than threads for operations, this approach might not align with its established architectural patterns.

Init Process Approach
- Pros:
  - No Thread Creation Overhead: This method avoids the overhead of creating new threads for each operation.
  - Alignment with `youki`'s Design: If `youki`'s core design favors process-based operations, this approach would be more consistent with its architectural philosophy.
- Cons:
  - Requires IPC for Data Sharing: Sharing data like address lists would necessitate implementing inter-process communication (IPC) mechanisms, such as socket communication.
  - Challenges with Large Data Sharing: Handling large address lists or other significant data sizes through IPC can become complex and less efficient.
```

utam0k: My concerns about using threads are performance and the potential for new security vulnerabilities. I don't think there is any other option besides the init process. Let's give it a try. Thank you for participating in the discussion. I feel like we've moved in a better direction.

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,263 @@
...
+use crate::network::serialize::SerializableAddress;
+
+/// dev_change_netns allows to move a device given by name to a network namespace given by nsPath
+/// and optionally change the device name.
+/// The device name will be kept the same if device.Name is None or an empty string.
+/// This function ensures that the move and rename operations occur atomically.
+/// It preserves existing interface attributes, including IP addresses.
+pub fn dev_change_net_namespace(
+    name: String,
+    netns_path: String,
```
utam0k: &str is enough for these vars. Others are the same.

crates/libcontainer/src/utils.rs
```diff
@@ -312,12 +315,94 @@ where
...
+        .ok_or(NetDevicesError::Spec(MissingSpecError::Linux))?;
+
+    if linux.net_devices().is_none() {
+        return Ok(());
+    }
+
+    if linux
+        .namespaces()
+        .as_ref()
+        .unwrap()
```
utam0k: Is this always safety?

crates/libcontainer/src/utils.rs
```diff
@@ -312,12 +315,94 @@ where
...
+    }
+
+    if let Some(devices) = linux.net_devices() {
+        for (name, net_dev) in devices {
+            if !dev_valid_name(name) {
+                return Err(NetDevicesError::InvalidDeviceName(name.to_string()));
+            }
+            if let Some(dev_name) = net_dev.name() {
+                if !dev_valid_name(dev_name) {
+                    return Err(NetDevicesError::InvalidDeviceName(dev_name.to_string()));
```
utam0k: Others too.
```suggestion
                    return Err(NetDevicesError::InvalidDeviceName(dev_name.into()));
```

crates/libcontainer/src/utils.rs
```diff
@@ -312,12 +315,94 @@ where
...
+                }
+            }
+        }
+    }
```
utam0k, line 345:
```suggestion
    if let Some(devices) = linux.net_devices() {
        devices.iter().try_for_each(|(name, net_dev)| {
            if !dev_valid_name(name) {
                return Err(NetDevicesError::InvalidDeviceName(name.to_string()));
            }
            if let Some(dev_name) = net_dev.name() {
                if !dev_valid_name(dev_name) {
                    return Err(NetDevicesError::InvalidDeviceName(dev_name.to_string()));
                }
            }
            Ok(())
        })?;
    }
```

crates/libcontainer/src/process/init/process.rs
```diff
@@ -276,6 +277,16 @@ pub fn container_init_process(
         InitProcessError::SyscallOther(err)
     })?;
 
+    // Setup some operations in the network namespace.
```
utam0k, line 301: Please make it clearly. What is `some operations`?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
 
+/// setup_network_device sets up and initializes any defined network interface inside the container.
+fn setup_network_device(
+    linux: &Linux,
+    init_pid: Pid,
+    main_receiver: &mut channel::MainReceiver,
+    init_sender: &mut channel::InitSender,
+) -> Result<()> {
+    let mut addrs_map = HashMap::new();
+    // host network pods does not move network devices.
```
utam0k: What does `pods` mean?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
 }
 
+/// setup_network_device sets up and initializes any defined network interface inside the container.
+fn setup_network_device(
+    linux: &Linux,
+    init_pid: Pid,
+    main_receiver: &mut channel::MainReceiver,
+    init_sender: &mut channel::InitSender,
+) -> Result<()> {
+    let mut addrs_map = HashMap::new();
```
utam0k: Define the address map where it is actually needed.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+    // host network pods does not move network devices.
+    if let Some(namespaces) = linux.namespaces() {
+        if !namespaces
+            .iter()
+            .any(|ns| ns.typ() == LinuxNamespaceType::Network)
```
utam0k: You can use `contains`
```suggestion
        if !namespaces.contains(LinuxNamespaceType::Network) {
```

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+        {
+            return Ok(());
+        }
+
+        // get the namespace defined by the config and fall back
+        // to the one created by youki to run the container process.
+        let fallback_ns_path = PathBuf::from(format!("/proc/{}/ns/net", init_pid.as_raw()));
```
utam0k: What is `fallback`? I guess it may be `default`...?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+        let ns_path = namespaces
+            .iter()
+            .find_map(|ns| {
+                if ns.typ() == LinuxNamespaceType::Network {
+                    ns.path().as_deref()
+                } else {
+                    None
+                }
+            })
+            .unwrap_or_else(|| &fallback_ns_path);
```
utam0k: You should find `ns_path` before L250, then return Ok if not found it.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
     Ok(())
 }
 
+/// setup_network_device sets up and initializes any defined network interface inside the container.
+fn setup_network_device(
```
utam0k: Could you please be more specific? As far as I know, this function is responsible for moving network devices.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +237,62 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+                )
+                .map_err(|err| {
+                    tracing::error!("failed to dev_change_net_namespace: {}", err);
+                    err
+                })?;
+                addrs_map.insert(name.clone(), addrs);
+            }
+            init_sender.move_network_device(addrs_map)?;
+        }
+    }
```
utam0k: How about this?
```suggestion
            let addrs_map = devices
                .iter()
                .map(|(name, net_dev)| {
                    let addrs = dev_change_net_namespace(
                        name.to_string(),
                        ns_path.to_string_lossy().to_string(),
                        net_dev,
                    )
                    .map_err(|err| {
                        tracing::error!("failed to dev_change_net_namespace: {}", err);
                        err
                    })?;
                    Ok((name.clone(), addrs))
                })
                .collect::<Result<HashMap<String, Vec<SerializableAddress>>>>()?;

            init_sender.move_network_device(addrs_map)?;

```

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,263 @@
...
+    let serialize_addrs: Vec<SerializableAddress> =
+        addrs.iter().map(SerializableAddress::from).collect();
+
+    Ok(serialize_addrs)
+}
+
+/// setup_network_device sets up a network device in a new namespace.
+/// It moves the device to the new namespace and adds the IP addresses to the device.
+/// It also sets the device up.
+pub fn setup_network_device(
```
utam0k: This function has two roles.
1. Assign an address.
2. Link up.

Can these parts be separated into two functions?

utam0k (commented):
I took a quick look at everything except the client.

crates/libcontainer/src/process/container_main_process.rs
nayuta723: I don't think it's possible to implement this using contains, because we're comparing different types.
```
/// LinuxNamespace is the configuration for a Linux namespace.
pub struct LinuxNamespace {
    #[serde(rename = "type")]
    #[getset(get_copy = "pub", set = "pub")]
    /// Type is the type of namespace.
    typ: LinuxNamespaceType,

    #[serde(default, skip_serializing_if = "Option::is_none")]
    #[getset(get = "pub", set = "pub")]
    /// Path is a path to an existing namespace persisted on disk that can
    /// be joined and is of the same type
    path: Option<PathBuf>,
}
```

crates/libcontainer/src/process/container_main_process.rs
nayuta723: fix: https://github.com/youki-dev/youki/pull/3163/commits/b9d62338ab7e011e8ff529930ad0993b240446a1

crates/libcontainer/src/network/network_device.rs
nayuta723: The implementation wasn't complex enough to warrant a separate function, so I called it from within `configure_container_network_devices`.
https://github.com/youki-dev/youki/pull/3163/commits/6de834c0aeeb0a391de0a169106f750bc2369382

crates/libcontainer/src/process/container_main_process.rs
nayuta723: ref: https://github.com/youki-dev/youki/pull/3163#discussion_r2273931004

crates/libcontainer/src/process/container_main_process.rs
nayuta723: fix: https://github.com/youki-dev/youki/pull/3163/commits/6de834c0aeeb0a391de0a169106f750bc2369382

crates/libcontainer/src/process/init/process.rs
nayuta723: https://github.com/youki-dev/youki/pull/3163/commits/6de834c0aeeb0a391de0a169106f750bc2369382

crates/libcontainer/src/utils.rs
nayuta723: fix: https://github.com/youki-dev/youki/pull/3163/commits/bb313743a782f06a09f65e25da5094e709bf1a9c

nayuta723:
@utam0k 
Please take a moment to review this when you have time.

scripts/contest.sh
```diff
@@ -23,10 +23,18 @@ if [ ! -f ${ROOT}/bundle.tar.gz ]; then
 fi
 touch ${LOGFILE}
 
+# TODO: This is a temporary change to exclude net_devices test for non-youki runtimes
```
saku3: You can skip tests individually by using the `is_runtime_runc` function within each test.

tests/contest/contest/src/tests/net_devices/net_devices_test.rs
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    if let Err(e) = cleanup_netns(&netns_name) {
+        return TestResult::Failed(anyhow!("Failed to cleanup netns: {}", e));
+    }
+
+    TestResult::Passed
+}
+
+pub fn get_net_devices_test() -> TestGroup {
```
saku3, line 474: If you want to skip tests when running with runc, you can use the `is_runtime_runc` function together with  `ConditionalTest`.

Please refer to:
https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/tests/fd_control/mod.rs#L94

tests/contest/contest/src/tests/net_devices/net_devices_test.rs
```diff
@@ -12,45 +11,72 @@ use test_framework::{ConditionalTest, TestGroup, TestResult, test_result};
...
 }
 
 pub fn create_unique_device_name(prefix: &str) -> String {
-    let count = DEVICE_COUNTER.fetch_add(1, Ordering::SeqCst);
-    format!("{}-{}", prefix, count)
+    let random_part: u16 = rand::random();
+    format!("{}{}", prefix, random_part)
 }
 
 fn create_netns(name: &str) -> Result<()> {
```
nayuta723, line 21: WIP: Mount propagation becomes private, resulting in files being inaccessible from the host side and failure to delete the netns.

nayuta723:
@saku3 
Thank you for your review. I have corrected the relevant points.

nayuta723:
@utam0k 
I'd like to check the status of the PR. Is my understanding correct that there's something blocking, and the review has been deferred/pushed back?

utam0k:
@nayuta723 Sorry for the delay 🙇  You're correct. I have a pretty big thing I need to wrap up that's been taking up my time.
That should be finished this week, so please bear with me a little longer.

Again, I apologize for holding up your great contribution.

nayuta723:
@utam0k 
No worries at all. Focus on wrapping up that big thing—that's the priority! I appreciate the update and happy to wait until you're ready.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +238,55 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+        // so we can use the process's net ns path directly.
+        let default_ns_path = PathBuf::from(format!("/proc/{}/ns/net", init_pid.as_raw()));
+        let ns_path = net_ns.path().as_deref().unwrap_or(&default_ns_path);
+
+        // If moving any of the network devices fails, we return an error immediately.
+        // The runtime spec requires that the kernel handles moving back any devices
+        // that were successfully moved before the failure occurred.
+        // See: https://github.com/opencontainers/runtime-spec/blob/27cb0027fd92ef81eda1ea3a8153b8337f56d94a/config-linux.md#namespace-lifecycle-and-container-termination
+        if let Some(devices) = linux.net_devices() {
+            main_receiver.wait_for_network_setup_ready()?;
```
utam0k: Is this waiting for the network namespace of init prcess is setup? What is network setup?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -229,6 +238,55 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+        // the container init process has already joined the provided net namespace,
+        // so we can use the process's net ns path directly.
+        let default_ns_path = PathBuf::from(format!("/proc/{}/ns/net", init_pid.as_raw()));
+        let ns_path = net_ns.path().as_deref().unwrap_or(&default_ns_path);
+
+        // If moving any of the network devices fails, we return an error immediately.
+        // The runtime spec requires that the kernel handles moving back any devices
+        // that were successfully moved before the failure occurred.
+        // See: https://github.com/opencontainers/runtime-spec/blob/27cb0027fd92ef81eda1ea3a8153b8337f56d94a/config-linux.md#namespace-lifecycle-and-container-termination
+        if let Some(devices) = linux.net_devices() {
```
utam0k: We can immidiately return before getting `ns_path`.

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+    tracing::debug!(
+        "attaching network device {} to network namespace {}",
+        name,
+        netns_path.display()
+    );
+
+    let mut link_client = LinkClient::new(create_network_client())?;
+    let mut addr_client = AddressClient::new(create_network_client())?;
+
+    let netns_file = File::open(netns_path)?;
```
utam0k: Please pass `netns_file` to dev_change_net_namespace() because the caller may want to validate whether the netns_path is real or not.

crates/libcontainer/src/network/link.rs
```diff
@@ -0,0 +1,477 @@
...
+    /// * `name` - The name of the network interface to retrieve
+    ///
+    /// # Returns
+    ///
+    /// A Result containing either the LinkMessage for the interface or an error
+    pub fn get_by_name(&mut self, name: &str) -> Result<LinkMessage> {
+        let mut message = LinkMessage::default();
+        message
+            .attributes
+            .push(LinkAttribute::IfName(name.to_string()));
```
utam0k, line 77: I prefer below:
```suggestion
        let mut message = LinkMessage::default()
            .attributes
            .push(LinkAttribute::IfName(name.to_string()));
```

crates/libcontainer/src/network/link.rs
```diff
@@ -0,0 +1,477 @@
...
+        message.header.index = index;
+
+        let mut req = NetlinkMessage::from(RouteNetlinkMessage::SetLink(message));
+        if let NetlinkPayload::InnerMessage(RouteNetlinkMessage::SetLink(ref mut link)) =
+            req.payload
+        {
+            link.header.change_mask |= LinkFlags::Up;
+            link.header.flags |= LinkFlags::Up;
+        }
+        req.header.flags = NLM_F_REQUEST | NLM_F_ACK | NLM_F_EXCL | NLM_F_CREATE;
```
utam0k: These flags are too complex to understand. Could you leave some comments to make them easier to understand?

crates/libcontainer/src/process/init/process.rs
```diff
@@ -879,6 +895,63 @@ fn sync_seccomp(
...
+
+    let addrs_map = init_receiver.wait_for_move_network_device()?;
+    for (name, net_dev) in net_device {
+        if let Some(serialize_addrs) = addrs_map.get(name) {
+            // Get the device's final name (use configured name if provided, otherwise use original name)
+            let new_name = net_dev
+                .name()
+                .as_ref()
+                .filter(|d| !d.is_empty())
+                .map_or(name.as_str(), |d| d);
```
utam0k: This logic is repeated at this link.
https://github.com/youki-dev/youki/pull/3163/files#diff-b8178a1348515af438aa0cd76216c344d312ea2292626d30c50b749862471d26R35-R39

crates/libcontainer/src/process/init/process.rs
```diff
@@ -879,6 +895,63 @@ fn sync_seccomp(
...
+                tracing::error!(?err, "failed to create address client");
+                err
+            })?;
+
+            // Get the device index
+            let ns_link = link_client.get_by_name(new_name).map_err(|err| {
+                tracing::error!(?err, "failed to get device by name: {}", new_name);
+                err
+            })?;
+            let ns_index = ns_link.header.index;
```
utam0k: `serialize_addrs` already includes `ns_index`. Can we utilize it?

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    let serialize_addrs: Vec<SerializableAddress> =
+        addrs.iter().map(SerializableAddress::from).collect();
+
+    Ok(serialize_addrs)
+}
+
+/// Core logic for setting up addresses in the new namespace
+/// This function is extracted to make it testable without system calls
+pub fn setup_addresses_in_namespace(
```
utam0k:
```suggestion
/// Core logic for setting up addresses in the new network namespace
/// This function is extracted to make it testable without system calls
pub fn setup_addresses_in_network_namespace(
```

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+    addrs: Vec<SerializableAddress>,
+    new_name: &str,
+    ns_index: u32,
+    addr_client: &mut AddressClient,
+) -> Result<()> {
+    // Re-add the original IP addresses to the interface in the new namespace.
+    // The kernel removes IP addresses when an interface is moved between network namespaces.
+    for addr in addrs {
+        let addr = AddressMessage::from(&addr);
```
utam0k:
```suggestion
    for addr: AddressMessage in addrs.iter().map(AddressMessage::from) {
```

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+        );
+        let mut ip_opts = None;
+        let mut flags_opts = None;
+        // Only move IP addresses with global scope because those are not host-specific, auto-configured,
+        // or have limited network scope, making them unsuitable inside the container namespace.
+        // Ref: https://www.ietf.org/rfc/rfc3549.txt
+        if addr.header.scope != AddressScope::Universe {
+            tracing::debug!(
+                "skipping address {:?} from network device {}",
+                addr.clone(),
```
utam0k:
```suggestion
                addr,
```

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+            "processing address {:?} from network device {}",
+            addr.clone(),
+            new_name
+        );
+        let mut ip_opts = None;
+        let mut flags_opts = None;
+        // Only move IP addresses with global scope because those are not host-specific, auto-configured,
+        // or have limited network scope, making them unsuitable inside the container namespace.
+        // Ref: https://www.ietf.org/rfc/rfc3549.txt
+        if addr.header.scope != AddressScope::Universe {
```
utam0k: Shouldn't this address be sent from the main process in the first place?

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+            );
+            continue;
+        }
+        for attr in &addr.attributes {
+            match attr {
+                AddressAttribute::Flags(flags) => flags_opts = Some(*flags),
+                AddressAttribute::Address(ip) => ip_opts = Some(*ip),
+                _ => {}
+            }
+        }
```
utam0k: What if `addr.attributes`has severl AddressAttribute::Address or AddressAttribute::Flags?

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        // Only move permanent IP addresses configured by the user, dynamic addresses are excluded because
+        // their validity may rely on the original network namespace's context and they may have limited
+        // lifetimes and are not guaranteed to be available in a new namespace.
+        // Ref: https://www.ietf.org/rfc/rfc3549.txt
+        if let Some(flag) = flags_opts {
+            if !flag.contains(AddressFlags::Permanent) {
+                tracing::debug!(
+                    "skipping address {:?} from network device {}",
+                    addr.clone(),
```
utam0k:
```suggestion
                    addr,
```

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        // Only move permanent IP addresses configured by the user, dynamic addresses are excluded because
+        // their validity may rely on the original network namespace's context and they may have limited
+        // lifetimes and are not guaranteed to be available in a new namespace.
+        // Ref: https://www.ietf.org/rfc/rfc3549.txt
+        if let Some(flag) = flags_opts {
+            if !flag.contains(AddressFlags::Permanent) {
+                tracing::debug!(
+                    "skipping address {:?} from network device {}",
```
utam0k: same as https://github.com/youki-dev/youki/pull/3163/files#r2506236583

crates/libcontainer/src/network/network_device.rs
```diff
@@ -0,0 +1,238 @@
...
+    let serialize_addrs: Vec<SerializableAddress> =
+        addrs.iter().map(SerializableAddress::from).collect();
+
+    Ok(serialize_addrs)
+}
+
+/// Core logic for setting up addresses in the new namespace
+/// This function is extracted to make it testable without system calls
+pub fn setup_addresses_in_namespace(
+    addrs: Vec<SerializableAddress>,
```
utam0k:
```suggestion
    addrs: &Vec<SerializableAddress>,
```

tests/contest/contest/src/tests/net_devices/net_devices_test.rs
```diff
@@ -0,0 +1,496 @@
...
+
+fn create_unique_netns_name(prefix: &str) -> String {
+    let random_part: u16 = rand::random();
+    format!("{}{}", prefix, random_part)
+}
+
+pub fn create_unique_device_name(prefix: &str) -> String {
+    let random_part: u16 = rand::random();
+    format!("{}{}", prefix, random_part)
+}
```
utam0k: Why did you separate these two functions?

tests/contest/contest/src/tests/net_devices/net_devices_test.rs
```diff
@@ -0,0 +1,496 @@
...
+        Err(e) => {
+            if let TestResult::Passed = result {
+                result = TestResult::Failed(anyhow!("Failed to check device2: {}", e));
+            }
+        }
+    }
+
+    // cleanup both devices regardless of test result
+    let _ = delete_dummy_device(&device_name1);
+    let _ = delete_dummy_device(&device_name2);
```
utam0k: How about using `Drop` trait?

tests/contest/runtimetest/src/tests.rs
```diff
@@ -1020,3 +1025,45 @@ pub fn validate_uid_mappings(spec: &Spec) {
...
+            req.serialize(&mut send_buf[..]);
+            socket.send(&send_buf[..], 0).unwrap();
+
+            let mut receive_buf = vec![0u8; 4096];
+            let n_received = socket.recv(&mut &mut receive_buf[..], 0).unwrap();
+            let bytes = &receive_buf[..n_received];
+            let rx_packet = <NetlinkMessage<RouteNetlinkMessage>>::deserialize(bytes).unwrap();
+
+            match rx_packet.payload {
+                NetlinkPayload::InnerMessage(RouteNetlinkMessage::NewLink(_link)) => {
```
utam0k: Shouldn't we check the address as well?

crates/libcontainer/src/network/link.rs
nayuta723: I implemented this in two steps because the chained notation on the second line would return `()` (the result of `.push()`) instead of the required `LinkMessage` struct, which I need to use in the subsequent lines.

crates/libcontainer/src/network/network_device.rs
nayuta723: Regarding `AddressAttribute::Flags`, it was more appropriate to use `AddressHeaderFlags`, so I made that change.

Also, concerning `AddressAttribute::Address`, reading the client-side implementations in Go, and C code suggests that they do not anticipate multiple instances of this attribute within a single message. 

However, since the **RFC does not explicitly state that only a single instance is allowed**, I am currently continuing my investigation into this constraint. For now, I have proceeded with the implementation by assuming the case where **multiplicity is not expected**.

Referenced Code:

* C (libnl): [https://github.com/thom311/libnl/blob/main/lib/route/addr.c#L238](https://github.com/thom311/libnl/blob/main/lib/route/addr.c#L238)
* Go (vishvananda/netlink): [https://github.com/vishvananda/netlink/blob/6e61cd407d1d66dd67fc66aaaf5c2a42e0614426/addr_linux.go#L224](https://github.com/vishvananda/netlink/blob/6e61cd407d1d66dd67fc66aaaf5c2a42e0614426/addr_linux.go#L224)

crates/libcontainer/src/network/network_device.rs
nayuta723: Is the implementation below suitable?
https://github.com/youki-dev/youki/pull/3163/commits/6eeefc7d9a7102be92ca93deb0c59ee785135f68

crates/libcontainer/src/process/init/process.rs
nayuta723: I have implemented the duplicate implementation into a function.
https://github.com/youki-dev/youki/pull/3163/commits/4ebd14ebbefc5ab400721f27859e8fdbc99969ec

crates/libcontainer/src/network/link.rs
nayuta723: Is the following sufficient?
[a07d087](https://github.com/youki-dev/youki/pull/3163/commits/a07d0876b90c2808037a2567105dcd712ff6bc48)

crates/libcontainer/src/process/container_main_process.rs
nayuta723: I commented here.
https://github.com/youki-dev/youki/pull/3163/commits/6eeefc7d9a7102be92ca93deb0c59ee785135f68#diff-4c3bee02d644d36778359cdb85ba23c194655461d95e34c298659d4633c2459eR269-R271

crates/libcontainer/src/network/network_device.rs
nayuta723: Regarding the current issue, since moving some SerializableAddress processing to the main process has enabled us to reduce the data required for inter-process communication, how about revising it as an intermediate type (a structure containing only prefix_len and IpAddr types)?

utam0k: That would be great.

crates/libcontainer/src/network/network_device.rs
utam0k: Thank you for your investigation. I believe it is reasonable to leave a comment at this time.

crates/libcontainer/src/process/init/process.rs
utam0k: Perfect.

utam0k:
Overall, it looks great! I replied to several comments.

crates/libcontainer/src/network/network_device.rs
nayuta723: I have addressed it.
09fd2db9b6eb618b36664de1e9520be8db249bdc

crates/libcontainer/src/network/network_device.rs
nayuta723: I left a comment.
[a239c34](https://github.com/youki-dev/youki/pull/3163/commits/a239c34629d144cb8d3dedce2c7d7d22205ad3a7)

nayuta723:
@utam0k 
Thank you! I've finished revising the comments section, so please take a look when you have a moment.

crates/libcontainer/src/process/init/process.rs
```diff
@@ -294,6 +298,18 @@ pub fn container_init_process(
         InitProcessError::SyscallOther(err)
     })?;
 
+    // Setup some operations in the network namespace.
+    // This is done here before dropping capabilities because we need to be able to add IP addresses to the device
+    // and set up the device.
+    if let Some(network_devices) = ctx.linux.net_devices() {
```
utam0k, line 304: Please case if network_device is empty because AFAI it leads to a deadlock.

crates/libcontainer/src/network/wrapper.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl Default for ClientWrapper {
+    fn default() -> Self {
+        if cfg!(test) {
+            ClientWrapper::Fake(FakeNetlinkClient::new())
+        } else {
+            ClientWrapper::Client(
+                NetlinkClient::new().unwrap_or_else(|_| panic!("Failed to create NetlinkClient")),
```
utam0k: Plase avoid using panic!

utam0k (commented):
last two comments 🙏

nayuta723:
@utam0k 
I've finished revising the comments section. Please take a look when you have a moment!

PR #3167, opened 2025-05-12
fix: remove println statements from contest tests

YJDoc2:
## Description
This cleans up the logs printed by the contest tests when running. Originally there were several debug-type logs in format of println, however I think in tests, we should only have the error logs indicating the failure. If someone is working on tests, they can add debug logs as they like, but remove before merge. This makes the failure o/p clear to understand when any test fails. Right now, the actual failure logs are mixed up with the debug logs, making it harder to pinpoint failing test.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Additional Context
There is also an issue with ns_itype_test where the tempdir it creates is not removed because it mounts it to actual root (expected behavior). This is same with runc, so still working on fixing it ; but opened this in meantime.


utam0k (approved):
+1

PR #3174, opened 2025-05-29
fix: youki does not conform with OCI runtime spec 

CheatCodeSam:
## Description

Hello, there was two issues here with nerdctl. The first one was the one described in the initial issue, which I fixed.

The second one had to do with nerdctl setting both `linux.resources.cpu.realtimeRuntime` and `linux.resources.cpu.realtimePeriod` to zero in the config.json despite youki trying to use cgroupsv2. This would cause a `V2CpuControllerError` to be thrown when the container is being created.

What I did is just check if cgroups2 is being used, and those two are set, if they are both set to zero then just continue as normal.

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [ ] Added new unit tests
- [ ] Added new integration tests
- [ ] Ran existing test suite
- [X] Tested manually (please provide steps)
I was able to run an alpine image using nerdctl and containerd after applying the fixes.

## Related Issues

Fixes #2994 

## Additional Context

N/A

Issue #2994, opened 2024-11-14
[Bug]: youki does not conform with OCI runtime spec


eternalphane:
### Bug Description

`builder_impl::ContainerBuilderImpl::run_container` calls `hooks::run_hook(hooks.create_runtime(), ...)` before container main process is created, while container is still in `creating` state. However, the spec states that

> * **`status`** (string, REQUIRED) is the runtime state of the container.
>     The value MAY be one of:
> 
>     * `creating`: the container is being created **(step 2 in the [lifecycle](#lifecycle))**
>     * `created`: the runtime has finished the [create operation](#create) **(after step 2 in the [lifecycle](#lifecycle))**, and the container process has neither exited nor executed the user-specified program

and

> ## Lifecycle
> The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
> 
> ...
> **2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).**
> ...
> **4. The [`createRuntime` hooks](config.md#createRuntime-hooks) MUST be invoked by the runtime.**

which indicates that `create_runtime` hook should only be executed when container is in `created` state.

---

I found this bug when trying to create a rootless container in nerdctl with youki as runtime. youki failed and complains:
```
error in executing command: hook command exited with non-zero exit code: 1
```
which is in turn caused by a fatal error in nerdctl `createRuntime` hook:
```
both state.Pid and the netNs annotation are unset
```

### Steps to Reproduce

1. install nerdctl v2.0.0 rootless and enable debug logging in `$XDG_CONFIG_HOME/nerdctl/nerdctl.toml`
   ```toml
   debug = true
   debug_full = true
   ```
2. install youki main with dev profile
   ```
   cargo install --git https://github.com/youki-dev/youki.git youki --locked --features v1,v2,systemd,seccomp --debug
   ```
3. enable youki debug logging in `$XDG_CONFIG_HOME/systemd/user/containerd.service.d/youki-debug.conf`
   ```
   [Service]
   Environment=RUST_BACKTRACE=full
   Environment=YOUKI_LOG_LEVEL=debug
   ```
4. create a container with nerdctl
   ```
   nerdctl run --rm -it --runtime=youki alpine
   ```
5. the following error occurs (note that I modified the nerdctl source code to log the received state in `createRuntime` hook)
   ```
   DEBU[0000] stateDir: /run/user/1000/containerd-rootless
   DEBU[0000] RootlessKit detach-netns mode: true
[...]

PR #3180, opened 2025-06-06
fix: Ignoring CPU realtime on cgroupsv2 if set to zero

CheatCodeSam:
## Description
Requested to break up #3174 into two seperate issues by @YJDoc2.

This fixes an auxiliary issue of #2994, where `nerdctl` will set both `linux.resources.cpu.realtimeRuntime` and `linux.resources.cpu.realtimePeriod` to zero in the config.json despite youki trying to use cgroupsv2. This would cause a V2CpuControllerError to be thrown when the container is being created.

My solution is just to ignore them if using cgroupsv2 and they are both set to zero.

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [X] Added new unit tests
- [X] Ran existing test suite
- [X] Tested manually (please provide steps)

Once the initial runtime spec was fixed, this was able to run with no problems on nerdctl.

## Related Issues

Fixes part of  #2994.

### DIFF
--- a/crates/libcgroups/src/v2/cpu.rs
+++ b/crates/libcgroups/src/v2/cpu.rs
@@ -86,7 +86,12 @@ impl StatsProvider for Cpu {
 impl Cpu {
     fn apply(path: &Path, cpu: &LinuxCpu) -> Result<(), V2CpuControllerError> {
         if Self::is_realtime_requested(cpu) {
-            return Err(V2CpuControllerError::RealtimeV2);
+            let realtime_runtime = cpu.realtime_runtime();
+            let runtime_period = cpu.realtime_period();
+
+            if !matches!((realtime_runtime, runtime_period), (Some(0), Some(0))) {
+                return Err(V2CpuControllerError::RealtimeV2);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


utam0k (changes requested):
Please add the unit test for your change 🙏

CheatCodeSam:
Tests have been added

crates/libcgroups/src/v2/cpu.rs
```diff
@@ -369,4 +374,36 @@ mod tests {
         let actual = fs::read_to_string(burst_file).expect("read burst file");
         assert_eq!(actual, expected.to_string());
     }
+
+    #[test]
```
utam0k, line 378: Add the unit test where both values are zero.

CheatCodeSam: Just set this.

utam0k (approved):
Thanks!

Issue #3178, opened 2025-06-03
Enhancing Compatibility with runc


saku3:
I’m interested in improving the compatibility between `youki` and `runc`.
Here are a few areas I’d like to focus on:

### CLI Command Compatibility

* The `exec` command in [`youki`](https://github.com/youki-dev/youki/blob/main/crates/youki/src/commands/exec.rs#L12) does not yet support some arguments available in [`runc`](https://github.com/opencontainers/runc/blob/main/man/runc-exec.8.md), such as `--preserve-fds` and `--ignore-paused` .
* The `update` command seems to support only `--pids-limit` and `-r`.
* I haven’t conducted an exhaustive review, so there may be differences in other commands as well.
* One point to consider is that commands like `exec` are not defined in the OCI runtime-spec.
* I plan to identify and implement the missing CLI options to align more closely with runc.

### Testing

* While youki already has integration tests using oci-runtime-tools [test](https://github.com/youki-dev/youki/issues/361), runc has its own [integration tests using bats](https://github.com/opencontainers/runc/tree/main/tests/integration#runc-integration-tests).
* I’d like to make it possible to run these bats-based tests on youki.
* Running these tests and analyzing failures could help improve compatibility.

If the maintainers are okay with this direction, I’d like to proceed with these efforts.
This may involve creating multiple PRs or sub-issues to track the work incrementally.

CheatCodeSam:
I would love to help with this.

YJDoc2:
Hey @saku3 @CheatCodeSam Thanks for your interest in doing this. This would be quite helpful. I'd suggest the following approach for this now -
1. I have some changes for update command locally but they are not usable. Let me open a draft PR, and maybe you can use it as a starting point to fix the update command. 
2. We have https://github.com/youki-dev/youki/issues/1305 opened for the runc tests. What I'll suggest is @saku3 should create a CI to run runc tests , but make it only manually triggered for now, not part of CI on PR. Once that is done, we can check what tests are failing, and you can divide the tests between yourselves and fix.
3. I remember something related to exec's preserve fds being discussed in past. Let me investigate more, but wait for that before starting work on it.

wdyt?

saku3:
@YJDoc2 
Thank you for your comment.

I apologize for overlooking issue #1305.
I’d like to proceed with the approach you suggested.

@CheatCodeSam 

I’m planning to first look into whether I can set up a CI that runs `runc`'s integration tests.
[...]

PR #3182, opened 2025-06-08
setup runc integration test

saku3:
## Description

This change allows `runc` integration tests to be executed using the `youki` runtime in CI.

### .github/workflows/runc_integration_tests.yaml

This file configures the CI workflow to run the `runc` integration tests. It performs the following steps:

- Build `youki`
- Run `runc` integration tests:
  - Install necessary packages
  - Set up `bats`
  - Execute tests by running `runc_integration_test.sh`

### scripts/runc_integration_test.sh

This script performs the following:

- Places the `youki` binary at the root of the `runc` repository, since `bats` expects the runtime binary to be in that location:
  https://github.com/opencontainers/runc/blob/main/tests/integration/helpers.bash#L15
- Runs tests using the `bats` command, targeting `*.bats` files under `tests/integration`:
  https://github.com/opencontainers/runc/tree/main/tests/integration
- To prevent test hangs in github actions, a mechanism was added to selectively skip individual test cases by injecting `skip` commands into `.bats` files:
  https://bats-core.readthedocs.io/en/stable/writing-tests.html#skip-easily-skip-tests

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Test updates
- [x] CI/CD related changes
- [ ] Other:

## Testing

 - [ ] Added new unit tests
 - [ ] Ran existing test suite

 I have confirmed that all tests pass with runc by executing the following command.(By specifying runc as the first argument, you can run the tests using the runc runtime.)
```
scripts/runc_integration_test.sh runc
```

## Related Issues

Part of #3178

### DIFF
--- a/.github/workflows/runc_integration_tests.yaml
+++ b/.github/workflows/runc_integration_tests.yaml
@@ -1,0 +1,49 @@
+name: 🧪 Runc integration test
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+  runc-integration-tests:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install just
+        uses: taiki-e/install-action@just
+      - name: install runc and youki  requirements
+        run: |
+          sudo apt update
+          sudo apt -y install libseccomp-dev sshfs uidmap \
+               pkg-config libsystemd-dev build-essential libelf-dev \
+               libclang-dev libssl-dev
+
+      - name: Build youki
+        run: just youki-release
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.23.x'
+          cache: true
+
+      - name: Add the permission to run
+        run: chmod +x ./youki
+      - name: Setup Bats and bats libs
+        uses: bats-core/bats-action@3.0.1
+        with:
+          bats-version: 1.9.0
+          support-install: false
+          assert-install: false
+          detik-install: false
+          file-install: false
+      - name: Allow userns for runc
+        # https://github.com/opencontainers/runc/pull/4286
+        run: |
+          sed "s;^profile runc /usr/sbin/;profile runc-test $(pwd)/tests/runc/src/github.com/opencontainers/runc/;" < /etc/apparmor.d/runc | sudo apparmor_parser
+      - name: Runc integration test
+        run: tests/runc/runc_integration_test.sh
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,8 @@
 	path = tests/oci-runtime-tests/src/github.com/opencontainers/runtime-tools
 	url = https://github.com/opencontainers/runtime-tools.git
     ignore = dirty
+
+[submodule "tests/runc/src/github.com/opencontainers/runc"]
+	path = tests/runc/src/github.com/opencontainers/runc
+	url = https://github.com/opencontainers/runc.git
+    ignore = dirty
--- a/tests/runc/runc_integration_test.sh
+++ b/tests/runc/runc_integration_test.sh
@@ -1,0 +1,30 @@
+#!/bin/bash -u
+
+RUNTIME=${1:-./youki}
+ROOT=$(git rev-parse --show-toplevel)
+RUNC_DIR="${ROOT}/tests/runc/src/github.com/opencontainers/runc"
+PATTERN_FILE="${ROOT}/${2:-tests/runc/runc_test_pattern}"
+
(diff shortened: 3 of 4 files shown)


saku3:
I might make further changes, but I’d appreciate any advice on whether the current direction is correct.

.github/workflows/runc_integration_tests.yaml
utam0k, line 1: It should be integrated with e2e.yaml.

YJDoc2:
Hey so a couple of points here,
1. code-wise looks ok to me apart from utam0k's comment
2. I see in the run you have linked that there are lot of debug logs from Youki, even though in PR we are compiling youki in release mode, can you check that? We should not have debug logs in CI as it will clutter o/p.
3. For now, we don't need to run it on both arch, only amd64 with gnu should be enough. Once we have all failing tests triaged, and we are adding it as a standard CI, then we can use both archs.
4. I saw in the run you linked that all tests were run in 3-4 minutes. Are those all the runc tests or you ran just a subset/fail fast mode?

Thanks for your efforts :)

saku3:
@utam0k @YJDoc2 

Thank you for the review!

> code-wise looks ok to me apart from utam0k's comment

I’ve updated it to run using e2e.yaml.

> I see in the run you have linked that there are lot of debug logs from Youki, even though in PR we are compiling youki in release mode, can you check that? We should not have debug logs in CI as it will clutter o/p.

The debug logs are output as part of the tests, specifically when `--debug` is passed as an option to `runc` during execution.


> For now, we don't need to run it on both arch, only amd64 with gnu should be enough. Once we have all failing tests triaged, and we are adding it as a standard CI, then we can use both archs.

I’ve updated the configuration accordingly.

> I saw in the run you linked that all tests were run in 3-4 minutes. Are those all the runc tests or you ran just a subset/fail fast mode?

Currently, it's set up to run all of the tests (although a few known-hanging tests are skipped within runc_integration_test.sh).
However, the tests sometimes hang randomly during execution.
 I'm not yet sure how best to handle those hangs.
 In some runs, like the one below, all tests do complete:
 https://github.com/saku3/youki/actions/runs/15792161456/job/44519468461?pr=13

I would appreciate any reviews or advice you could provide.

scripts/runc_integration_test.sh
```diff
@@ -0,0 +1,64 @@
...
+RUNC_TEST_DIR="${ROOT}/tests/runc/src/github.com/opencontainers/runc/tests/integration"
+
+if [[ "$RUNTIME" == "youki" ]]; then
+  if [[ ! -x ./youki ]]; then
+    echo "youki binary not found"
+    exit 1
+  fi
+  cp ./youki "$RUNC_DIR/runc"
+  chmod +x "$RUNC_DIR/runc"
+fi
```
utam0k: How about passing a path of a binary for runtime?

YJDoc2:
@saku3 , the CI seems to be failing. Or is it just a flake?

Also @utam0k , I think for now it is better if we put these tests in a separate CI job like podman and run it daily. Currently these are still failing, so maybe we will fix them first and then add them as a normal job in our CI. wdyt?

saku3:
For now, I’ve set all failing tests to be skipped.
However, even with this change, the tests occasionally hang during execution. The reason for the hang is still under investigation.
It's still difficult to integrate this into the CI at this stage.

YJDoc2:
Hey @saku3 , thanks for the changes. Let us do this - can we make the CI into a separate CI workflow like podman, I think similar to what you had done originally? That should be the only change from my side, and I'll merge the PR after that. Apologies it is taking so long from our side to get this finished :sweat:

saku3:
@YJDoc2 
Thank you for the review!
Sorry for the late response, but I’ve made the changes.

The updates since the last PR are as follows:

* As you pointed out, separated the tests, similar to the Podman CI workflow
* Prepared a file to list the test cases

saku3:
Thanks! After your comments, I noticed a few parts I’d like to revise.
I’d appreciate it if you could hold off on merging for a little while.

YJDoc2:
Sure @saku3 , I'll wait on merging. Let me know when you're ready :)

saku3:
@YJDoc2 

Sorry for the delay.

* I prefixed failing or hanging tests with `[skip]` at the beginning of the line, so these tests will be skipped.
* For tests that hang, I added the following comment:
  `# This test hangs on Youki`

If this looks good (no concerns), please merge.

YJDoc2 (approved):
Looks great to me!

One more update for future we can do is have a script which will scan the runc dir and generate the test_pattern file, not sure how we can handle skip there, but that way we'll have easier time to keep up with runc when they add a new test. 

Thanks for the PR!

PR #3200, opened 2025-07-20
fix(3197): fix youki version command Part of Enhancing Compatibility with runc

### LINKED ISSUE #3197: youki version command

The short `version` command for `youki` is different from that of `runc`.

youki 
The "v" is capitalized.
```
$ youki -V
youki version 0.5.4
commit: 0.5.4-0-6a1a32a624e526ff9e69eee5bf48a1b7347cd6d9

$ youki -v
error: unexpected argument '-v' found

Usage: youki [OPTIONS] <COMMAND>

For more information, try '--help'.
```

runc
```
$ runc -v
runc version 1.3.0-rc.1+dev
commit: v1.3.0-rc.1-153-gb04031d7
spec: 1.2.1
go: go1.23.0
libseccomp: 2.5.5
```

tommady:
## Description

Fix Youki version command
This is Part of Enhancing Compatibility with runc

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

```bash
./target/debug/youki -v                                                                                                                                  
youki version: 0.5.4                                                                                                                                          
commit: 0.5.4-81806f8a9175bbed7cb176cb66f527ff89a187ac    
```

## Related Issues

Fixes #3197

### DIFF
--- a/crates/youki/src/commands/info.rs
+++ b/crates/youki/src/commands/info.rs
@@ -27,6 +27,7 @@ pub fn info(_: Info) -> Result<()> {
 }
 
 /// print Version of Youki in Moby compatible format
+/// https://github.com/moby/moby/blob/65cc84abc522a564699bb171ca54ea1857256d10/daemon/info_unix.go#L280
 pub fn print_youki() {
     println!("youki version: {}", env!("CARGO_PKG_VERSION"));
     println!(
--- a/crates/youki/src/commands/info.rs
+++ b/crates/youki/src/commands/info.rs
@@ -26,10 +26,14 @@ pub fn info(_: Info) -> Result<()> {
     Ok(())
 }
 
-/// print Version of Youki
+/// print Version of Youki in Moby compatible format
 pub fn print_youki() {
-    println!("{:<18}{}", "Version", env!("CARGO_PKG_VERSION"));
-    println!("{:<18}{}", "Commit", env!("VERGEN_GIT_SHA"));
+    println!("youki version: {}", env!("CARGO_PKG_VERSION"));
+    println!(
+        "commit: {}-{}",
+        env!("CARGO_PKG_VERSION"),
+        env!("VERGEN_GIT_SHA")
+    );
 }
 
 /// Print Kernel Release, Version and Architecture
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -142,7 +142,7 @@ fn main() -> Result<()> {
         Some(SubCommand::Completion(completion)) => {
             commands::completion::completion(completion, &mut app)
         }
-        None => Opts::command()
+        None => app
             .print_help()
             .map_err(|e| anyhow::anyhow!("failed to print help: {e}")),
     };
--- a/crates/youki/src/main.rs
+++ b/crates/youki/src/main.rs
@@ -7,7 +7,7 @@ mod rootpath;
 mod workload;
 
 use anyhow::{Context, Result};
-use clap::{crate_version, CommandFactory, Parser};
+use clap::{CommandFactory, Parser};
 use libcontainer::syscall::syscall::create_syscall;
 use liboci_cli::{CommonCmd, GlobalOpts, StandardCmd};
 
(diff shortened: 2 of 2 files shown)


utam0k:
@saku3 May I ask you to review this P?

saku3:
Sure!

crates/youki/src/main.rs
```diff
@@ -145,10 +138,13 @@ fn main() -> Result<()> {
             CommonCmd::Update(update) => commands::update::update(update, root_path),
         },
 
-        SubCommand::Info(info) => commands::info::info(info),
-        SubCommand::Completion(completion) => {
+        Some(SubCommand::Info(info)) => commands::info::info(info),
+        Some(SubCommand::Completion(completion)) => {
             commands::completion::completion(completion, &mut app)
         }
+        None => Opts::command()
```
saku3:
```suggestion
        None => app.print_help().map_err(|e| anyhow::anyhow!(":{e}"))
```

saku3:
Thanks. I think this modification looks good.

crates/youki/src/commands/info.rs
```diff
@@ -26,10 +26,14 @@ pub fn info(_: Info) -> Result<()> {
     Ok(())
 }
 
-/// print Version of Youki
+/// print Version of Youki in Moby compatible format
```
utam0k, line 29: Don't remove the reference.
https://github.com/moby/moby/blob/65cc84abc522a564699bb171ca54ea1857256d10/daemon/info_unix.go#L280

utam0k:
Could you add the tests to keep this compatibility with runc?

crates/youki/src/commands/info.rs
tommady: Added the link you provided.

tommady:
> Could you add the tests to keep this compatibility with runc?

sorry I am not quite sure how, could you guide me?

I thought the issue's goal is to let youki accept `-v` instead of `-V`

YJDoc2:
>sorry I am not quite sure how, could you guide me?
> I thought the issue's goal is to let youki accept -v instead of -V

Hey, so that is correct, utam0k is asking to add test, so we can be safe against regression with this. The test for this would be running youki with `-v`  . So  in the test we can run the youki with `-v` flag and make sure that it does not error, and if we want the o/p in a certain format, we can also check that. Currently the best place for this kind of test would be in the e2e tests, but this seems a bit out-of-context for that (e2e are focused on behavioral , this seems more like an interface test) . Maybe for now we can have a simple sh script like what we have for rootless podman test, and check in that?

Also do should we support `-V` as well for backward compatibility?
Can you add a comment in code for why we set `disable_version_flag=true` ?

Thanks :)

tommady:
> >sorry I am not quite sure how, could you guide me?
> 
> > I thought the issue's goal is to let youki accept -v instead of -V
> 
> 
> 
> Hey, so that is correct, utam0k is asking to add test, so we can be safe against regression with this. The test for this would be running youki with `-v`  . So  in the test we can run the youki with `-v` flag and make sure that it does not error, and if we want the o/p in a certain format, we can also check that. Currently the best place for this kind of test would be in the e2e tests, but this seems a bit out-of-context for that (e2e are focused on behavioral , this seems more like an interface test) . Maybe for now we can have a simple sh script like what we have for rootless podman test, and check in that?
> 
> 
> 
> Also do should we support `-V` as well for backward compatibility?
> 
> Can you add a comment in code for why we set `disable_version_flag=true` ?
> 
> 
> 
> Thanks :)

I think you’ve highlighted two key points:
	1.	Should we maintain backward compatibility?
	2.	What should the testing strategy be?

These two directly affect your suggestions:
	1.	For e2e, shell script, and interface tests, they all rely on exact output matching. Without that, we lack a clear criterion for validation. However, I believe that level of thoroughness goes beyond the scope of this issue.
	2.	Regarding the disable_version_flag=true comment—whether or not we add it depends on the decision about backward compatibility.
	•	If we decide to support backward compatibility, the comment might not be necessary.
	•	If we don’t, this becomes a breaking change and should be mentioned in the release notes.
	•	Also, since it’s a Clap-only argument, I assumed it was already documented sufficiently.

In any case, I think we should keep this PR simple and focused, as requested by the original issue.

utam0k:
> 1. Should we maintain backward compatibility?

In my opinion, we don't need backward compatibility in this case. Not many people currently depend on this behavior.

> 2. What should the testing strategy be?

After reading YJDoc's comments, I reconsidered. I don't think it's necessary to test using new methods. I vote against no testing.

tommady:
> > 1. Should we maintain backward compatibility?
> 
> In my opinion, we don't need backward compatibility in this case. Not many people currently depend on this behavior.
> 
> > 2. What should the testing strategy be?
> 
> After reading YJDoc's comments, I reconsidered. I don't think it's necessary to test using new methods. I vote against no testing.

Do you think the manual test is OK for you?

utam0k:
> Do you think the manual test is OK for you?

Yes, I think, but wait for @YJDoc2

YJDoc2:
Hey, I'm also fine with manual test for this.

Issue #3209, opened 2025-07-26
In youki, the exec command only works when the container status is Running.


saku3:
Part of [Enhancing Compatibility with runc](https://github.com/youki-dev/youki/issues/3178)

In Youki, the `exec` command currently only works when the target container is in the `Running` state.

```sh
$ youki create -b tutorial/ testcon 

$ youki exec testcon sh
DEBUG youki: started by user 0 with ArgsOs { inner: ["./youki", "exec", "testcon", "sh"] }
ERROR libcontainer::container::tenant_builder: cannot exec as container status=Created
ERROR youki: error in executing command: failed to perform operation due to incorrect container status
exec failed : failed to perform operation due to incorrect container statu
```

In contrast, `runc` allows executing commands in a container even when it is in the `Created` state:

```sh
$ runc create -b tutorial/ testcon
$ runc exec testcon sh
```

Additionally, by using the `--ignore-paused` flag, it is possible to execute commands in a container that is in the `Paused` state.

tommady:
Hi, I'd like to take this, thank you.

tommady:
I think I can fix this in this issue's fixing PR also.
https://github.com/youki-dev/youki/issues/3207
WDYT?

because:
1. To avoid conflicts
2. Those two are somehow related because the behavior is not the same as runc

saku3:
From my perspective, the issue #3207 is about the command args, while this issue is about when exec is allowed, so the concerns were a bit different.
But if you can handle both at once, feel free to go ahead. Thanks!

tommady:
Sure, I'd be glad to handle them. Since I read the runc go code, I felt these two issues can be handled together.

saku3:
This has already been addressed in the following PR, so I’m closing this.

https://github.com/youki-dev/youki/pull/3210

PR #3210, opened 2025-07-29
fix(3207, 3209) Difference between the exec command in runc and youki

### LINKED ISSUE #3207: Difference between the exec command in runc and youki

There are differences between the `exec` command in `runc` and `youki`.

Please refer to the following:
https://github.com/opencontainers/runc/blob/main/man/runc-exec.8.md

The `exec` command in Youki currently does not support several options that are available in `runc`, such as `--ignore-paused` and `--preserve-fds`.

tommady:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3207 #3209

### DIFF
--- a/crates/libcontainer/src/container/tenant_builder.rs
+++ b/crates/libcontainer/src/container/tenant_builder.rs
@@ -450,14 +450,15 @@ impl TenantContainerBuilder {
 
     fn load_container_state(&self, container_dir: PathBuf) -> Result<Container, LibcontainerError> {
         let container = Container::load(container_dir)?;
-        if !container.can_exec()
-            && (container.status() == ContainerStatus::Paused && !self.ignore_paused)
-        {
-            tracing::error!(status = ?container.status(), "cannot exec as container");
-            return Err(LibcontainerError::IncorrectStatus(container.status()));
-        }
 
-        Ok(container)
+        match container.status() {
+            ContainerStatus::Running => Ok(container),
+            ContainerStatus::Paused if self.ignore_paused => Ok(container),
+            _ => {
+                tracing::error!(status = ?container.status(), "cannot exec: invalid container state");
+                Err(LibcontainerError::IncorrectStatus(container.status()))
+            }
+        }
     }
 
     fn adapt_spec_for_tenant(


tommady:
hi @saku3 
sorry for the very long developing time.
please do the review while you have time.
thanks.

saku3:
@tommady 
Thanks for creating the PR! 
It looks like the tests you added are failing. Could you fix them?

tommady:
> @tommady Thanks for creating the PR! It looks like the tests you added are failing. Could you fix them?

thanks for the mention.
but I am curious why that case only failed on gnu case but succeeded on the musl one.

I run the case on my local
```bash
 uname --all
Linux amd64-node-2 6.16.8-arch3-1 #1 SMP PREEMPT_DYNAMIC Mon, 22 Sep 2025 22:08:35 +0000 x86_64 GNU/Linux
```
it will succeed...
so I am wondering the reason why
could you help me do a re-run?

saku3:
Sorry, it looks like I don’t have permission to re-run the tests.
However, I also tried running it locally and got an error. (I don’t know the reason yet either.)
It looks like it’s failing due to a CapAmb mismatch, same as in CI.

```
$ just test-contest exec
/home/ubuntu/workspace/youki/scripts/build.sh -o /home/ubuntu/workspace/youki -r -c youki
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
    Finished `release` profile [optimized] target(s) in 0.26s
/home/ubuntu/workspace/youki/scripts/build.sh -o /home/ubuntu/workspace/youki -r -c contest
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
removed '/home/ubuntu/workspace/youki/contest'
   Compiling contest v0.1.0 (/home/ubuntu/workspace/youki/tests/contest/contest)
    Finished `release` profile [optimized] target(s) in 47.93s
removed '/home/ubuntu/workspace/youki/runtimetest'
    Finished `release` profile [optimized] target(s) in 0.16s
sudo /home/ubuntu/workspace/youki/scripts/contest.sh /home/ubuntu/workspace/youki/youki exec
# Start group exec
1 / 8 : capabilities_by_flag_test_case1 : ok
2 / 8 : capabilities_by_flag_test_case2 : not ok
        CapAmb unexpected output: Name: cat
Umask:  0022
State:  R (running)
Tgid:   3
Ngid:   0
Pid:    3
PPid:   0
TracerPid:      0
Uid:    0       0       0       0
Gid:    0       0       0       0
FDSize: 64
Groups: 0 
NStgid: 3
NSpid:  3
NSpgid: 3
NSsid:  3
Kthread:        0
VmPeak:     1308 kB
VmSize:     1308 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:       520 kB
VmRSS:       520 kB
RssAnon:               0 kB
RssFile:             520 kB
RssShmem:              0 kB
VmData:       36 kB
VmStk:       132 kB
VmExe:       884 kB
VmLib:         8 kB
VmPTE:        28 kB
VmSwap:        0 kB
HugetlbPages:          0 kB
CoreDumping:    0
THP_enabled:    1
untag_mask:     0xffffffffffffffff
Threads:        1
SigQ:   3/62216
SigPnd: 0000000000000000
ShdPnd: 0000000000000000
SigBlk: 0000000000000000
SigIgn: 0000000000001000
SigCgt: 0000000000000000
CapInh: 0000000000000001
CapPrm: 0000000400000021
CapEff: 0000000400000021
CapBnd: 0000000400000021
CapAmb: 0000000000000000
NoNewPrivs:     1
Seccomp:        0
Seccomp_filters:        0
Speculation_Store_Bypass:       vulnerable
SpeculationIndirectBranch:      always enabled
Cpus_allowed:   f
Cpus_allowed_list:      0-3
Mems_allowed:   00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000001
Mems_allowed_list:      0
voluntary_ctxt_switches:        6
nonvoluntary_ctxt_switches:     2
x86_Thread_features:
x86_Thread_features_locked:

3 / 8 : cgroup_test : ok
4 / 8 : ignore_paused_test : ok
5 / 8 : new_privileges_test : ok
6 / 8 : no_capabilities_test : ok
7 / 8 : preserve_fds_test : ok
8 / 8 : some_capabilities_test : ok
# End group exec
```

Did you base the tests on the following?

https://github.com/opencontainers/runc/blob/80486a207f54fb9e5e79ad2b43f11e675aff6371/tests/integration/capabilities.bats

tommady:
> Sorry, it looks like I don’t have permission to re-run the tests.
> 
> However, I also tried running it locally and got an error. (I don’t know the reason yet either.)
> 
> It looks like it’s failing due to a CapAmb mismatch, same as in CI.
> 
> 
> 
> ```
> 
> $ just test-contest exec
> 
> /home/ubuntu/workspace/youki/scripts/build.sh -o /home/ubuntu/workspace/youki -r -c youki
> 
> * FEATURES: <default>
> 
> * TARGET: x86_64-unknown-linux-gnu
> 
>     Finished `release` profile [optimized] target(s) in 0.26s
> 
> /home/ubuntu/workspace/youki/scripts/build.sh -o /home/ubuntu/workspace/youki -r -c contest
> 
> * FEATURES: <default>
> 
> * TARGET: x86_64-unknown-linux-gnu
> 
> removed '/home/ubuntu/workspace/youki/contest'
> 
>    Compiling contest v0.1.0 (/home/ubuntu/workspace/youki/tests/contest/contest)
> 
>     Finished `release` profile [optimized] target(s) in 47.93s
> 
> removed '/home/ubuntu/workspace/youki/runtimetest'
> 
>     Finished `release` profile [optimized] target(s) in 0.16s
> 
> sudo /home/ubuntu/workspace/youki/scripts/contest.sh /home/ubuntu/workspace/youki/youki exec
> 
> # Start group exec
> 
> 1 / 8 : capabilities_by_flag_test_case1 : ok
> 
> 2 / 8 : capabilities_by_flag_test_case2 : not ok
> 
>         CapAmb unexpected output: Name: cat
> 
> Umask:  0022
> 
> State:  R (running)
> 
> Tgid:   3
> 
> Ngid:   0
> 
> Pid:    3
> 
> PPid:   0
> 
> TracerPid:      0
> 
> Uid:    0       0       0       0
> 
> Gid:    0       0       0       0
> 
> FDSize: 64
> 
> Groups: 0 
> 
> NStgid: 3
> 
> NSpid:  3
> 
> NSpgid: 3
> 
> NSsid:  3
> 
> Kthread:        0
> 
> VmPeak:     1308 kB
> 
> VmSize:     1308 kB
> 
> VmLck:         0 kB
> 
> VmPin:         0 kB
> 
> VmHWM:       520 kB
> 
> VmRSS:       520 kB
> 
> RssAnon:               0 kB
> 
> RssFile:             520 kB
> 
> RssShmem:              0 kB
> 
> VmData:       36 kB
> 
> VmStk:       132 kB
> 
> VmExe:       884 kB
> 
> VmLib:         8 kB
> 
> VmPTE:        28 kB
> 
> VmSwap:        0 kB
> 
> HugetlbPages:          0 kB
> 
> CoreDumping:    0
> 
> THP_enabled:    1
> 
> untag_mask:     0xffffffffffffffff
> 
> Threads:        1
> 
> SigQ:   3/62216
> 
> SigPnd: 0000000000000000
> 
> ShdPnd: 0000000000000000
> 
> SigBlk: 0000000000000000
> 
> SigIgn: 0000000000001000
> 
> SigCgt: 0000000000000000
> 
> CapInh: 0000000000000001
> 
> CapPrm: 0000000400000021
> 
> CapEff: 0000000400000021
> 
> CapBnd: 0000000400000021
> 
> CapAmb: 0000000000000000
> 
> NoNewPrivs:     1
> 
> Seccomp:        0
> 
> Seccomp_filters:        0
> 
> Speculation_Store_Bypass:       vulnerable
> 
> SpeculationIndirectBranch:      always enabled
> 
> Cpus_allowed:   f
> 
> Cpus_allowed_list:      0-3
> 
> Mems_allowed:   00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000001
> 
> Mems_allowed_list:      0
> 
> voluntary_ctxt_switches:        6
> 
> nonvoluntary_ctxt_switches:     2
> 
> x86_Thread_features:
> 
> x86_Thread_features_locked:
> 
> 
> 
> 3 / 8 : cgroup_test : ok
> 
> 4 / 8 : ignore_paused_test : ok
> 
> 5 / 8 : new_privileges_test : ok
> 
> 6 / 8 : no_capabilities_test : ok
> 
> 7 / 8 : preserve_fds_test : ok
> 
> 8 / 8 : some_capabilities_test : ok
> 
> # End group exec
> 
> ```
> 
> 
> 
> Did you base the tests on the following?
> 
> 
> 
> https://github.com/opencontainers/runc/blob/80486a207f54fb9e5e79ad2b43f11e675aff6371/tests/integration/capabilities.bats
> 
> 

Interesting!
So your local can reproduce.

Yes I followed the test case the link you pasted.

saku3:
Sorry to keep you waiting. I’ll review it, so please give me a little more time.

saku3:
I’ve figured out the cause of the error above.

In the implementation of `drop_privileges`, when setting ambient capabilities, if an error occurs midway, the function exits without setting the remaining capabilities.
Also, because we loop over a `HashSet` when setting the capabilities, the order in which capabilities are applied is non-deterministic, so the final result is not stable.

As a reproduction method, if you run `get_test_capabilities_by_flag_case2` multiple times, you’ll see that the value of `CapAmb` sometimes becomes `0000000000000001` and other times `0000000000000000`.

Here is the reproduction procedure:

Start youki with the following `capabilities` section in `config.json`:

```json
"capabilities": {
  "bounding": [
    "CAP_KILL",
    "CAP_CHOWN",
    "CAP_SYSLOG"
  ],
  "effective": [
  ],
  "inheritable": [
    "CAP_CHOWN"
  ],
  "permitted": [
    "CAP_KILL",
    "CAP_CHOWN"
  ],
  "ambient": [
    "CAP_CHOWN"
  ]
},
```

then

```
$ youki run -b tutorial/ container-1
$ youki exec  --cap=CAP_SYSLOG container-1 cat /proc/self/status
```

In this test case, we newly add the `CAP_SYSLOG` capability.
Since `CAP_SYSLOG` does not exist in the original `inheritable` set, attempts to set `CAP_SYSLOG` in the ambient set will fail (this is due to the kernel’s behavior/spec).

https://man7.org/linux/man-pages/man7/capabilities.7.html

Now let’s look at the capability setting flow:

`drop_privileges` → `syscall.set_capability` → (caps crate) `caps::set(None, cset, value)?;` → (caps crate) `ambient::set(value)`

Here is the relevant code. This sets capabilities in order.
For the ambient set, it uses a `for` loop in  runtime::thread_all_supported (thread_all_supported are also HashSet):

```rust
pub type CapsHashSet = std::collections::HashSet<Capability>;

pub fn thread_all_supported() -> CapsHashSet {
    let mut supported = super::all();
    for c in super::all() {
        if super::has_cap(None, CapSet::Bounding, c).is_err() {
            supported.remove(&c);
        }
    }
    supported
}

pub fn set(value: &super::CapsHashSet) -> Result<(), CapsError> {
    for c in runtime::thread_all_supported() {
        if value.contains(&c) {
            raise(c)?;
        } else {
            drop(c)?;
        };
    }
    Ok(())
}
```

In this case, we are trying to set `CAP_SYSLOG` and `CAP_CHOWN`.

Because a `HashSet` does not guarantee iteration order, there are two possible patterns:

1. `CAP_CHOWN` → `CAP_SYSLOG`
2. `CAP_SYSLOG` → `CAP_CHOWN`

**Pattern 1**
`CAP_CHOWN` is successfully set first. Then, when trying to set `CAP_SYSLOG`, it fails and returns a `CapsError` up the call stack.

Result: `CapAmb: 0000000000000001`

**Pattern 2**
Setting `CAP_SYSLOG` is attempted first → this fails and returns a `CapsError`.
(`CAP_CHOWN` is never set.)

Result: `CapAmb: 0000000000000000`

tests/contest/contest/src/tests/exec/cap_test.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        let (stdout, _) = exec_container(
+            id,
+            dir,
+            &["--cap=CAP_SYSLOG", "cat", "/proc/self/status"],
+            None,
+        )
+        .expect("exec failed");
+
+        if !stdout.contains("CapInh:\t0000000000000001") {
```
saku3, line 310: Please add a comment explaining why the expected result looks like this.

It would be preferable to also include a comment here explaining how the bitmask values correspond to each capability.

tests/contest/contest/src/tests/exec/cap_test.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        let start_result = start_container(id, dir).unwrap().wait().unwrap();
+        if !start_result.success() {
+            return TestResult::Failed(anyhow!("container start failed"));
+        }
+
+        let (stdout, _) =
+            exec_container(id, dir, &["cat", "/proc/self/status"], None).expect("exec failed");
+
+        if !stdout.contains("CapInh:\t0000000000000000") {
```
saku3, line 141: Please add a comment noting that the bounding set limits the capabilities, so only CAP_SYS_ADMIN is actually granted.

tests/contest/contest/src/tests/exec/cgroup_test.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::anyhow;
```
saku3, line 3: (TODO)I’d like to implement the following test, but it’s currently not implemented because it can’t pass due to the issue described below.

https://github.com/youki-dev/youki/issues/3342

Ideally, the following test cases should exist.
(Please remove `ro` from the spec so that we can create directories in the cgroup.
I’ll use runc’s tests as a reference.)
https://github.com/opencontainers/runc/blob/main/tests/integration/exec.bats

* **Now move init to a subcgroup, and check it was moved.**
```
youki exec a sh -euc "mkdir /sys/fs/cgroup/foobar && echo 1 > /sys/fs/cgroup/foobar/cgroup.procs && grep -w foobar /proc/1/cgroup"
```
* **The init process is now in "/foo", but an exec process can still join "/" because we haven't enabled any domain controller yet.**
```
 youki exec a grep '^0::/$' /proc/self/cgroup
```
* **Turn on a domain controller (memory).**
```
 youki exec a sh -euc 'echo $$ > /sys/fs/cgroup/foobar/cgroup.procs; echo +memory > /sys/fs/cgroup/cgroup.subtree_control'
```
* **An exec process can no longer join "/" after turning on a domain  controller.  Check that cgroup v2 fallback to init cgroup works.**
    * This command is expected to succeed. 
```
 youki exec a sh -euc "cat /proc/self/cgroup && grep '^0::/foobar$' /proc/self/cgroup"
```
* **Check that --cgroup / disables the init cgroup fallback.**
  * This command is expected to fail.
```
youki exec --cgroup / a true
```
* **Check that explicit --cgroup foobar works.**
 ```
youki exec --cgroup foobar a grep '^0::/foobar$' /proc/self/cgroup
```
* **Check all processes is in foobar (this check is redundant).**
```
youki exec --cgroup foobar a sh -euc '! grep -vwH foobar /proc/*/cgroup'
```
* **Add a second subcgroup, check we're in it.**
```
youki exec --cgroup foobar a mkdir /sys/fs/cgroup/second
youki exec --cgroup second a grep -w second /proc/self/cgroup
```

File: crates/libcontainer/src/container/builder_impl.rs [hunk 4]
```diff
@@ -85,9 +91,38 @@ impl ContainerBuilderImpl {
...
     fn run_container(&mut self) -> Result<Pid, LibcontainerError> {
         let linux = self.spec.linux().as_ref().ok_or(MissingSpecError::Linux)?;
-        let cgroups_path = utils::get_cgroup_path(linux.cgroups_path(), &self.container_id);
+        let base_cgroups_path = utils::get_cgroup_path(linux.cgroups_path(), &self.container_id);
+        let mut final_cgroups_path = base_cgroups_path;
+
+        if let Some(sub_cgroup_path) = &self.sub_cgroup_path {
+            if sub_cgroup_path != "/" {
+                let potential_path = final_cgroups_path.join(sub_cgroup_path);
+                let normalized =
```
saku3: Can we use the `normalize` function?

https://github.com/youki-dev/youki/blob/eaf8ca518a2faf60fd4ccf8cee9c9f4f0605e38e/crates/libcontainer/src/utils.rs#L94

saku3 (commented):
I haven’t been able to review everything yet, but I did check the cgroup-related parts.

I’m not sure whether the deletion logic works correctly when a sub-cgroup is created.

```
$ youki  run -b tutorial container
```

another terminal

```
$ youki exec container sh -euc "mkdir /sys/fs/cgroup/foobar && echo 1 > /sys/fs/cgroup/foobar/cgroup.procs && grep -w foobar /proc/1/cgroup"
```

the original terminal

```
$ exit
ERROR libcontainer::container::container_delete: failed to remove cgroup due to: V2(WrappedIo(Other { err: Custom { kind: TimedOut, error: "could not delete" }, path: "/sys/fs/cgroup/:youki:container" })) cgroup_path=":youki:container"
ERROR youki: error in executing command: io error: at /sys/fs/cgroup/:youki:container: could not delete

Caused by:
    at /sys/fs/cgroup/:youki:container: could not delete
run failed : io error: at /sys/fs/cgroup/:youki:container: could not delete
```

tommady:
The changes in https://github.com/youki-dev/youki/pull/3347 will help address and enhance the testcase requested by saku3.

Just a memo for myself — turning this PR back to draft.

saku3:
@tommady 
I haven’t had much time to review this PR recently — sorry about that.

For this PR, if we can’t move it forward until issue #3347 is resolved, I think we could proceed by leaving TODO comments (or create sub-issue) for the parts related to #3347 for now.

What approach would be easiest for you to proceed?

tommady:
> @tommady I haven’t had much time to review this PR recently — sorry about that.
> 
> For this PR, if we can’t move it forward until issue #3347 is resolved, I think we could proceed by leaving TODO comments (or create sub-issue) for the parts related to #3347 for now.
> 
> What approach would be easiest for you to proceed?

Thanks for the suggestion!

From my point of view, this PR is ready for review.  
Aside from the parts related to issue #3347, there was also an earlier comment from you:
https://github.com/youki-dev/youki/pull/3210#discussion_r2554961182

Because of that, some of the cgroup tests can’t be implemented right now.

So yes, please review it when you have time 🙂

saku3:
Thank! I’ll review it this week.

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -580,8 +620,10 @@ impl TenantContainerBuilder {
         env
     }
 
-    fn get_no_new_privileges(&self) -> Option<bool> {
+    fn get_no_new_privileges(&self, spec: &Spec) -> Option<bool> {
         self.no_new_privs
+            .filter(|&is_set| is_set)
+            .or_else(|| spec.process().clone().and_then(|p| p.no_new_privileges()))
```
saku3: It looks like we can avoid using `clone()`.

```suggestion
            .or_else(|| spec.process().as_ref().and_then(|p| p.no_new_privileges()))
```

tests/contest/contest/src/tests/exec/ignore_paused_test.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        let pause_result = pause_container(id, dir).unwrap().wait().unwrap();
+        if !pause_result.success() {
+            return TestResult::Failed(anyhow!("container pause failed"));
+        }
+
+        let id2 = id.clone();
+        let dir2 = dir.to_path_buf();
+        std::thread::spawn(move || {
```
saku3, line 30: This means we want to resume the paused container at the end, right?

This might be prone to flakiness.
Would you consider using `scopeguard::defer!`?

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -85,9 +92,26 @@ impl ContainerBuilderImpl {
...
+                let potential_path = final_cgroups_path.join(sub_cgroup_path);
+                let normalized = potential_path.normalize();
+
+                if !normalized.starts_with(&final_cgroups_path) {
+                    return Err(LibcontainerError::OtherCgroup(format!(
+                        "{} is not a sub cgroup path",
+                        sub_cgroup_path
+                    )));
+                }
+                final_cgroups_path = potential_path;
```
saku3: Shouldn’t the final value we want to set be `normalized`?

```suggestion
                final_cgroups_path = normalized;
```

tommady: good catcha!

tests/contest/contest/src/tests/exec/ignore_paused_test.rs
tommady: `scopeguard` was removed earlier (see commit below), and in this case it would also resume the container too late since it only runs at scope exit. I’ll use an `mpsc` channel instead to explicitly synchronize the resume during `exec --ignore-paused`. Thanks for the suggestion.

https://github.com/youki-dev/youki/commit/123b51a4ee054fb94d9843d4a5bdb5210b73ca5a

crates/youki/src/commands/exec.rs
```diff
@@ -9,8 +9,6 @@ use nix::sys::wait::{WaitStatus, waitpid};
 use crate::workload::executor::default_executor;
 
 pub fn exec(args: Exec, root_path: PathBuf) -> Result<i32> {
```
saku3, line 11: with_apparmor seems to be missing.

tests/contest/contest/src/tests/exec/ignore_paused_test.rs
saku3: Sorry, I understand now.

In this case, I think the previous implementation (adding a short sleep) is acceptable.
It aligns with runc’s behavior, so I don’t expect it to cause issues in practice.

Also, since the spawn order doesn’t guarantee the actual execution order, `resume` may run first, meaning the test might not actually verify that `exec` works while the container is paused.

saku3:
I think everything looks mostly fine aside from the parts I commented on.

PTAL @utam0k

tommady:
it’s strange that both integration verifications failed
```
# Start group example
1 / 1 : hello world : not ok
	container stderr was not empty, found : 
thread 'main' (1) panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
```
I couldn’t reproduce this on my local machine

saku3:
This is tough...

I checked it in my repository’s GitHub Actions as well. 
* Running only the `example` and `exec` tests → fails.
* Running the tests with `batch_size` set to 1 → succeeds.
  * [batch_size](https://github.com/youki-dev/youki/blob/main/tests/contest/test_framework/src/test_manager.rs#L71)

It’s possible the exec tests aren’t cleaning up properly.

tommady:
> This is tough...
> 
> I checked it in my repository’s GitHub Actions as well.
> 
> * Running only the `example` and `exec` tests → fails.
> * Running the tests with `batch_size` set to 1 → succeeds.
>   
>   * [batch_size](https://github.com/youki-dev/youki/blob/main/tests/contest/test_framework/src/test_manager.rs#L71)
> 
> It’s possible the exec tests aren’t cleaning up properly.

Hi @saku3, @utam0k,

You were absolutely spot on with your observation about batch_size=1. 
The issue was indeed caused by tests interfering with each other when run concurrently! 

I've pushed a fix for this. The root cause was that preserve_fds_test was modifying the global file descriptor table in the parent contest test runner process by manually clearing the FD_CLOEXEC flag and running dup2(fd, 3). 

Because contest runs tests concurrently by default, this led to two distinct issues:
1. Race condition in hello_world: Overwriting fd=3 in the parent silently closed the pipe stdout used by other concurrently running tests (like hello_world), causing them to panic with a "Broken pipe" (OS error 32).
2. FD Leak causing EPERM: Clearing the O_CLOEXEC flag on the host file descriptor caused it to leak into child processes spawned by other tests, which interfered with mount/remount operations in new namespaces and caused EPERM errors.

To fix this, the file descriptor modifications have been confined entirely to the child process specifically spawned for preserve_fds_test by using CommandExt::pre_exec. 
This allows us to run dup2(fd, 3) safely after fork() but before exec(), preventing any side effects on the test runner or other concurrent tests.

Please feel free to review the changes when you have time.
Thanks.

saku3:
@tommady 
Sorry for the delay.
I haven’t had much time since my last review.
I’ll review this within the next three days.
In the meantime, could you please resolve the conflicts?

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -419,7 +450,9 @@ impl TenantContainerBuilder {
 
     fn load_container_state(&self, container_dir: PathBuf) -> Result<Container, LibcontainerError> {
         let container = Container::load(container_dir)?;
-        if !container.can_exec() {
+        if !container.can_exec()
```
saku3: I missed that.

With the current condition, statuses that should not allow exec, such as Stopped, would end up allowing exec.

I think it would be better to handle it like this.
(I wrote it this way for readability first, though it may be possible to simplify it further.)

```
        if container.status() == ContainerStatus::Paused {
            if !self.ignore_paused {
                return Err(...);
            }
        } else if !container.can_exec() {
            return Err(...);
        }
```

saku3 (commented):
Thank you for working through this patiently.

I added one more comment.

I think the preserve_fds_test looks good.

crates/libcontainer/src/container/tenant_builder.rs
tommady: Great catch on that logic bug!

I think using **match** is the way to go here. It’s a lot more readable than the nested if statements and makes it much easier to see exactly which statuses allow execution and which don't. 

Thanks for pointing that out!

PR #3214, opened 2025-08-03
add parse capabilities for duration on event command

### LINKED ISSUE #3206: Difference between the events command in runc and youki

There are differences between the `events` command in `runc` and `youki`.

Please refer to the following:
https://github.com/opencontainers/runc/blob/main/man/runc-events.8.md

In `runc`, options like `--interval 1s` or `--interval 100ms` are supported:

```sh
runc events testcon --interval 1s

{"type":"stats","id":"testcon","data": ... }
```

However, in Youki, the `--interval` argument currently expects a digit-only value:

```sh
$ youki events testcon --interval 1s
error: invalid value '1s' for '--interval <INTERVAL>': invalid digit found in string

For more information, try '--help'.
```

This behavior differs from `runc` and should be addressed for compatibility.

mdaffad:
## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Close #3206


YJDoc2:
Hey, can you rebase and fix the clippy issues? Thanks :)

mdaffad:
Updated

utam0k:
@mdaffad Please sign following the instructions in the link
https://github.com/youki-dev/youki/pull/3214/checks?check_run_id=47689260136

crates/liboci-cli/src/events.rs
```diff
@@ -1,11 +1,35 @@
...
+    }
+
+    let (num, unit) = s
+        .trim_end_matches(|c: char| !c.is_ascii_alphabetic())
+        .chars()
+        .partition::<String, _>(|c| c.is_ascii_digit());
+
+    let num = num.parse::<u64>().map_err(|_| "Invalid number")?;
+
+    let dur = match unit.as_str() {
```
utam0k: What if 1h30m?

mdaffad: updated

utam0k:
@mdaffad May I ask you to address DCO?
https://github.com/youki-dev/youki/pull/3214/checks?check_run_id=48316005413

crates/liboci-cli/Cargo.toml
```diff
@@ -10,6 +10,9 @@ authors = ["youki team"]
 edition = "2021"
 keywords = ["youki", "container", "oci"]
 
+[dependencies]
+parse_duration = "2.1.1"
```
utam0k: Can't we implement it ourselves?

crates/liboci-cli/src/events.rs
```diff
@@ -1,11 +1,82 @@
...
+    }
+
+    Ok(Duration::from_secs(total_seconds))
+}
+
+fn parse_interval(s: &str) -> Result<Duration, String> {
+    let s = s.trim();
+    if let Ok(secs) = s.parse::<u64>() {
+        return Ok(Duration::from_secs(secs));
+    }
```
nayuta723: It looks like runc is not supported.
```
$ sudo runc events --interval 1 tutorial_container
Incorrect Usage: invalid value "1" for flag -interval: parse error

NAME:
   runc events - display container events such as OOM notifications, cpu, memory, and IO usage statistics

USAGE:
   runc events [command options] <container-id>

Where "<container-id>" is the name for the instance of the container.

DESCRIPTION:
   The events command displays information about the container. By default the
information is displayed once every 5 seconds.

OPTIONS:
   --interval value  set the stats collection interval (default: 5s)
   --stats           display the container's stats then exit
   
ERRO[0000] invalid value "1" for flag -interval: parse error 
```

crates/liboci-cli/src/events.rs
```diff
@@ -1,11 +1,82 @@
...
+
+    if total_seconds == 0 {
+        return Err("Duration cannot be zero".to_string());
+    }
+
+    Ok(Duration::from_secs(total_seconds))
+}
+
+fn parse_interval(s: &str) -> Result<Duration, String> {
```
nayuta723: Please update the return value to use an `Error` type instead of a `String`, to stay consistent with the implementation in `exec.rs`.
```suggestion
fn parse_interval(s: &str) -> Result<Duration, Box<dyn Error + Send + Sync + 'static>> {
```

ref: https://github.com/youki-dev/youki/blob/main/crates/liboci-cli/src/exec.rs#L67-L92

crates/liboci-cli/src/events.rs
```diff
@@ -1,11 +1,82 @@
+use std::time::Duration;
+
 use clap::Parser;
 
+fn parse_duration_string(s: &str) -> Result<Duration, String> {
+    let s = s.trim().to_lowercase();
+    let mut total_seconds = 0u64;
+    let mut current_number = String::new();
```
nayuta723: Using `current_number.push(ch)` in a loop allocates a `String` and can cause repeated reallocations. Prefer a slice-based approach: work on `&[u8]` (or `&str`), use `split_at` and a remainder slice so you don't need a digit buffer. That avoids extra allocations and is more efficient.

crates/liboci-cli/src/events.rs
```diff
@@ -1,11 +1,82 @@
+use std::time::Duration;
+
 use clap::Parser;
 
+fn parse_duration_string(s: &str) -> Result<Duration, String> {
```
nayuta723: Could you also add some unit tests for this logic?

nayuta723 (commented):
@mdaffad 
Thank you for your contribution! 
To confirm that this change resolves Issue #3206, could you please post the test results in this PR?

crates/liboci-cli/src/events.rs
mdaffad: I see. I think we should drop the digit only then.

mdaffad:
Thanks for the review @nayuta723. I have pushed an update addressing all the points:

- **Dropped the digit-only path**: a unit-less value such as `1` is now rejected, matching runc (which relies on Go's `time.ParseDuration`).
- **Return type**: `parse_interval` now returns `Box<dyn Error + Send + Sync + 'static>` instead of `String`, consistent with `exec.rs`.
- **Slice-based parsing**: replaced the per-character `String` buffer with `split_at` over a remainder slice, so no digit buffer / repeated allocations.
- **Sub-second units**: added `ms`, `us`/`µs`, `ns` so the `100ms` case from #3206 works, and compound durations like `1h30m` (per @utam0k's earlier question) are still supported.
- **Unit tests**: added tests for single units, compound durations, whitespace trimming, and the rejected/invalid cases.

### Test results

```
$ cargo test -p liboci-cli --lib
running 5 tests
test events::tests::parses_compound_durations ... ok
test events::tests::rejects_digit_only_value ... ok
test events::tests::parses_single_units ... ok
test events::tests::trims_surrounding_whitespace ... ok
test events::tests::rejects_invalid_values ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Behaviour now matches runc:

| input | result |
|-------|--------|
| `5s` (default) | 5s |
| `1s` | 1s |
| `100ms` | 100ms |
| `1h30m` | 1h30m |
| `1` | error (unit required) |

`cargo fmt --check` and `cargo clippy` are both clean.

crates/libcontainer/src/container/container_events.rs
```diff
@@ -23,11 +24,11 @@ impl Container {
     /// .as_init("/var/run/docker/bundle")
     /// .build()?;
     ///
-    /// container.events(5000, false)?;
+    /// container.events(Duration::from_secs(5000), false)?;
```
nayuta723:
```suggestion
    /// container.events(Duration::from_secs(5), false)?;
```

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,137 @@
...
+/// before `m` and `s`). Each entry maps a unit to its length in nanoseconds.
+const UNITS: &[(&str, u128)] = &[
+    ("ns", 1),
+    ("us", 1_000),
+    ("µs", 1_000),
+    ("ms", 1_000_000),
+    ("h", 3_600_000_000_000),
+    ("m", 60_000_000_000),
+    ("s", 1_000_000_000),
+];
```
nayuta723: The following order seems better to me. Are there any potential issues with it?

```suggestion
const UNITS: &[(&str, u128)] = &[
    ("ns", 1),
    ("us", 1_000),
    ("µs", 1_000),
    ("ms", 1_000_000),
    ("s", 1_000_000_000),
    ("m", 60_000_000_000),
    ("h", 3_600_000_000_000),
];
```

nayuta723 (changes requested):
The following behavior also does not appear to match `runc`, so could you please take a look?

When `0` is specified:

```console
$ sudo runc events --interval 0s test01
ERRO[0000] duration interval must be greater than 0
```

When a floating-point value is specified:

```console
$ sudo runc events --interval 1.5s test03
```

nayuta723:
@mdaffad Please add a DCO sign-off.

mdaffad:
@nayuta723 signed-off DCO. Kindly check, thanks

crates/libcontainer/src/container/container_events.rs
```diff
@@ -23,11 +24,18 @@ impl Container {
...
     /// .build()?;
     ///
-    /// container.events(5000, false)?;
+    /// container.events(Duration::from_secs(5), false)?;
     /// # Ok(())
     /// # }
     /// ```
-    pub fn events(&mut self, interval: u32, stats: bool) -> Result<(), LibcontainerError> {
+    pub fn events(&mut self, interval: Duration, stats: bool) -> Result<(), LibcontainerError> {
+        if !stats && interval.is_zero() {
```
nayuta723, line 32:
```suggestion
        if interval.is_zero() {
```

crates/libcontainer/src/container/container_events.rs
```diff
@@ -23,11 +24,18 @@ impl Container {
     /// .as_init("/var/run/docker/bundle")
     /// .build()?;
     ///
-    /// container.events(5000, false)?;
+    /// container.events(Duration::from_secs(5), false)?;
     /// # Ok(())
     /// # }
     /// ```
-    pub fn events(&mut self, interval: u32, stats: bool) -> Result<(), LibcontainerError> {
+    pub fn events(&mut self, interval: Duration, stats: bool) -> Result<(), LibcontainerError> {
```
nayuta723, line 31:
```suggestion
    pub fn events(&mut self, interval: Duration, once: bool) -> Result<(), LibcontainerError> {
```

crates/libcontainer/src/container/container_events.rs
```diff
@@ -56,7 +64,7 @@ impl Container {
                     serde_json::to_string_pretty(&stats)
                         .map_err(LibcontainerError::OtherSerialization)?
                 );
-                thread::sleep(Duration::from_secs(interval as u64));
+                thread::sleep(interval);
             },
         }
```
nayuta723, line 69: This is out of scope for this PR, but please address the nitpicks.

```suggestion
if once {
    print_stats(&cgroup_manager)?;
} else {
    loop {
        print_stats(&cgroup_manager)?;
        thread::sleep(interval);
    }
}
```

crates/libcontainer/src/container/container_events.rs
```diff
@@ -23,11 +24,18 @@ impl Container {
...
     /// # Ok(())
     /// # }
     /// ```
-    pub fn events(&mut self, interval: u32, stats: bool) -> Result<(), LibcontainerError> {
+    pub fn events(&mut self, interval: Duration, stats: bool) -> Result<(), LibcontainerError> {
+        if !stats && interval.is_zero() {
+            tracing::error!(id = ?self.id(), "stats collection interval must be greater than 0");
+            return Err(LibcontainerError::InvalidInput(
+                "duration interval must be greater than 0".into(),
+            ));
```
nayuta723, line 36: The log message and the returned error message are inconsistent, so please fix them.

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,287 @@
...
+        }
+    }
+    (x, scale, &s[end..])
+}
+
+/// Parses `[-+]?([0-9]*(\.[0-9]*)?unit)+` into nanoseconds, matching Go's
+/// `time.ParseDuration`, which is what runc uses for `--interval`.
+///
+/// Signed, and bounded by `i64` like Go's `time.Duration`, so that youki
+/// accepts and rejects exactly the values runc does.
```
nayuta723, line 63:
```suggestion
/// Parses `[-+]?([0-9]*(\.[0-9]*)?unit)+` into nanoseconds, transliterated
/// from Go's `time.ParseDuration` (what runc parses `--interval` with) so the
/// accepted grammar and its edge cases match. Keep in sync with Go rather
/// than restyling; the quirks are intentional.
```

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,287 @@
...
+fn parse_duration_nanos(orig: &str) -> Result<i64, ParseError> {
```
nayuta723, line 64:
```suggestion
fn parse_duration_nanos(input: &str) -> Result<i64, ParseError> {
```

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,287 @@
...
+    let mut s = orig;
```
nayuta723, line 65:
```suggestion
    let mut s = input;
```

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,287 @@
...
+
+    let mut neg = false;
+    if let Some(rest) = s.strip_prefix('-') {
+        neg = true;
+        s = rest;
+    } else if let Some(rest) = s.strip_prefix('+') {
+        s = rest;
+    }
```
nayuta723, line 73:
```suggestion
    let (neg, mut s) = match orig.strip_prefix('-') {
        Some(rest) => (true, rest),
        None => (false, orig.strip_prefix('+').unwrap_or(orig)),
    };
```

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,287 @@
...
+            .map(|(_, nanos)| *nanos)
+            .ok_or_else(|| format!("unknown unit {name:?} in duration {orig:?}"))?;
+
+        let mut nanos = value.checked_mul(unit).ok_or_else(|| invalid(orig))?;
+        if fraction > 0 {
+            // f64 keeps fractions of an hour nanosecond-accurate.
+            let scaled = (fraction as f64 * (unit as f64 / scale)) as i64;
+            nanos = nanos.checked_add(scaled).ok_or_else(|| invalid(orig))?;
+        }
+        total = total.checked_add(nanos).ok_or_else(|| invalid(orig))?;
```
nayuta723, line 130: It might be more readable to extract this into a function.

crates/liboci-cli/src/events.rs
```diff
@@ -1,15 +1,287 @@
...
+    }
+    if s.is_empty() {
+        return Err(invalid(orig));
+    }
+
+    let mut total: i64 = 0;
+    while !s.is_empty() {
+        if !(s.starts_with('.') || s.starts_with(|c: char| c.is_ascii_digit())) {
+            return Err(invalid(orig));
+        }
```
nayuta723, line 87: Isn’t this code effectively doing the same thing as the code below? Could we consolidate it into an `is_number_start()` function?
```
let unit_end = s
            .find(|c: char| c == '.' || c.is_ascii_digit())
            .unwrap_or(s.len());
```

```
fn is_number_start(c: char) -> bool {
    c == '.' || c.is_ascii_digit()
}
```

nayuta723 (changes requested):
Sorry for the late review. I left some comments.
Also, to confirm that this change resolves Issue https://github.com/youki-dev/youki/issues/3206, could you please post the test results in this PR description?

PR #3230, opened 2025-08-27
Implement set_mempolicy syscall and related error handling

n4mlz:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #

### DIFF
--- a/crates/libcontainer/src/process/init/error.rs
+++ b/crates/libcontainer/src/process/init/error.rs
@@ -61,6 +61,8 @@ pub enum InitProcessError {
     IoPriorityClass(String),
     #[error("call exec sched_setattr error: {0}")]
     SchedSetattr(String),
+    #[error("invalid memory policy: {0}")]
+    InvalidMemoryPolicy(String),
     #[error("failed to verify if current working directory is safe")]
     InvalidCwd(#[source] nix::Error),
     #[error("missing linux section in spec")]
--- a/crates/libcontainer/src/process/init/process.rs
+++ b/crates/libcontainer/src/process/init/process.rs
@@ -45,6 +45,8 @@ pub fn container_init_process(
 
     setup_scheduler(ctx.process.scheduler())?;
 
+    setup_memory_policy(ctx.linux.memory_policy(), ctx.syscall.as_ref())?;
+
     // set up tty if specified
     if let Some(csocketfd) = args.console_socket {
         tty::setup_console(csocketfd).map_err(|err| {
@@ -831,6 +833,299 @@ fn setup_scheduler(sc_op: &Option<Scheduler>) -> Result<()> {
     Ok(())
 }
 
+fn setup_memory_policy(
+    memory_policy: &Option<oci_spec::runtime::LinuxMemoryPolicy>,
+    syscall: &dyn Syscall,
+) -> Result<()> {
+    if let Some(policy) = memory_policy {
+        // Memory policy mode constants from Linux UAPI (include/uapi/linux/mempolicy.h)
+        const MPOL_DEFAULT: i32 = 0;
+        const MPOL_PREFERRED: i32 = 1;
+        const MPOL_BIND: i32 = 2;
+        const MPOL_INTERLEAVE: i32 = 3;
+        const MPOL_LOCAL: i32 = 4;
+        const MPOL_PREFERRED_MANY: i32 = 5;
+        const MPOL_WEIGHTED_INTERLEAVE: i32 = 6;
+
+        // Memory policy flag constants from Linux UAPI
+        const MPOL_F_NUMA_BALANCING: u32 = 1 << 13; // 0x2000
+        const MPOL_F_RELATIVE_NODES: u32 = 1 << 14; // 0x4000
+        const MPOL_F_STATIC_NODES: u32 = 1 << 15; // 0x8000
+
+        let base_mode = match policy.mode() {
+            oci_spec::runtime::MemoryPolicyModeType::MpolDefault => MPOL_DEFAULT,
+            oci_spec::runtime::MemoryPolicyModeType::MpolPreferred => MPOL_PREFERRED,
+            oci_spec::runtime::MemoryPolicyModeType::MpolBind => MPOL_BIND,
+            oci_spec::runtime::MemoryPolicyModeType::MpolInterleave => MPOL_INTERLEAVE,
+            oci_spec::runtime::MemoryPolicyModeType::MpolLocal => MPOL_LOCAL,
+            oci_spec::runtime::MemoryPolicyModeType::MpolPreferredMany => MPOL_PREFERRED_MANY,
+            oci_spec::runtime::MemoryPolicyModeType::MpolWeightedInterleave => {
+                MPOL_WEIGHTED_INTERLEAVE
+            }
+        };
+
+        let mut flags_value: u32 = 0;
+        if let Some(flags) = policy.flags() {
+            let mut has_static = false;
+            let mut has_relative = false;
+
+            for flag in flags {
+                match flag {
+                    oci_spec::runtime::MemoryPolicyFlagType::MpolFNumaBalancing => {
+                        // MPOL_F_NUMA_BALANCING is only valid with MPOL_BIND
+                        if base_mode != MPOL_BIND {
+                            return Err(InitProcessError::InvalidMemoryPolicy(
+                                "MPOL_F_NUMA_BALANCING can only be used with MPOL_BIND".to_string(),
+                            ));
+                        }
+                        flags_value |= MPOL_F_NUMA_BALANCING;
+                    }
+                    oci_spec::runtime::MemoryPolicyFlagType::MpolFRelativeNodes => {
+                        has_relative = true;
+                        flags_value |= MPOL_F_RELATIVE_NODES;
+                    }
+                    oci_spec::runtime::MemoryPolicyFlagType::MpolFStaticNodes => {
+                        has_static = true;
+                        flags_value |= MPOL_F_STATIC_NODES;
+                    }
+                }
+            }
+
+            // STATIC_NODES and RELATIVE_NODES are mutually exclusive
+            if has_static && has_relative {
+                return Err(InitProcessError::InvalidMemoryPolicy(
+                    "MPOL_F_STATIC_NODES and MPOL_F_RELATIVE_NODES are mutually exclusive"
+                        .to_string(),
+                ));
+            }
+        }
+
+        // Combine mode and flags
+        let mode_with_flags = base_mode | (flags_value as i32);
+
+        // Handle special cases for MPOL_DEFAULT and MPOL_LOCAL
+        match base_mode {
+            MPOL_DEFAULT => {
+                // MPOL_DEFAULT
+                // For MPOL_DEFAULT, nodes must be empty and no flags allowed
+                if !policy.nodes().trim().is_empty() {
+                    return Err(InitProcessError::InvalidMemoryPolicy(
+                        "MPOL_DEFAULT does not accept node specification".to_string(),
+                    ));
+                }
+                if flags_value != 0 {
+                    return Err(InitProcessError::InvalidMemoryPolicy(
+                        "MPOL_DEFAULT does not accept flags".to_string(),
+                    ));
+                }
+                syscall
+                    .set_mempolicy(mode_with_flags, &[], 0)
+                    .map_err(|err| {
+                        tracing::error!(?err, "failed to set memory policy (MPOL_DEFAULT)");
+                        InitProcessError::SyscallOther(err)
+                    })?;
+            }
+            MPOL_LOCAL => {
+                // MPOL_LOCAL
+                // For MPOL_LOCAL, nodes must be empty and no flags allowed
+                if !policy.nodes().trim().is_empty() {
+                    return Err(InitProcessError::InvalidMemoryPolicy(
+                        "MPOL_LOCAL does not accept node specification".to_string(),
+                    ));
+                }
+                if flags_value != 0 {
+                    return Err(InitProcessError::InvalidMemoryPolicy(
+                        "MPOL_LOCAL does not accept flags".to_string(),
+                    ));
+                }
+                syscall
+                    .set_mempolicy(mode_with_flags, &[], 0)
+                    .map_err(|err| {
+                        tracing::error!(?err, "failed to set memory policy (MPOL_LOCAL)");
+                        InitProcessError::SyscallOther(err)
+                    })?;
+            }
+            MPOL_PREFERRED => {
+                // MPOL_PREFERRED
+                // For MPOL_PREFERRED with empty node string, use empty set (local allocation)
+                if policy.nodes().trim().is_empty() {
+                    // Empty nodemask for PREFERRED with STATIC/RELATIVE flags is invalid
+                    if flags_value & (MPOL_F_RELATIVE_NODES | MPOL_F_STATIC_NODES) != 0 {
+                        return Err(InitProcessError::InvalidMemoryPolicy(
+                            "MPOL_PREFERRED with empty nodes cannot use MPOL_F_STATIC_NODES or MPOL_F_RELATIVE_NODES flags".to_string(),
+                        ));
+                    }
+                    syscall
+                        .set_mempolicy(mode_with_flags, &[], 0)
+                        .map_err(|err| {
+                            tracing::error!(
+                                ?err,
+                                "failed to set memory policy (MPOL_PREFERRED with empty nodes)"
+                            );
+                            InitProcessError::SyscallOther(err)
+                        })?;
+                } else {
+                    // Parse nodes and build nodemask
+                    let (nodemask, maxnode) = build_nodemask(policy.nodes())?;
+
+                    // Double-check: if the parsed result is empty (e.g., whitespace-only input),
+                    // treat as empty nodemask
+                    if maxnode == 0 {
+                        if flags_value & (MPOL_F_RELATIVE_NODES | MPOL_F_STATIC_NODES) != 0 {
+                            return Err(InitProcessError::InvalidMemoryPolicy(
+                                "MPOL_PREFERRED with empty nodes cannot use MPOL_F_STATIC_NODES or MPOL_F_RELATIVE_NODES flags".to_string(),
+                            ));
+                        }
+                        syscall
+                            .set_mempolicy(mode_with_flags, &[], 0)
+                            .map_err(|err| {
+                                tracing::error!(
+                                    ?err,
+                                    "failed to set memory policy (MPOL_PREFERRED with empty nodes)"
+                                );
+                                InitProcessError::SyscallOther(err)
+                            })?;
+                    } else {
+                        syscall
+                            .set_mempolicy(mode_with_flags, &nodemask, maxnode)
+                            .map_err(|err| {
+                                tracing::error!(
+                                    ?err,
+                                    "failed to set memory policy (MPOL_PREFERRED)"
+                                );
+                                InitProcessError::SyscallOther(err)
+                            })?;
+                    }
+                }
+            }
+            _ => {
+                // For other modes (BIND, INTERLEAVE, PREFERRED_MANY, WEIGHTED_INTERLEAVE),
+                // nodes must not be empty
+                if policy.nodes().trim().is_empty() {
+                    return Err(InitProcessError::InvalidMemoryPolicy(format!(
+                        "Mode {} requires non-empty node specification",
+                        base_mode
+                    )));
+                }
+
+                // Parse nodes and build nodemask
+                let (nodemask, maxnode) = build_nodemask(policy.nodes())?;
+
+                // Double-check: verify the parsed result is not empty
+                if maxnode == 0 {
+                    return Err(InitProcessError::InvalidMemoryPolicy(format!(
+                        "Mode {} requires non-empty node specification (parsed result is empty)",
+                        base_mode
+                    )));
+                }
+                syscall
+                    .set_mempolicy(mode_with_flags, &nodemask, maxnode)
+                    .map_err(|err| {
+                        tracing::error!(?err, "failed to set memory policy");
+                        InitProcessError::SyscallOther(err)
+                    })?;
+            }
+        }
+    }
+    Ok(())
+}
+
+// Build a proper nodemask for set_mempolicy
+fn build_nodemask(nodes: &str) -> Result<(Vec<u64>, u64)> {
+    let node_ids = parse_node_string(nodes)?;
+
+    if node_ids.is_empty() {
+        // Empty nodemask - return NULL equivalent (empty vector)
+        return Ok((Vec::new(), 0));
+    }
+
+    // Find the highest node ID
+    let highest_node = node_ids.iter().max().copied().unwrap_or(0) as usize;
+
+    // Calculate maxnode as highest_node + 1 (number of nodes to consider)
+    let maxnode = (highest_node + 1) as u64;
+
+    // Calculate how many u64 values we need to store the bitmask
+    let u64_bits = 64;
+    let num_u64s = (highest_node / u64_bits) + 1;
+
+    // Build the nodemask array as Vec<u64>
+    let mut nodemask = vec![0u64; num_u64s];
+
+    // Set bits for each node ID
+    for node_id in node_ids {
+        let node_id = node_id as usize;
+        let word_index = node_id / u64_bits;
+        let bit_index = node_id % u64_bits;
+
+        if word_index < nodemask.len() {
+            nodemask[word_index] |= 1u64 << bit_index;
+        }
+    }
+
+    Ok((nodemask, maxnode))
+}
+
+fn parse_node_string(nodes: &str) -> Result<Vec<u32>> {
+    let mut node_ids = Vec::new();
+
+    // Trim whitespace and check for empty string
+    let nodes = nodes.trim();
+    if nodes.is_empty() {
+        return Ok(node_ids);
+    }
+
+    for range in nodes.split(',') {
+        let range = range.trim();
+        if range.is_empty() {
+            continue; // Skip empty entries caused by multiple commas
+        }
+        let range = range.trim();
+        if range.is_empty() {
+            continue;
+        }
+
+        if let Some(dash_pos) = range.find('-') {
+            // Range format: "node1-node2"
+            let start_str = range[..dash_pos].trim();
+            let end_str = range[dash_pos + 1..].trim();
+
+            let start: u32 = start_str.parse().map_err(|_| {
+                InitProcessError::InvalidMemoryPolicy(format!(
+                    "Invalid node range start: {}",
+                    start_str
+                ))
+            })?;
+            let end: u32 = end_str.parse().map_err(|_| {
+                InitProcessError::InvalidMemoryPolicy(format!(
+                    "Invalid node range end: {}",
+                    end_str
+                ))
+            })?;
+
+            if start > end {
+                return Err(InitProcessError::InvalidMemoryPolicy(format!(
+                    "Invalid node range: {}-{}",
+                    start, end
+                )));
+            }
+
+            for node in start..=end {
+                node_ids.push(node);
+            }
+        } else {
+            // Single node
+            let node: u32 = range.parse().map_err(|_| {
+                InitProcessError::InvalidMemoryPolicy(format!("Invalid node: {}", range))
+            })?;
+
+            node_ids.push(node);
+        }
+    }
+
+    Ok(node_ids)
+}
+
 #[cfg(feature = "libseccomp")]
 fn sync_seccomp(
     fd: Option<i32>,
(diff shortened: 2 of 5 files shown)


crates/libcontainer/src/process/init/process.rs
```diff
@@ -831,6 +833,299 @@ fn setup_scheduler(sc_op: &Option<Scheduler>) -> Result<()> {
...
+        const MPOL_PREFERRED_MANY: i32 = 5;
+        const MPOL_WEIGHTED_INTERLEAVE: i32 = 6;
+
+        // Memory policy flag constants from Linux UAPI
+        const MPOL_F_NUMA_BALANCING: u32 = 1 << 13; // 0x2000
+        const MPOL_F_RELATIVE_NODES: u32 = 1 << 14; // 0x4000
+        const MPOL_F_STATIC_NODES: u32 = 1 << 15; // 0x8000
+
+        let base_mode = match policy.mode() {
+            oci_spec::runtime::MemoryPolicyModeType::MpolDefault => MPOL_DEFAULT,
```
saku3: import them like this:

```
use oci_spec::runtime::{ /* other */, LinuxMemoryPolicy, MemoryPolicyFlagType, MemoryPolicyModeType };

```

Please refer to line 11.

crates/libcontainer/src/process/init/process.rs
```diff
@@ -831,6 +833,299 @@ fn setup_scheduler(sc_op: &Option<Scheduler>) -> Result<()> {
...
+    if node_ids.is_empty() {
+        // Empty nodemask - return NULL equivalent (empty vector)
+        return Ok((Vec::new(), 0));
+    }
+
+    // Find the highest node ID
+    let highest_node = node_ids.iter().max().copied().unwrap_or(0) as usize;
+
+    // Calculate maxnode as highest_node + 1 (number of nodes to consider)
+    let maxnode = (highest_node + 1) as u64;
```
saku3: Could you explain the rationale for how `maxnode` is calculated?

When I run an example like the one in runc’s[ test code](https://github.com/opencontainers/runc/pull/4726/files#diff-eab28bb7cc15c949e829d02adfa5ad5869d5cea2dde63d0c2c5f4c52225967e6), youki fails with an error. I suspect the cause is a difference in how `maxnode` is implemented, but I don’t fully understand it yet.

https://github.com/opencontainers/runc/pull/4726/files#diff-1c1e2ce4c942d6847ab0885e1ff1b92c28f8502fc536cc66e33493a8cbe13109R172

n4mlz: In `set_mempolicy(2)`, *`nodemask` points to a bit mask of node IDs that contains up to `maxnode` bits* and *the kernel will use bits only up to `maxnode`*. Also, *a `maxnode` value of zero specifies the empty set of nodes* (in which case `nodemask` is ignored) .
Therefore, to make the highest set bit in `nodemask` visible to the kernel, `maxnode` should be one greater than the highest node ID we want to include.

The failure you encountered is likely not caused by the `maxnode` calculation itself. Right now, `oci-spec-rs` has already merged support for Linux memory policy (including mode/nodes/flags), but no new crate release has been published yet. If your local environment is still pointing to the crates.io release instead of the git revision with that PR, you may hit errors when parsing or building the spec. That may explain the discrepancy you observed when comparing with runc’s tests.

n4mlz:
Since oci-spec-rs crate—which includes support for Linux memory policy—hasn't yet been released, I couldn't perform tests via CI. However, I have added integration tests. By cloning the branch containing [these changes](https://github.com/youki-dev/oci-spec-rs/pull/292) in oci-spec-rs, editing youroki's Cargo.toml to point to the local crate, and running tests locally, we were able to verify that the tests pass.

crates/libcontainer/src/process/init/process.rs
saku3: Sorry for the long message.
Please let me know if my analysis is incorrect.
If it’s working on your local environment, then something might be wrong with mine.

I first confirmed the machine has two NUMA nodes and the kernel version:

```
cat /sys/devices/system/node/online
0-1

uname -a
Linux ip-172-31-11-219 6.14.0-1012-aws #12~24.04.1-Ubuntu SMP Fri Aug 15 00:16:05 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

I also switched oci-spec to a branch that includes memoryPolicy support:
```
[dependencies]
- oci-spec = { version = "0.8.2", features = ["runtime"] }
+ oci-spec = { git = "https://github.com/n4mlz/oci-spec-rs.git", branch = "fix/mempolicy", features = ["runtime"] }
```

Then I ran the contest tests:

failed

```
just test-contest memory_policy
/home/ubuntu/workspace/youki/scripts/build.sh -o /home/ubuntu/workspace/youki -r -c youki
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
   Compiling libcontainer v0.5.5 (/home/ubuntu/workspace/youki/crates/libcontainer)
   Compiling youki v0.5.5 (/home/ubuntu/workspace/youki/crates/youki)
    Finished `release` profile [optimized] target(s) in 52.06s
/home/ubuntu/workspace/youki/scripts/build.sh -o /home/ubuntu/workspace/youki -r -c contest
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
removed '/home/ubuntu/workspace/youki/contest'
   Compiling contest v0.1.0 (/home/ubuntu/workspace/youki/tests/contest/contest)
    Finished `release` profile [optimized] target(s) in 33.22s
removed '/home/ubuntu/workspace/youki/runtimetest'
    Finished `release` profile [optimized] target(s) in 0.12s
sudo /home/ubuntu/workspace/youki/scripts/contest.sh /home/ubuntu/workspace/youki/youki memory_policy
# Start group memory_policy
1 / 10 : bind_static : not ok
        container stderr was not empty, found : ERROR libcontainer::process::init::process: failed to set memory policy err=Nix(EINVAL)
ERROR youki: error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
Error: failed to create container: exec process failed with error error in executing process : failed syscall

2 / 10 : bind_way_too_large_node_number : ok
3 / 10 : default_with_missing_nodes_ok : ok
4 / 10 : empty_memory_policy : ok
5 / 10 : interleave_without_flags : not ok
        container stderr was not empty, found : ERROR libcontainer::process::init::process: failed to set memory policy err=Nix(EINVAL)
ERROR youki: error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
Error: failed to create container: exec process failed with error error in executing process : failed syscall

6 / 10 : invalid_flag_string : ok
7 / 10 : invalid_mode_string : ok
8 / 10 : missing_mode_but_nodes_present : ok
9 / 10 : preferred_relative : not ok
        container stderr was not empty, found : ERROR libcontainer::process::init::process: failed to set memory policy (MPOL_PREFERRED) err=Nix(EINVAL)
ERROR youki: error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
Error: failed to create container: exec process failed with error error in executing process : failed syscall

10 / 10 : syscall_invalid_arguments : ok
# End group memory_policy

error: Recipe `test-contest` failed on line 62 with exit code 1
```

A simple manual run with the following OCI config also fails:
```
"linux": {
  "memoryPolicy" : {
    "mode": "MPOL_INTERLEAVE",
    "nodes": "1"
  }
}
```

```
sudo strace -f -tt -s 128 -e trace=set_mempolicy -o /tmp/youki.strace \
  ./youki run -b tutorial/ container

ERROR libcontainer::process::init::process: failed to set memory policy err=Nix(EINVAL)
 INFO libcgroups::common: cgroup manager V2 will be used
DEBUG libcgroups::v2::manager: remove cgroup "/sys/fs/cgroup/:youki:container"
ERROR youki: error in executing command: failed to create container: exec process failed with error error in executing process : failed syscall
run failed : failed to create container: exec process failed with error error in executing process : failed syscall
```

trace result 
```
50065 06:03:37.212423 set_mempolicy(MPOL_INTERLEAVE, [0x00000000000002], 2) = -1 EINVAL (Invalid argument)
50064 06:03:37.212710 +++ exited with 0 +++
50063 06:03:37.212746 --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=50064, si_uid=0, si_status=0, si_utime=0, si_stime=0} ---
50065 06:03:37.213411 +++ exited with 255 +++
50063 06:03:37.213466 --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=50065, si_uid=0, si_status=255, si_utime=0, si_stime=0} ---
50063 06:03:37.221851 +++ exited with 255 +++
```


The root cause is explained by the Linux source code:

* set_mempolicy
* kernel_set_mempolicy
* get_nodes
* get_bitmap
* do_set_mempolicy
* mpol_new

See mm/mempolicy.c around here: https://github.com/torvalds/linux/blob/v6.14/mm/mempolicy.c


Why `set_mempolicy(MPOL_INTERLEAVE, [0x00000000000002], 2)` fails:

In the kernel’s `get_nodes()`, maxnode is decremented first (`--maxnode`).
With maxnode=2 from userspace, it becomes 1 inside the kernel.

Then `get_bitmap()` applies a tail mask to keep only the lowest maxnode bits:

`mask[nlongs - 1] &= (1UL << (maxnode % BITS_PER_LONG)) - 1;`


With `maxnode=1` and `BITS_PER_LONG=64`, the mask is `0x1`.

nodemask word is `0x2` (bit1 set = node1).
0x2 & 0x1 = 0x0 → the effective nodemask becomes empty.

MPOL_INTERLEAVE does not allow an empty nodemask, so the kernel returns -EINVAL.

mpol_new
```
} else if (nodes_empty(*nodes))
	return ERR_PTR(-EINVAL);
```

###  runc

With the above in mind, running runc’s code related to memory_policy succeeds: https://github.com/opencontainers/runc/pull/4726

```
 sudo strace -f -tt -s 128   -e trace=set_mempolicy   -o /tmp/youki.strace runc run  -b tutorial/ 
container
```

strace

```
38296 06:57:27.319222 set_mempolicy(MPOL_INTERLEAVE, [0x00000000000002], 64) = 0
```

Therefore, I think it’s preferable to implement the maxnode specification the same way as runc.

### numactl

```
strace -f -tt -s 128 -e trace=set_mempolicy -o /tmp/numa.strace   numactl --interleave=1 sh
```

```
cat  /tmp/numa.strace 
38676 06:59:11.357283 set_mempolicy(MPOL_INTERLEAVE, [0x00000000000002, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000, 0000000000000000], 1025) = 0
38676 06:59:14.501691 +++ exited with 0 +++
```

For your reference, when I run numactl, it uses a much larger value for maxnode.

crates/libcontainer/src/process/init/process.rs
```diff
@@ -831,6 +833,330 @@ fn setup_scheduler(sc_op: &Option<Scheduler>) -> Result<()> {
     Ok(())
 }
 
+fn setup_memory_policy(
```
saku3: (Before we proceed with the overall review.)
For maintainability, please consider moving setup_memory_policy, build_nodemask, and parse_node_string into a separate file like memorypolicy.rs.

crates/libcontainer/src/process/init/process.rs
n4mlz: @saku3 
Thank you very much for this thorough response!
I realize now that my initial verification was insufficient. When I set up a virtual machine with NUMA node number 2 and retested the functionality, I encountered exactly the same error you pointed out.

As you've correctly noted, the `maxnode` parameter in `set_mempolicy(2)` requires the total number of bits in the `nodemask` array. I've made the required modifications to properly prepare the `nodemask` in the same format as runc. Could you please verify this again?

cf: In my environment, the test results looked like this:

```
$ cat /sys/devices/system/node/online
0-1

$ uname -a
Linux kernelhack 6.1.0-37-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux

$ just test-contest memory_policy
/home/n4mlz/youki/scripts/build.sh -o /home/n4mlz/youki -r -c youki
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
    Finished `release` profile [optimized] target(s) in 0.30s
/home/n4mlz/youki/scripts/build.sh -o /home/n4mlz/youki -r -c contest
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
removed '/home/n4mlz/youki/contest'
    Finished `release` profile [optimized] target(s) in 0.30s
removed '/home/n4mlz/youki/runtimetest'
    Finished `release` profile [optimized] target(s) in 0.27s
sudo /home/n4mlz/youki/scripts/contest.sh /home/n4mlz/youki/youki memory_policy
Validation successful for runtime /home/n4mlz/youki/youki
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+    Syscall(#[from] SyscallError),
+}
+
+type Result<T> = std::result::Result<T, MemoryPolicyError>;
+
+pub fn setup_memory_policy(
+    memory_policy: &Option<oci_spec::runtime::LinuxMemoryPolicy>,
+    syscall: &dyn Syscall,
+) -> Result<()> {
+    if let Some(policy) = memory_policy {
```
saku3: How about using early returns? The `if` nesting is deep.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+        };
+
+        let mut flags_value: u32 = 0;
+        if let Some(flags) = policy.flags() {
+            let mut has_static = false;
+            let mut has_relative = false;
+
+            for flag in flags {
+                match flag {
+                    oci_spec::runtime::MemoryPolicyFlagType::MpolFNumaBalancing => {
```
saku3: Please add a use import so `MemoryPolicyFlagType` can be referenced directly.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+                        // MPOL_F_NUMA_BALANCING is only valid with MPOL_BIND
+                        if base_mode != MPOL_BIND {
+                            return Err(MemoryPolicyError::IncompatibleFlagMode(
+                                "MPOL_F_NUMA_BALANCING can only be used with MPOL_BIND".to_string(),
+                            ));
+                        }
+                        flags_value |= MPOL_F_NUMA_BALANCING;
+                    }
+                    oci_spec::runtime::MemoryPolicyFlagType::MpolFRelativeNodes => {
```
saku3: Same here `MemoryPolicyFlagType`

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+                        has_relative = true;
+                        flags_value |= MPOL_F_RELATIVE_NODES;
+                    }
+                    oci_spec::runtime::MemoryPolicyFlagType::MpolFStaticNodes => {
```
saku3: Same here `MemoryPolicyFlagType`

tests/contest/runtimetest/src/main.rs
```diff
@@ -44,6 +44,34 @@ fn main() {
         "io_priority_class_rt" => tests::test_io_priority_class(&spec, IoprioClassRt),
         "io_priority_class_be" => tests::test_io_priority_class(&spec, IoprioClassBe),
         "io_priority_class_idle" => tests::test_io_priority_class(&spec, IoprioClassIdle),
+        "memory_policy_default" => {
+            tests::validate_memory_policy(&spec, Some(MemoryPolicyModeType::MpolDefault))
+        }
+        "memory_policy_empty" => tests::validate_memory_policy(&spec, None),
```
saku3: I think the `memory_policy_empty` test case is checking the same thing as `hello_world`, so it’s fine to omit it.

tests/contest/runtimetest/src/main.rs
```diff
@@ -44,6 +44,34 @@ fn main() {
         "io_priority_class_rt" => tests::test_io_priority_class(&spec, IoprioClassRt),
         "io_priority_class_be" => tests::test_io_priority_class(&spec, IoprioClassBe),
         "io_priority_class_idle" => tests::test_io_priority_class(&spec, IoprioClassIdle),
+        "memory_policy_default" => {
```
saku3: Can `expected_mode` be derived from `spec`?
If so, the code below should be reducible to a single line.

`"memory_policy" => tests::validate_memory_policy(&spec)`

(In `tests/contest/contest/src/tests/memory_policy/memory_policy_test.rs`, we can assume that the spec created there is passed through unchanged. Verifying that behavior is not within the scope of the memory\_policy tests.
)

tests/contest/contest/src/tests/memory_policy/memory_policy_test.rs
```diff
@@ -0,0 +1,302 @@
...
+    LinuxBuilder, LinuxMemoryPolicyBuilder, MemoryPolicyFlagType, MemoryPolicyModeType,
+    ProcessBuilder, Spec, SpecBuilder,
+};
+use serde_json::json;
+use test_framework::{ConditionalTest, Test, TestGroup, TestResult};
+
+use crate::utils::test_inside_container;
+use crate::utils::test_utils::CreateOptions;
+
+fn has_multi_numa_nodes() -> bool {
```
saku3: The tests here verify whether the NUMA policy is configured, not the behavior across NUMA nodes, so they’re meaningful even on a single-node system. Therefore, I don’t think `has_multi_numa_nodes` is necessary.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+    #[error("Mutually exclusive flags: {0}")]
+    MutuallyExclusiveFlags(String),
+
+    #[error("Syscall error: {0}")]
+    Syscall(#[from] SyscallError),
+}
+
+type Result<T> = std::result::Result<T, MemoryPolicyError>;
+
+pub fn setup_memory_policy(
```
saku3: Please include this link as a reference: [https://man7.org/linux/man-pages/man2/set\_mempolicy.2.html](https://man7.org/linux/man-pages/man2/set_mempolicy.2.html)

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,6 +545,139 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+                }
+            }
+        }
+        (nodes_is_empty, has_static, has_relative)
+    } else {
+        (true, false, false)
+    };
+
+    // Verify the policy matches expected mode
+    match expected_mode {
```
saku3, line 620: Please ensure that, in the normal (successful) path, when nodes is specified, you verify that the allocation is on NUMA node 0; this information should be obtainable from the spec.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+            MemoryPolicyModeType::MpolDefault => MPOL_DEFAULT,
+            MemoryPolicyModeType::MpolPreferred => MPOL_PREFERRED,
+            MemoryPolicyModeType::MpolBind => MPOL_BIND,
+            MemoryPolicyModeType::MpolInterleave => MPOL_INTERLEAVE,
+            MemoryPolicyModeType::MpolLocal => MPOL_LOCAL,
+            MemoryPolicyModeType::MpolPreferredMany => MPOL_PREFERRED_MANY,
+            MemoryPolicyModeType::MpolWeightedInterleave => MPOL_WEIGHTED_INTERLEAVE,
+        };
+
+        let mut flags_value: u32 = 0;
```
saku3: Personally, I think it would be easier to read if the validation and the construction of flags_value were separated.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+                        .to_string(),
+                ));
+            }
+        }
+
+        // Combine mode and flags
+        let mode_with_flags = base_mode | (flags_value as i32);
+
+        // Handle special cases for MPOL_DEFAULT and MPOL_LOCAL
+        match base_mode {
```
saku3: Personally, I think it would be clearer to separate validation from the application of set_mempolicy.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+            MPOL_DEFAULT => {
```
saku3: Can the validation for MPOL_DEFAULT and MPOL_LOCAL be consolidated?

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+                            })?;
+                    }
+                    Some(nodes) => {
+                        // Parse nodes and build nodemask
+                        let (nodemask, maxnode) = build_nodemask(nodes)?;
+
+                        // Double-check: if the parsed result is empty (e.g., whitespace-only input),
+                        // treat as empty nodemask
+                        if maxnode == 0 {
+                            if flags_value & (MPOL_F_RELATIVE_NODES | MPOL_F_STATIC_NODES) != 0 {
```
saku3: memo: https://docs.kernel.org/admin-guide/mm/numa_memory_policy.html

```
It is possible for the user to specify that local allocation is always preferred by passing an empty nodemask with this mode. If an empty nodemask is passed, the policy cannot use the MPOL_F_STATIC_NODES or MPOL_F_RELATIVE_NODES flags described below.
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,542 @@
...
+            .as_any()
+            .downcast_ref::<TestHelperSyscall>()
+            .unwrap()
+            .get_mempolicy_args();
+
+        assert_eq!(got_args.len(), 1);
+        assert_eq!(got_args[0].mode, 2); // MPOL_BIND (corrected value)
+        assert_eq!(got_args[0].nodemask.len(), 1); // One c_ulong needed
+        assert_eq!(got_args[0].nodemask[0], 3); // 2^0 + 2^1 = 1 + 2 = 3
+        assert_eq!(got_args[0].maxnode, 2); // highest node ID (1) + 1 = 2
```
saku3: Since `maxnode` is calculated as `(num_u64s * u64_bits) as u64` above, the test may be incorrect.

saku3 (commented):
I’ve reviewed it, and the overall structure looks good.
I left comments on the finer details—when you have a moment, please consider addressing them. If anything is unclear, feel free to ask anytime.

n4mlz:
Thank you for the thorough code review.
Sorry for the delay!
I've made all the requested changes. Please take a look.

On a machine with two NUMA nodes, I have verified that the tests pass as follows:

```
$ cat /sys/devices/system/node/online
0-1

$ just test-contest memory_policy
/home/n4mlz/youki/scripts/build.sh -o /home/n4mlz/youki -r -c youki
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
    Finished `release` profile [optimized] target(s) in 0.29s
/home/n4mlz/youki/scripts/build.sh -o /home/n4mlz/youki -r -c contest
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
removed '/home/n4mlz/youki/contest'
    Finished `release` profile [optimized] target(s) in 0.30s
removed '/home/n4mlz/youki/runtimetest'
    Finished `release` profile [optimized] target(s) in 0.27s
sudo /home/n4mlz/youki/scripts/contest.sh /home/n4mlz/youki/youki memory_policy
Validation successful for runtime /home/n4mlz/youki/youki

$
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,514 @@
...
+    if node_ids.is_empty() {
+        // Empty nodemask - return NULL equivalent (empty vector)
+        return Ok((Vec::new(), 0));
+    }
+
+    // Find the highest node ID
+    let highest_node = node_ids.iter().max().copied().unwrap_or(0) as usize;
+
+    // Calculate how many c_ulong values we need to store the bitmask
+    let bits_per_ulong = std::mem::size_of::<libc::c_ulong>() * 8;
```
n4mlz, line 279: u_long may have different bit widths depending on the environment, so it is treated as libc::u_long rather than u64.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,514 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum MemoryPolicyError {
+    #[error("Invalid memory policy flag: {0}")]
+    InvalidFlag(String),
+
+    #[error("Invalid node specification: {0}")]
+    InvalidNodes(String),
+
+    #[error("Invalid memory policy mode: {0}")]
+    InvalidMode(String),
```
saku3: Is this flag used anywhere?

crates/libcontainer/src/process/init/process.rs
```diff
@@ -20,6 +20,7 @@ use crate::error::MissingSpecError;
 use crate::namespaces::Namespaces;
 use crate::process::args::{ContainerArgs, ContainerType};
 use crate::process::channel;
+use crate::process::memory_policy;
```
saku3: Running `just lint` shows the following:
```suggestion
use crate::process::{channel, memory_policy};
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,514 @@
+use oci_spec::runtime::MemoryPolicyFlagType;
```
saku3: Running `just lint` shows the following:

```suggestion
use oci_spec::runtime::{MemoryPolicyFlagType, MemoryPolicyModeType};
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,514 @@
...
+                    nodemask,
+                    maxnode,
+                }))
+            }
+        },
+        _ => {
+            let nodes = match policy.nodes() {
+                None => {
+                    return Err(MemoryPolicyError::InvalidNodes(format!(
+                        "Mode {} requires non-empty node specification",
```
saku3, line 219:
```
    "memoryPolicy": {
      "mode": "MPOL_INTERLEAVE",
      "nodes": ""
    },
```

Here is the error you get when specifying the above.

```
ERROR youki: error in executing command: failed to create container: exec process failed with error error in executing process : Invalid node specification: Mode 3 requires non-empty node specification
```

Outputting raw numbers as-is makes it hard to understand.
It would be preferable to output each mode.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,6 +545,208 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+            }
+            // Optional: check flag display
+            if has_static_flag && !policy_field.contains("static") {
+                eprintln!("expected bind static, but found: {}", policy_field);
+            }
+            if expected_nodes.as_ref().is_some_and(|v| v.contains(&0))
+                && !numa_maps_indicates_node0(first_line, policy_field)
+            {
+                eprintln!(
+                    "expected interleave including node0, but got: {}",
```
saku3: bind?

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,6 +545,208 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+    let linux = spec.linux().as_ref()?;
+    let mp = linux.memory_policy().as_ref()?;
+    let nodes = mp.nodes().as_deref()?;
+    Some(parse_node_string(nodes))
+}
+
+fn numa_maps_indicates_node0(first_line: &str, policy_field: &str) -> bool {
+    policy_field.contains("bind:0")
+        || policy_field.contains("prefer:0")
+        || first_line.contains("N0=")
```
saku3: I think this line is unnecessary.

```
|| first_line.contains("N0=")
```

tests/contest/runtimetest/src/tests.rs
```diff
@@ -545,6 +545,208 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+    }
+    let expected = expected_mode.unwrap();
+
+    if memory_policy.is_none() {
+        eprintln!("memory policy expected but not found in spec");
+        return;
+    }
+
+    let policy = memory_policy.as_ref().unwrap();
+    if policy.mode() != expected {
```
saku3: It’s unnecessary because it always matches.
```
let expected_mode = memory_policy.as_ref().map(|p| p.mode());
let expected = expected_mode.unwrap();
let policy = memory_policy.as_ref().unwrap();
policy.mode()
```

saku3 (commented):
Thanks for updating the changes. I really appreciate your thoughtful handling. I left a few additional notes—when you have time, please check them.

n4mlz:
Thank you for your always thorough reviews. The corrections have been completed. Please check them.

saku3:
Please wait a little while until oci-spec-rs is released🙏

saku3:
Please resolve the conflict.

n4mlz:
@saku3 
I apologize for the delay🙏
When I updated the version of runc used in integration tests to one that supports Linux memory policy (v1.4.0-rc.2), it seems there was a change in how cpu.weight is calculated in cgroup v2, causing the CI to fail (cf: https://github.com/n4mlz/youki/actions/runs/18583292135/job/52982229747?pr=3#step:9:64).
I have created a PR (https://github.com/youki-dev/youki/pull/3274) to update to the first version where this change occurred (v1.3.2), so I would appreciate it if you could merge this into main first.

n4mlz:
@saku3 
Thank you for waiting. I have updated the version of runc for CI, confirmed that the CI completes successfully in my forked repository, and am ready to run the tests. I would appreciate it if you could check it.

saku3:
Sorry for the delay.
I’ll review it in the coming days.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -561,6 +561,201 @@ pub fn test_io_priority_class(spec: &Spec, io_priority_class: IOPriorityClass) {
...
+        }
+        Some(MemoryPolicyModeType::MpolBind) => {
+            let (policy_field, full_line) = find_with_substring("bind");
+            if !policy_field.contains("bind") {
+                eprintln!("expected bind policy, but found: {}", policy_field);
+            }
+            if has_static_flag && !policy_field.contains("static") {
+                eprintln!("expected bind static, but found: {}", policy_field);
+            }
+            if expected_nodes.as_ref().is_some_and(|v| v.contains(&0))
```
saku3: I’d like it to support checking values other than 0, too.

example image
```
const EXPECT_NODE = 1;
if expected_nodes.as_ref().is_some_and(|v| v.contains(EXPECT_NODE))
 && !numa_maps_expected_node(policy_field) // check for EXPECT_NODE
```

I know we’ll specify only 0 on GitHub Actions, but I want to be able to specify `1`,`0-2`,  (or etc.) locally and still run valid tests.
Since you’ve also implemented functions `parse_node_string`, I believe it’s feasible—what do you think?

.github/workflows/integration_tests_validation.yaml
```diff
@@ -92,9 +92,11 @@ jobs:
         run: sudo add-apt-repository -y ppa:criu/ppa
       - name: Install requirements
         run: sudo env PATH=$PATH just ci-prepare
-      - name: Install runc 1.3.2
```
saku3: I’m sorry for the trouble.
`validate-contest-runc` is intended to verify the validity of newly added integration tests by running them with runc.
Thinking ahead to other integration tests we might add, I believe it’s best to use the stable version here.
If the memory policy test fails, we can skip it.
Please refer to the following.

https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/tests/fd_control/mod.rs#L100

tests/contest/runtimetest/src/tests.rs
n4mlz: Thanks for the suggestion.
Supporting nodes other than `0` makes sense. However, I realized that the current test logic itself wasn’t actually correct; even the original implementation didn’t properly handle the intended semantics when `MPOL_F_RELATIVE_NODES` is used.

When this flag is set, the nodemask is interpreted relative to the process’s cpuset, not as absolute physical node IDs.
So a nodemask containing `0` doesn’t necessarily mean physical node 0 — it could mean “the first node in `Mems_allowed_list`”.
In that case, `/proc/self/numa_maps` will legitimately show a different node number, so our current assumption is not correct.

Reference:

* [`set_mempolicy(2)` — MPOL_F_RELATIVE_NODES / MPOL_F_STATIC_NODES](https://man7.org/linux/man-pages/man2/set_mempolicy.2.html)
* [Kernel docs: *NUMA Memory Policy* (flags section)](https://docs.kernel.org/admin-guide/mm/numa_memory_policy.html)

I’ll update the test so that it maps relative nodemasks to the actual physical nodes (using `Mems_allowed_list` from `/proc/self/status`) before comparing them with `/proc/self/numa_maps`.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -592,27 +593,80 @@ fn parse_node_string(nodes: &str) -> Vec<u32> {
     out
 }
 
-fn expected_nodes_from_spec(spec: &Spec) -> Option<Vec<u32>> {
+fn mems_allowed_list() -> Option<Vec<u32>> {
```
n4mlz, line 600: I’ve added logic to handle the `MPOL_F_RELATIVE_NODES` case by reading `Mems_allowed_list` from `/proc/self/status` and comparing the translated physical nodes against `/proc/self/numa_maps`.
This makes the test technically correct, but the logic became quite large and complex.
If you think this level of validation is too heavy for an integration test, I can remove or trim it down. What do you think?

n4mlz: In an environment with two NUMA nodes, I confirmed that the tests pass as follows:
```
$ cat /sys/devices/system/node/online
0-1
$ uname -a
Linux kernelhack 6.1.0-37-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux
$ just test-contest memory_policy
/home/n4mlz/youki/scripts/build.sh -o /home/n4mlz/youki -r -c youki
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
    Finished `release` profile [optimized] target(s) in 0.29s
/home/n4mlz/youki/scripts/build.sh -o /home/n4mlz/youki -r -c contest
* FEATURES: <default>
* TARGET: x86_64-unknown-linux-gnu
removed '/home/n4mlz/youki/contest'
    Finished `release` profile [optimized] target(s) in 0.28s
removed '/home/n4mlz/youki/runtimetest'
    Finished `release` profile [optimized] target(s) in 0.27s
sudo /home/n4mlz/youki/scripts/contest.sh /home/n4mlz/youki/youki memory_policy
Validation successful for runtime /home/n4mlz/youki/youki
$
```

saku3: Reflecting on the comment above, it might be worth adding the following cases:

```
{ "mode": "MPOL_LOCAL" }
{ "mode": "MPOL_WEIGHTED_INTERLEAVE", "nodes": "0" }
{ "mode": "MPOL_PREFERRED_MANY", "nodes": "0" }
```

saku3: Thank you for the thoughtful feedback.

When `MPOL_F_RELATIVE_NODES` or `MPOL_F_STATIC_NODES` is specified, I think it’s sufficient to verify that `relative` or `static` appears in `/proc/self/numa_maps`.

If we design tests to verify that, when multiple nodes are specified, the nodes are honored exactly as specified, it becomes quite complex—not only due to `MPOL_F_RELATIVE_NODES` but also features like `MPOL_WEIGHTED_INTERLEAVE`. I don’t think we need to account for that level of complexity.

For multiple nodes, I think we can stick to the simplest case. For example:

```json
"mode": "MPOL_BIND",
"nodes": "0-1"
```

If we can verify that this is reflected in `/proc/self/numa_maps`, that should be sufficient to ensure Youki’s handling of “multiple nodes passed to the syscall” works correctly for other modes and flags as well.  

Rather than testing every (node (one-or-multiple) × mode × flag) combination, we’ll split coverage by dimension: 
* Provide test cases for multiple nodes.
* Provide test cases for each mode. 
* Provide test cases for each flag. 

This gives us confidence in the behavior without enumerating all cross-product combinations.
So as long as the test covers a simple case where multiple nodes work correctly, there’s no problem.

saku3: If we have enough capacity, it might be good to add the following test cases as well:

```json
{ "mode": "MPOL_LOCAL" }
{ "mode": "MPOL_WEIGHTED_INTERLEAVE", "nodes": "0" }
{ "mode": "MPOL_PREFERRED_MANY", "nodes": "0" }
{ "mode": "MPOL_BIND", "nodes": "0", "flags": ["MPOL_F_NUMA_BALANCING"] }
```

saku3 (approved):
@n4mlz 
Please resolve the conflicts.

Overall, I think it's mostly fine from my side.
Thank you for patiently working through all the changes.

saku3:
@utam0k 
PTAL 🙏

crates/libcgroups/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libcgroups"
-version = "0.5.7" # MARK: Version
+version = "0.5.7"                                 # MARK: Version
```
utam0k: I'm unsure about the reason for this change.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,525 @@
...
+    const MPOL_BIND: i32 = 2;
+    const MPOL_INTERLEAVE: i32 = 3;
+    const MPOL_LOCAL: i32 = 4;
+    const MPOL_PREFERRED_MANY: i32 = 5;
+    const MPOL_WEIGHTED_INTERLEAVE: i32 = 6;
+
+    // Memory policy flag constants from Linux UAPI
+    const MPOL_F_NUMA_BALANCING: u32 = 1 << 13; // 0x2000
+    const MPOL_F_RELATIVE_NODES: u32 = 1 << 14; // 0x4000
+    const MPOL_F_STATIC_NODES: u32 = 1 << 15; // 0x8000
```
utam0k: An `enum` is best suited for two scenarios.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,525 @@
...
+
+    let base_mode = match policy.mode() {
+        MemoryPolicyModeType::MpolDefault => MPOL_DEFAULT,
+        MemoryPolicyModeType::MpolPreferred => MPOL_PREFERRED,
+        MemoryPolicyModeType::MpolBind => MPOL_BIND,
+        MemoryPolicyModeType::MpolInterleave => MPOL_INTERLEAVE,
+        MemoryPolicyModeType::MpolLocal => MPOL_LOCAL,
+        MemoryPolicyModeType::MpolPreferredMany => MPOL_PREFERRED_MANY,
+        MemoryPolicyModeType::MpolWeightedInterleave => MPOL_WEIGHTED_INTERLEAVE,
+    };
```
utam0k: You can use the `From` trait.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,525 @@
...
+
+    let mut has_static = false;
+    let mut has_relative = false;
+    if let Some(flags) = policy.flags() {
+        for flag in flags.iter() {
```
utam0k:
```suggestion
    if let Some(flags) = policy.flags() {
        let mut has_static = false;
        let mut has_relative = false;
        for flag in flags.iter() {
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,525 @@
...
+                            "MPOL_F_NUMA_BALANCING can only be used with MPOL_BIND".to_string(),
+                        ));
+                    }
+                }
+                MemoryPolicyFlagType::MpolFRelativeNodes => {
+                    has_relative = true;
+                }
+                MemoryPolicyFlagType::MpolFStaticNodes => {
+                    has_static = true;
+                }
```
utam0k: Using flags.contains(...) makes this logic simpler.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,544 @@
...
+                MemoryPolicyFlagType::MpolFRelativeNodes => {
+                    flags_value |= u32::from(MemoryPolicyFlag::RelativeNodes);
+                }
+                MemoryPolicyFlagType::MpolFStaticNodes => {
+                    flags_value |= u32::from(MemoryPolicyFlag::StaticNodes);
+                }
+            }
+        }
+    }
```
utam0k:
```suggestion
  let (flags_value, has_static, has_relative) = policy
      .flags()
      .map(|flags| {
          flags.iter().fold((0u32, false, false), |(val, s, r), flag| {
              match flag {
                  MpolFNumaBalancing => (val | MPOL_F_NUMA_BALANCING, s, r),
                  MpolFStaticNodes => (val | MPOL_F_STATIC_NODES, true, r),
                  MpolFRelativeNodes => (val | MPOL_F_RELATIVE_NODES, s, true),
              }
          })
      })
      .unwrap_or((0, false, false));

```

utam0k: Simplify it and remove one iteration over flags.

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    for range in nodes.split(',') {
+        let range = range.trim();
+        if range.is_empty() {
+            continue; // Skip empty entries caused by multiple commas
+        }
+        let range = range.trim();
+        if range.is_empty() {
+            continue;
+        }
```
utam0k:
```suggestion
        let range = range.trim();
        if range.is_empty() {
            continue; // Skip empty entries caused by multiple commas
        }
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    let mode_with_flags = i32::from(base_mode) | (flags_value as i32);
+
+    match base_mode {
+        MemoryPolicyMode::Default | MemoryPolicyMode::Local => {
+            let mode_name = if base_mode == MemoryPolicyMode::Default {
+                "MPOL_DEFAULT"
+            } else {
+                "MPOL_LOCAL"
+            };
```
utam0k: Please implement the ToString trait in MemoryPolicyMode.

crates/libcontainer/src/process/init/error.rs
```diff
@@ -64,6 +65,10 @@ pub enum InitProcessError {
     IoPriorityClass(String),
     #[error("call exec sched_setattr error: {0}")]
     SchedSetattr(String),
+    #[error("invalid memory policy: {0}")]
+    InvalidMemoryPolicy(String),
+    #[error(transparent)]
+    MemoryPolicy(#[from] MemoryPolicyError),
```
utam0k, line 69: Is it used?

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -932,6 +932,30 @@ impl Syscall for LinuxSyscall {
...
+                mode as libc::c_long,
+                libc_nodemask,
+                libc_maxnode,
+            )
+        } {
+            0 => Ok(()),
+            -1 => Err(nix::Error::last()),
+            _ => Err(nix::Error::UnknownErrno),
+        }?;
+        Ok(())
```
utam0k:
```suggestion
        match unsafe {
            libc::syscall(
                libc::SYS_set_mempolicy,
                mode as libc::c_long,
                libc_nodemask,
                libc_maxnode,
            )
        } {
            0 => Ok(()),
            -1 => Err(nix::Error::last()),
            _ => Err(nix::Error::UnknownErrno),
        }
```

crates/libcontainer/src/process/memory_policy.rs
```diff
@@ -0,0 +1,544 @@
...
+        MemoryPolicyMode::Preferred => {
+            let relative_or_static: u32 = u32::from(MemoryPolicyFlag::RelativeNodes)
+                | u32::from(MemoryPolicyFlag::StaticNodes);
+            match policy.nodes() {
+                None => {
+                    if flags_value & relative_or_static != 0u32 {
+                        return Err(MemoryPolicyError::IncompatibleFlagMode(
+                            "MPOL_PREFERRED with empty nodes cannot use MPOL_F_STATIC_NODES or MPOL_F_RELATIVE_NODES flags".to_string(),
+                        ));
+                    }
```
utam0k: This logic appears three times. Follow DRY.

crates/libcontainer/src/process/init/error.rs
n4mlz: `MemoryPolicy(#[from] MemoryPolicyError)` is being used.

It is utilized on line 57 of `crates/libcontainer/src/process/init/process.rs` as follows:

```
memory_policy::setup_memory_policy(ctx.linux.memory_policy(), ctx.syscall.as_ref())?;
```

The `setup_memory_policy` function returns `Result<(), MemoryPolicyError>`.

`InvalidMemoryPolicy(String)` was not in use, so I have removed it.

n4mlz:
Thank you for your thorough review. I apologize for the delay in making the corrections. Please take a look.

PR #3252, opened 2025-09-22
Handle frozen cgroup properly when creating container

### LINKED ISSUE #3233: [Bug]: youki hangs when linux.cgroupsPath points to a pre-frozen cgroup
### Bug Description

If config.json’s linux.cgroupsPath points to a pre-frozen cgroup and you run youki run, youki appears to hang .
In contrast, runc detects this case and fails with an error, but it seems youki doesn’t have an equivalent check.

### Steps to Reproduce

Create a frozen cgroup:

```
REL=/runc-cgroups-integration-test/test-cgroup-$$ 
CG=/sys/fs/cgroup$REL
sudo mkdir -p "$CG"
echo 1 | sudo tee "$CG/cgroup.freeze" >/dev/null 
cat "$CG/cgroup.events" 
populated 0
frozen 1 
```

Update config.json’s cgroupsPath:

```
ls 
config.json

 jq '.linux.cgroupsPath = "'$REL'"' config.json | sudo tee config.json >/dev/null
```

It should look like this:
```
  "linux": {
...
"cgroupsPath": "/runc-cgroups-integration-test/test-cgroup-4069"
}
```

When starting youki, you can see it stalls mid-run:

```
./youki run -b tutorial/ container
DEBUG youki: started by user 0 with ArgsOs { inner: ["./youki", "run", "-b", "tutorial/", "container"] }
DEBUG libcontainer::user_ns: this container does NOT create a new user namespace
DEBUG libcontainer::container::init_builder: container directory will be "/run/youki/container"
DEBUG libcontainer::container::container: Save container status: Container { state: State { oci_version: "v1.0.2", id: "container", status: Creating, pid: None, bundle: "/home/ubuntu/workspace/youki/tutorial", annotations: Some({}), created: None, creator: None, use_systemd: false, clean_up_intel_rdt_subdirectory: None }, root: "/run/youki/container" } in "/run/youki/container"
DEBUG libcontainer::user_ns: this container does NOT create a new user namespace
DEBUG libcontainer::notify_socket: create notify listener socket_path="/run/youki/container/notify.sock"
DEBUG libcontainer::notify_socket: the cwd to create the notify socket cwd="/home/ubuntu/workspace/youki"
 INFO libcgroups::common: cgroup manager V2 will be used
DEBUG libcontainer::process::cpu_affinity: affinity: 0x3
```

With runc, container creation fails like this:

```
 runc run -b tutorial/ container
ERRO[0000] runc run failed: container's cgroup unexpectedly frozen 
```

### Expectation

Fail fast on run and on create if linux.cgroupsPath points to a cgroup that is already frozen.

Return a non-zero exit code and print a clear error similar to runc:
` container's cgroup unexpectedly frozen`

https://github.com/opencontainers/runc/blob/main/tests/integration/cgroups.bats#L518

### System and Setup Info

_No response_

### Additional Context

wyhao31:
## Description
- Added `get_freezer_state` in `CgroupManager` trait. Similar to `GetFreezerState` in https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/cgroups.go#L76
- Implemented `get_freezer_state` for `v1/v2/systemd` in `libcgroups` crate.
  - For `v2/systemd`, reused the existing `read_freezer_state` function with some small change.
  - For `v1`, it's a little bit tricky. I referenced the implementation from https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/fs/freezer.go#L114.
- Added cgroup frozen check in `container_intermediate_process`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite
I have run unit tests locally.
Tested using the reproduce steps mentioned in the issue(#3233).
```
$ sudo ./youki run -b tutorial/ container
ERROR libcontainer::process::container_main_process: failed to run intermediate process cgroup error: container's cgroup unexpectedly frozen
ERROR libcontainer::container::builder_impl: failed to run container process intermediate process error cgroup error: container's cgroup unexpectedly frozen
ERROR youki: error in executing command: failed to create container: intermediate process error cgroup error: container's cgroup unexpectedly frozen
run failed : failed to create container: intermediate process error cgroup error: container's cgroup unexpectedly frozen
```

## Related Issues

Fixes #3233

### DIFF
--- a/crates/libcgroups/src/v1/freezer.rs
+++ b/crates/libcgroups/src/v1/freezer.rs
@@ -1,4 +1,5 @@
 use std::fs::OpenOptions;
+use std::fs::read_to_string;
 use std::io::Read;
 use std::path::Path;
 use std::{thread, time};
@@ -7,6 +8,7 @@ use super::controller::Controller;
 use crate::common::{self, ControllerOpt, FreezerState, WrapIoResult, WrappedIoError};
 
 const CGROUP_FREEZER_STATE: &str = "freezer.state";
+const CGROUP_FREEZER_SELF_STATE: &str = "freezer.self_freezing";
 const FREEZER_STATE_THAWED: &str = "THAWED";
 const FREEZER_STATE_FROZEN: &str = "FROZEN";
 const FREEZER_STATE_FREEZING: &str = "FREEZING";
@@ -79,9 +81,15 @@ impl Freezer {
                         if i % 25 == 24 {
                             thread::sleep(time::Duration::from_millis(10));
                         }
-
-                        let r = Self::read_freezer_state(cgroup_root)?;
-                        match r.trim() {
+                        let mut state = String::new();
+                        OpenOptions::new()
+                            .create(false)
+                            .read(true)
+                            .open(cgroup_root.join(CGROUP_FREEZER_STATE))
+                            .wrap_open(cgroup_root)?
+                            .read_to_string(&mut state)
+                            .wrap_read(cgroup_root)?;
+                        match state.trim() {
                             FREEZER_STATE_FREEZING => {
                                 continue;
                             }
@@ -93,7 +101,7 @@ impl Freezer {
                             }
                             _ => {
                                 // should not reach here.
-                                return Err(V1FreezerControllerError::UnexpectedState { state: r });
+                                return Err(V1FreezerControllerError::UnexpectedState { state: state });
                             }
                         }
                     }
@@ -114,17 +122,43 @@ impl Freezer {
         Ok(())
     }
 
-    fn read_freezer_state(cgroup_root: &Path) -> Result<String, WrappedIoError> {
+    pub fn read_freezer_state(cgroup_root: &Path) -> FreezerState {
         let path = cgroup_root.join(CGROUP_FREEZER_STATE);
-        let mut content = String::new();
-        OpenOptions::new()
-            .create(false)
-            .read(true)
-            .open(path)
-            .wrap_open(cgroup_root)?
-            .read_to_string(&mut content)
-            .wrap_read(cgroup_root)?;
-        Ok(content)
+        loop {
+            let state = match read_to_string(&path) {
+                Ok(state) => state,
+                Err(_) => {
+                    return FreezerState::Undefined;
+                }
+            };
+            match state.trim() {
+                FREEZER_STATE_THAWED => return FreezerState::Thawed,
+                FREEZER_STATE_FROZEN => {
+                    // Find out whether the cgroup is frozen directly,
+                    // or indirectly via an ancestor.
+                    let self_state = match read_to_string(path.join(CGROUP_FREEZER_SELF_STATE)) {
+                        Ok(self_state) => self_state,
+                        Err(_) => {
+                            return FreezerState::Frozen;
+                        }
+                    };
+                    return match self_state.as_str() {
+                        "0\n" => FreezerState::Thawed,
+                        "1\n" => FreezerState::Frozen,
+                        _ => FreezerState::Undefined,
+                    };
+                }
+                FREEZER_STATE_FREEZING => {
+                    // Make sure we get a stable freezer state, so retry if the cgroup
+                    // is still undergoing freezing. This should be a temporary delay.
+                    thread::sleep(time::Duration::from_millis(1));
+                    continue;
+                }
+                _ => {
+                    return FreezerState::Undefined;
+                }
+            }
+        }
     }
 }
 
--- a/crates/libcgroups/src/v1/freezer.rs
+++ b/crates/libcgroups/src/v1/freezer.rs
@@ -1,5 +1,4 @@
-use std::fs::OpenOptions;
-use std::fs::read_to_string;
+use std::fs::{OpenOptions, read_to_string};
 use std::io::Read;
 use std::path::Path;
 use std::{thread, time};
@@ -101,7 +100,9 @@ impl Freezer {
                             }
                             _ => {
                                 // should not reach here.
-                                return Err(V1FreezerControllerError::UnexpectedState { state: state });
+                                return Err(V1FreezerControllerError::UnexpectedState {
+                                    state: state,
+                                });
                             }
                         }
                     }
--- a/crates/libcgroups/src/v1/freezer.rs
+++ b/crates/libcgroups/src/v1/freezer.rs
@@ -100,9 +100,7 @@ impl Freezer {
                             }
                             _ => {
                                 // should not reach here.
-                                return Err(V1FreezerControllerError::UnexpectedState {
-                                    state: state,
-                                });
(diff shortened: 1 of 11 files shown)


crates/libcgroups/src/v1/freezer.rs
```diff
@@ -114,17 +122,43 @@ impl Freezer {
...
+                        "1\n" => FreezerState::Frozen,
+                        _ => FreezerState::Undefined,
+                    };
+                }
+                FREEZER_STATE_FREEZING => {
+                    // Make sure we get a stable freezer state, so retry if the cgroup
+                    // is still undergoing freezing. This should be a temporary delay.
+                    thread::sleep(time::Duration::from_millis(1));
+                    continue;
+                }
```
YJDoc2, line 155: Technically this will exit after a couple of tries, but I feel we should still have a counter and return undefined/frozen after n-retries, just so we can be sure we'll never get stuck in infinite loop

crates/libcgroups/src/v1/freezer.rs
```diff
@@ -114,17 +122,43 @@ impl Freezer {
...
+                        Err(_) => {
+                            return FreezerState::Frozen;
+                        }
+                    };
+                    return match self_state.as_str() {
+                        "0\n" => FreezerState::Thawed,
+                        "1\n" => FreezerState::Frozen,
+                        _ => FreezerState::Undefined,
+                    };
+                }
```
YJDoc2, line 149: I'm not sure this is what we should follow. From kernel docs
```
The cgroup freezer is hierarchical. Freezing a cgroup freezes all tasks belonging to 
the cgroup and all its descendant cgroups. Each cgroup has its own state (self-state) 
and the state inherited from the parent (parent-state). Iff both states are THAWED, 
the cgroup is THAWED.
```
https://docs.kernel.org/admin-guide/cgroup-v1/freezer-subsystem.html

So, if we are getting frozen, we should return frozen even if self_state is thawed, right?

crates/libcgroups/src/v1/freezer.rs
```diff
@@ -79,9 +81,15 @@ impl Freezer {
...
-                        match r.trim() {
+                        let mut state = String::new();
+                        OpenOptions::new()
+                            .create(false)
+                            .read(true)
+                            .open(cgroup_root.join(CGROUP_FREEZER_STATE))
+                            .wrap_open(cgroup_root)?
+                            .read_to_string(&mut state)
+                            .wrap_read(cgroup_root)?;
+                        match state.trim() {
```
YJDoc2, line 91: is it possible to re used the read_freezer_state here instead of doing a read again?

crates/libcgroups/src/v1/freezer.rs
wyhao31: Thanks for the suggestion. But I feel like I don't want to introduce another parameter (n-retries) for this simple case. This is also aligned with runc's implementation(https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/fs/freezer.go#L149-L151).

crates/libcgroups/src/v1/freezer.rs
wyhao31: Glad that you raised the question. This is what runc's implementation does. I'm also confused when I first read runc's implementation[1]. After some thinking, I realized that this is to make sure the cgroup v1 implementation has the same behavior as v2. When using cgroup v2, if the specified cgroup is frozen, runc will fail immediately with error message `container's cgroup unexpectedly frozen`. If the specified cgroup is not frozen, but its parent cgroup is frozen, runc will stuck there. This is because when runc checks whether the cgroup is frozen, it calls `GetFreezerState` method[2]. In its cgroup v2 implementation, it only read the value from current cgroup's `cgroup.freeze` file, not from its ancestors' file[3].

[1] https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/fs/freezer.go#L128-L147
[2] https://github.com/opencontainers/runc/blob/77ead42c9f197e1db8494839de34277ce7d79bc5/libcontainer/factory_linux.go#L82-L88
[3] https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/fs2/freezer.go#L54

crates/libcgroups/src/v1/freezer.rs
wyhao31: I also thought about that. The logic here is simpler, the wait/retry are handled in outside for-loop. While in read_freezer_state, it keeps waiting if the state is freezing. If we want to reuse read_freezer_state, we need an additional flag to control whether to wait on freezing state. I feel like it's fine to have a little bit duplication here. The go implementation also doesn't reuse GetState method in Set method. https://github.com/opencontainers/cgroups/blob/6a793b6ea778415b3c0f7a3d2efdb70c786fad92/fs/freezer.go#L86-L97

wyhao31:
@YJDoc2 Can you please add the label to this PR? There's a check failure due to missing label.

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -53,6 +53,14 @@ pub fn container_intermediate_process(
...
     let cgroup_manager = libcgroups::common::create_cgroup_manager(args.cgroup_config.to_owned())
         .map_err(|e| IntermediateProcessError::Cgroup(e.to_string()))?;
+    let state = cgroup_manager.get_freezer_state().map_err(|e| {
+        IntermediateProcessError::Cgroup(format!("unable to get cgroup freezer state: {}", e))
+    })?;
+    if state == FreezerState::Frozen {
+        return Err(IntermediateProcessError::Cgroup(String::from(
+            "container's cgroup unexpectedly frozen",
+        )));
+    }
```
utam0k, line 63: Why don't we check it in the main process?

crates/libcgroups/src/v1/freezer.rs
```diff
@@ -114,17 +121,43 @@ impl Freezer {
...
+                        Ok(self_state) => self_state,
+                        Err(_) => {
+                            return FreezerState::Frozen;
+                        }
+                    };
+                    return match self_state.as_str() {
+                        "0\n" => FreezerState::Thawed,
+                        "1\n" => FreezerState::Frozen,
+                        _ => FreezerState::Undefined,
+                    };
```
utam0k, line 148: Since there's no clear way to determine the number of levels, why focus on just one? Also, why do we need to know the direct or indirect?

utam0k (changes requested):

utam0k:
Thanks for your PR! I left some comments.

utam0k:
@wyhao31 Is there anything I can help you with?

Issue #3186, opened 2025-06-18
[Bug]: Check if hook order is correct


CheatCodeSam:
### Bug Description

Just parroting what [was mentioned here](https://github.com/youki-dev/youki/pull/3174#issuecomment-2948103123) in the comments on my PR so that we don't lose track of it. I poked around the code a little and it definitely looks like `hooks.create_container` is ran when we build the container, while `hooks.prestart` is ran when we run the container. `prestart` needs to be in the create operation, so I'm guessing it needs to be moved somewhere inside `ContainerBuilderImpl::run_container`?

### Steps to Reproduce

Just walking through the `youki run` seems to give a lot of insight on the order the hooks are being called.

### Expectation

The OCI hooks should be ran in the proper order [specified here](https://github.com/youki-dev/youki/pull/3174#issuecomment-2948103123).

YJDoc2:
Hey, I did a small test and this is the current state of hook order. We should confirm to whatever oci says as first priority, and then try to follow runc behavior. If you're willing to take this up, I'd be glad to help you out. 

```
youki: 
	before_create
	create_container
	create_runtime
	after_create
	before_start
	prestart
	poststart
	after_start
	before_delete
	poststop
	after_delete

runc: 
	before_create
	prestart
	create_runtime
	create_container
	poststart
	after_create
	before_start
	after_start
	before_delete
	poststop
	after_delete
```

CheatCodeSam:
I'm definitely thinking of giving it a shot, and I'll start but just moving prestart up to the top. I'll just poke around and see how runc does it. From what it looks like, all that's required to fire off the hook is just something like this.

https://github.com/youki-dev/youki/blob/2c361a8bc8c6a723948efefb37fa2923272ff9f4/crates/libcontainer/src/container/builder_impl.rs#L196

CheatCodeSam:
Hello, I'm currently hacking away at this, and I just had a question about your comment. @YJDoc2 

I'm having difficulty seeing the hook order for the container I'm running. I'm on Ubuntu 24.04 using Docker 28.3.1, I have this in my `/etc/docker/daemon.json`:

```json
{
  "log-level": "trace",
  "default-runtime": "runc",
  "runtimes": {
    "youki": {
      "path": "/home/carson/repos/youki/youki",
      "runtimeArgs": [
          "--log-level",
          "debug",
          "--systemd-log"
      ]
    }
  }
}
```

[...]

PR #3256, opened 2025-09-24
fix hooks order

saku3:
### 1. Fix order
 I’ll modify it to execute the hooks in the following order:
(refer runtime-spec: https://github.com/opencontainers/runtime-spec/blob/62106c19fd422dc70b92c2abbeb4c51d9e39b5d0/runtime.md#lifecycle)

```
prestart
create_runtime
create_container
poststart
poststop
```

### 2. Implemented a channel for hooks.
`prestart` and `createRuntime` are defined as follows:
  * Execute in the runtime namespace.
  * Execute before pivot_root.

To achieve this, we introduced a `channel` so that, right before `pivot_root`, the init process sends a message to the main process, which then executes the hooks.
This is based on the way runc’s `syncParentHooks` handles communication between the child and parent processes.

(refer syncParentHooks: https://github.com/opencontainers/runc/blob/77ead42c9f197e1db8494839de34277ce7d79bc5/libcontainer/rootfs_linux.go#L183C12-L183C27)

(refer oci-spec: https://github.com/opencontainers/runtime-spec/blob/62106c19fd422dc70b92c2abbeb4c51d9e39b5d0/config.md#posix-platform-hooks)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes
- [x] Other (please describe):

I’ve confirmed it runs with nerdctl.
Looking back through the issue, it originated because it didn’t work with nerdctl at first.

https://github.com/youki-dev/youki/issues/2994

```
nerdctl run --rm -it --runtime=youki alpine
```

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes https://github.com/youki-dev/youki/issues/3186

### DIFF
--- a/crates/libcontainer/src/container/builder_impl.rs
+++ b/crates/libcontainer/src/container/builder_impl.rs
@@ -16,8 +16,8 @@ use crate::process::intel_rdt::delete_resctrl_subdirectory;
 use crate::process::{self};
 use crate::syscall::syscall::SyscallType;
 use crate::user_ns::UserNamespaceConfig;
+use crate::utils;
 use crate::workload::Executor;
-use crate::{hooks, utils};
 
 pub(super) struct ContainerBuilderImpl {
     /// Flag indicating if an init or a tenant container should be created
@@ -165,6 +165,7 @@ impl ContainerBuilderImpl {
             stdout: self.stdout.as_ref().map(|x| x.as_raw_fd()),
             stderr: self.stderr.as_ref().map(|x| x.as_raw_fd()),
             as_sibling: self.as_sibling,
+            pid_file: self.pid_file.to_owned(),
         };
 
         let (init_pid, need_to_clean_up_intel_rdt_dir) =
@@ -175,14 +176,6 @@ impl ContainerBuilderImpl {
                 },
             )?;
 
-        // if file to write the pid to is specified, write pid of the child
-        if let Some(pid_file) = &self.pid_file {
-            fs::write(pid_file, format!("{init_pid}")).map_err(|err| {
-                tracing::error!("failed to write pid to file: {}", err);
-                LibcontainerError::OtherIO(err)
-            })?;
-        }
-
         if let Some(container) = &mut self.container {
             // update status and pid of the container process
             container
@@ -191,16 +184,6 @@ impl ContainerBuilderImpl {
                 .set_pid(init_pid.as_raw())
                 .set_clean_up_intel_rdt_directory(need_to_clean_up_intel_rdt_dir)
                 .save()?;
-        }
-
-        if matches!(self.container_type, ContainerType::InitContainer) {
-            if let Some(hooks) = self.spec.hooks() {
-                hooks::run_hooks(
-                    hooks.create_runtime().as_ref(),
-                    self.container.as_ref(),
-                    None,
-                )?
-            }
         }
 
         Ok(init_pid)
--- a/crates/libcontainer/src/hooks.rs
+++ b/crates/libcontainer/src/hooks.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::io::{ErrorKind, Write};
 use std::os::unix::prelude::CommandExt;
@@ -8,7 +9,7 @@ use nix::sys::signal;
 use nix::unistd::Pid;
 use oci_spec::runtime::Hook;
 
-use crate::container::Container;
+use crate::container::{Container, State};
 use crate::utils;
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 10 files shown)


crates/libcontainer/src/hooks.rs
```diff
@@ -70,6 +70,8 @@ pub fn run_hooks(
                 .env_clear()
                 .envs(envs)
                 .stdin(process::Stdio::piped())
+                .stdout(std::process::Stdio::null())
```
saku3, line 85: This change was made based on the results of running runc and containerd tests. We suppress stdout and allow stderr to be emitted.

There may be better ways to control this, but since this PR focuses on fixing the hook order, we chose the simplest approach. 
(For example, crun appears to let you specify hook output destinations via annotations. https://github.com/containers/crun/blob/270c76ad3b7fd185d82d8067cbc5a6375fb4a34c/src/libcrun/container.c#L1463)

saku3:
The `delete` test in oci-validate-rust is failing, but it passes locally.

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -147,7 +147,7 @@ impl ContainerBuilderImpl {
         // This container_args will be passed to the container processes,
         // therefore we will have to move all the variable by value. Since self
         // is a shared reference, we have to clone these variables here.
-        let container_args = ContainerArgs {
+        let mut container_args = ContainerArgs {
```
utam0k: I prefer that `container_args` be immutable. Do we really need `mut`?

saku3: I apologize for the delayed response.

I removed `mut` from `container_args`.
Since I need to pass the `container` inside `container_args` to run_hooks, I decided to call clone() at that point.

```rust
if let Some(mut container_for_hooks) = container_args.container.clone() {
    container_for_hooks.set_pid(init_pid.as_raw());
}
```

The reason for cloning is that the `container` at that moment doesn’t have its PID set, so I need a mutable clone to set the PID before passing it to the hook.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -127,10 +131,52 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+
+                hooks::run_hooks(
+                    hooks.create_runtime().as_ref(),
+                    Some(&container_for_hooks),
+                    None,
+                )
+                .map_err(|err| {
+                    tracing::error!("failed to run create runtime hooks: {}", err);
+                    err
+                })?;
```
utam0k: You mentioned in the PR description that `prestart()` is invoked after `create_runtime()`. However, your implementation does the opposite. Which sequence is correct?

crates/libcontainer/src/process/args.rs
```diff
@@ -52,4 +52,7 @@ pub struct ContainerArgs {
     pub stderr: Option<RawFd>,
     // Indicate if the init process should be a sibling of the main process.
     pub as_sibling: bool,
+    /// File which will be used to communicate the pid of the
```
utam0k: This is not a file; it's a path.

crates/libcontainer/src/process/init/process.rs
```diff
@@ -107,6 +97,23 @@ pub fn container_init_process(
...
+        if let Some(hooks) = ctx.hooks {
+            // send a request to the main process to run prestart and create_runtime hooks.
+            // prestart and create_runtime hook needs to be called after the namespace setup, but
+            // before pivot_root is called. This runs in the runtime(not container) namespaces.
+            main_sender.hook_request()?;
+            init_receiver.wait_for_hook_request_done()?;
+
+            // create_container hook needs to be called after the namespace setup, but
+            // before pivot_root is called. This runs in the container namespaces.
+            hooks::run_hooks(hooks.create_container().as_ref(), ctx.container, None).map_err(
```
utam0k: `create_container` should be invoked first, right?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -127,10 +131,52 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+    if let Some(pid_file) = &container_args.pid_file {
+        if let Err(err) = fs::write(pid_file, format!("{init_pid}")) {
+            tracing::warn!("failed to write pid to file: {err}");
+        }
+    }
+
+    if matches!(container_args.container_type, ContainerType::InitContainer) {
+        if let Some(hooks) = container_args.spec.hooks() {
+            main_receiver.wait_for_hook_request()?;
+            if let Some(mut container_for_hooks) = container_args.container.clone() {
```
utam0k: This `clone` operation is quite heavy. We should consider using Arc<Container> or passing a pid to hookState when running hooks.

crates/libcontainer/src/process/container_main_process.rs
saku3: I’m sorry — my description was incorrect. I’ve updated the description.
As defined in the OCI spec, the following order is correct. The implementation order is correct.

```
prestart
create_runtime
create_container
poststart
poststop
```

crates/libcontainer/src/process/args.rs
saku3: fixed.

crates/libcontainer/src/process/init/process.rs
saku3: I’m sorry — my description was incorrect. I’ve updated the description.
As defined in the OCI spec, the following order is correct. The implementation order is correct.

```
prestart
create_runtime
create_container
poststart
poststop
```

crates/libcontainer/src/process/container_main_process.rs
saku3: Since using something like Arc would require changes across a wide area, I changed the approach to pass the pid to the hook instead.

However, because the pid ultimately needs to be set on the state, I’m cloning the state inside the hook handling.

https://github.com/youki-dev/youki/pull/3256/files#diff-f8d87cd2792f7b186f8955cde622a297e504876104fa29a91c6ec5557e6c8e96R48

If this approach is not acceptable, I’ll look into using Arc or other options.
Any advice or suggestions you might have would be greatly appreciated.

utam0k: Your change is reasonable. 👍

utam0k (approved):

PR #3280, opened 2025-10-26
mount info provider

### LINKED ISSUE #3279: `test_make_parent_mount_private` fails on non-shared parent mounts
## Overview

The unit test test_make_parent_mount_private assumes that the parent of the temporary directory is a shared mount and  unconditionally asserts result.is_some(). On many local setups /tmp (or /) is already private, causing Mount::make_parent_mount_private to return Ok(None) and the test to panic. GitHub Actions hides the issue because scripts/cargo.sh mounts /tmp with bind-propagation=shared before running  tests, but running cargo test directly on a typical Linux host reproduces the failure.
https://github.com/youki-dev/youki/blob/61614cc82c9920974f482be98e1c71ba2a6fe0c9/crates/libcontainer/src/rootfs/mount.rs#L1040-L1068

## Possible Fixes

1. Inject a mockable MountInfo source into make_parent_mount_private so the test can supply deterministic mountinfo entries (covering both shared and private cases).
2. Or, adjust the test to branch on result: assert the remount path only when Some is returned and ensure no mount calls occur when None is returned.
3. Optionally keep an integration-style test (run via scripts/cargo.sh) to verify the shared case end-to-end.

I prefer no1 but welcome to discussion.

CheatCodeSam:
## Description

Seemed easy enough, just created a Mount Info Provider that does the same thing as the function used to do, and swapping it out with a bogus one in the test. I've also created a new test to test both routes.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [X] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [X] Ran existing test suite

## Related Issues

Fixes #3279

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -1267,8 +1267,6 @@ mod tests {
         assert_eq!(got.flags, MsFlags::MS_PRIVATE);
         assert_eq!(got.data, None);
 
-        // This can be either depending on the system, some systems mount tmpfs at /tmp others it's
-        // a plain directory. See https://github.com/containers/youki/issues/471
         assert_eq!(got.target, parent);
 
         Ok(())


CheatCodeSam:
I have a dumb question on this, why not make the `make_parent_mount_private` function return the parent `MountInfo` regardless of rather it changed the parent mount or not?

```rust
    pub fn make_parent_mount_private(&self, rootfs: &Path) -> Result<MountInfo> {
        let mount_infos = self.mountinfo_provider.mountinfo()?;
        let parent_mount = find_parent_mount(rootfs, mount_infos)?;

        // check parent mount has 'shared' propagation type
        if parent_mount
            .opt_fields
            .iter()
            .any(|field| matches!(field, MountOptFields::Shared(_)))
        {
            self.syscall.mount(
                None,
                &parent_mount.mount_point,
                None,
                MsFlags::MS_PRIVATE,
                None,
            )?;
            
        }
        Ok(parent_mount)
    }
```

utam0k:
> I have a dumb question on this, why not make the make_parent_mount_private function return the parent MountInfo regardless of rather it changed the parent mount or not?

Your idea is better than the current implementation 👍 AFAI, we don't have a specific reason.

saku3:
I’d appreciate it if you could resolve the conflicts.

CheatCodeSam:
Good to go, this is a dead simple PR, I'm really looking to get this merged.

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -1195,33 +1219,91 @@ mod tests {
...
+    }
+    impl MountInfoProvider for FakeMountInfo {
+        fn mountinfo(&self) -> std::result::Result<Vec<MountInfo>, MountError> {
+            std::result::Result::Ok(self.entries.clone())
+        }
+    }
+
     #[test]
     fn test_make_parent_mount_private() -> Result<()> {
         let tmp_dir = tempfile::tempdir()?;
```
saku3: Wouldn’t it be enough to use a `PathBuf` instead of `tempdir`?

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -1195,33 +1219,91 @@ mod tests {
...
+        assert_eq!(got.flags, MsFlags::MS_PRIVATE);
+        assert_eq!(got.data, None);
 
-            // This can be either depending on the system, some systems mount tmpfs at /tmp others it's
-            // a plain directory. See https://github.com/youki-dev/youki/issues/471
-            assert!(got.target == PathBuf::from("/") || got.target == PathBuf::from("/tmp"));
-        }
+        // This can be either depending on the system, some systems mount tmpfs at /tmp others it's
+        // a plain directory. See https://github.com/containers/youki/issues/471
+        assert!(got.target == PathBuf::from("/") || got.target == PathBuf::from("/tmp"));
```
saku3: If we avoid using `tempdir`, couldn’t we write it like this instead?

```
assert_eq!(got.target, parent);
```

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -1195,33 +1219,91 @@ mod tests {
...
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_not_make_parent_mount_private_if_already_private() -> Result<()> {
+        let tmp_dir = tempfile::tempdir()?;
```
saku3: Wouldn’t it be enough to use a `PathBuf` instead of `tempdir`?

saku3 (commented):
Thank you.
I think it’s already good enough as it is, but I left a comment anyway.
Please take a look.

crates/libcontainer/src/rootfs/mount.rs
CheatCodeSam: I could make those changes, the reason I had it like that is simply because that's how it was before. Since we're not exactly making any changes to a real directory, it's probably okay to just use a pathbuf.

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -1195,33 +1220,91 @@ mod tests {
...
-            // a plain directory. See https://github.com/youki-dev/youki/issues/471
-            assert!(got.target == PathBuf::from("/") || got.target == PathBuf::from("/tmp"));
-        }
+        let got = &set[0];
+        assert_eq!(got.source, None);
+        assert_eq!(got.fstype, None);
+        assert_eq!(got.flags, MsFlags::MS_PRIVATE);
+        assert_eq!(got.data, None);
+
+        // This can be either depending on the system, some systems mount tmpfs at /tmp others it's
```
saku3: One more small thing, sorry about that.
It’s fine to delete the comment.

CheatCodeSam:
good to go

saku3:
@utam0k @YJDoc2
PTAL

saku3:
Sorry. 
Since the return type of `make_parent_mount_private` changes from `Option<MountInfo>` to `MountInfo`, it will likely be a breaking change.

https://docs.rs/libcontainer/0.5.7/libcontainer/rootfs/struct.Mount.html

CheatCodeSam:
> Sorry. Since the return type of `make_parent_mount_private` changes from `Option<MountInfo>` to `MountInfo`, it will likely be a breaking change.
> 
> https://docs.rs/libcontainer/0.5.7/libcontainer/rootfs/struct.Mount.html

I could just make this always return a `Some(MountInfo)` instead and keep the `Option<MountInfo>` type, but I was just making changes based off the previous comment. This fixes the breaking change, but it's a little misleading since the function will never return a `None`.

CheatCodeSam:
I could also split these into two PRs and we can discuss that breaking change later.

PR #3283, opened 2025-10-30
linux_ns_path contest integration test

cody-herbst:
## Description
part of #361 

Implemented the linux_ns_path integration test

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [X] Added new integration tests
- [ ] Ran existing test suite

## Related Issues

Fixes #

### DIFF
--- a/tests/contest/contest/src/tests/linux_ns_path/ns_path_test.rs
+++ b/tests/contest/contest/src/tests/linux_ns_path/ns_path_test.rs
@@ -34,7 +34,7 @@ impl Drop for WithCleanup {
     }
 }
 
-fn create_spec(namespace_path: &Vec<NamespacePath>) -> Spec {
+fn create_spec(namespace_path: &[NamespacePath]) -> Spec {
     let mut linux_namespace_types = Vec::new();
     for namespace_path in namespace_path {
         linux_namespace_types.push(
@@ -150,7 +150,7 @@ fn test_namespace_paths(mut linux_namespace_types: Vec<LinuxNamespaceType>) -> T
     let child = match child {
         Ok(child) => child,
         Err(e) => {
-            return TestResult::Failed(anyhow!(format!("could not spawn unshare: {}", e)));
+            return TestResult::Failed(anyhow!("could not spawn unshare: {}", e));
         }
     };
     let g_child = WithCleanup::new(child);
@@ -168,10 +168,10 @@ fn test_namespace_paths(mut linux_namespace_types: Vec<LinuxNamespaceType>) -> T
         )
         .err();
         if err.is_some() {
-            return TestResult::Failed(anyhow!(format!(
+            return TestResult::Failed(anyhow!(
                 "could not wait for path {}",
                 &namespace_path.path.display()
-            )));
+            ));
         }
     }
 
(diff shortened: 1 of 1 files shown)


cody-herbst:
Apologies for all the fix commits. I just notice the "just lint" command

tests/contest/contest/src/tests/linux_ns_path/ns_path_test.rs
```diff
@@ -0,0 +1,275 @@
...
+    fn new(child: Child) -> Self {
+        Self { child }
+    }
+}
+
+impl Drop for WithCleanup {
+    fn drop(&mut self) {
+        match kill(Pid::from_raw(-(self.child.id() as i32)), Signal::SIGKILL) {
+            Ok(()) => {}
+            Err(_e) => {} // Not sure what else can be done here
```
YJDoc2: log the error at error level, we can't really do anything, but good to have an error log. We can also structure this as `if let Err(e) = ...` as we don't care about Ok

tests/contest/contest/src/tests/linux_ns_path/ns_path_test.rs
```diff
@@ -0,0 +1,275 @@
...
+        }
```
utam0k, line 33: Should we wait for the child processes is killed to avoid making zombi processes?

cody-herbst: Well we can... but if for whatever reason the process status doesn't change we will hold up the test for a long time. I made the sleep function match the original test duration which is 10000 seconds (honestly I don't think we need anywhere near that). Maybe a compromise would bring the sleep duration down to 120 secs and in the absolute worst case a few processes are there for 2 mins then gone. I could go either way.

utam0k: > Maybe a compromise would bring the sleep duration down to 120 secs and in the absolute worst case a few processes are there for 2 mins then gone. I could go either way.

+1

cody-herbst:
@utam0k I made the updates copilot requested. Any way we can move forward?

tests/contest/contest/src/tests/linux_ns_path/ns_path_test.rs
```diff
@@ -0,0 +1,275 @@
...
+    test_namespace_paths(namespaces)
+}
+
+fn test_namespace_paths(mut linux_namespace_types: Vec<LinuxNamespaceType>) -> TestResult {
+    if !linux_namespace_types.contains(&LinuxNamespaceType::Mount) {
+        // to prevent mounting issues with the new container, we will always
+        // create a new mount namespace. This was added for the runc runtime as the container would
+        // fail to be created without it
+        linux_namespace_types.push(LinuxNamespaceType::Mount);
+    }
```
utam0k, line 134: Forcing a Mount namespace is not necessary. If it's absent, both masked and read-only paths should be empty.

cody-herbst: I'll look into this when I get back from vacation. 

I do have a memory that this portion was necessary to pass the runc integration test though. I remember even trying to create a runc container manually and it fought me until I gave it mount ns. But I'll triple check everything when I get back

utam0k: > I do have a memory that this portion was necessary to pass the runc integration test though. I remember even trying to create a runc container manually and it fought me until I gave it mount ns. But I'll triple check everything when I get back

I didn't know that. If that's the case, please leave a comment.

cody-herbst: Yeah I'll triple check everything when I get back. But that is what my existing comment is referencing. The test I coded would pass for youki... but fail when testing against runc

cody-herbst: I just tagged you in the discord convo I started regarding this

PR #3292, opened 2025-11-12
Add poststart hook test

fspv:
## Description

This implements a test similar to https://github.com/opencontainers/runtime-tools/blob/master/validation/poststart/poststart.go as a part of the https://github.com/youki-dev/youki/issues/361

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

https://github.com/youki-dev/youki/issues/361

## Additional Context

The Youki test differs from the Go test to handle a race condition. According to the OCI spec, the poststart hook must run after the container process starts [2], but there's no guarantee about when the container process will write to the output file.

The Go test appears to rely on timing assumptions - it expects the process to write first, then the hook. However, this isn't guaranteed by the spec. In practice, Youki correctly starts both the process and hook in the right order, but the hook often writes to the file before the process gets a chance to.

To make the test deterministic, I've added an explicit wait after the start returns. It waits until it sees "process called" in the output file before asserting the contents of the file. This ensures we're testing the actual requirement (hook runs after process starts) rather than relying on unpredictable write timing.

Another issue here is a bug in `runc` https://github.com/opencontainers/runc/issues/4347 which incorrectly calls post-start hook on create [1] [2]. So for the `runc` case I explicitly skip the assertion that validates that the hook did not run. Youki implements the spec correctly.

[1] https://github.com/opencontainers/runtime-spec/blob/main/config.md#poststart
[2] https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle

### DIFF
--- a/tests/contest/contest/src/tests/poststart/mod.rs
+++ b/tests/contest/contest/src/tests/poststart/mod.rs
@@ -170,8 +170,6 @@ fn get_test(test_name: &'static str) -> Test {
                     "process called\npost-start called\n" => TestResult::Passed,
                     "post-start called\nprocess called\n" => TestResult::Passed,
                     "process called\n" => {
-                        // TODO: in the original test this is allowed for some reason, but I don't
-                        // think this is correct
                         TestResult::Failed(anyhow!("The runtime MUST run the post-start hook"))
                     }
                     "post-start called\n" => TestResult::Failed(anyhow!(


YJDoc2:
Hey there are some format and lint errors : https://github.com/youki-dev/youki/actions/runs/19300493934/job/55408290569?pr=3292

fspv:
should be fine now. Forgot to run clippy one more time before commit

tests/contest/contest/src/tests/poststart/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+            return Ok(());
+        }
+        std::thread::sleep(poll_interval);
+    }
+
+    Err(anyhow!(
+        "Timed out waiting for file {:?} to contain '{}'",
+        file_path,
+        expected_content
+    ))
```
YJDoc2, line 70: Can we also add the actual content in error message, so debugging is easier when it fails?

tests/contest/contest/src/tests/poststart/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+                    return match (has_poststart, has_process) {
+                        (true, _) => TestResult::Failed(anyhow!(
+                            "The post-start hooks MUST NOT be called before the `start` operation"
+                        )),
+                        (false, true) => TestResult::Failed(anyhow!(
+                            "The user-specified program (from process) MUST NOT be run before the `start` operation"
+                        )),
+                        (false, false) => TestResult::Failed(anyhow!(
+                            "file {:?} should not exist after create",
+                            host_output_file
```
YJDoc2: Maybe also add content of file here. If the file should not exist at all, an it still exists, its content might be helpful to know why.

tests/contest/contest/src/tests/poststart/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+                "process called",
+                std::time::Duration::from_secs(5),
+                std::time::Duration::from_millis(100),
+            );
+
+            let result = if let Err(e) = wait_result {
+                TestResult::Failed(anyhow!(
+                    "The poststart hooks MUST be invoked by the runtime\n\
+                     The runtime MUST run the user-specified program, as specified by `process`\n\
+                     Error: {}",
```
YJDoc2:
```suggestion
                    "The container process did not ran correctly : Error: {}",
```

Technically this error is when we do not get the container process o/p, so mentioning poststart hook might be confusing.

tests/contest/contest/src/tests/poststart/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+                    e
+                ))
+            } else if !host_output_file.exists() {
+                TestResult::Failed(anyhow!(
+                    "The poststart hooks MUST be invoked by the runtime\n\
+                     The runtime MUST run the user-specified program, as specified by `process`"
```
YJDoc2: Here also, we should mention the exact error that expected o/p file did not exist

fspv:
Thanks for the review! I'll fix error messages a bit later.

Regarding the ordering, I thought about that as well. We can add a sleep to the poststart hook to make it more likely to execute after the main process writes to the file. However, due to the nature of the OS scheduler there is still no 100% guarantee that the process will execute first. If we get super unlucky and the test will get onto a very slow machine, it could be flaky. This was the main reason I didn't do it this way.

Another option here would be in the post-start hook to explicitly wait until the first process writes to the file. But this makes the test a bit complex, which I also didn't want to do because it will be harder to read such a test.

So, in the end I decided that just saying that any order is correct is a simpler solution which is also not flaky. However, if you think any of the two approaches above are better and the tradeoffs are acceptable, I'm happy to implement any of them.

YJDoc2:
Hey, so after thinking on your comment, I agree with you that for now, we should accept both o/p. While making post-start wait for the process is possible, I'm not sure how sensible that would be. Basically what we want to test in the end is that the hook must not be called before the process starts ; but is allowed to be called any time, including immediately, after the process is spawned. Which means having almost no time between process start and hook is correct behavior. And given that actual scheduling of process will depend on os scheduler and cpu availability , I cannot think of any reasonable way to detect this order in current approach.

Another thing we can try is make the process and hook print the current time to maybe nano-sec precision and verify in test code that hook time >= process time. But even then it is still possible that os decides to not schedule container main process, and for some reason hook gets schedules slightly before, even though we created them in correct order.

@saku3 would like to have your opinion as well, Do you think doing time based checked with occasional flakes is better, or should we allow the current behavior while accepting that we would not catch a regression if somehow the hook is started before the process (in youki/runtime itself)?

saku3:
For this test case, I think it’s fine to allow both outputs.

My concern is more that, in youki’s code, it doesn’t seem like we actually guarantee that the poststart hook runs after the process. I tried observing the poststart behavior locally, and the ordering between the process and the poststart hook was not deterministic at all. Sometimes the process wrote to the file first, and sometimes the poststart hook wrote first.

With the current source code, what we’re really guaranteeing is only that the parent process instructs the init process to start the container (`notify_socket.notify_container_start()`) before the poststart hook is executed. That ordering is guaranteed, but nothing more.

https://github.com/youki-dev/youki/blob/main/crates/libcontainer/src/container/container_start.rs#L61

I couldn’t immediately come up with a good way to fully guarantee the ordering between the process and the poststart hook, especially because the process will eventually call `execvep`. (We can’t use the channel we’re currently using to synchronize the order between processes for this.)

fspv:
Hey, I'm not sure if anything is expected here from me, so let me know if I need to change something else here.

Regarding this point:

> I couldn’t immediately come up with a good way to fully guarantee the ordering between the process and the poststart hook

I don't think we should do it, even if it is possible. If we guarantee the ordering, which is not required by the specification, there will be people who will start relying on it. And then if for whatever reason it has to be changed, it will be a breaking change.

saku3:
Sorry if that was confusing. If I’d like you to change something, I’ll leave a comment on the code.

The comment above is only meant to say that the poststart hook must run after the container process starts,(in accordance with the OCI spec).

YJDoc2 (approved):
LGTM. I'm gonna wait for few hours to see if there are any comments (just to make sure) otherwise will go ahead and merge.

@fspv Thanks you for the contribution! The test is really well written :)

fspv:
Thanks! Just removed one TODO comment that I have left earlier. Please run the checks again if you're going to merge.

PR #3299, opened 2025-11-23
closes 3286 wait for job remove signal

### LINKED ISSUE #3286: Wait for a JobRemoved signal from Systemd after creating Scope, before intermediate process ends
### Background

_No response_

### Feature Request

I mentioned this as a more complete solution to the fix in #3262, the current implementation just runs in a loop waiting to see if the cgroup has been created. 

_No response_

### Proposed Solution

I'm thinking we start waiting for the JobRemoved signal before we even send the signal to create the cgroup, then we just use a mutex to prevent the intermediate process from continuing until the signal has been received and the mutex unlocks. Seems simple enough.

### Considerations

_No response_

### Additional Context

_No response_

cody-herbst:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #3286


cody-herbst:
@YJDoc2 / @utam0k / @CheatCodeSam 

There is another option instead of spinning up a thread and creating a brand new dbus connection.

In theory we could call AddMatch and Subscribe on the existing dbus connection right before calling start_transient_unit. We read for the signal after the call finishes. Finally call Unsubscribe to stop systemd from sending us signals (I don't believe we need to remove the AddMatch rule). I'm a little nervous that the resulting signal could get mixed up with the start_transient_unit call but I don't think that can happen. 

I'm not sure the above approach will be faster but I can test it out when I get home. 

I didn't end up going the convar/mutex approach because the thread signaling involved became complicated. In order to avoid the unshared limitation, the thread had to be more or less confined to the add_task method. When creating the DBusConnection and subscribing, I ran into several cases were race conditions occurred without a lot thread signaling/waiting for other things to happen.

cody-herbst:
I wrote up the synchronous implementation real quick. Going to run some perf tests to compare the two and ill post here. The we can figure out the better approach

cody-herbst:
@YJDoc2 / @utam0k / @CheatCodeSam 

I removed the threaded approach. From some testing doing it synchronously is both faster and easier on memory

cody-herbst:
@saku3 so not sure why the previous checks have failed.... Linting seems to work just fine for me on local. I merged with main and pushed. Also it seems like the "enhancement" tag is causing an issue. I don't seem to have the permissions to change the tag.

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -274,6 +274,26 @@ impl DbusConnection {
         Ok(ret)
     }
 
+    /// function to read messages off the socket
+    pub fn read_messages(&self) -> Result<Vec<Message>> {
```
utam0k, line 278: If the message doesn’t send, the manager waits forever, right?

utam0k (changes requested):
May I ask you to add the e2e test?

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -341,6 +342,41 @@ impl Manager {
...
     }
 
+    fn wait_for_remove_job_signal(&self) -> Result<(), SystemdManagerError> {
+        let start = Instant::now();
+        while start.elapsed() < self.cgroup_wait_timeout_duration {
+            let messages = self.client.read_messages()?;
+            for message in messages {
+                if message.preamble.mtype != MessageType::Signal {
+                    continue;
+                }
```
utam0k, line 352:
```suggestion
            let messages = self.client.read_messages()?.into_iter().filter(|m| m.preamble.mtype == MessageType::Signal);
            for message in messages {
```

crates/libcgroups/src/systemd/dbus_native/dbus.rs
cody-herbst: That is correct. Honestly that the nature of all of our communication with systemd right now from my understanding. receive_complete_response() blocks until message can be read if I recall correctly

cody-herbst:
Not sure why the tests timed out.... Locally all of the integration tests passed and consistently passed. I guess its possible theres a race case but I'd have to look into it.

PR #3313, opened 2025-12-07
Add poststart_fail hook test

fspv:
## Description
This implements a test similar to https://github.com/opencontainers/runtime-tools/blob/master/validation/poststart_fail/poststart_fail.go as a part of the https://github.com/youki-dev/youki/issues/361

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

https://github.com/youki-dev/youki/issues/361

## Additional Context

The test is similar to the happy case test landed in https://github.com/youki-dev/youki/pull/3292 but it  focuses on the error handling.

I made it a bit different to the original go test to make sure we also test a couple of more things. I create three hooks, with the second hook failing. We expect that the first two hooks to run and exit with an error. The hook after the failed hook should not run as implied by the lifecycle spec https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle.

One assumption I make here is that the hooks are executed in the same order they're defined in a spec, which is not explicitly stated anywhere in the runtime spec but anyway I think it is a relatively safe assumption and it is nice to guarantee this order even if the spec doesn't require it.

Also, this test is affected by the same bug as a happy case scenario https://github.com/opencontainers/runc/issues/4347 so an extra branch has been added to handle the incorrect runc behaviour.

### DIFF
--- a/tests/contest/contest/src/tests/poststart_fail/mod.rs
+++ b/tests/contest/contest/src/tests/poststart_fail/mod.rs
@@ -92,6 +92,10 @@ fn get_spec(host_output_file: &str) -> Spec {
 /// Validates that the runtime stops executing remaining poststart hooks after one fails,
 /// and returns an error (exit code 1). This test creates 3 hooks where `hook_2` fails,
 /// then verifies that `hook_1` and `hook_2` ran but `hook_3` did not.
+///
+/// The test is skipped for runc, because the runc behaviour is incorrect and doesn't match the
+/// OCI spec behaviour which is implemented in youki
+/// <https://github.com/opencontainers/runc/issues/4347>
 fn get_test(test_name: &'static str) -> ConditionalTest {
     ConditionalTest::new(
         test_name,


fspv:
Hey @YJDoc2

Can you help me with [containerd-integration-tests](https://github.com/youki-dev/youki/actions/runs/20004655320/job/57365115298?pr=3313#logs) failure?

It fails on
```
--- FAIL: TestContainerPids (0.19s)
```
I can't see how it  could be related to the change I made here.

saku3:
As you said, this PR only changes test code, so the containerd-test error is unrelated.

tests/contest/contest/src/tests/poststart_fail/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+
+            let create_failed = match create_result {
+                Err(_) => true,
+                Ok(Ok(status)) if !status.success() => true,
+                _ => false,
+            };
+
+            // runc doesn't follow spec and runs hooks at the create stage
+            // https://github.com/opencontainers/runc/issues/4347
+            if create_failed && !is_runtime_runc() {
```
saku3: If runc’s behavior is incorrect, I think it’s fine to skip the test altogether.

Please refer to the code below. (use ConditionalTest)
https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/tests/misc_props/misc_props_test.rs#L95

tests/contest/contest/src/tests/poststart_fail/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::utils::test_utils::CreateOptions;
+use crate::utils::{
+    create_container, delete_container, generate_uuid, is_runtime_runc, prepare_bundle, set_config,
+    start_container,
+};
+
+const CONTAINER_OUTPUT_FILE: &str = "output";
```
saku3: Since poststart hooks are executed by the runtime rather than inside the container, I find the name CONTAINER_OUTPUT_FILE a bit confusing.
```suggestion
const HOOK_OUTPUT_FILE: &str = "output";
```

fspv:
I've addressed the comments, but one of the tests timed out. Pretty sure it is not related to my change, so will be happy if someone can trigger a re-run.

tests/contest/contest/src/tests/poststart_fail/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+/// Tests that when a poststart hook fails, subsequent hooks are not executed.
+///
+/// Validates that the runtime stops executing remaining poststart hooks after one fails,
+/// and returns an error (exit code 1). This test creates 3 hooks where `hook_2` fails,
+/// then verifies that `hook_1` and `hook_2` ran but `hook_3` did not.
+fn get_test(test_name: &'static str) -> ConditionalTest {
+    ConditionalTest::new(
+        test_name,
+        Box::new(|| !is_runtime_runc()),
```
utam0k, line 102: Could you add a comment explaining why we need to skip it when using runc?

utam0k:
LGTM w/ one nit

fspv:
Hey! I have addressed the comment, let me know if anything else is needed here. I have a couple more of the tests in my stack already, so will be nice to close this one.

utam0k (approved):
Thanks!

Issue #2335, opened 2023-09-01
Implement container restore functionality


YJDoc2:
ref : continues from https://github.com/containers/youki/issues/142
Currently youki supports checkpointing, (with command name `checkpointt`) , but the restore part has not been implemented yet. We should do that.

__Note__ Before anyone starts with this, make sure the following commands are working as expected on your system:
```bash
# run a loop which keeps printing numbers, with runc runtime ; in background
sudo podman run --runtime runc -dt fedora bash -c "v=0;while true;do sleep 1; echo \"\$v\"; let \"v++\";done;"
# get the name/id of the launched container
sudo podman ps

# this will attach current console to the container. DO NOT do ctrl+c to exit, instead use `a` key
# keep running for some time, let the number increase
sudo podman attach <container-id/name> --detach-keys=a

# enter a, and detach again

# checkpoint and shut-down container
sudo podman container checkpoint <container-id/name>

# wait for a bit

# restore the container
sudo podman container restore <container-id/name>

# attach again immediately 
sudo podman attach <container-id/name> --detach-keys=a

# in the output you see should print numbers >greater than what we saw in previous attach with considerable range
```

After that implement restore in youki, rename the `checkpointt` to `checkpoint` and make the above work with youki instead of runc.

__Another Note__ criu library is quite specific with which kernel versions it supports and need. If you run into criu failure with seg-fault , check [previous issues](https://github.com/checkpoint-restore/criu/issues?q=is%3Aissue+segmentation+fault+is%3Aclosed) on criu and check if you need to upgrade/downgrade library version for your kernel.

I ran above on ubuntu-based, kernel version 6.4.6 , criu v3.17.1 (3.16 does not work)

YJDoc2:
You can check the crun code for help :
checkpoint : https://github.com/opencontainers/runc/blob/main/checkpoint.go
restore : https://github.com/opencontainers/runc/blob/main/restore.go

anti-entropy123:
Hey, I'm trying to research `checkpoint` and `restore`. However, I've noticed that there seem to be some problems with the current `checkpoint` implementation.
```bash
> sudo podman run --runtime ~/rust_project/youki/youki -dt fedora bash -c "v=0;while true;do sleep 1; echo \"\$v\"; let \"v++\";done;"

> sudo podman container --runtime ~/rust_project/youki/youki ps                 
CONTAINER ID  IMAGE                                     COMMAND               CREATED       STATUS           PORTS       NAMES
[...]

PR #3333, opened 2025-12-21
refactor(tty): call setup_console after pivot_root, use syscall for mount_console

nayuta723:
## Description

CRIU checkpoint fails for TTY-enabled containers because the PTY belongs to the host's devpts, not the container's.
CRIU requires that all FDs can be traced back to mounts **within the container's mount namespace**.

**Solution**

Align `setup_console` with runc's approach: 

1. Call `setup_console` after `pivot_root`
   - PTY is created from container's `/dev/pts/ptmx`, not host's
   - FDs now belong to a mount CRIU can track

2. Unify init/exec console setup
   - `setup_console(syscall, fd, mount)` with `mount` param (like runc)
   - `mount=true` for init, `false` for exec

3. Use `syscall.mount()` for consistency

**Result**

| Before | After |
|--------|-------|
| FD 0 → host's `/dev/pts/x` | FD 0 → container's `/dev/pts/0` |
| CRIU: mount not found | CRIU: checkpoint success |

**Reference**
- https://github.com/opencontainers/runc/blob/v1.4.0/libcontainer/standard_init_linux.go
- https://github.com/opencontainers/runc/blob/v1.4.0/libcontainer/rootfs_linux.go

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

**Setup**
```
sudo podman run -d --name test-youki-tty --runtime $RUNTIME -t alpine sh -c 'i=0; while true; do echo $i; i=$((i+1)); sleep 1; done' 2>&1
```

**Result**
- youki（main branch）
```
=== YOUKI (main branch) Container ===
--- FD links ---
total 0
dr-x------ 2 root root  3 Dec 21 05:03 .
dr-xr-xr-x 9 root root  0 Dec 21 05:03 ..
lrwx------ 1 root root 64 Dec 21 05:04 0 -> /dev/pts/8
lrwx------ 1 root root 64 Dec 21 05:04 1 -> /dev/pts/8
lrwx------ 1 root root 64 Dec 21 05:04 2 -> /dev/pts/8
--- mountinfo devpts ---
546 544 0:101 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666

```
- runc
```
=== RUNC Container ===
--- FD links ---
total 0
dr-x------ 2 root root  3 Dec 21 04:58 .
dr-xr-xr-x 9 root root  0 Dec 21 04:58 ..
lrwx------ 1 root root 64 Dec 21 04:58 0 -> /dev/pts/0
lrwx------ 1 root root 64 Dec 21 04:58 1 -> /dev/pts/0
lrwx------ 1 root root 64 Dec 21 04:58 2 -> /dev/pts/0
--- mountinfo devpts ---
509 507 0:94 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
434 507 0:94 /0 /dev/console rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
```
- youki（This PR）
```
=== YOUKI Container ===
--- FD links ---
total 0
dr-x------ 2 root root  3 Dec 21 04:51 .
dr-xr-xr-x 9 root root  0 Dec 21 04:51 ..
lrwx------ 1 root root 64 Dec 21 04:51 0 -> /dev/pts/0
lrwx------ 1 root root 64 Dec 21 04:51 1 -> /dev/pts/0
lrwx------ 1 root root 64 Dec 21 04:51 2 -> /dev/pts/0
--- mountinfo devpts ---
456 454 0:66 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
379 454 0:66 /0 /dev/console rw,nosuid,noexec,relatime - devpts devpts rw,gid=5,mode=620,ptmxmode=666
```

## Related Issues

related to #2335 

## Additional Context

I have verified that the checkpoint functions correctly, as demonstrated by the tests in this Draft PR: https://github.com/nayuta723/youki/pull/6

### DIFF
--- a/crates/libcontainer/src/tty.rs
+++ b/crates/libcontainer/src/tty.rs
@@ -166,6 +166,8 @@ const DEVPTS_SUPER_MAGIC: u64 = 0x1cd1;
 const DEVPTS_SUPER_MAGIC: i64 = 0x1cd1;
 /// Path to the PTY master device
 const PTMX_PATH: &[u8] = b"/dev/ptmx";
+/// Path to the console device
+const CONSOLE_PATH: &str = "/dev/console";
 
 /// Verify that the ptmx handle points to a real /dev/pts/ptmx device.
 ///
@@ -333,13 +335,11 @@ pub fn setup_console(syscall: &dyn Syscall, console_fd: RawFd, mount: bool) -> R
 fn mount_console(syscall: &dyn Syscall, slave: &OwnedFd) -> Result<()> {
     use std::fs::OpenOptions;
 
-    // After pivot_root, the target is /dev/console
-    let console_path = Path::new("/dev/console");
+    let console_path = Path::new(CONSOLE_PATH);
 
     tracing::debug!(
         slave_fd = slave.as_raw_fd(),
-        ?console_path,
-        "mounting PTY on /dev/console"
+        "mounting PTY on {CONSOLE_PATH}"
     );
 
     // Create /dev/console mount target.
@@ -353,7 +353,7 @@ fn mount_console(syscall: &dyn Syscall, slave: &OwnedFd) -> Result<()> {
         .mode(0o666)
         .open(console_path)
         .map_err(|err| {
-            tracing::error!(?err, ?console_path, "failed to create /dev/console");
+            tracing::error!(?err, "failed to create {CONSOLE_PATH}");
             TTYError::CreateDevConsole { source: err }
         })?;
 
@@ -363,16 +363,14 @@ fn mount_console(syscall: &dyn Syscall, slave: &OwnedFd) -> Result<()> {
         tracing::error!(
             ?err,
             slave_fd = slave.as_raw_fd(),
-            ?console_path,
-            "failed to bind mount pty on /dev/console"
+            "failed to bind mount pty on {CONSOLE_PATH}"
         );
         TTYError::MountConsole { source: err }
     })?;
 
     tracing::debug!(
         slave_fd = slave.as_raw_fd(),
-        ?console_path,
-        "mounted PTY on /dev/console"
+        "mounted PTY on {CONSOLE_PATH}"
     );
     Ok(())
 }


crates/libcontainer/src/tty.rs
```diff
@@ -120,34 +141,128 @@ pub fn setup_console_socket(
...
+///
+/// # Arguments
+/// * `console_fd` - The console socket file descriptor
+/// * `mount` - Whether to mount PTY slave on /dev/console (true for init, false for exec)
+///
+/// By creating PTY from container's devpts, the PTY belongs to a mount that
+/// exists within the container's namespace, which is required for CRIU checkpoint.
+///
+/// See: https://github.com/opencontainers/runc/blob/v1.4.0/libcontainer/rootfs_linux.go
+pub fn setup_console(syscall: &dyn Syscall, console_fd: RawFd, mount: bool) -> Result<()> {
```
nayuta723, line 286: Since I'm changing the arguments of a public function, should I keep the original function as deprecated and add a new one instead to avoid a breaking change?

utam0k: No need. We'll release v0.6.0 in the next version, so we're not too concerned about breaking changes.

nayuta723:
@utam0k 
> This change addresses https://github.com/advisories/GHSA-qw9x-cqr3-wc7r. Please verify its safety.

What is the recommended approach for verifying this fix? I want to ensure we have adequate test coverage for this specific edge case.

utam0k:
@nayuta723 As far as I know, it's hard to cover this with e2e tests because it would be caused by TOCTOU. The most common safety method is to add verification and use [mount_from_fd](https://github.com/youki-dev/youki/blob/93951d2bedd732e8a02e4976a486b080a519379a/crates/libcontainer/src/syscall/syscall.rs#L48). 
runc's code:
https://github.com/opencontainers/runc/commit/2f74f4ae1b483b3a4d197f574a37c7b768baf96f#diff-af76b9486c10ca0b6b87e8b8229fa12da73d8ece65e42a0358bec25bfbc21cd9

For example, it might be possible to test this by placing a malicious symbolic link within the rootfs from the start, but it might not be that simple.

nayuta723:
Thanks. In that case, can I assume that the goal of this PR is to demonstrate that the CVE-2025-52565 vulnerability can no longer be reproduced? I will set up a reproduction environment and verify that the updated implementation effectively prevents the exploit.

utam0k:
> Thanks. In that case, can I assume that the goal of this PR is to demonstrate that the [CVE-2025-52565](https://github.com/advisories/GHSA-qw9x-cqr3-wc7r) vulnerability can no longer be reproduced? I will set up a reproduction environment and verify that the updated implementation effectively prevents the exploit.

The goal is not only to reproduce it, but also to avoid directly using `mount`. In other words, we aim to verify the filepath (file descriptor) before mounting and then mount the path using the file descriptor.

nayuta723:
@utam0k 

To mitigate TOCTOU (Time-of-Check to Time-of-Use) attacks, I have introduced `verify_ptmx_handle` and `verify_pty_slave` to validate the file descriptors of the link targets. I have also added unit tests to cover cases where the targets are not a PTMX or PTY:

Valid cases:

- `test_verify_pty_slave_with_real_pty`
- `test_verify_ptmx_handle_with_real_pty`

Invalid cases:

- `test_verify_ptmx_handle_with_regular_file`
- `test_verify_pty_slave_with_regular_file`

Furthermore, I have updated `mount_console` to use `mount_from_fd`. This ensures the operation remains secure even if the path is replaced by a symbolic link.

Do you think these tests are enough, or should I try any other specific cases to make sure the fix is complete?

crates/libcontainer/src/tty.rs
```diff
@@ -120,34 +144,261 @@ pub fn setup_console_socket(
...
+#[cfg(not(target_env = "musl"))]
+const DEVPTS_SUPER_MAGIC: i64 = 0x1cd1;
+
+/// Verify file descriptor using stat and statfs, similar to runc's VerifyInode.
+///
+/// This is a helper function that gets stat/statfs for a file descriptor and
+/// calls the provided verification function with the results.
+///
+/// Ref: https://github.com/opencontainers/runc/blob/v1.4.0/libcontainer/system/linux.go
+fn verify_inode<F>(fd: &OwnedFd, verify: F) -> Result<()>
```
utam0k: How about making it util function not only for tty?

crates/libcontainer/src/tty.rs
```diff
@@ -120,34 +144,261 @@ pub fn setup_console_socket(
...
+        if let Err(err) = mount_console(syscall, slave) {
+            tracing::warn!(
+                ?err,
+                "failed to mount /dev/console, CRIU checkpoint may not work"
+            );
+        }
+    }
+
+    // Send PTY master to console socket
+    let pty_name: &[u8] = b"/dev/ptmx";
```
utam0k: It can modify a constant.

crates/libcontainer/src/tty.rs
```diff
@@ -120,34 +144,261 @@ pub fn setup_console_socket(
...
+    verify_ptmx_handle(master)?;
+    verify_pty_slave(slave)?;
+
+    // Mount PTY slave on /dev/console (only for init container)
+    if mount {
+        if let Err(err) = mount_console(syscall, slave) {
+            tracing::warn!(
+                ?err,
+                "failed to mount /dev/console, CRIU checkpoint may not work"
+            );
```
utam0k: Could you explain why we should skip this error

utam0k:
> Do you think these tests are enough, or should I try any other specific cases to make sure the fix is complete?

They are addressing our needs.

crates/libcontainer/src/tty.rs
nayuta723: Fixed the missing parts. Sorry for the oversight.

Issue #3309, opened 2025-12-03
[FEATURE]: support for OCI Runtime Spec compliant types (State, ContainerProcessState)


nayuta723:
### Background

I'm working on integrating OCI Runtime Specification compliant types into `youki` as part of addressing https://github.com/youki-dev/oci-spec-rs/issues/301. Currently, youki uses its own internal type definitions for container state management, hooks, and seccomp configuration. However, `oci-spec-rs` provides standardized OCI-compliant types that should be used for external interfaces to ensure specification compliance and interoperability with other OCI tooling. Following the pattern used by `runc` (ref: [runc's specs.State usage](https://github.com/opencontainers/runc/search?q=specs.State&type=code)), we should adopt these standardized types from `oci-spec-rs` for external consumption while maintaining youki-specific internal types for optimized runtime operations.

### Feature Request

I'm proposing that `youki` should integrate OCI-compliant types from `oci-spec-rs` for state management, POSIX platform hooks, and seccomp configuration.

### Proposed Solution

Youki should use `oci-spec-rs` OCI-compliant types for external interfaces while maintaining youki-specific internal types for core runtime logic. 
This requires:

- Integrating `oci-spec-rs::State` and `ContainerProcessState` for container state management
- Using `oci-spec-rs::State` for POSIX platform hooks (https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks)
- Adopting o`ContainerProcessState` for seccomp types (https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#seccomp)

### Additional Context

Related to: https://github.com/youki-dev/oci-spec-rs/issues/301 

https://github.com/youki-dev/oci-spec-rs/pull/302 is currently open and adds State and ContainerState types to `oci-spec-rs`. This youki integration work is planned to proceed after that PR is merged.

nayuta723:
Since PR https://github.com/youki-dev/oci-spec-rs/pull/302 has been merged, I would like to proceed with this work. Would it be okay to release `oci-spec-rs` as version v0.8.4?

→ Done

nayuta723:
This involves changes to seccomp and hooks, but since the scope would be quite large, I’m planning to split them into two separate PRs.

- [x] seccomp: https://github.com/youki-dev/youki/pull/3330
- [x] hooks: https://github.com/youki-dev/youki/pull/3346

nayuta723:
Closing this issue as it has been completed.

PR #3346, opened 2026-01-02
refactor(hooks): pass OCI-compliant state to lifecycle hooks

nayuta723:
## Description

Since `hook` is an external API, we should use types that comply with the OCI Runtime Spec instead of youki-specific types. I'm refactoring the code to align with this standard.

This follows the same approach taken by `runc`.
ref: https://github.com/search?q=repo%3Aopencontainers%2Frunc+specs.State+hook&type=code

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

related to #3309 
## Additional Context

Scope: Focused on aligning the State struct used in hooks with OCI specifications. 
Out of Scope: Refactoring `run_hooks` arguments. Given the high impact on existing functionality and test cases, this will be addressed in a separate effort to minimize regression risks.

### DIFF
--- a/crates/libcontainer/src/container/state.rs
+++ b/crates/libcontainer/src/container/state.rs
@@ -250,7 +250,7 @@ pub enum StateConversionError {
 /// Convert internal State to OCI-compliant State (by value, no cloning).
 ///
 /// Based on runc's implementation:
-/// <https://github.com/opencontainers/runc/blob/v2.2.1/libcontainer/container_linux.go#L961>
+/// https://github.com/opencontainers/runc/blob/v2.2.1/libcontainer/container_linux.go#L961
 impl TryFrom<State> for OciState {
     type Error = StateConversionError;
 


crates/libcontainer/src/hooks.rs
```diff
@@ -28,27 +30,59 @@ pub enum HookError {
...
 type Result<T> = std::result::Result<T, HookError>;
 
+fn to_oci_state(state: &State) -> Result<OciState> {
+    let status = to_oci_status_for_hooks(state.status)?;
+
+    let oci_state = OciStateBuilder::default()
+        .version(state.oci_version.clone())
+        .id(state.id.clone())
+        .status(status)
+        .pid(state.pid.unwrap_or_default())
```
saku3: I haven’t looked at runc’s code in detail, but wouldn’t it be better to leave pid unset when it doesn’t exist?  pid is an Option. 
With the current code, pid is set to 0 when it’s missing.

nayuta723:
The current CI failure will be addressed in the following PR: #3348

crates/libcontainer/src/hooks.rs
```diff
@@ -28,27 +30,63 @@ pub enum HookError {
...
+    InvalidStatus(ContainerStatus),
 }
 
 type Result<T> = std::result::Result<T, HookError>;
 
+/// Convert internal State to OCI-compliant State for hooks.
+///
+/// Based on runc's implementation:
+/// <https://github.com/opencontainers/runc/blob/v2.2.1/libcontainer/container_linux.go#L961>
+fn to_oci_state(state: &State, pid: Option<i32>) -> Result<OciState> {
```
utam0k: Could we use `TryFrom` trait?
https://doc.rust-lang.org/std/convert/trait.TryFrom.html

crates/libcontainer/src/hooks.rs
```diff
@@ -28,27 +30,63 @@ pub enum HookError {
...
+        .bundle(state.bundle.clone())
+        .annotations(state.annotations.clone().unwrap_or_default())
+        .build()?;
+
+    oci_state.set_pid(pid);
+
+    Ok(oci_state)
+}
+
+fn to_oci_status_for_hooks(status: ContainerStatus) -> Result<OciContainerState> {
```
utam0k: This too: https://github.com/youki-dev/youki/pull/3346/changes#r2659156200

crates/libcontainer/src/hooks.rs
```diff
@@ -28,27 +30,63 @@ pub enum HookError {
...
+/// <https://github.com/opencontainers/runc/blob/v2.2.1/libcontainer/container_linux.go#L961>
+fn to_oci_state(state: &State, pid: Option<i32>) -> Result<OciState> {
+    let status = to_oci_status_for_hooks(state.status)?;
+
+    let mut oci_state = OciStateBuilder::default()
+        .version(state.oci_version.clone())
+        .id(state.id.clone())
+        .status(status)
+        .bundle(state.bundle.clone())
+        .annotations(state.annotations.clone().unwrap_or_default())
```
utam0k: Annotations is defined as an optional map. Should `None` be empty or nil?

utam0k: We'd like to avoid `clone()`. If we will only pass State to run_hooks, can we remove it?

nayuta723: Regarding the `clone()`, we can eliminate it by passing `State` to `run_hooks` instead of `Option<&Container>`. Since this would involve changing the function signature, should I include this refactoring in this PR, or would you prefer to keep it out of scope for now?

utam0k: In my opinion, we shouldn't include this PR right now. Could you please create an issue for it?

nayuta723: #3351

nayuta723:
I've fixed it. Please take a look.

crates/libcontainer/src/container/state.rs
```diff
@@ -234,6 +238,58 @@ impl State {
...
+    #[error("failed to build OCI state: {0}")]
+    OciStateBuild(#[from] OciSpecError),
+    #[error("invalid container status for OCI conversion: {0}")]
+    InvalidStatus(ContainerStatus),
+}
+
+/// Convert internal State to OCI-compliant State (by value, no cloning).
+///
+/// Based on runc's implementation:
+/// <https://github.com/opencontainers/runc/blob/v2.2.1/libcontainer/container_linux.go#L961>
```
utam0k:
```suggestion
/// https://github.com/opencontainers/runc/blob/v2.2.1/libcontainer/container_linux.go#L961
```

utam0k: Is this a kind of rustdoc style?

nayuta723: It was a simple mistake on my part. After reading the official Rust documentation, I found that this specific syntax is not supported.

PR #3355, opened 2026-01-07
fix(libcgroups): pass `full_path` to Devices controller instead of `cgroup_path`

sou1118:
## Description

Fix `ENOENT` error in Devices controller when using cgroups v2 with BPF device filtering.

The `Devices::apply` function was receiving `cgroup_path` (a relative path like `youki/<container-id>`) instead of `full_path` (an absolute path like `/sys/fs/cgroup/youki/<container-id>`).

This caused `nix::dir::Dir::open()` to fail with `ENOENT` because it interpreted the relative path from the current working directory, which doesn't exist.

### Root Cause

In `manager.rs:170`, the Devices controller was the only controller receiving `cgroup_path`:

```rust
// Other controllers use full_path
ControllerType::Cpu => Cpu::apply(controller_opt, &self.full_path)?,
ControllerType::CpuSet => CpuSet::apply(controller_opt, &self.full_path)?,
// ...

// Devices controller was using cgroup_path (bug)
Devices::apply(controller_opt, &self.cgroup_path)?;
```

## Fix
```diff
- Devices::apply(controller_opt, &self.cgroup_path)?;
+ Devices::apply(controller_opt, &self.full_path)?;
```
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## How to Reproduce (Before Fix)

1. Build youki with v2 and cgroupsv2_devices features:
  `cargo build --package youki --features "v2,cgroupsv2_devices"`
2. Create a minimal OCI bundle with device rules in config.json:
```
{
  "linux": {
    "resources": {
      "devices": [
        { "allow": false, "access": "rwm" },
        { "allow": true, "type": "c", "major": 1, "minor": 3, "access": "rwm" }
      ]
    }
  }
}
```
  3. Run a container on a cgroups v2 system:
  `sudo ./target/debug/youki run --bundle /path/to/bundle test-container`
  4. Error observed:
  `ERROR failed to apply cgroup err=V2(DevicesController(Nix(ENOENT)))`

## After Fix

The container starts successfully:

```
DEBUG Apply Devices cgroup config
DEBUG apply user defined rule: LinuxDeviceCgroup { allow: false, ... }
...
hello from container
```

## Related Issues

Fixes #

## Additional Context

This bug only affects builds with the `cgroupsv2_devices` feature enabled, which uses BPF for device access control on cgroups v2 systems.

### DIFF
--- a/crates/libcgroups/src/v2/manager.rs
+++ b/crates/libcgroups/src/v2/manager.rs
@@ -167,7 +167,7 @@ impl CgroupManager for Manager {
         }
 
         #[cfg(feature = "cgroupsv2_devices")]
-        Devices::apply(controller_opt, &self.cgroup_path)?;
+        Devices::apply(controller_opt, &self.full_path)?;
 
         for pseudoctlr in PSEUDO_CONTROLLER_TYPES {
             if let PseudoControllerType::Unified = pseudoctlr {


utam0k (approved):
Thanks!

utam0k:
We need an e2e test... 😅

PR #3358, opened 2026-01-09
feat(libcgroups): ensure user-defined device cgroup rules take precedence over defaults

sou1118:
## Summary

- Fix device cgroup rule ordering so user-defined rules take precedence over defaults
- Add integration test (`cgroup_rule_precedence`) to verify the fix

## Description
User-defined device cgroup rules in cgroupsv2 were being ignored because default rules were added to the emulator after user rules. Since the BPF program checks rules in reverse order (last added = first checked), default rules took precedence.

### Investigation of Other Runtimes

OCI Runtime Spec:
> "The runtime MUST apply entries in the listed order."

ref: https://specs.opencontainers.org/runtime-spec/config-linux/?v=v1.3.0#RUNTIME-SPEC-CONFIG-LINUX-62

crun ([`src/libcrun/cgroup-resources.c`](https://github.com/containers/crun/blob/main/src/libcrun/cgroup-resources.c)):
- Adds default devices first (reverse iteration)
- Then adds user-specified devices (reverse iteration)
- Default rules are at the start of BPF program → checked first → default rules take precedence

runc ([`vendor/github.com/opencontainers/cgroups/devices/devicefilter.go`](https://github.com/opencontainers/runc/blob/main/vendor/github.com/opencontainers/cgroups/devices/devicefilter.go)):
- Based on crun's ebpf.c implementation
- Uses emulator to optimize/merge rules before BPF generation
- First matching rule wins after emulator processing

youki (`crates/libcgroups/src/v2/devices/program.rs`):
- Uses `rules.iter().rev()` when generating BPF bytecode
- Rules added last to emulator are checked first by BPF

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #

### Manual Testing Steps

Prerequisites:
- Linux environment with cgroup v2 (unified hierarchy)
- youki built with `cgroupsv2_devices` feature

Build:
```bash
cargo build --release --features "v2,cgroupsv2_devices"
```

1. Create test bundle:
```bash
mkdir -p /tmp/test-device-rule/rootfs/bin
cp /bin/busybox /tmp/test-device-rule/rootfs/busybox
ln -s /busybox /tmp/test-device-rule/rootfs/bin/sh
cd /tmp/test-device-rule
youki spec
```

2. Modify config.json:
```bash
# Add device rule to deny write to /dev/zero
cat config.json | jq '
.process.args = ["/busybox", "sh", "-c", "echo test > /dev/zero 2>&1 && echo write_success || echo write_failed"] |
.root.readonly = false |
.linux.resources.devices = [{ "allow": false, "type": "c", "major": 1, "minor": 5, "access": "w" }]
' > config_new.json
mv config_new.json config.json
```

3. Run container:
```bash
sudo youki run test-device-rule
```

Expected result:
- Before fix: `write_success` (user rule ignored, default allows write)
- After fix: `write_failed` (user rule properly denies write)

4. Run integration tests:
```bash
sudo ./scripts/contest.sh ./target/release/youki devices
```

Expected result:
```
# Start group devices
1 / 3 : cgroup_rule_precedence : ok
2 / 3 : device_default_permissions : ok
3 / 3 : device_test : ok
# End group devices
```

## Additional Context

The BPF device filter (`program.rs`) iterates rules in reverse order when generating bytecode:
```rust
for rule in rules.iter().rev() {
    prog.add_rule(rule)?;
}
```
This means rules added last to the emulator are checked first by the BPF program.
The fix ensures user-defined rules are added last, giving them priority over defaults.

### The Fix

This PR aligns youki with crun's approach: add default rules first, then user rules,
so user-defined rules take precedence in BPF's LIFO evaluation.

## Note on Integration Test

The `cgroup_rule_precedence` integration test requires the `cgroupsv2_devices` feature to be enabled.
Since CI does not currently build with this feature, the test is automatically skipped.

To run this test locally:
```bash
cargo build --release --features "v2,cgroupsv2_devices"
sudo ./scripts/contest.sh ./target/release/youki devices
```

Enabling cgroupsv2_devices in CI is tracked as a separate issue.

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -731,3 +731,8 @@ impl Display for MustBePowerOfTwo {
         f.write_str("page size must be in the format of 2^(integer)")
     }
 }
+
+/// Returns true if cgroupsv2_devices feature is enabled.
+pub fn is_cgroupsv2_devices_available() -> bool {
+    cfg!(feature = "cgroupsv2_devices")
+}
--- a/crates/libcgroups/src/v2/devices/controller.rs
+++ b/crates/libcgroups/src/v2/devices/controller.rs
@@ -51,17 +51,12 @@ impl Devices {
     ) -> Result<(), DevicesControllerError> {
         tracing::debug!("Apply Devices cgroup config");
 
-        // FIXME: should we start as "deny all"?
+        // Start with "deny all" as the default policy
         let mut emulator = emulator::Emulator::with_default_allow(false);
 
-        // FIXME: apply user-defined and default rules in which order?
-        if let Some(devices) = linux_devices {
-            for d in devices {
-                tracing::debug!("apply user defined rule: {:?}", d);
-                emulator.add_rule(d);
-            }
-        }
-
+        // Apply default rules first, so user-defined rules take precedence.
+        // BPF program checks rules in reverse order (last added = first checked),
+        // so we add defaults first, then user rules on top.
         for d in [
             default_devices().iter().map(|d| d.into()).collect(),
             default_allow_devices(),
@@ -70,6 +65,14 @@ impl Devices {
         {
             tracing::debug!("apply default rule: {:?}", d);
             emulator.add_rule(&d);
+        }
+
+        // Apply user-defined rules last, so they take precedence over defaults
+        if let Some(devices) = linux_devices {
+            for d in devices {
+                tracing::debug!("apply user defined rule: {:?}", d);
+                emulator.add_rule(d);
+            }
         }
 
         let prog = program::Program::from_rules(&emulator.rules, emulator.default_allow)?;
--- a/tests/contest/contest/src/tests/devices/devices_test.rs
+++ b/tests/contest/contest/src/tests/devices/devices_test.rs
@@ -1,9 +1,13 @@
 use anyhow::{Context, Ok, Result};
+use libcgroups::common::{self as cgroup_common, CgroupSetup, is_cgroupsv2_devices_available};
 use oci_spec::runtime::{
-    LinuxBuilder, LinuxDeviceBuilder, LinuxDeviceType, ProcessBuilder, Spec, SpecBuilder,
+    LinuxBuilder, LinuxDeviceBuilder, LinuxDeviceCgroupBuilder, LinuxDeviceType,
+    LinuxResourcesBuilder, ProcessBuilder, Spec, SpecBuilder,
 };
-use test_framework::{Test, TestGroup, TestResult, test_result};
+use test_framework::{ConditionalTest, Test, TestGroup, TestResult, test_result};
+use tracing::debug;
 
+use crate::utils::support::is_runtime_runc;
 use crate::utils::test_inside_container;
 use crate::utils::test_utils::CreateOptions;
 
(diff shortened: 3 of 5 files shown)


crates/libcgroups/src/v2/devices/controller.rs
```diff
@@ -72,6 +67,14 @@ impl Devices {
             emulator.add_rule(&d);
         }
 
+        // Apply user-defined rules last, so they take precedence over defaults
```
utam0k, line 70: Since the above comment has already mentioned it, it's redundant.
```suggestion
```

crates/libcgroups/src/v2/devices/controller.rs
```diff
@@ -72,6 +67,14 @@ impl Devices {
...
+        if let Some(devices) = linux_devices {
+            for d in devices {
```
utam0k, line 72:
```suggestion
        linux_devices.map(|devices| devices.iter().for_each(|d| {
```

crates/libcgroups/src/common.rs
```diff
@@ -731,3 +731,8 @@ impl Display for MustBePowerOfTwo {
         f.write_str("page size must be in the format of 2^(integer)")
     }
 }
+
+/// Returns true if cgroupsv2_devices feature is enabled.
+pub fn is_cgroupsv2_devices_available() -> bool {
+    cfg!(feature = "cgroupsv2_devices")
+}
```
utam0k, line 738: It's okay for now, but we should make this feature the default.

tests/contest/contest/src/tests/devices/devices_test.rs
```diff
@@ -94,6 +98,73 @@ fn devices_default_permissions_test() -> TestResult {
...
+/// This test requires cgroupsv2_devices feature and cgroup v2.
+fn can_run_cgroup_v2_devices() -> bool {
+    // Skip if cgroupsv2_devices feature is not enabled
+    if !is_cgroupsv2_devices_available() {
+        debug!("cgroupsv2_devices feature is not enabled");
+        return false;
+    }
+    // Skip for runc (this test is specific to youki's implementation)
+    if is_runtime_runc() {
+        debug!("skipping test for runc runtime");
```
utam0k, line 111: We should comment on the reason for skipping this and remove the above comment.

crates/libcgroups/src/common.rs
utam0k: BTW, will it check youki itself? I wonder if this checks youki or contest.

tests/contest/contest/src/tests/devices/devices_test.rs
```diff
@@ -94,6 +98,73 @@ fn devices_default_permissions_test() -> TestResult {
...
+        return false;
+    }
+    let setup_result = cgroup_common::get_cgroup_setup();
+    if !matches!(setup_result, std::result::Result::Ok(CgroupSetup::Unified)) {
+        debug!("cgroup setup is not v2, was {:?}", setup_result);
+        return false;
+    }
```
utam0k, line 118: It's a bit surprising that the setup will be called in the `can_*` function.

tests/contest/contest/src/tests/devices/devices_test.rs
```diff
@@ -94,6 +98,73 @@ fn devices_default_permissions_test() -> TestResult {
...
+    true
+}
+
+/// Create spec for testing device cgroup rule precedence.
+/// This spec sets a rule to deny write access to /dev/zero.
+fn create_spec_cgroup_rule_precedence() -> Result<Spec> {
```
utam0k, line 124: How about this? It makes it clear the purpose of the generated spec.
```suggestion
fn create_spec_deny_dev_zero) -> Result<Spec> {
```

tests/contest/runtimetest/src/tests.rs
```diff
@@ -1378,3 +1378,42 @@ pub fn validate_net_devices(spec: &Spec) {
         }
     }
 }
+
+/// Validates that user-defined device cgroup rules take precedence over defaults.
+/// This test expects /dev/zero write to be denied by a user-defined rule.
+pub fn validate_devices_cgroup_rule_precedence() {
```
utam0k, line 1384: See also https://github.com/youki-dev/youki/pull/3358/changes#r2678063749 for a similar point.

saku3:
@sou1118 

Thank you for the PR.

Looking at the crun code below, I can see that both the `default_devices` array and the user-defined device rules are iterated in reverse order.
https://github.com/containers/crun/blob/main/src/libcrun/cgroup-resources.c#L635C1-L635C15

However, from what I can tell, the BPF program is still constructed in the order 「default」 → 「user」.
Is there anywhere after `create_dev_bpf` where the order is reversed? I couldn’t find such a place.

I didn’t understand the phrase “in BPF’s LIFO evaluation.”
Are you referring specifically to the following code? (Isn’t that about how we generate the BPF program, rather than a property of BPF itself?)

```rust
for rule in rules.iter().rev() {
    prog.add_rule(rule)?;
}
```

Also, what should the result of running `sudo ./scripts/contest.sh crun devices` look like?

<details>
<summary>As a bit of a hack, if I modify `create_container_command` as shown below, it becomes runnable.</summary>
https://github.com/youki-dev/youki/blob/41bbbf545f21ff09a45586ece165b23a7df7b43d/tests/contest/contest/src/utils/test_utils.rs#L66

(crun seems to be stricter than youki about the ordering of command-line arguments.)
```
fn create_container_command<P: AsRef<Path>>(id: &str, dir: P, options: &CreateOptions) -> Command {
    let mut command = Command::new(get_runtime_path());
    command
        .stdout(Stdio::piped())
        .stderr(Stdio::piped())
        .arg("--root")
        .arg(dir.as_ref().join("runtime"))
        .arg("create")
        .arg("--bundle")
        .arg(dir.as_ref().join("bundle"))
        .args(options.extra_args)
        .arg(id);
    if options.no_pivot {
        command.arg("--no-pivot");
    }
    command
}
```
</details>

Sorry for all the questions.

sou1118:
@saku3 Thank you for the detailed review and sorry for the confusion.

You're absolutely right about the BPF evaluation. "BPF's LIFO evaluation" was a poor choice of words - it's not a property of BPF itself, just how youki generates the program with `rules.iter().rev()`.

I also need to correct my PR description. After re-reading the crun code more carefully, I realized I misunderstood its behavior. crun constructs the BPF program in "default → user" order, which means default rules are checked first. There's no reversal after `create_dev_bpf`. So crun actually prioritizes defaults over user rules, not the other way around as I claimed.

The original code had a FIXME comment asking about the rule ordering, and I tried to resolve it by choosing "user takes precedence". But my investigation was clearly insufficient - I should have tested the actual behavior instead of just reading the code.

Regarding the argument ordering issue with crun, I'll submit a separate PR for that fix since it's not related to this change.

Given that my understanding of crun was wrong, would you prefer that youki matches crun's behavior (default takes precedence)? I'm happy to update the PR either way.

saku3:
@sou1118 
Thank you for addressing the FIXME comment.
I’m fine with following crun’s behavior. (Just to be safe, it would be helpful if you could also confirm runc’s behavior as well.)
Thanks in advance.

PR #3360, opened 2026-01-13
Pass State directly to `run_hooks` instead of Container reference

### LINKED ISSUE #3351: refactor `run_hooks` to avoid cloning State
## Background
In #3346, a `clone()` of State is currently required when calling `run_hooks`.

## Proposal
We could eliminate this clone by passing `State` directly to run_hooks
instead of `Option<&Container>`.

## Considerations
This would require changing the `run_hooks` function signature.
Such an API change may affect existing call sites.

## Current Decision
To keep #3346 focused, this refactoring is considered out of scope for now
and will be tracked in this issue.

IrvingMg:
## Description

Refactor `run_hooks` to pass `State` directly instead of `&Container`.

/kind cleanup

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3351

### DIFF
--- a/crates/libcontainer/src/hooks.rs
+++ b/crates/libcontainer/src/hooks.rs
@@ -35,7 +35,7 @@ type Result<T> = std::result::Result<T, HookError>;
 
 pub fn run_hooks(
     hooks: Option<&Vec<Hook>>,
-    state: Option<State>,
+    state: Option<&State>,
     // TODO: Remove the following parameters. To comply with the OCI State, hooks should only depend on structures defined in oci-spec-rs. Cleaning these up ensures proper functional isolation.
     cwd: Option<&Path>,
     pid: Option<Pid>,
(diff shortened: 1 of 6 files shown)


crates/libcontainer/src/hooks.rs
```diff
@@ -35,16 +35,12 @@ type Result<T> = std::result::Result<T, HookError>;
 
 pub fn run_hooks(
     hooks: Option<&Vec<Hook>>,
-    container: Option<&Container>, // TODO: instead of passing the container, we should pass only the state.
+    state: Option<State>,
```
tommady: I’m considering whether this should use Option<&State> instead. Most existing APIs in this area prefer borrowing when ownership isn’t required, and this would let callers avoid cloning the state. Curious if others think this would be a reasonable change here.

saku3: It would be better to use Option<&State>.

IrvingMg: Done. I've made the changes.

PR #3365, opened 2026-01-18
Refactor checkpoint

nayuta723:
## Description

This PR improves the checkpoint functionality by aligning directory handling behavior with crun/runc and cleaning up unimplemented CLI options.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

```
# setup
mkdir -p /tmp/checkpoint-test/rootfs
cd /tmp/checkpoint-test
docker export $(docker create busybox) | tar -C rootfs -xf -
youki spec
sed -i 's/"terminal": true/"terminal": false/' config.json
sed -i 's/"sh"/"sh", "-c", "i=0; while true; do echo $i; i=$((i+1)); sleep 1; done"/' config.json

# run container
sudo youki run -d test_checkpoint </dev/null >/dev/null 2>&1

# checkpoint
sudo youki checkpointt --work-path ./work-dir --image-path ./image-dir --shell-job test_checkpoint
```

## Related Issues

related to #2335

## Additional Context

- Redirecting stdout and stderr to /dev/null during run -d is done to comply with the implementation described below.
  - https://github.com/containerd/containerd/blob/main/cmd/containerd-shim-runc-v2/process/io.go#L89
- I have kept it as 'checkpoint' for now because the restore implementation does not exist yet

### DIFF
--- a/crates/liboci-cli/src/checkpoint.rs
+++ b/crates/liboci-cli/src/checkpoint.rs
@@ -4,6 +4,7 @@ use clap::Parser;
 
 /// Checkpoint a running container
 /// Reference: https://github.com/opencontainers/runc/blob/main/man/runc-checkpoint.8.md
+/// Unimplemented options vs runc: https://github.com/youki-dev/youki/issues/3394
 #[derive(Parser, Debug)]
 pub struct Checkpoint {
     /// Path for saving criu image files


crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -25,13 +25,22 @@ impl Container {
...
+        // checkpointing. is_running() would make more sense here, but let's
         // just reuse existing functions.
         if !self.can_pause() {
             tracing::error!(status = ?self.status(), id = ?self.id(), "cannot checkpoint container because it is not running");
             return Err(LibcontainerError::IncorrectStatus);
         }
 
+        // Create checkpoint image directory if it doesn't exist.
+        // Like crun, allow EEXIST (directory may already exist).
+        if let Err(err) = create_dir(&opts.image_path) {
```
saku3: What permissions does `create_dir` use?

It looks like `crun` creates the directory with mode `0700`.

https://github.com/containers/crun/blob/main/src/libcrun/criu.c#L550

nayuta723: I've looked into it and found that the default for `create_dir` is `0o755`, so I'll update the code accordingly.
[a9a3267](https://github.com/youki-dev/youki/pull/3365/commits/a9a32674d9dcd2bafe282867309a36fdb16bc295)

crates/liboci-cli/src/checkpoint.rs
```diff
@@ -12,45 +12,51 @@ pub struct Checkpoint {
     /// Path for saving work files and logs
     #[clap(long)]
     pub work_path: Option<PathBuf>,
-    /// Path for previous criu image file in pre-dump
-    #[clap(long)]
-    pub parent_path: Option<PathBuf>,
+    /// TODO: Path for previous criu image file in pre-dump
```
saku3, line 16: I think this is fine as it is, but I'd like others to review how we handle unimplemented CLI options. @youki-dev/youki

saku3 (approved):
LGTM.
Thanks!

crates/liboci-cli/src/checkpoint.rs
utam0k: Good catch! Create the issue and make sure to reference it.

nayuta723: I've created the issue and added references to it.

PR #3385, opened 2026-02-02
Sync the state to confirm hooks execution

### LINKED ISSUE #3007: [Bug]: Rust OCI test is maybe flacky
### Bug Description

See https://github.com/youki-dev/youki/actions/runs/12028493866/job/33580623551?pr=2950
This is the first time I have seen rust oci tests failing at all. We need to validate what is the issue and fix it. The Go tests are already quiet flacky I would really prefer the rust test to not be flacky unless the fix is really hard or the frequency is really really small.

### Steps to Reproduce

Not sure, maybe run the failed test from CI https://github.com/youki-dev/youki/actions/runs/12028493866/job/33580623551?pr=2950 repeatedly.

### Expectation

_No response_

### System and Setup Info

_No response_

### Additional Context

_No response_

donkomura:
## Description

The hooks test was flaky because it does not wait for delete operation unlike [the runtime-tools test](https://github.com/opencontainers/runtime-tools/blob/master/validation/hooks/hooks.go). This adds:
- `wait_for_state()` in utils to poll container state with timeout
  - this implementation comes from `lifecycle` tests, but it useful for other tests, so I implement it on utils
- Hook output file path changed to bundle directory to match runtime tool behavior

This ensures the test waits for the container to reach the expected state before proceeding, eliminating timing-related failures.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #3007

### DIFF
--- a/tests/contest/contest/src/tests/hooks/invoke.rs
+++ b/tests/contest/contest/src/tests/hooks/invoke.rs
@@ -1,13 +1,14 @@
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 use anyhow::{anyhow, bail};
 use oci_spec::runtime::{Hook, HookBuilder, HooksBuilder, ProcessBuilder, Spec, SpecBuilder};
 use test_framework::{Test, TestGroup, TestResult};
 
-use crate::utils::test_utils::{CreateOptions, LifecycleStatus, start_container};
 use crate::utils::{
-    create_container, delete_container, generate_uuid, prepare_bundle, set_config, wait_for_state,
+    CreateOptions, LifecycleStatus, WaitTarget, create_container, delete_container, generate_uuid,
+    prepare_bundle, set_config, start_container, wait_for_state,
 };
 
 const STATE_WAIT_TIMEOUT_SECS: u64 = 5;
@@ -17,7 +18,7 @@ fn get_hook_output_path(bundle: &tempfile::TempDir) -> PathBuf {
     bundle.as_ref().join("bundle").join("rootfs").join("output")
 }
 
-fn delete_hook_output_file(path: &PathBuf) -> anyhow::Result<()> {
+fn delete_hook_output_file(path: &Path) -> anyhow::Result<()> {
     match fs::remove_file(path) {
         Ok(()) => Ok(()),
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()),
@@ -80,6 +81,17 @@ fn get_spec(host_output_file: &str) -> Spec {
         .unwrap()
 }
 
+fn wait_for_target(id: &str, bundle_path: &Path, target: WaitTarget) {
+    wait_for_state(
+        id,
+        bundle_path,
+        target,
+        Duration::from_secs(STATE_WAIT_TIMEOUT_SECS),
+        Duration::from_millis(STATE_POLL_INTERVAL_MILLIS),
+    )
+    .unwrap();
+}
+
 fn get_test(test_name: &'static str) -> Test {
     Test::new(
         test_name,
@@ -97,24 +109,14 @@ fn get_test(test_name: &'static str) -> Test {
                 .unwrap()
                 .wait()
                 .unwrap();
-            wait_for_state(
+            wait_for_target(
                 &id_str,
                 bundle.path(),
-                LifecycleStatus::Created,
-                std::time::Duration::from_secs(STATE_WAIT_TIMEOUT_SECS),
-                std::time::Duration::from_millis(STATE_POLL_INTERVAL_MILLIS),
-            )
-            .unwrap();
+                WaitTarget::Status(LifecycleStatus::Created),
+            );
             start_container(&id_str, &bundle).unwrap().wait().unwrap();
             delete_container(&id_str, &bundle).unwrap().wait().unwrap();
-            wait_for_state(
-                &id_str,
-                bundle.path(),
-                LifecycleStatus::Stopped,
-                std::time::Duration::from_secs(STATE_WAIT_TIMEOUT_SECS),
-                std::time::Duration::from_millis(STATE_POLL_INTERVAL_MILLIS),
-            )
-            .unwrap();
+            wait_for_target(&id_str, bundle.path(), WaitTarget::Deleted);
             let log = fs::read_to_string(&host_output_file).expect("cannot read output file");
             delete_hook_output_file(&host_output_file).unwrap();
             let expected = "pre-start1 called\n\
--- a/tests/contest/contest/src/utils/test_utils.rs
+++ b/tests/contest/contest/src/utils/test_utils.rs
@@ -56,6 +56,14 @@ pub enum LifecycleStatus {
     Stopped,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum WaitTarget {
+    Status(LifecycleStatus),
+    // the state after the container is deleted
+    // this state isn't in the runtime spec, but is useful for tests that wait for deletion
+    Deleted,
+}
+
 #[derive(Debug)]
 pub struct ContainerData {
     pub id: String,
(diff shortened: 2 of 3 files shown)


tests/contest/contest/src/tests/hooks/invoke.rs
```diff
@@ -1,38 +1,43 @@
...
+    bundle.as_ref().join("bundle").join("rootfs").join("output")
 }
 
-fn delete_hook_output_file() {
-    std::fs::remove_file(HOOK_OUTPUT_FILE).expect("fail to remove hook output file");
+fn delete_hook_output_file(path: &PathBuf) {
+    if path.exists() {
+        fs::remove_file(path).expect("failed to remove output file");
+    }
 }
```
nayuta723: Instead of checking the path before deleting, please just delete the file and ignore the error if it’s a "NotFound" error.


```suggestion
fn delete_hook_output_file(path: &PathBuf) -> Result<()> {
    match fs::remove_file(path) {
        Ok(()) => Ok(()),
        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()),
        Err(e) => bail!("failed to remove output file: {}", e),
    }
}
```

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -127,6 +127,76 @@ pub fn get_state<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
...
+    poll_interval: Duration,
+) -> Result<()> {
+    let start = std::time::Instant::now();
+
+    while start.elapsed() < timeout {
+        match is_in_state(id, &dir, expected_state) {
+            Ok(true) => return Ok(()),
+            Ok(false) => std::thread::sleep(poll_interval),
+            Err(e) => {
+                if e.to_string().contains("does not exist") && expected_state == "stopped" {
```
nayuta723: Since `is_in_state` already handles the "does not exist" case, is this condition still necessary?

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -127,6 +127,76 @@ pub fn get_state<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
...
+                    Err(err) => bail!("Failed to parse state output as JSON: {} - {}", stdout, err),
+                }
+            }
+        }
+        Err(e) => Err(e.context("failed to get container state")),
+    }
+}
+
+/// Check if a container is in a specific state
+pub fn is_in_state<P: AsRef<Path>>(id: &str, dir: P, expected_state: &str) -> Result<bool> {
```
nayuta723: It's a bit unclear what this boolean value represents. Could you define an Enum or add a docstring to clarify it?

nayuta723 (commented):
@donkomura 
Thanks, I’ve left a few minor comments.

tests/contest/contest/src/utils/test_utils.rs
donkomura: As you've pointed out, this is indeed unnecessary. I'll remove it.

tests/contest/contest/src/utils/test_utils.rs
donkomura: In the runtime-tools implementation, it was a string, but the correct state should be defined.

Well, I'll implement that as Enum.

nayuta723: I apologize for the confusion. My point was that the meaning of the true/false returned by `is_in_state` is unclear. As a proposed fix to make it clearer, how about adding a docstring explaining that the “does not exist” case is treated as `Stopped`?

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -127,6 +136,78 @@ pub fn get_state<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
...
+) -> Result<bool> {
+    match get_container_status(id, &dir) {
+        Ok(status) => Ok(status == expected_state),
+        Err(e) => {
+            if e.to_string().contains("does not exist") {
+                // regard non-existent container as stopped (delete after stopping)
+                if expected_state == LifecycleStatus::Stopped {
+                    return Ok(true);
+                }
+                return Ok(false);
```
nayuta723:
```suggestion
                // regard non-existent container as stopped (delete after stopping)
                return Ok(expected_state == LifecycleStatus::Stopped);
```

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -127,6 +136,78 @@ pub fn get_state<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
...
+    timeout: Duration,
+    poll_interval: Duration,
+) -> Result<()> {
+    let start = std::time::Instant::now();
+
+    while start.elapsed() < timeout {
+        match is_in_state(id, &dir, expected_state) {
+            Ok(true) => return Ok(()),
+            Ok(false) => std::thread::sleep(poll_interval),
+            Err(_) => std::thread::sleep(poll_interval),
```
nayuta723: Could you explain why we are ignoring errors and continuing to wait? If there is a specific reason for this behavior, please add a comment to clarify.

donkomura: [The runtime spec](https://specs.opencontainers.org/runtime-spec/runtime/) says, 'Additional values MAY be defined by the runtime.' So runtimes can optionally add their own states, which might cause JSON parse errors. However, it's spec-compliant, so no error should happen — we should just keep waiting for the expected state.

For other errors, we probably shouldn't ignore them (I'm not confident about this since I'm not reading the runtime spec).

tests/contest/contest/src/utils/test_utils.rs
donkomura: I've got it. I'll make some explanations. Thank you for the suggestion!

tests/contest/contest/src/utils/test_utils.rs
nayuta723: I agree. We should only ignore parse errors. For any other types of errors, we may return them immediately as errors, even before the timeout.

donkomura: I introduced error types to distinguish between parsing errors and other errors.

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -127,6 +147,100 @@ pub fn get_state<P: AsRef<Path>>(id: &str, dir: P) -> Result<(String, String)> {
...
+/// Get the container status as a LifecycleStatus
+pub fn get_container_status<P: AsRef<Path>>(
+    id: &str,
+    dir: P,
+) -> Result<LifecycleStatus, ContainerStateError> {
+    let (stdout, stderr) = get_state(id, &dir).map_err(|e| {
+        if e.to_string().contains("does not exist") {
+            ContainerStateError::ContainerNotFound
+        } else {
+            ContainerStateError::Other(e.context("failed to get container state"))
```
nayuta723:
```suggestion
            ContainerStateError::Other(e)
```

nayuta723 (commented):
I’ve left one comment.

tests/contest/contest/src/tests/hooks/invoke.rs
```diff
@@ -77,27 +84,39 @@ fn get_test(test_name: &'static str) -> Test {
...
-                output
-                    .read_to_string(&mut log)
-                    .expect("fail to read hook log");
-                log
-            };
-            delete_hook_output_file();
+            wait_for_state(
+                &id_str,
+                bundle.path(),
+                LifecycleStatus::Stopped,
```
saku3: Looking at this section in isolation, it may be confusing for future readers.
After delete, the container state no longer exists.

I think it would be better to introduce an enum like `WaitTarget` for `expected_state` in `is_in_state` and `wait_for_state`, so we can explicitly handle a `Deleted` (or `NotFound`) target.

donkomura: Yeah, the current implementation is a bit too complex. I'm trying to clarify it along these lines like bellow. Does this seem reasonable?

```rust
pub enum WaitTarget {
    Status(LifecycleStatus),
    Deleted,
}
```

saku3: Sorry for the late reply. This seems reasonable to me.

donkomura: Thanks for the review. Fixed.

PR #3396, opened 2026-02-08
Add create_runtime hook test

fspv:
## Description

This implements a test similar to a test for a deprecated `prestart` hook test https://github.com/youki-dev/youki/pull/3382

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [ ] Added new unit tests
- [x] Added new integration tests
- [ ] Ran existing test suite
- [ ] Tested manually (please provide steps)

## Related Issues

https://github.com/youki-dev/youki/issues/361

## Additional Context
In https://github.com/youki-dev/youki/pull/3382 it was requested to also create a test  for a `create_runtime` hook in addition to the derecated `prestart` hook.

PR #3406, opened 2026-02-15
Add prestart_fail hook test

fspv:
## Description
This implements a test similar to https://github.com/opencontainers/runtime-tools/blob/master/validation/prestart_fail/prestart_fail.go as a part of the https://github.com/youki-dev/youki/issues/361

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues
#361 

## Additional Context
The Go test tests only one piece of behaviour: when the prestart hook fails the runtime must generate an error. This test also tests that:
* Hooks before the failed hook execute successfully
* Hooks after the failed hook don't execute

### DIFF
--- a/tests/contest/contest/src/tests/prestart_fail/mod.rs
+++ b/tests/contest/contest/src/tests/prestart_fail/mod.rs
@@ -10,14 +10,14 @@ use test_framework::{Test, TestGroup, TestResult};
 use crate::utils::test_utils::CreateOptions;
 use crate::utils::{create_container, delete_container, generate_uuid, prepare_bundle, set_config};
 
-const HOOK_OUTPUT_FILE: &str = "output";
+const OUTPUT_FILE: &str = "output";
 
 fn get_output_file_path(bundle: &tempfile::TempDir) -> PathBuf {
     bundle
         .as_ref()
         .join("bundle")
         .join("rootfs")
-        .join(HOOK_OUTPUT_FILE)
+        .join(OUTPUT_FILE)
 }
 
 fn delete_output_file(path: &PathBuf) {
@@ -40,7 +40,7 @@ fn get_spec(host_output_file: &str) -> Spec {
                 .args(vec![
                     "/bin/sh".to_string(),
                     "-c".to_string(),
-                    "true".to_string(),
+                    format!("echo 'process called' >> {OUTPUT_FILE}"),
                 ])
                 .cwd("/")
                 .build()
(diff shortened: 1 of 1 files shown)


fspv:
Also added a create_runtime failed test https://github.com/youki-dev/youki/pull/3408 similar to https://github.com/youki-dev/youki/pull/3396 which has been added in addition to the prestart test

tests/contest/contest/src/tests/prestart_fail/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+    HookBuilder, HooksBuilder, ProcessBuilder, RootBuilder, Spec, SpecBuilder,
+};
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::utils::test_utils::CreateOptions;
+use crate::utils::{create_container, delete_container, generate_uuid, prepare_bundle, set_config};
+
+const HOOK_OUTPUT_FILE: &str = "output";
+
+fn get_output_file_path(bundle: &tempfile::TempDir) -> PathBuf {
```
YJDoc2, line 15: A nit for separate PR : I see this fn in your other PRs, and I believe this logic would also use used in some other existing tests. Would be better if we could extract the function in the utils and use it from there instead of re-defining every time. Not required change for this PR, but would be good to have in other PRs.

YJDoc2 (commented):
Hey, thanks for the PR. I have something I'd like your opinion on -
- In original go test we would explicitly have a file created v/s not created check. Here we don't use file for that, but just use container creation. I'd prefer if the container command also wrote to the file similar to hooks, instead of just `true`
- If I'm not wrong, then OCI spec imposes order on hooks, so here we can also check that hook 1 ran before hook 2, not just that their output is present or not.

Apart from that, overall looks good to me.

tests/contest/contest/src/tests/prestart_fail/mod.rs
utam0k: +1

fspv: Hey, yes will try to get back to this today or tomorrow, should be easy to do

fspv:
@YJDoc2 replying to your comments.

> In original go test we would explicitly have a file created v/s not created check. Here we don't use file for that, but just use container creation. I'd prefer if the container command also wrote to the file similar to hooks, instead of just true

I'm not sure it makes sense to test container startup logic here, as it is part of the "start" command, not "create". We already test that hooks execute during the "create" stage here: https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/tests/prestart/mod.rs#L103 So it kind of guaranteed that we don't execute start. So testing start functionality here will be redundant.

I hope I understood your question correctly, so let me know if I didn't.

> If I'm not wrong, then OCI spec imposes order on hooks, so here we can also check that hook 1 ran before hook 2, not just that their output is present or not.

Yeah, this makes sense. Will also add order check.

YJDoc2:
@fspv , let me try to elaborate more, I think the way I wrote was confusing.

The original test sets the container process to create a file, and in assertions we check that the file does not exist
```go
g.SetProcessArgs([]string{"sh", "-c", fmt.Sprintf("touch %s", "/output")})
```

In our test, we only have `true` as the container command -
```rust
            ProcessBuilder::default()
                .args(vec![
                    "/bin/sh".to_string(),
                    "-c".to_string(),
                    "true".to_string(),
                ])
```

What I'm suggesting is that we should have the container command similar to hooks, which writes something to the same output file, and assert explicitly that that string is not present. 

My reasoning is that, as of now youki behaves correctly, and will not start container process, but we might have some regression in future (or some other runtime might use this test suit) where it might start the container process even though a prestart hook fails. Adding a check here of the container process o/p would also assert that runtime is not starting the container process if one of the prestart hook fails. I checked the test you are mentioning, but I don't think that considers the prestart fail case, right?

What do you think?

fspv:
> My reasoning is that, as of now youki behaves correctly, and will not start container process

I still think this is covered by other tests, so we will duplicate the existing coverage here. But anyway having more tests never hurt anyone, so added this check as well.

edit: also added an ordering assertion

tests/contest/contest/src/tests/prestart_fail/mod.rs
fspv: Started a separate draft PR for this https://github.com/youki-dev/youki/pull/3412

PR #3384, opened 2026-02-01
[WIP] id mapped mount

YamasouA:
## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [ ] Added new unit tests
- [ ] Added new integration tests
- [ ] Ran existing test suite
- [ ] Tested manually (please provide steps)

## Related Issues

Fixes #2307

PR #3409, opened 2026-02-15
[ID-Mapped Mount] Impl Channel & Message

### LINKED ISSUE #2307: Support idmapped mount
relate
- https://github.com/opencontainers/runtime-spec/pull/1222

YamasouA:
## Description

“I implemented ID-mapped mount support based on runc’s approach. In this PR, I extended the channel between Main and Init to allow passing mount FDs, and added unit tests.”

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #2307 

## Additional Context

The following PR has too many changes, so it will be split and submitted for review.
https://github.com/youki-dev/youki/pull/3384


crates/libcontainer/src/process/channel.rs
```diff
@@ -1,18 +1,18 @@
...
-use crate::process::message::Message;
+use crate::process::message::{Message, MountMsg};
 
 #[derive(Debug, thiserror::Error)]
 pub enum ChannelError {
     #[error("received unexpected message: {received:?}, expected: {expected:?}")]
     UnexpectedMessage {
-        expected: Message,
-        received: Message,
+        expected: Box<Message>,
```
YamasouA: Boxed to keep ChannelError small and avoid clippy::result_large_err.

utam0k:
Could you please share the final design plan using a sequence diagram or similar documentation?

YamasouA:
@utam0k 
I’ve added the sequence diagram and a brief summary to the issue. Please let me know if it covers the information you were looking for.
https://github.com/youki-dev/youki/issues/2307#issuecomment-3977305214

crates/libcontainer/src/process/message.rs
```diff
@@ -35,8 +39,25 @@ impl fmt::Display for Message {
...
             Message::OtherError(s) => write!(f, "OtherError({})", s),
         }
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct MountMsg {
+    pub source: String,
+    pub idmap: Option<MountIdMap>,
+    pub recursive: bool,
```
nayuta723: How is this field intended to be used in relation to `recursive` in `MountIdMap`?

crates/libcontainer/src/process/channel.rs
```diff
@@ -490,6 +569,53 @@ mod tests {
         Ok(())
     }
```
nayuta723, line 541: Could you add a test for `wait_for_mount_fd_request`?

crates/libcontainer/src/process/channel.rs
```diff
@@ -148,12 +158,43 @@ impl MainReceiver {
...
+            })?;
+
+        match msg {
+            Message::MountFdPlease(req) => Ok(req),
+            msg => Err(ChannelError::UnexpectedMessage {
+                expected: Box::new(Message::MountFdPlease(MountMsg {
+                    source: String::new(),
+                    idmap: None,
+                    recursive: false,
+                })),
```
nayuta723: I have a couple of questions on UnexpectedMessage:

- Is the dummy expected value necessary?
Since we don't have the actual "expected" info here, does this dummy value provide any real benefit for debugging?

- Can we simplify it?
Would it be better to just store the received message alone? I'm concerned that dummy values might be confusing for future maintenance.

I know the existing code uses this pattern, but I’d love to make this new part a bit cleaner if possible. What do you think?

nayuta723 (commented):
Thank you for your contribution! I've left a few comments, so please take a look when you have a chance.

crates/libcontainer/src/process/message.rs
YamasouA: @nayuta723 
This is an implementation I initially considered but no longer need.
I will delete it.

crates/libcontainer/src/process/channel.rs
YamasouA: That's right.
I'll correct it.

nayuta723:
@YamasouA 
The DCO check failed because of a missing sign-off. Could you please amend your commit with the `-s` flag?

YamasouA:
@nayuta723 
Thanks for the heads-up. I’ve amended the commit with `-s` and force-pushed the branch.

crates/libcontainer/src/process/channel.rs
```diff
@@ -1,18 +1,24 @@
...
+        "received unexpected message: {received:?}{expected_suffix}",
+        expected_suffix = .expected
+            .as_ref()
+            .map(|msg| format!(", expected: {msg:?}"))
+            .unwrap_or_default()
+    )]
     UnexpectedMessage {
-        expected: Message,
-        received: Message,
+        expected: Option<Box<Message>>,
```
nayuta723: How about this approach? I want to avoid Option as it leaves the meaning of None unclear.
```suggestion
        expected: &'static str,
```

nayuta723 (commented):
I've left a few comments, so please take a look when you have a chance.

crates/libcontainer/src/process/channel.rs
YamasouA: @nayuta723 
That’s certainly true.

nayuta723 (approved):
LGTM, Thanks!

@youki-dev/youki 
Please take a look when you have a moment！

YamasouA:
Thank you for your review @nayuta723.
I think I may have found an issue—could you please bear with me for a moment?

crates/libcontainer/src/process/message.rs
```diff
@@ -16,8 +17,11 @@ pub enum Message {
     SeccompNotifyDone,
     SetupNetworkDeviceReady,
     MoveNetworkDevice(HashMap<String, Vec<CidrAddress>>),
+    MountFdPlease(MountMsg),
```
utam0k: How about this?
```suggestion
    AskMountFd(MountMsg),
```

crates/libcontainer/src/process/message.rs
```diff
@@ -35,8 +39,24 @@ impl fmt::Display for Message {
             Message::SeccompNotifyDone => write!(f, "SeccompNotifyDone"),
             Message::HookRequest => write!(f, "HookRequest"),
             Message::HookDone => write!(f, "HookDone"),
+            Message::MountFdPlease(_) => write!(f, "MountFdPlease"),
```
utam0k: Is there a reason why you aren't showing the contents of `MountFdPlease`?

crates/libcontainer/src/process/message.rs
```diff
@@ -35,8 +39,24 @@ impl fmt::Display for Message {
...
+            Message::MountFdError(err) => write!(f, "MountFdError({})", err),
             Message::ExecFailed(s) => write!(f, "ExecFailed({})", s),
             Message::OtherError(s) => write!(f, "OtherError({})", s),
         }
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct MountMsg {
+    pub source: String,
```
utam0k: This is a path, right? If so, it'd be better to use Path.

crates/libcontainer/src/process/message.rs
```diff
@@ -35,8 +39,24 @@ impl fmt::Display for Message {
...
+    pub idmap: Option<MountIdMap>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct MountIdMap {
+    pub uid_mappings: Vec<LinuxIdMapping>,
+    pub gid_mappings: Vec<LinuxIdMapping>,
+    pub recursive: bool,
```
utam0k: We should use an Enum here since we might need to add more types in the future.

crates/libcontainer/src/process/channel.rs
```diff
@@ -131,8 +141,8 @@ impl MainReceiver {
             Message::ExecFailed(err) => Err(ChannelError::ExecError(err)),
             Message::OtherError(err) => Err(ChannelError::OtherError(err)),
             msg => Err(ChannelError::UnexpectedMessage {
-                expected: Message::IntermediateReady(0),
-                received: msg,
+                expected: "IntermediateReady",
+                received: Box::new(msg),
```
utam0k: This pattern is frequently used. Let's make it a common component.
```rust
impl ChannelError {
      fn unexpected(expected: &'static str, received: Message) -> Self {
          Self::UnexpectedMessage { expected, received: Box::new(received) }
     }
}
```

crates/libcontainer/src/process/channel.rs
```diff
@@ -490,6 +565,81 @@ mod tests {
...
+
+        sender.request_mount_fd(request.clone())?;
+        let received = receiver.wait_for_mount_fd_request()?;
+
+        assert_eq!(received.source, request.source);
+        let received_idmap = received.idmap.context("missing idmap in mount request")?;
+        let request_idmap = request.idmap.context("missing idmap in mount request")?;
+        assert_eq!(received_idmap.recursive, request_idmap.recursive);
+        assert_eq!(received_idmap.uid_mappings, request_idmap.uid_mappings);
+        assert_eq!(received_idmap.gid_mappings, request_idmap.gid_mappings);
```
utam0k: It's a good idea to add PartialEq and Eq.
```rust
assert_eq!(received, request);
```

YamasouA:
@utam0k @nayuta723 
 I split the recursive flags because they control different syscalls. `clone_mount_tree_recursively` controls recursive cloning in `open_tree(2)`, while `apply_idmap_recursively` controls recursive idmap application in `mount_setattr(2)` with `MOUNT_ATTR_IDMAP`.

I also introduced `MountIdMapUsernsSource` because the user namespace used for `MOUNT_ATTR_IDMAP` is not always created from mount specific uid/gid mappings.
When a mount has explicit `uidMappings` / `gidMappings`, we use `Mappings` and create a user namespace from those mappings. 
When the mount only requests `idmap` / `ridmap` without mount-specific mappings, we need to use the container's user namespace instead.
That case is represented by `ContainerUserns`.

utam0k (approved):
LGTM

PR #3408, opened 2026-02-15
Add create_runtime fail hook test

fspv:
## Description
Implements a create_runtime fail hook test as a part of the https://github.com/youki-dev/youki/issues/361. create_runtime hook replaces the deprecated prestart hook.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [ ] Added new unit tests
- [x] Added new integration tests
- [ ] Ran existing test suite
- [ ] Tested manually (please provide steps)

## Related Issues
https://github.com/youki-dev/youki/issues/361
Similar prestart fail test https://github.com/youki-dev/youki/pull/3406

PR #3412, opened 2026-02-18
Deduplicate e2e test hook helpers

fspv:
## Description
There is some duplication in the e2e tests for hooks. This PR moves similar functions to the helpers library so we can reuse them across the tests.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues
https://github.com/youki-dev/youki/pull/3406
https://github.com/youki-dev/youki/pull/3408

## Additional Context
Was raised here https://github.com/youki-dev/youki/pull/3406 and here https://github.com/youki-dev/youki/pull/3408
The functions in the tests in those PRs will be also deduplicated once they're landed and this PR is rebased on main.

### DIFF
--- a/tests/contest/contest/src/tests/hooks/invoke.rs
+++ b/tests/contest/contest/src/tests/hooks/invoke.rs
@@ -4,7 +4,6 @@ use std::time::Duration;
 
 use anyhow::{anyhow, bail};
 use oci_spec::runtime::{Hook, HookBuilder, HooksBuilder, ProcessBuilder, Spec, SpecBuilder};
-use tempfile::TempDir;
 use test_framework::{Test, TestGroup, TestResult};
 
 use crate::utils::{
@@ -17,7 +16,7 @@ const STATE_POLL_INTERVAL_MILLIS: u64 = 100;
 
 const HOOK_OUTPUT_FILE: &str = "output";
 
-pub fn get_hook_output_path(bundle: &TempDir) -> PathBuf {
+pub fn get_hook_output_path(bundle: &tempfile::TempDir) -> PathBuf {
     bundle
         .as_ref()
         .join("bundle")


tests/contest/contest/src/utils/support.rs
```diff
@@ -126,3 +126,31 @@ pub fn wait_for_file_content(
...
+    }
+}
+
+pub fn build_hook(message: &str, host_output_file: &str) -> Hook {
+    HookBuilder::default()
+        .path("/bin/sh")
+        .args(vec![
+            "sh".to_string(),
+            "-c".to_string(),
+            format!("echo '{message}' >> {host_output_file}"),
```
utam0k: What is `{message}`?

tests/contest/contest/src/utils/support.rs
utam0k, line 1: Since it's only used in related hooks tests, how about creating a `tests/hooks` directory and moving the utils there?

fspv: the problem is that `tests/hooks` already exists. And there is a test already defined there https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/tests/hooks/invoke.rs

I agree that it probably makes sense to put these functions somewhere separately, but not sure where. Does something like `tests/contest/contest/src/tests/hook_utils` look good?

tests/contest/contest/src/utils/support.rs
fspv: well, for different tests we want to write different messages. Like "prestart hook called" or whatever. Maybe the name of the function is not very obvious and I should call it something `build_write_message_to_file_hook`

tests/contest/contest/src/utils/support.rs
utam0k: Maybe you could just put it in the hooks dir along with the existing ones? Is there any reason you want to keep it separate from `invoke.rs`?

fspv: moved it to `invoke.rs` in the hooks dir. In fact those functions are already there, I have just re-exported them and reused in the other tests

saku3:
If there are no comments from @utam0k , I’ll merge it this week.

PR #3429, opened 2026-02-27
Add basic restore implementation

nayuta723:
## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [x] Added new unit tests
- [x] Added new integration tests
- [x] Ran existing test suite
- [x] Tested manually (please provide steps)
```
$ sudo podman run --runtime youki -dt fedora bash -c "v=0;while true;do sleep 1; echo \"\$v\"; let \"v++\";do
ne;"
DEBUG youki: started by user 0 with ArgsOs { inner: ["/usr/local/bin/youki", "start", "cb2f992ddbbe56435225c177f9ad5dd6d8eccec92bffcfd88737a9a77dd81d32"] }
DEBUG libcontainer::notify_socket: notify container start
DEBUG libcontainer::notify_socket: notify finished
DEBUG libcontainer::container::container: Save container status: Container { state: State { oci_version: "v1.0.2", id: "cb2f992ddbbe56435225c177f9ad5dd6d8eccec92bffcfd88737a9a77dd81d32", status: Running, pid: Some(80124), bundle: "/var/lib/containers/storage/overlay-containers/cb2f992ddbbe56435225c177f9ad5dd6d8eccec92bffcfd88737a9a77dd81d32/userdata", annotations: Some({"org.opencontainers.image.stopSignal": "15", "io.container.manager": "libpod"}), created: Some(2026-01-12T07:43:28.114251896Z), creator: Some(0), use_systemd: true, clean_up_intel_rdt_subdirectory: Some(false) }, root: "/run/youki/cb2f992ddbbe56435225c177f9ad5dd6d8eccec92bffcfd88737a9a77dd81d32", criu_version: None } in "/run/youki/cb2f992ddbbe56435225c177f9ad5dd6d8eccec92bffcfd88737a9a77dd81d32"
cb2f992ddbbe56435225c177f9ad5dd6d8eccec92bffcfd88737a9a77dd81d32
$ sudo podman attach cb2f992ddbbe --detach-keys=a
[...]

Issue #3394, opened 2026-02-06
Difference between the checkpoint command in runc and youki


nayuta723:
Part of [Implement container restore functionality](https://github.com/youki-dev/youki/issues/2335#issuecomment-3765092260)

There are differences between the checkpoint command in runc and youki.

Please refer to the following:
https://github.com/opencontainers/runc/blob/main/man/runc-checkpoint.8.md

The checkpoint command in youki currently does not support several options that are available in runc, such as --tcp-skip-in-flight and --pre-dump.

donkomura:
I'm interested in implementing checkpoints. Would it be possible for me to collaborate?

nayuta723:
Thank you! However, since the Restore function is still under implementation, we're currently unable to perform the verification. Please give me some time to figure out the best way to handle the splitting units.

nayuta723:
I would like to share two ways you can contribute:

One is to help add components that are currently listed as TODOs for the checkpoint feature in youki.
- I am currently working on a minimal implementation of checkpoint/restore, which you can use as a reference: https://github.com/youki-dev/youki/pull/3429

The other is to contribute by extending criu-rust to enable more options to be used from within youki.
- https://github.com/checkpoint-restore/rust-criu

@donkomura 
Which of these implementations are you interested in?

donkomura:
@nayuta723 
Thank you for your suggestions. Extending rust-criu sounds interesting, but I'd like to implement the former TODOs first.

nayuta723:
#3429 is a simplified summary, and I haven't fully identified which parts can be implemented without extending `rust-criu` yet. Could you help investigate which tasks we should prioritize for the initial implementation?

Also, I’d like to move forward using an Umbrella Issue to track our progress. Could you help organize the tasks into a ToDo list or a set of issues? Feel free to structure it however works best for you!

nayuta723:
To avoid potential conflicts, I've made sure #3429 doesn't depend on the checkpoint functionality. I will split the checkpoint feature into a separate PR from #3429.

donkomura:
Ok, Here is a list of cli options which is implemented on runc but not implemented on youki (or unimplemented on rust-criu).

### Not implemented on youki
- [x] `--manage-cgroups-mode` https://github.com/youki-dev/youki/pull/3452

### Not implemented on rust-criu (needs to expand rust-criu)
- [ ] `--pre-dump`
  - [ ] support in rust-criu
  - [ ] support in youki
- [ ] `--parent-path`
[...]

PR #3438, opened 2026-03-01
feat(checkpoint): add CRIU version validation as prerequisite for checkpoint/restore

nayuta723:
## Description
Add `check_criu_version()` to a new `container_criu` module as a prerequisite
validation before invoking CRIU operations. Before any checkpoint can proceed,
youki now verifies that the installed CRIU binary meets the minimum version
requirement (>= 3.15.0), matching crun's `LIBCRIU_MIN_VERSION`. This ensures
a clear and early failure with a descriptive error when the environment is not
suitable for CRIU-based checkpoint/restore, rather than encountering obscure
failures deep in the CRIU RPC flow.

This is part of the checkpoint feature split described in #3394.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing
- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues
Part of #3394

## Additional Context
This PR is intentionally scoped to the CRIU version validation only, split from
#3429 to avoid conflicts with the restore implementation. The `container_criu`
module will serve as the shared foundation for both checkpoint and restore CRIU
utilities in subsequent PRs.

### DIFF
--- a/crates/libcontainer/src/container/container_criu.rs
+++ b/crates/libcontainer/src/container/container_criu.rs
@@ -1,0 +1,51 @@
+//! Common CRIU utilities for checkpoint and restore operations.
+//!
+//! This module provides shared functionality between container checkpoint and restore,
+//! following the patterns established by runc's CRIU integration.
+
+use crate::error::LibcontainerError;
+
+/// Minimum CRIU version required for checkpoint/restore functionality.
+/// This matches crun's LIBCRIU_MIN_VERSION requirement (3.15.0).
+/// Version format: MAJOR * 10000 + MINOR * 100 + PATCH
+pub const CRIU_VERSION_MINIMUM: u32 = 31500; // 3.15.0
+
+fn compare_criu_version(version: u32, min_version: u32) -> Result<(), LibcontainerError> {
+    if version < min_version {
+        return Err(LibcontainerError::Other(format!(
+            "CRIU version {} is below minimum required version {}",
+            version, min_version,
+        )));
+    }
+    Ok(())
+}
+
+/// Check if CRIU version is greater than or equal to min_version.
+pub fn check_criu_version(min_version: u32) -> Result<(), LibcontainerError> {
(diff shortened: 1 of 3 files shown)


crates/libcontainer/src/container/container_criu.rs
```diff
@@ -0,0 +1,54 @@
...
+pub const CRIU_VERSION_MINIMUM: u32 = 31500; // 3.15.0
+
+fn compare_criu_version(version: u32, min_version: u32) -> Result<(), LibcontainerError> {
+    if version < min_version {
+        return Err(LibcontainerError::Other(format!(
+            "CRIU version {} is below minimum required version {}.{}.{}",
+            version,
+            min_version / 10000,
+            (min_version % 10000) / 100,
+            min_version % 100,
```
YJDoc2: we might as well print min_version without the % , as version itself will be in the `1000*...` format

nayuta723: `runc` also uses the integer format for min_version without the % formatting, as seen here:
https://github.com/opencontainers/runc/blob/a51b20a37bfb8095c22c1db7beb6bd4ff4b67706/libcontainer/criu_linux.go#L83

I'll update the code to match this.

nayuta723:
This might be a flaky test failure. Would you mind rerunning the CI?

crates/libcontainer/src/container/container_criu.rs
YJDoc2: Just for future ref : In my opinion, we don't * have *  to follow runc in each decision. In this case both us and runc reached the same conclusion, but if we could show both versions in a `x.y.z` format, we could have differed from runc.

The only thing we have to align is external interface, so other tools can use youki or runc interchangeably as much as possible, but otherwise it is fine to  have a different behavior than runc if that is appropriate.

PR #3439, opened 2026-03-01
fix: inherit config.json env vars in exec processes

### LINKED ISSUE #3428: [Bug]: config.json env is not inherited into processes spawned with exec
### Bug Description

youki version: 0.5.7
commit: 0.5.7-bd54457ba9de1629f21fe678687f433aec63e7c7

### Steps to Reproduce

- generate spec using `youki spec`
- add `env` entry `AAA=bbb`
- add ubuntu roofs (does not really matter)

<details><summary>Details</summary>
<p>

```
{
  "ociVersion": "1.0.2-dev",
  "root": {
    "path": "rootfs",
    "readonly": true
  },
  "mounts": [
    {
      "destination": "/proc",
      "type": "proc",
      "source": "proc"
    },
    {
      "destination": "/dev",
      "type": "tmpfs",
      "source": "tmpfs",
      "options": [
        "nosuid",
        "strictatime",
        "mode=755",
        "size=65536k"
      ]
    },
    {
      "destination": "/dev/pts",
      "type": "devpts",
      "source": "devpts",
      "options": [
        "nosuid",
        "noexec",
        "newinstance",
        "ptmxmode=0666",
        "mode=0620",
        "gid=5"
      ]
    },
    {
      "destination": "/dev/shm",
      "type": "tmpfs",
      "source": "shm",
      "options": [
        "nosuid",
        "noexec",
        "nodev",
        "mode=1777",
        "size=65536k"
      ]
    },
    {
      "destination": "/dev/mqueue",
      "type": "mqueue",
      "source": "mqueue",
      "options": [
        "nosuid",
        "noexec",
        "nodev"
      ]
    },
    {
      "destination": "/sys",
      "type": "sysfs",
      "source": "sysfs",
      "options": [
        "nosuid",
        "noexec",
        "nodev",
        "ro"
      ]
    },
    {
      "type": "cgroup",
      "source": "cgroup",
      "options": [
        "nosuid",
        "noexec",
        "nodev",
        "relatime",
        "ro"
      ]
    }
  ],
  "process": {
    "terminal": false,
    "user": {
      "uid": 0,
      "gid": 0
    },
    "args": [
      "sh"
    ],
    "env": [
      "TERM=xterm",
      "AAA=bbb"
    ],
    "cwd": "/",
    "capabilities": {
      "bounding": [
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE",
        "CAP_AUDIT_WRITE"
      ],
      "effective": [
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE",
        "CAP_AUDIT_WRITE"
      ],
      "inheritable": [
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE",
        "CAP_AUDIT_WRITE"
      ],
      "permitted": [
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE",
        "CAP_AUDIT_WRITE"
      ],
      "ambient": [
        "CAP_KILL",
        "CAP_NET_BIND_SERVICE",
        "CAP_AUDIT_WRITE"
      ]
    },
    "rlimits": [
      {
        "type": "RLIMIT_NOFILE",
        "hard": 1024,
        "soft": 1024
      }
    ],
    "noNewPrivileges": true
  },
  "hostname": "youki",
  "annotations": {},
  "linux": {
    "resources": {
      "devices": []
    },
    "namespaces": [
      {
        "type": "pid"
      },
      {
        "type": "network"
      },
      {
        "type": "ipc"
      },
      {
        "type": "uts"
      },
      {
        "type": "mount"
      },
      {
        "type": "cgroup"
      }
    ],
    "maskedPaths": [
      "/proc/acpi",
      "/proc/asound",
      "/proc/kcore",
      "/proc/keys",
      "/proc/latency_stats",
      "/proc/timer_list",
      "/proc/timer_stats",
      "/proc/sched_debug",
      "/sys/firmware",
      "/proc/scsi"
    ],
    "readonlyPaths": [
      "/proc/bus",
      "/proc/fs",
      "/proc/irq",
      "/proc/sys",
      "/proc/sysrq-trigger"
    ]
  }
}
```

</p>
</details> 

- `youki run yy`
- `youki exec yy sh -c 'echo $AAA'`

prints nothing.

Do the same, but use `runc`, `crun` or `runsc` exec:

- `runc run yy`
- `runc exec yy sh -c 'echo $AAA'`

All of `runc`, `crun` and `runsc` print `bbb`.

### Expectation

[Spec](https://github.com/opencontainers/runtime-spec/blob/main/config.md#process) is unclear.

But given that `runc`, `crun` and `runsc` inherit env, I guess youki needs to be patched.

### System and Setup Info

_No response_

### Additional Context

_No response_

KevinKickass:
Closes #3428

## Problem

`youki exec` does not inherit environment variables from the container's `config.json` `process.env`. Only variables passed via `--env` on the CLI (plus `PATH` from the spec) are available to the exec process. This differs from runc, crun, and runsc, which all inherit the full `process.env`.

**Reproduction:**
```
youki spec  # add AAA=bbb to process.env
youki run yy
youki exec yy sh -c "echo $AAA"  # prints nothing (expected: bbb)
```

## Root cause

In `crates/libcontainer/src/container/tenant_builder.rs`, the function `get_path_from_spec()` only extracted the `PATH` variable from the spec's env. This single value was passed to `get_environment()`, which built the exec env from CLI `--env` args + that one PATH fallback. All other spec env vars were silently dropped.

## Changes

**`crates/libcontainer/src/container/tenant_builder.rs`** (only file changed):

1. **Replaced `get_path_from_spec()`** with **`get_env_from_spec()`** — returns *all* env vars from the container spec instead of just PATH. The old function was only used in one place, so nothing else is affected.

2. **Rewrote `get_environment()`** — now takes the full spec env as a baseline. CLI `--env` overrides by variable name (matching by the key before `=`). This means:
   - Spec env vars are inherited (matches runc/crun/runsc behavior)
   - `--env FOO=bar` on exec still overrides `FOO` from the spec
   - New vars via `--env` are appended as before

3. **Added 5 unit tests** for `get_environment()`:
   - `env_inherits_spec_vars` — spec vars are passed through
   - `cli_env_overrides_spec` — CLI `--env` takes precedence by name
   - `cli_env_adds_new_vars` — new CLI vars are appended
   - `empty_spec_env_uses_cli_only` — no spec env, CLI only
   - `no_env_at_all` — empty in, empty out

## Backward compatibility

Fully backward compatible. The `--env` flag still works exactly as before (overrides by name). The only difference is that spec env vars are now inherited as the baseline, which is what users expect from the [OCI runtime spec](https://github.com/opencontainers/runtime-spec/blob/main/config.md#process) and how every other runtime handles it.

### DIFF
--- a/tests/contest/contest/src/tests/exec_env/exec_env_test.rs
+++ b/tests/contest/contest/src/tests/exec_env/exec_env_test.rs
@@ -114,10 +114,11 @@ fn test_cli_env_adds_new_var() -> TestResult {
         }
 
         // Exec with a new env var, print all env
-        let (stdout, _) = match exec_container(id, dir, &["/bin/env"], None, &[("NEW_VAR", "hello")]) {
-            Ok(output) => output,
-            Err(e) => return TestResult::Failed(e),
-        };
+        let (stdout, _) =
+            match exec_container(id, dir, &["/bin/env"], None, &[("NEW_VAR", "hello")]) {
+                Ok(output) => output,
+                Err(e) => return TestResult::Failed(e),
+            };
 
         if !stdout.contains("NEW_VAR=hello") {
             return TestResult::Failed(anyhow!(
@@ -175,7 +176,8 @@ fn test_env_from_process_json() -> TestResult {
             return TestResult::Failed(anyhow!("failed to write process.json: {}", e));
         }
 
-        let (stdout, _) = match exec_container(id, dir, &[OsStr::new("")], Some(&process_path), &[]) {
+        let (stdout, _) = match exec_container(id, dir, &[OsStr::new("")], Some(&process_path), &[])
+        {
             Ok(output) => output,
             Err(e) => return TestResult::Failed(e),
         };
--- a/tests/contest/contest/src/tests/exec/mount_test.rs
+++ b/tests/contest/contest/src/tests/exec/mount_test.rs
@@ -23,8 +23,8 @@ pub(crate) fn get_mount_test() -> TestResult {
             return TestResult::Failed(anyhow!("container start failed"));
         }
 
-        let (stdout, _) =
-            exec_container(id, dir, &["cat", "/proc/self/mountinfo"], None, &[]).expect("exec failed");
+        let (stdout, _) = exec_container(id, dir, &["cat", "/proc/self/mountinfo"], None, &[])
+            .expect("exec failed");
 
         let rootfs_lines: Vec<&str> = stdout
             .lines()
--- a/tests/contest/contest/src/tests/exec_cpu_affinity/exec_cpu_affinity_test.rs
+++ b/tests/contest/contest/src/tests/exec_cpu_affinity/exec_cpu_affinity_test.rs
@@ -52,10 +52,11 @@ fn test_cpu_affinity_only_initial_set_from_process_json() -> TestResult {
             return TestResult::Failed(anyhow!("failed to write process.json: {}", e));
         }
 
-        let (_stdout, stderr) = match exec_container(id, dir, &["/bin/true"], Some(&process_path), &[]) {
-            Ok(output) => output,
-            Err(e) => return TestResult::Failed(e),
-        };
+        let (_stdout, stderr) =
+            match exec_container(id, dir, &["/bin/true"], Some(&process_path), &[]) {
+                Ok(output) => output,
+                Err(e) => return TestResult::Failed(e),
+            };
 
         let mask = affinity_mask_from_str(process_affinity_initial);
         let pattern = format!(r".*affinity: 0x{:x}", mask);
--- a/tests/contest/contest/src/tests/personality/mod.rs
+++ b/tests/contest/contest/src/tests/personality/mod.rs
@@ -51,7 +51,8 @@ fn personality_for_linux(domain: LinuxPersonalityDomain, expect: &str) -> TestRe
             return TestResult::Failed(anyhow!("container start failed"));
         }
 
-        let (stdout, _) = exec_container(id, dir, &["uname", "-m"], None, &[]).expect("exec failed");
+        let (stdout, _) =
+            exec_container(id, dir, &["uname", "-m"], None, &[]).expect("exec failed");
 
         if !stdout.contains(expect) {
             return TestResult::Failed(anyhow!("unexpected personality: {}", stdout));


crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -33,24 +33,14 @@ const NAMESPACE_TYPES: &[&str] = &["ipc", "uts", "net", "pid", "mnt", "cgroup"];
...
-    // specified value overriding
-    env.iter()
-        .find(|e| e.starts_with("PATH"))
-        .iter()
-        .next_back()
-        .map(|s| s.to_string())
+/// Returns the environment variables from the container's original spec.
+/// These are inherited by exec processes, matching the behavior of runc,
+/// crun, and runsc. See https://github.com/youki-dev/youki/issues/3428.
+fn get_env_from_spec(spec: &Spec) -> Vec<String> {
```
nayuta723: Is `get_env_from_spec` necessary? We can also avoid using a `match` expression as shown below.

```
spec.process().and_then(|p| p.env().as_ref()).cloned().unwrap_or_default()
```

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -556,27 +546,29 @@ impl TenantContainerBuilder {
...
-                    env_exists = true;
-                }
-                format!("{k}={v}")
+    /// Builds the environment for an exec process. The spec's env vars are
+    /// inherited as a baseline, then any vars passed via `--env` on the CLI
+    /// override them by name. This matches the behavior of runc, crun, and
+    /// runsc. See <https://github.com/youki-dev/youki/issues/3428>.
+    fn get_environment(&self, spec_env: Vec<String>) -> Vec<String> {
+        // Collect CLI overrides into a set of variable names for quick lookup.
+        let cli_keys: std::collections::HashSet<&str> =
```
nayuta723: Since `self.env` is already a `HashMap`, we don't need the extra heap allocation of creating a `HashSet`.

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -795,3 +787,74 @@ mod test {
...
+            detached: false,
+            as_sibling: false,
+            additional_gids: Vec::new(),
+            user: None,
+            group: None,
+        };
+        for (k, v) in env {
+            b.env.insert(k.to_string(), v.to_string());
+        }
+        b
```
nayuta723:
```suggestion
    let env_map: HashMap<String, String> = env
        .iter()
        .map(|(k, v)| (k.to_string(), v.to_string()))
        .collect();

    TenantContainerBuilder::new(base)
        .with_env(env_map)
```

crates/libcontainer/src/container/tenant_builder.rs
KevinKickass: yeah, sure we can do that.

crates/libcontainer/src/container/tenant_builder.rs
KevinKickass: you are right. would be better. saving unnecessary heap alloc

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -795,3 +777,63 @@ mod test {
         Ok(())
     }
 }
+
+#[cfg(test)]
```
saku3: There are both a `mod test` and a `mod tests` module. Could we consolidate them into a single `mod tests` module?

saku3:
Could you please make sure the CI passes?

Also, it would be great if we could add an e2e test for this.

Ref: https://github.com/youki-dev/youki/tree/main/tests/contest/contest

crates/libcontainer/src/container/tenant_builder.rs
KevinKickass: yes. you are right missed that one. should be only one mod test - will fix that

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -455,10 +435,16 @@ impl TenantContainerBuilder {
         let process = if let Some(process) = &self.process {
             self.get_process(process)?
         } else {
-            let original_path_env = get_path_from_spec(spec);
+            // Inherit spec env vars for exec processes (matches runc/crun/runsc).
```
saku3: `// Use the spec's process env as the baseline for exec.`

is enough. The same point is already described in get_environment(), so if the behavior changes later, having it in both places would hurt maintainability.

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -556,27 +542,25 @@ impl TenantContainerBuilder {
...
-        let mut env_exists = false;
-        let mut env: Vec<String> = self
-            .env
-            .iter()
-            .map(|(k, v)| {
-                if k == "PATH" {
-                    env_exists = true;
-                }
-                format!("{k}={v}")
+    /// Builds the environment for an exec process. The spec's env vars are
```
saku3:
```suggestion
    /// Builds the environment for an exec process.
    /// The spec's env vars are used as the baseline, and env vars provided to the
    /// builder, such as those from the CLI, override entries with the same key.
    /// This follows runc's behavior.
    /// See <https://github.com/youki-dev/youki/issues/3428>.
```

I checked the behavior of `crun` and `runsc` as well.

That said, mentioning them in the comment would make us implicitly responsible for keeping that comparison accurate, and there may still be subtle behavioral differences.
Since youki primarily aims for compatibility with `runc`, I think it would be better to mention only `runc` here.

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -794,4 +789,48 @@ mod test {
...
+    fn env_inherits_spec_vars() {
+        let b = builder_with_env(&[]);
+        let spec_env = vec!["PATH=/usr/bin".to_string(), "AAA=bbb".to_string()];
+        let result = b.get_environment(spec_env);
+        assert!(result.contains(&"PATH=/usr/bin".to_string()));
+        assert!(result.contains(&"AAA=bbb".to_string()));
+    }
+
+    #[test]
+    fn cli_env_overrides_spec() {
```
saku3: When I think of this as a unit test for `TenantContainerBuilder`, it is not clear that this is specifically about the CLI.

How about the following name instead?

`builder_env_overrides_spec`

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -794,4 +789,48 @@ mod test {
...
+        let b = builder_with_env(&[("AAA", "override")]);
+        let spec_env = vec!["PATH=/usr/bin".to_string(), "AAA=bbb".to_string()];
+        let result = b.get_environment(spec_env);
+        assert!(result.contains(&"PATH=/usr/bin".to_string()));
+        assert!(result.contains(&"AAA=override".to_string()));
+        assert!(!result.contains(&"AAA=bbb".to_string()));
+    }
+
+    #[test]
+    fn cli_env_adds_new_vars() {
```
saku3: Similarly, how about:

`builder_env_adds_new_vars`

tests/contest/contest/src/tests/exec_env/exec_env_test.rs
```diff
@@ -0,0 +1,258 @@
...
+use test_framework::{Test, TestGroup, TestResult, test_result};
+
+use crate::utils::test_utils::check_container_created;
+use crate::utils::{exec_container, get_runtime_path, start_container, test_outside_container};
+
+fn create_spec_with_env(env: Vec<String>) -> anyhow::Result<Spec> {
+    SpecBuilder::default()
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["sleep".to_string(), "10000".to_string()])
```
saku3:
```suggestion
                .args(vec!["sleep".to_string(), "1000".to_string()])
```

tests/contest/contest/src/tests/exec_env/exec_env_test.rs
```diff
@@ -0,0 +1,258 @@
...
+                .env(env)
+                .build()?,
+        )
+        .build()
+        .context("failed to create spec")
+}
+
+/// Helper: run exec with --env flags via the CLI.
+fn exec_with_env<P: AsRef<Path>>(
```
saku3: It might be better to consolidate this into the existing `exec_container` helper, but that does not need to be done right now.

crates/libcontainer/src/container/tenant_builder.rs
```diff
@@ -794,4 +789,48 @@ mod test {
...
+        let b = builder_with_env(&[("NEW_VAR", "hello")]);
+        let spec_env = vec!["PATH=/usr/bin".to_string()];
+        let result = b.get_environment(spec_env);
+        assert!(result.contains(&"PATH=/usr/bin".to_string()));
+        assert!(result.contains(&"NEW_VAR=hello".to_string()));
+    }
+
+    #[test]
+    fn empty_spec_env_uses_cli_only() {
```
saku3:
```suggestion
    fn empty_spec_env_uses_builder_env_only() {
```

KevinKickass:
Thanks for taking the time to review this so thoroughly — really good catches all around.

Pushed the updates:
- Shortened the inline comment and updated the docstring to reference only runc
- Renamed the unit tests from `cli_env_*` to `builder_env_*`
- Reduced sleep from 10000 to 1000
- Fixed the exec_env integration tests — they were using `PATH=/usr/bin` which doesn't include `/bin` where `sleep` lives in the busybox rootfs. Switched to the full default PATH.
- Consolidated `exec_with_env` into the existing `exec_container` helper by adding an `env` parameter directly, and updated all callers

saku3:
Thank you for the update.
Could you please fix the CI issue?

Running `cargo fmt` should resolve it.

PR #3452, opened 2026-03-13
Support manage-cgroup-mode option for checkpoint

donkomura:
## Description

Add support for CRIU's `--manage-cgroup-mode` option in checkpoint command. This primarily affects restore behavior, and with "ignore" mode, cgroup information will not be dumped.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

## Additional Context

For detail about the option, see https://criu.org/CGroups.

### DIFF
--- a/crates/libcontainer/src/container/container.rs
+++ b/crates/libcontainer/src/container/container.rs
@@ -223,6 +223,7 @@ pub struct CheckpointOptions {
     pub shell_job: bool,
     pub tcp_established: bool,
     pub work_path: Option<PathBuf>,
+    pub manage_cgroups_mode: rust_criu::CgMode,
 }
 
 #[cfg(test)]
--- a/crates/libcontainer/src/container/container_checkpoint.rs
+++ b/crates/libcontainer/src/container/container_checkpoint.rs
@@ -168,6 +168,7 @@ impl Container {
                 .into_string()
                 .unwrap(),
         );
+        criu.cgroups_mode(opts.manage_cgroups_mode.clone());
 
         criu.dump().map_err(|err| {
             tracing::error!(?err, id = ?self.id(), logfile = ?opts.image_path.join(CRIU_CHECKPOINT_LOG_FILE), "checkpointing container failed");
--- a/crates/liboci-cli/src/checkpoint.rs
+++ b/crates/liboci-cli/src/checkpoint.rs
@@ -49,16 +49,14 @@ pub struct Checkpoint {
     // TODO: Do a pre-dump
     // #[clap(long)]
     // pub pre_dump: bool,
-    // TODO: Cgroups mode
-    // #[clap(long)]
-    // pub manage_cgroups_mode: Option<String>,
+    #[clap(long, default_value = "soft", value_parser = clap::builder::PossibleValuesParser::new(["ignore", "full", "strict", "soft"]))]
+    pub manage_cgroups_mode: String,
     // TODO: Checkpoint a namespace, but don't save its properties
     // #[clap(long)]
     // pub empty_ns: bool,
     // TODO: Enable auto-deduplication
     // #[clap(long)]
     // pub auto_dedup: bool,
-    /// Container identifier
     #[clap(value_parser = clap::builder::NonEmptyStringValueParser::new(), required = true)]
     pub container_id: String,
 }
--- a/crates/youki/src/commands/checkpoint.rs
+++ b/crates/youki/src/commands/checkpoint.rs
@@ -17,8 +17,52 @@ pub fn checkpoint(args: Checkpoint, root_path: PathBuf) -> Result<()> {
         shell_job: args.shell_job,
         tcp_established: args.tcp_established,
         work_path: args.work_path,
+        manage_cgroups_mode: parse_cgroups_mode(&args.manage_cgroups_mode)?,
     };
     container
         .checkpoint(&opts)
         .with_context(|| format!("failed to checkpoint container {}", args.container_id))
+}
+
+fn parse_cgroups_mode(s: &str) -> Result<rust_criu::CgMode, anyhow::Error> {
+    match s {
+        "ignore" => Ok(rust_criu::CgMode::IGNORE),
+        "full" => Ok(rust_criu::CgMode::FULL),
+        "strict" => Ok(rust_criu::CgMode::STRICT),
+        "soft" => Ok(rust_criu::CgMode::SOFT),
+        _ => Err(anyhow::anyhow!("invalid manage-cgroups-mode: {s}")),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_cgroups_mode_ok() {
+        assert!(matches!(
+            parse_cgroups_mode("ignore"),
+            Ok(rust_criu::CgMode::IGNORE)
+        ));
+        assert!(matches!(
+            parse_cgroups_mode("full"),
+            Ok(rust_criu::CgMode::FULL)
+        ));
+        assert!(matches!(
+            parse_cgroups_mode("strict"),
+            Ok(rust_criu::CgMode::STRICT)
+        ));
+        assert!(matches!(
+            parse_cgroups_mode("soft"),
+            Ok(rust_criu::CgMode::SOFT)
+        ));
+    }
+
+    #[test]
+    fn test_parse_cgroups_mode_ng() {
+        assert!(parse_cgroups_mode("IGNORE").is_err());
+        assert!(parse_cgroups_mode("Ignore").is_err());
+        assert!(parse_cgroups_mode("unknown").is_err());
+        assert!(parse_cgroups_mode("").is_err());
+    }
 }
--- a/tests/contest/contest/src/tests/lifecycle/checkpoint.rs
+++ b/tests/contest/contest/src/tests/lifecycle/checkpoint.rs
@@ -79,29 +79,12 @@ fn setup_network_namespace(project_path: &Path, id: &str) -> Result<(), TestResu
 fn checkpoint(
     project_path: &Path,
     id: &str,
+    image_path: &Path,
     args: Vec<&str>,
     work_path: Option<&str>,
 ) -> TestResult {
     if let Err(e) = setup_network_namespace(project_path, id) {
         return e;
-    }
-
-    let temp_dir = match tempfile::tempdir() {
-        Ok(td) => td,
-        Err(e) => {
-            return TestResult::Failed(anyhow::anyhow!(
-                "failed creating temporary directory {:?}",
-                e
-            ));
-        }
-    };
-    let checkpoint_dir = temp_dir.as_ref().join("checkpoint");
-    if let Err(e) = std::fs::create_dir(&checkpoint_dir) {
-        return TestResult::Failed(anyhow::anyhow!(
-            "failed creating checkpoint directory ({:?}): {}",
-            &checkpoint_dir,
-            e
-        ));
     }
 
     let additional_args = match work_path {
@@ -111,8 +94,8 @@ fn checkpoint(
 
     let runtime_path = get_runtime_path();
 
-    let checkpoint = Command::new(runtime_path)
-        .stdout(Stdio::piped())
+    let mut cmd = Command::new(runtime_path);
+    cmd.stdout(Stdio::piped())
         .stderr(Stdio::piped())
         .arg("--root")
         .arg(project_path.join("runtime"))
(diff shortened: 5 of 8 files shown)


utam0k (commented):
Is it hard to add the unit or e2e test?

donkomura:
I missed that checkpoint tests were implemented in the contest. I have added test cases.

crates/liboci-cli/src/checkpoint.rs
```diff
@@ -50,8 +50,8 @@ pub struct Checkpoint {
     /// #[clap(long)]
     /// pub pre_dump: bool,
     /// TODO: Cgroups mode
```
nayuta723: Please delete TODO comment.

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -151,6 +165,11 @@ impl Container {
         )
         .map_err(LibcontainerError::OtherIO)?;
 
+        let cgroups_mode = match &opts.manage_cgroup_mode {
```
nayuta723: It should validate the `manage-cgroup-mode` arg on `liboci-cli` pacakge, because invalid arguments currently lead to orphaned directories and inconsistent internal states.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -81,27 +81,25 @@ fn checkpoint(
     id: &str,
     args: Vec<&str>,
     work_path: Option<&str>,
-) -> TestResult {
-    if let Err(e) = setup_network_namespace(project_path, id) {
-        return e;
-    }
+) -> Result<tempfile::TempDir, TestResult> {
```
nayuta723: The checkpoint function shouldn't be responsible for creating and returning a `tempfile::TempDir`. It's better to decouple directory management. We should allow the caller to pass an image_path (e.g., via a new option), which makes the function more predictable and easier to test.

nayuta723: BTW, I'm on board with using a Result type for the return value.

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
```diff
@@ -106,6 +106,28 @@ impl ContainerLifecycle {
         )
     }
 
+    pub fn checkpoint_manage_cgroups_mode_ignore(&self) -> TestResult {
```
nayuta723, line 111: Is there a reason why `full`, `strict`, and `props` are excluded from the test cases?

crates/libcontainer/src/container/container_checkpoint.rs
```diff
@@ -22,6 +22,20 @@ pub enum CheckpointError {
     CriuError(String),
 }
 
+fn parse_cgroups_mode(s: &str) -> Result<rust_criu::CgMode, LibcontainerError> {
+    match s {
+        "ignore" => Ok(rust_criu::CgMode::IGNORE),
+        "none" => Ok(rust_criu::CgMode::NONE),
+        "props" => Ok(rust_criu::CgMode::PROPS),
```
nayuta723: `props` mode should be excluded since it is not supported by `runc` or `crun`.

nayuta723 (commented):
Thanks! I've left some comments. Please take a look when you have a chance.

crates/libcontainer/src/container/container_checkpoint.rs
donkomura: That's a good catch. We should disable this to make the responsibilities of youki and CRIU distinct.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
donkomura: Fix it to create TempDir at caller.

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
donkomura: Since checkpoints show different behavior depending on whether 'ignore' or other options are chosen, I've included tests for 'ignore' and 'soft' as representative cases to clarify the meaning of the tests in the code.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -79,30 +79,11 @@ fn setup_network_namespace(project_path: &Path, id: &str) -> Result<(), TestResu
 fn checkpoint(
     project_path: &Path,
     id: &str,
+    image_path: &Path,
```
nayuta723, line 82:
```suggestion
    image_path: Option<&Path>,
```
I suggest wrapping image_path in an Option, since some use cases don't require an image path to be specified.

crates/liboci-cli/src/checkpoint.rs
nayuta723: I would like to put this on hold until #3455 is merged.

crates/libcontainer/src/container/container.rs
```diff
@@ -223,6 +223,7 @@ pub struct CheckpointOptions {
     pub shell_job: bool,
     pub tcp_established: bool,
     pub work_path: Option<PathBuf>,
+    pub manage_cgroup_mode: String,
```
nayuta723:
```suggestion
    pub manage_cgroup_mode: rust_criu::CgMode,
```

nayuta723: `parse_cgroups_mode` is currently called inside `container_checkpoint.rs`, which means an invalid value error surfaces deep within libcontainer during checkpoint execution. This makes it harder to diagnose and handle for callers.

Since `liboci-cli` is intentionally kept dependency-free (only clap), I'd suggest moving the parsing to `youki/src/commands/checkpoint.rs` and changing `CheckpointOptions.manage_cgroup_mode` from `String` to `rust_criu::CgMode`.

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
nayuta723: Thanks. Could you please include this intention in the doc string for future reference?

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -130,43 +111,200 @@ fn checkpoint(
...
+        }
+    };
+    let image_path = temp_dir.path().join("checkpoint");
+    if let Err(e) = std::fs::create_dir(&image_path) {
+        return TestResult::Failed(anyhow::anyhow!(
+            "failed creating checkpoint directory ({:?}): {}",
+            &image_path,
+            e
+        ));
+    }
```
nayuta723: Please refactor the common logic into a shared function. BTW, is it necessary to create the `image_path` there?

crates/liboci-cli/src/checkpoint.rs
```diff
@@ -49,9 +49,8 @@ pub struct Checkpoint {
     /// TODO: Do a pre-dump
     /// #[clap(long)]
     /// pub pre_dump: bool,
-    /// TODO: Cgroups mode
-    /// #[clap(long)]
-    /// pub manage_cgroups_mode: Option<String>,
+    #[clap(long, default_value = "soft", value_parser = clap::builder::PossibleValuesParser::new(["ignore", "none", "full", "strict", "soft"]))]
```
nayuta723: We should avoid showing `none`, to stay consistent with `runc`.
https://github.com/opencontainers/runc/blob/main/checkpoint.go#L43

nayuta723 (changes requested):
Thanks for the update! I've reviewed your changes and left a few comments. Please take a look when you have a chance.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
donkomura: Without specifying image_path, we can't control where CRIU writes the image files, and thus can't inspect them.

crates/libcontainer/src/container/container.rs
donkomura: I see. I'll parse it in youki/src/commands/checkpoint.rs.

crates/liboci-cli/src/checkpoint.rs
donkomura: Is the purpose to prevent conflicts? If so, I will leave it as is.

crates/libcontainer/src/container/container_checkpoint.rs
donkomura: validating on clap

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
donkomura: fixed

crates/liboci-cli/src/checkpoint.rs
donkomura: fixed, enum as well

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
nayuta723: I understand that the `checkpoint` function must have an `image_path` to operate. If it remains an `Option<&Path>`, the function should internally generate a default path when None is provided. Alternatively, we could remove the `Option` and make it a required argument passed from the outside.
I recommend the first approach, as providing a default path internally seems more convenient.

donkomura: Ok, I will set the default path.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -79,30 +77,21 @@ fn setup_network_namespace(project_path: &Path, id: &str) -> Result<(), TestResu
 fn checkpoint(
     project_path: &Path,
     id: &str,
+    image_path: Option<&Path>,
```
nayuta723: I’m sorry to backtrack on my previous comment, but I think it’s better to remove the `Option` after all, as there is processing inside the function that relies on `image_path`.
```suggestion
    image_path: &Path,
```

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -79,30 +77,21 @@ fn setup_network_namespace(project_path: &Path, id: &str) -> Result<(), TestResu
...
     args: Vec<&str>,
     work_path: Option<&str>,
-) -> TestResult {
-    if let Err(e) = setup_network_namespace(project_path, id) {
-        return e;
-    }
+) -> Result<(), TestResult> {
```
nayuta723: I’m also sorry to backtrack again, but I think it’s better to remove the `Result` as well. Requiring the caller to handle errors with match every time feels a bit excessive/clunky in this case.
```suggestion
) -> TestResult {
```

donkomura:
Thank you for reviewing, and sorry for the late reply. (I was on vacation.)

crates/youki/src/commands/checkpoint.rs
```diff
@@ -17,8 +17,52 @@ pub fn checkpoint(args: Checkpoint, root_path: PathBuf) -> Result<()> {
...
         .with_context(|| format!("failed to checkpoint container {}", args.container_id))
 }
+
+fn parse_cgroups_mode(s: &str) -> Result<rust_criu::CgMode, anyhow::Error> {
+    match s {
+        "ignore" => Ok(rust_criu::CgMode::IGNORE),
+        "full" => Ok(rust_criu::CgMode::FULL),
+        "strict" => Ok(rust_criu::CgMode::STRICT),
+        "soft" => Ok(rust_criu::CgMode::SOFT),
+        _ => Err(anyhow::anyhow!("manage-cgroup-mode: {s}")),
```
saku3:
```suggestion
        _ => Err(anyhow::anyhow!("invalid manage-cgroups-mode: {s}")),
```

Alternatively, since this is already validated by clap::builder::PossibleValuesParser, I think the following would also be reasonable.

```
_ => unreachable!("invalid manage-cgroups-mode: {s}"),
```

crates/libcontainer/src/container/container.rs
```diff
@@ -223,6 +223,7 @@ pub struct CheckpointOptions {
     pub shell_job: bool,
     pub tcp_established: bool,
     pub work_path: Option<PathBuf>,
+    pub manage_cgroup_mode: rust_criu::CgMode,
```
saku3: Overall, I think manage_cgroup_mode should be renamed to manage_cgroups_mode consistently throughout the codebase.

donkomura: Good catch, I'll fix it.

PR #3475, opened 2026-03-30
ci: install CRIU in containerd integration tests and skip checkpoint tests

nayuta723:
## Description
In the containerd integration test job, some tests were being silently skipped because CRIU was not installed in the CI environment. Silent skips are easy to overlook and can lead to reduced test coverage going unnoticed.

This PR makes the skipping explicit: install CRIU via `ppa:criu/ppa` so it is available in the environment, and add `-skip TestCheckpoint` to `TESTFLAGS` so that the checkpoint/restore tests — which happen to require a CRIU-compatible kernel configuration not present in CI — are visibly excluded rather than silently dropped.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [x] CI/CD related changes

## Testing
- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues
Related to https://github.com/youki-dev/youki/issues/3476

## Additional Context
The `-skip` flag is available from Go 1.20+; this job uses Go `1.20.12`.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -56,8 +56,10 @@ jobs:
         with:
           go-version: '1.20.12'
           cache: true
+      - name: Add CRIU PPA
+        run: sudo add-apt-repository -y ppa:criu/ppa
       - run: sudo apt-get -y update
-      - run: sudo apt-get install -y pkg-config libsystemd-dev libelf-dev libseccomp-dev btrfs-progs libbtrfs-dev
+      - run: sudo apt-get install -y pkg-config libsystemd-dev libelf-dev libseccomp-dev btrfs-progs libbtrfs-dev criu
       - name: Build containerd
         run: |
           make build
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -76,7 +78,7 @@ jobs:
           sudo cp youki /usr/bin/runc
           runc --version
       - name: Integration Test
-        run: sudo make RUNC_FLAVOR=crun TEST_RUNTIME=io.containerd.runc.v2 TESTFLAGS="-timeout 40m" integration
+        run: sudo make RUNC_FLAVOR=crun TEST_RUNTIME=io.containerd.runc.v2 TESTFLAGS="-timeout 40m -skip TestCheckpoint" integration
```
saku3: `For tests that youki does not yet support, add explicit -skip entries with comments explaining why they are excluded`

You mentioned that in the issue, but does that mean it does not need to be commented on here?

nayuta723: You're right, I missed mentioning that here, so I'll add it now. Thank you for pointing it out!

PR #3484, opened 2026-04-06
Checking source file type with `is_dir()` in bind mount implementation

logica0419:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #3483

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -1022,9 +1022,9 @@ pub fn find_parent_mount(
 mod tests {
     #[cfg(feature = "v1")]
     use std::fs;
-    use std::fs::OpenOptions;
     use std::os::unix::fs::symlink;
     use std::str::FromStr;
+    use std::{fs::OpenOptions, os::unix::net::UnixListener};
 
     use anyhow::{Context, Ok, Result};
 
@@ -1117,6 +1117,39 @@ mod tests {
                     data: None,
                 },
             ];
+            let got = &m
+                .syscall
+                .as_any()
+                .downcast_ref::<TestHelperSyscall>()
+                .unwrap()
+                .get_mount_args();
+            assert_eq!(want, *got);
+            assert_eq!(got.len(), 2);
+        }
+        {
+            // Socket file bind-mount
+            // https://github.com/youki-dev/youki/issues/3483
+            let m = Mount::new();
+            let mount = &SpecMountBuilder::default()
+                .destination(PathBuf::from("/tmp.sock"))
+                .typ("bind")
+                .source(tmp_dir.path().join("tmp.sock"))
+                .build()?;
+            let mount_option_config = parse_mount(mount)?;
+            UnixListener::bind(tmp_dir.path().join("tmp.sock"))?;
+
+            assert!(
+                m.mount_into_container(mount, tmp_dir.path(), &mount_option_config, None)
+                    .is_ok()
+            );
+
+            let want = vec![MountArgs {
+                source: Some(tmp_dir.path().join("null")),
+                target: tmp_dir.path().join("dev/null"),
+                fstype: Some("bind".to_string()),
+                flags: MsFlags::empty(),
+                data: Some("".to_string()),
+            }];
             let got = &m
                 .syscall
                 .as_any()
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -1022,9 +1022,10 @@ pub fn find_parent_mount(
 mod tests {
     #[cfg(feature = "v1")]
     use std::fs;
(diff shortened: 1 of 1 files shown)


logica0419:
I added a test case, but it looks like the fd-based test is currently being skipped.
What should I do about that?

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -1126,6 +1127,39 @@ mod tests {
...
+                .build()?;
+            let mount_option_config = parse_mount(mount)?;
+            UnixListener::bind(tmp_dir.path().join("tmp.sock"))?;
+
+            assert!(
+                m.mount_into_container(mount, tmp_dir.path(), &mount_option_config, None)
+                    .is_ok()
+            );
+
+            let want = vec![MountArgs {
```
saku3, line 1147: The expected value here seems inconsistent with the test case.

This test creates a mount for tmp.sock, so expecting /dev/null in want does not appear to match the setup.

saku3 (approved):
Thanks.

This fix looks good to me.

runc also branches on whether the source is a directory, and treats non-directories as file-like mount targets.

https://github.com/opencontainers/runc/blob/80fc1cd34e1e1f3417ed8add4b60121ac06492bd/libcontainer/rootfs_linux.go#L557

saku3:
>I added a test case, but it looks like the fd-based test is currently being skipped.

That was my oversight when I changed the mount handling to use the fd-based path.

The test will also need to be updated to cover the fd-based path, but I think it is fine to leave that out of scope for this PR.

crates/libcontainer/src/rootfs/mount.rs
logica0419: Oh, I missed that, thank you so much!

logica0419:
> The test will also need to be updated to cover the fd-based path, but I think it is fine to leave that out of scope for this PR.

I see. I also think this should be fixed in a separate PR.

I’m wondering if there’s a way to run the existing test cases.
When I run tests in `mount.rs`, it fails with the error:
```
running 1 test

thread 'rootfs::mount::tests::test_mount_into_container' (116078) panicked at crates/libcontainer/src/syscall/test.rs:312:9:
not yet implemented
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/panicking.rs:80:14
   2: core::panicking::panic
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/panicking.rs:150:5
   3: <libcontainer::syscall::test::TestHelperSyscall as libcontainer::syscall::syscall::Syscall>::fsopen
             at ./src/syscall/test.rs:312:9
   4: libcontainer::rootfs::mount::Mount::mount_into_container::{{closure}}
             at ./src/rootfs/mount.rs:705:47
   5: libcontainer::rootfs::mount::Mount::mount_into_container
             at ./src/rootfs/mount.rs:794:19
   6: libcontainer::rootfs::mount::tests::test_mount_into_container
             at ./src/rootfs/mount.rs:1057:19
   7: libcontainer::rootfs::mount::tests::test_mount_into_container::{{closure}}
             at ./src/rootfs/mount.rs:1037:39
   8: core::ops::function::FnOnce::call_once
             at /home/logica0419/.rustup/toolchains/1.92.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test rootfs::mount::tests::test_mount_into_container ... FAILED

failures:

failures:
    rootfs::mount::tests::test_mount_into_container
```

saku3:
We need to implement mock versions of fsopen, fsconfig, fsmount, and open_tree for the tests :cry:

https://github.com/youki-dev/youki/blob/main/crates/libcontainer/src/syscall/test.rs#L311

If you’re interested, feel free to rewrite the tests. If not, I’m happy to take care of it on my side as well.

>I’m wondering if there’s a way to run the existing test cases.

The existing test cases cannot be run unless you are using a version earlier than v0.5.6.

saku3:
We will handle the tests in the following issue.

https://github.com/youki-dev/youki/issues/3485

I will merge this PR.

logica0419:
I got it. Thank you!

PR #3493, opened 2026-04-12
Add checkpoint/restore integration tests to contest

### LINKED ISSUE #3447: Add checkpoint/restore integration tests to contest
## Summary

Add checkpoint/restore integration tests to `tests/contest` equivalent to [`tests/integration/checkpoint.bats`](https://github.com/opencontainers/runc/blob/main/tests/integration/checkpoint.bats) in runc.

## Motivation

The `contest` framework (Rust-based OCI runtime integration tests) currently has no coverage for checkpoint/restore functionality. The existing BATS-based tests in runc's `checkpoint.bats` cover this area but are not available through `contest`.

## Tests to implement

All 16 tests from runc's `checkpoint.bats`:

- [ ] checkpoint and restore
- [ ] checkpoint and restore (bind mount, destination is symlink)
- [ ] checkpoint and restore (with --debug)
- [ ] checkpoint and restore (cgroupns)
- [ ] checkpoint and restore with netdevice
- [ ] checkpoint and restore with netdevice (bind mount, destination is symlink)
- [ ] checkpoint and restore with netdevice (with --debug)
- [ ] checkpoint and restore with netdevice (cgroupns)
- [ ] checkpoint --pre-dump (bad --parent-path)
- [ ] checkpoint --pre-dump and restore
- [ ] checkpoint --lazy-pages and restore
- [ ] checkpoint and restore in external network namespace
- [ ] checkpoint and restore with container specific CRIU config
- [ ] checkpoint and restore with nested bind mounts
- [ ] checkpoint then restore into a different cgroup (via --manage-cgroups-mode ignore)
- [ ] checkpoint/restore and exec

## Notes

- All tests are skipped when running with youki (checkpoint/restore not yet implemented) and when CRIU is not installed.
- Some tests have additional preconditions (e.g. `mem_dirty_track` / `uffd-noncoop` CRIU features, cgroups v1 + cgroupns).
- Requires adding `DummyDevice` / `NetNamespace` utilities to `tests/contest/contest/src/utils/net.rs` to support netdevice tests.

tommady:
## Description

| Upstream runc Test| Ported to youki? | Notes |
|-------------------|------------------|-------|
| checkpoint and restore with netdevice | ✅ Yes | Uses NetNamespace and DummyDevice utilities |
| checkpoint and restore | ✅ Yes | |
| checkpoint and restore (bind mount, destination is symlink) | ✅ Yes | |
| checkpoint and restore (with --debug) | ✅ Yes | |
| checkpoint and restore (cgroupns) | ✅ Yes | Checks cgroups v1 and cgroupns |
| checkpoint --pre-dump (bad --parent-path) | ✅ Yes | |
| checkpoint --pre-dump and restore | ✅ Yes | |
| checkpoint --lazy-pages and restore | ✅ Yes | Uses status-fd IPC and background criu daemon |
| checkpoint and restore in external network namespace | ✅ Yes | Uses NetNamespace |
| checkpoint and restore with container specific CRIU config | ✅ Yes | Creates custom_criu.conf and injects org.criu.config annotation |
| checkpoint and restore with nested bind mounts | ✅ Yes | Uses host directory disruption |
| checkpoint then restore into a different cgroup (via --manage-cgroups-mode ignore) | ✅ Yes | Dynamically verifies cgroup paths via /proc |
| checkpoint/restore and exec | ✅ Yes | Runs background process and verifies pid restoration |

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3447 phase1

### DIFF
--- a/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
+++ b/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
@@ -175,8 +175,14 @@ fn simple_cr(
             ));
         }
 
-        if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), global_args)
-        {
+        if let Err(e) = restore_container(
+            bundle.path(),
+            id,
+            image_dir,
+            Some(work_dir),
+            &[],
+            global_args,
+        ) {
             return TestResult::Failed(anyhow!("restore failed: {e}"));
         }
 
@@ -488,7 +494,7 @@ fn checkpoint_pre_dump_and_restore() -> TestResult {
     }
 
     // Restore the container from the final dump directory
-    if let Err(e) = restore_container(bundle.path(), id, &final_dump_dir, None, &[]) {
+    if let Err(e) = restore_container(bundle.path(), id, &final_dump_dir, None, &[], &[]) {
         return TestResult::Failed(anyhow!("restore failed: {e}"));
     }
 
@@ -547,6 +553,9 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         &[],
     );
 
+    checkpoint_cmd.stdout(std::process::Stdio::piped());
+    checkpoint_cmd.stderr(std::process::Stdio::piped());
+
     let pipe_w_raw = pipe_w.as_raw_fd();
 
     // Ensure the child process inherits the write end of the pipe
@@ -566,6 +575,20 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         Err(e) => return TestResult::Failed(anyhow!("failed to spawn checkpoint: {e}")),
     };
 
+    // Close the write end in the parent immediately so read() will get EOF if CRIU exits
+    let _ = nix::unistd::close(pipe_w.as_raw_fd());
+
+    // Set read pipe to non-blocking so the timeout loop isn't blocked forever if no data arrives
+    let flags = nix::fcntl::OFlag::from_bits_truncate(
+        nix::fcntl::fcntl(pipe_r.as_raw_fd(), nix::fcntl::FcntlArg::F_GETFL)
+            .expect("F_GETFL failed"),
+    );
+    nix::fcntl::fcntl(
+        pipe_r.as_raw_fd(),
+        nix::fcntl::FcntlArg::F_SETFL(flags | nix::fcntl::OFlag::O_NONBLOCK),
+    )
+    .expect("F_SETFL failed");
+
     // Wait for CRIU to become ready by reading from the pipe
     let mut buf = [0u8; 1];
     let mut ready = false;
@@ -576,15 +599,22 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
                 ready = true;
                 break;
             }
+            Ok(0) => {
+                // EOF - CRIU died or closed its end early
+                break;
+            }
+            Err(nix::errno::Errno::EAGAIN) => {
+                std::thread::sleep(Duration::from_millis(100));
+            }
             _ => std::thread::sleep(Duration::from_millis(100)),
         }
     }
 
-    // Cleanup pipes
+    // Cleanup read pipe
     let _ = nix::unistd::close(pipe_r.as_raw_fd());
-    let _ = nix::unistd::close(pipe_w.as_raw_fd());
 
     if !ready {
+        let _ = checkpoint_child.kill();
         let _ = checkpoint_child.wait();
         return TestResult::Failed(anyhow!("lazy-page server readiness timeout"));
     }
@@ -610,6 +640,15 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         }
     };
 
+    let restore_id = format!("{id}-restore");
+    let new_cgroup = format!("/runtime-test/cgroup-2-{restore_id}");
+    let mut spec =
+        oci_spec::runtime::Spec::load(bundle.path().join("bundle").join("config.json")).unwrap();
+    if let Some(linux) = spec.linux_mut() {
+        linux.set_cgroups_path(Some(PathBuf::from(&new_cgroup)));
+    }
+    set_config(bundle.path(), &spec).unwrap();
+
     // Restore lazily from checkpoint
     let restore_result = restore_container(
         bundle.path(),
@@ -617,6 +656,7 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         image_dir,
         Some(work_dir),
         &["--lazy-pages", "--manage-cgroups-mode=ignore"],
+        &[],
     );
 
     // Wait for background jobs to finish
@@ -700,7 +740,7 @@ fn checkpoint_and_restore_in_external_netns() -> TestResult {
         return TestResult::Failed(anyhow!("not deleted after checkpoint: {e}"));
     }
 
-    if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[]) {
+    if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[], &[]) {
         return TestResult::Failed(anyhow!("restore failed: {e}"));
     }
 
@@ -779,7 +819,7 @@ fn checkpoint_and_restore_with_container_specific_criu_config() -> TestResult {
         return TestResult::Failed(anyhow!("container not deleted after checkpoint: {e}"));
     }
 
-    if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[]) {
+    if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[], &[]) {
         return TestResult::Failed(anyhow!("restore failed: {e}"));
     }
 
@@ -863,7 +903,7 @@ fn checkpoint_and_restore_with_nested_bind_mounts() -> TestResult {
     std::fs::remove_dir_all(&bind1).unwrap();
     std::fs::create_dir_all(&bind1).unwrap();
 
-    if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[]) {
+    if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[], &[]) {
         return TestResult::Failed(anyhow!("restore failed: {e}"));
     }
 
@@ -933,7 +973,7 @@ fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
         id,
         image_dir,
         Some(work_dir),
-        &["--manage-cgroups-mode", "ignore"],
+        &["--manage-cgroups-mode=ignore"],
         &[],
     );
     match cp_result {
@@ -969,7 +1009,8 @@ fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
 
     // Update config to a DIFFERENT cgroup
     let new_cgroup = format!("/runtime-test/cgroup-2-{}", id);
-    let mut spec = oci_spec::runtime::Spec::load(bundle.path().join("config.json")).unwrap();
+    let mut spec =
+        oci_spec::runtime::Spec::load(bundle.path().join("bundle").join("config.json")).unwrap();
     spec.linux_mut()
         .as_mut()
         .unwrap()
@@ -982,7 +1023,8 @@ fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
         id,
         image_dir,
         Some(work_dir),
-        &["--manage-cgroups-mode", "ignore"],
+        &["--manage-cgroups-mode=ignore"],
+        &[],
     ) {
         return TestResult::Failed(anyhow!("restore failed: {e}"));
     }
@@ -1002,7 +1044,7 @@ fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
     let state_restored: oci_spec::runtime::State = serde_json::from_str(&stdout_restored).unwrap();
     let pid_restored = state_restored.pid().unwrap();
     let cgroup_data_restored =
-        std::fs::read_to_string(format!("/proc/{}/cgroup", pid_restored)).unwrap();
+        std::fs::read_to_string(format!("/proc/{pid_restored}/cgroup")).unwrap();
     let cgroup_path_suffix_restored = cgroup_data_restored
         .lines()
         .next()
@@ -1014,16 +1056,14 @@ fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
 
     if !new_host_cgroup_path.exists() {
         return TestResult::Failed(anyhow!(
-            "new cgroup path does not exist after restore: {:?}",
-            new_host_cgroup_path
+            "new cgroup path does not exist after restore: {new_host_cgroup_path:?}",
         ));
     }
 
     // Verify the old one still does NOT exist
     if host_cgroup_path.exists() {
         return TestResult::Failed(anyhow!(
-            "initial cgroup path magically reappeared after restore: {:?}",
-            host_cgroup_path
+            "initial cgroup path magically reappeared after restore: {host_cgroup_path:?}",
         ));
     }
 
@@ -1059,7 +1099,7 @@ fn checkpoint_and_restore_and_exec() -> TestResult {
             ));
         }
 
-        if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[]) {
+        if let Err(e) = restore_container(bundle.path(), id, image_dir, Some(work_dir), &[], &[]) {
             return TestResult::Failed(anyhow!("restore failed: {e}"));
         }
 
@@ -1168,12 +1208,10 @@ pub fn get_checkpoint_restore_tests() -> TestGroup {
         "checkpoint_and_restore_with_nested_bind_mounts",
         checkpoint_and_restore_with_nested_bind_mounts
     ))]);
-
     tg.add(vec![Box::new(cr_test!(
         "checkpoint_then_restore_into_a_different_cgroup",
         checkpoint_then_restore_into_a_different_cgroup
     ))]);
-
     tg.add(vec![Box::new(cr_test!(
         "checkpoint_and_restore_and_exec",
         checkpoint_and_restore_and_exec
--- a/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
+++ b/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
@@ -872,6 +872,103 @@ fn checkpoint_and_restore_with_container_specific_criu_config() -> TestResult {
     TestResult::Passed
 }
 
+// Test: checkpoint and restore with nested bind mounts
+// (runc: @test "checkpoint and restore with nested bind mounts")
+fn checkpoint_and_restore_with_nested_bind_mounts() -> TestResult {
+    let id = generate_uuid().to_string();
+    let bundle = prepare_bundle().unwrap();
+
+    let mut spec = oci_spec::runtime::Spec::default();
+    let mut process = oci_spec::runtime::Process::default();
+    process.set_args(Some(vec!["sleep".into(), "10".into()]));
+    spec.set_process(Some(process));
+
+    // Create bind mount source directories
+    let bind1 = bundle.path().join("bind1");
+    let bind2 = bundle.path().join("bind2");
+    std::fs::create_dir_all(&bind1).unwrap();
+    std::fs::create_dir_all(&bind2).unwrap();
+
+    let mnt1 = MountBuilder::default()
+        .typ("bind".to_string())
+        .source(&bind1)
+        .destination(PathBuf::from("/test"))
+        .options(vec!["rw".to_string(), "bind".to_string()])
+        .build()
+        .unwrap();
+
+    let mnt2 = MountBuilder::default()
+        .typ("bind".to_string())
+        .source(&bind2)
+        .destination(PathBuf::from("/test/for/nested"))
+        .options(vec!["rw".to_string(), "bind".to_string()])
+        .build()
+        .unwrap();
+
+    let mut mounts = spec.mounts().clone().unwrap_or_default();
+    mounts.push(mnt1);
+    mounts.push(mnt2);
+    spec.set_mounts(Some(mounts));
+
+    set_config(&bundle, &spec).unwrap();
+
+    let run_result = (|| -> anyhow::Result<()> {
+        let status = run_container(&id, &bundle)?.wait()?;
+        if !status.success() {
+            anyhow::bail!("run -d failed ({})", status);
+        }
+        wait_container_running(&id, &bundle)
+    })();
+    if let Err(e) = run_result {
+        return TestResult::Failed(anyhow!("container did not reach running state: {e}"));
+    }
+
+    let _guard = ContainerGuard {
+        bundle: &bundle,
+        id: &id,
+    };
+
+    let (image_dir, work_dir) = match make_cr_dirs(bundle.path()) {
+        Ok(d) => d,
+        Err(r) => return r,
+    };
+
+    if let Err(e) = checkpoint_container(bundle.path(), &id, &image_dir, Some(&work_dir), &[]) {
+        return TestResult::Failed(anyhow!("checkpoint failed: {e}"));
+    }
+
+    if let Err(e) = wait_for_state(
+        &id,
+        &bundle,
+        WaitTarget::Deleted,
+        Duration::from_secs(10),
+        Duration::from_millis(100),
+    ) {
+        return TestResult::Failed(anyhow!("container not deleted after checkpoint: {e}"));
+    }
+
+    // cleanup mountpoints created by runc/youki during creation
+    // the mountpoints should be recreated during restore - that is the actual thing tested here
+    std::fs::remove_dir_all(&bind1).unwrap();
+    std::fs::create_dir_all(&bind1).unwrap();
+
+    if let Err(e) = restore_container(bundle.path(), &id, &image_dir, Some(&work_dir), &[]) {
+        return TestResult::Failed(anyhow!("restore failed: {e}"));
+    }
+
+    if let Err(e) = wait_for_state(
+        &id,
+        &bundle,
+        WaitTarget::Status(LifecycleStatus::Running),
+        Duration::from_secs(10),
+        Duration::from_millis(100),
+    ) {
+        return TestResult::Failed(anyhow!("container not running after restore: {e}"));
+    }
+
+    TestResult::Passed
+}
+
 pub fn get_checkpoint_restore_tests() -> TestGroup {
     let mut tg = TestGroup::new("checkpoint_restore");
     // Run sequentially: CRIU uses global kernel resources and parallel
@@ -923,6 +1020,10 @@ pub fn get_checkpoint_restore_tests() -> TestGroup {
     tg.add(vec![Box::new(cr_test!(
         "checkpoint_and_restore_with_container_specific_criu_config",
         checkpoint_and_restore_with_container_specific_criu_config
+    ))]);
+    tg.add(vec![Box::new(cr_test!(
+        "checkpoint_and_restore_with_nested_bind_mounts",
+        checkpoint_and_restore_with_nested_bind_mounts
     ))]);
 
     tg
--- a/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
+++ b/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
@@ -19,9 +19,9 @@ use test_framework::{ConditionalTest, TestGroup, TestResult};
 use crate::utils::{
     LifecycleStatus, WaitTarget, build_checkpoint_command, checkpoint_container, criu_has_feature,
     criu_installed, delete_container, exec_container, generate_uuid, get_container_pid,
-    get_runtime_path, get_state, handle_console_socket, is_runtime_youki, kill_container, net,
-    prepare_bundle, restore_container, set_config, try_checkpoint_container,
-    wait_container_running, wait_for_state,
+    get_runtime_path, handle_console_socket, is_runtime_youki, kill_container, net, prepare_bundle,
+    restore_container, set_config, try_checkpoint_container, wait_container_running,
+    wait_for_state,
 };
 
 /// Used as check_fn for all ConditionalTests in this module:
@@ -760,7 +760,7 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
     unsafe {
         checkpoint_cmd.pre_exec(move || {
             let flags = FdFlag::from_bits_truncate(
-                fcntl(pipe_w_raw, FcntlArg::F_GETFD).expect("from_bits_truncate failed"),
+                fcntl(pipe_w_raw, FcntlArg::F_GETFD).expect("fcntl failed"),
             );
             fcntl(pipe_w_raw, FcntlArg::F_SETFD(flags & !FdFlag::FD_CLOEXEC))
                 .expect("fcntl failed");
@@ -774,14 +774,13 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
     };
 
     // Close the write end in the parent immediately so read() will get EOF if CRIU exits
-    let _ = nix::unistd::close(pipe_w.as_raw_fd());
+    drop(pipe_w);
 
     // Set read pipe to non-blocking so the timeout loop isn't blocked forever if no data arrives
     let flags = nix::fcntl::OFlag::from_bits_truncate(
-        nix::fcntl::fcntl(pipe_r.as_raw_fd(), nix::fcntl::FcntlArg::F_GETFL)
-            .expect("F_GETFL failed"),
+        fcntl(pipe_r.as_raw_fd(), nix::fcntl::FcntlArg::F_GETFL).expect("F_GETFL failed"),
     );
-    nix::fcntl::fcntl(
+    fcntl(
         pipe_r.as_raw_fd(),
         nix::fcntl::FcntlArg::F_SETFL(flags | nix::fcntl::OFlag::O_NONBLOCK),
     )
@@ -808,9 +807,6 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
             _ => std::thread::sleep(Duration::from_millis(100)),
         }
     }
-
-    // Cleanup read pipe
-    let _ = nix::unistd::close(pipe_r.as_raw_fd());
 
     if !ready {
         let _ = checkpoint_child.kill();
@@ -1000,9 +996,7 @@ fn checkpoint_and_restore_in_external_netns() -> TestResult {
             return TestResult::Failed(anyhow!("ping container failed after restore: {e}"));
         }
 
-        let (stdout, _) = get_state(id, bundle.path()).unwrap();
-        let state: crate::utils::State = serde_json::from_str(&stdout).unwrap();
-        let new_pid = state.pid.unwrap();
+        let new_pid = get_container_pid(id, bundle.path()).unwrap();
         let new_inode = std::fs::read_link(format!("/proc/{new_pid}/ns/net")).unwrap();
 
         if original_inode != new_inode {
@@ -1208,9 +1202,7 @@ fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
     let work_dir = &ctx.work_dir;
 
     // Get the container PID and use it to find the real cgroup path on the host
-    let (stdout, _) = get_state(id, bundle.path()).unwrap();
-    let state: oci_spec::runtime::State = serde_json::from_str(&stdout).unwrap();
-    let pid = state.pid().unwrap();
+    let pid = get_container_pid(id, bundle.path()).unwrap();
     let cgroup_data = std::fs::read_to_string(format!("/proc/{pid}/cgroup")).unwrap();
     // Parse the /proc/<pid>/cgroup file.
     // Format is typically `0::/path/to/cgroup` for v2, or `N:name:/path` for v1.
--- a/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
+++ b/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
@@ -4,7 +4,6 @@
 // not yet implemented in youki.  They are also skipped when CRIU is not
 // installed on the host.
 
-use std::cell::RefCell;
 use std::os::unix::fs::symlink;
 use std::os::unix::io::AsRawFd;
 use std::os::unix::process::CommandExt;
@@ -42,7 +41,7 @@ fn has_cgroupns() -> bool {
 
 struct CrTestContext {
     id: String,
-    restore_id: RefCell<Option<String>>,
+    restore_id: Option<String>,
     bundle: tempfile::TempDir,
     image_dir: PathBuf,
     work_dir: PathBuf,
@@ -65,15 +64,15 @@ impl Drop for CrTestContext {
         // Clean up primary container
         cleanup(&self.id, self.bundle.path());
         // Clean up the dynamically registered restore ID (if any)
-        if let Some(rid) = self.restore_id.borrow().as_ref() {
+        if let Some(rid) = self.restore_id.as_ref() {
             cleanup(rid, self.bundle.path());
         }
     }
 }
 
 impl CrTestContext {
-    fn register_restore_id(&self, rid: String) {
-        *self.restore_id.borrow_mut() = Some(rid);
+    fn register_restore_id(&mut self, rid: String) {
+        self.restore_id = Some(rid);
     }
 
     // TODO: Consider extracting this into test_utils.rs as run_container_with_console (see issue #3529)
@@ -247,7 +246,7 @@ fn setup_cr_test(
 
     Ok(CrTestContext {
         id,
-        restore_id: std::cell::RefCell::new(None),
+        restore_id: None,
         bundle,
         image_dir,
         work_dir,
@@ -702,7 +701,7 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         return TestResult::Skipped;
     }
 
-    let ctx = match setup_cr_test(|_, _| {}) {
+    let mut ctx = match setup_cr_test(|_, _| {}) {
         Ok(c) => c,
         Err(e) => return e,
     };
@@ -710,7 +709,9 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         return e;
     }
 
-    let id = &ctx.id;
+    let id = ctx.id.clone();
+    let restore_id = format!("{id}-restore");
+    ctx.register_restore_id(restore_id.clone());
     let bundle = &ctx.bundle;
     let image_dir = &ctx.image_dir;
     let work_dir = &ctx.work_dir;
@@ -736,7 +737,7 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
     let port_str = format!("0.0.0.0:{}", port);
     let mut checkpoint_cmd = build_checkpoint_command(
         bundle.path(),
-        id,
+        &id,
         image_dir,
         Some(work_dir),
         &[
@@ -857,9 +858,14 @@ fn checkpoint_lazy_pages_and_restore() -> TestResult {
         }
     };
 
-    let restore_id = format!("{id}-restore");
-    ctx.register_restore_id(restore_id.clone());
-
+    // We must restore the container into a different cgroup because the original
+    // container is still running. If we don't use a different cgroup, the restored
+    // container and the original container will conflict.
+    // This is particularly due to the behavior when systemd is used as the cgroup
+    // manager, as it can cause `systemd` to become confused about the container's
+    // state and potentially send SIGTERM to the process.
+    // See also runc's checkpoint tests:
+    // https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/tests/integration/checkpoint.bats#L303
     let new_cgroup = format!("/runtime-test/cgroup-2-{restore_id}");
     let mut spec =
         oci_spec::runtime::Spec::load(bundle.path().join("bundle").join("config.json")).unwrap();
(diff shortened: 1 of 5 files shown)


tommady:
Hi @nayuta723 
please help review while you have time.
thx.

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -550,13 +535,67 @@ pub fn checkpoint_container(
...
+        global_args,
+    )
+    .spawn()
+    .context("failed to spawn checkpoint")?
+    .wait_with_output()
+    .context("failed to wait for checkpoint")
+}
+
 /// Restore a checkpointed container from `image_dir` using `restore -d`.
 /// `global_args` are passed before `--root` (e.g. `&["--debug"]`).
```
nayuta723, line 592: Please add a doc string for the difference between `global_args` and `restore_args`.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -126,21 +226,34 @@ fn simple_cr(
...
+            id,
+            bundle,
             WaitTarget::Status(LifecycleStatus::Running),
             Duration::from_secs(10),
             Duration::from_millis(100),
         ) {
             return TestResult::Failed(anyhow!("not running after restore: {e}"));
         }
+
+        if let Err(e) = ping_container(bundle.path()) {
```
nayuta723, line 250: Why is the FIFO ping-pong test being performed for all tests? In runc, it’s only used for pre-dump and lazy-pages tests.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -70,43 +189,24 @@ fn make_cr_dirs(bundle_path: &Path) -> Result<(PathBuf, PathBuf), TestResult> {
 fn simple_cr(
     global_args: &[&str],
     setup: impl Fn(&tempfile::TempDir, &mut oci_spec::runtime::Spec),
+    verify: impl Fn(&str, &Path) -> anyhow::Result<()>,
```
nayuta723, line 192: The execution timing of verify is a bit unclear. Could you please add a docstring to clarify it, or rename the argument.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +302,1008 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    )
+    .unwrap();
+
+    if output1.status.success() {
+        return TestResult::Failed(anyhow!(
+            "expected checkpoint to fail with absolute --parent-path"
+        ));
+    }
+    let stderr1 = String::from_utf8_lossy(&output1.stderr);
+    if !stderr1.contains("--parent-path must be relative") {
```
nayuta723, line 468: The exact match seems like overkill. Wouldn't it be sufficient to just check if the string `--parent-path` appears in stderr, similar to how runc does it?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +302,1008 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+            "expected 'invalid --parent-path' but got stderr: {stderr2}"
+        ));
+    }
+
+    TestResult::Passed
+}
+
+// Test: checkpoint --pre-dump and restore
+// (runc: @test "checkpoint --pre-dump and restore")
+fn checkpoint_pre_dump_and_restore() -> TestResult {
```
nayuta723, line 501: We need to check if the `mem_dirty_track` feature is enabled.

https://criu.org/Memory_changes_tracking

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +302,1008 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        return TestResult::Failed(anyhow!("ping container failed after restore: {e}"));
+    }
+
+    TestResult::Passed
+}
+
+// Test: checkpoint --lazy-pages and restore
+// (runc: @test "checkpoint --lazy-pages and restore")
+fn checkpoint_lazy_pages_and_restore() -> TestResult {
+    let ctx = match setup_cr_test(|_, _| {}) {
```
nayuta723, line 599: We need to check if the uffd-noncoop feature is enabled.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +302,1008 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    let _ = kill_container(&restore_id, bundle).map(|mut c| c.wait());
+    let _ = delete_container(&restore_id, bundle).map(|mut c| c.wait());
+
+    TestResult::Passed
+}
+
+// Test: checkpoint and restore in external network namespace
+// (runc: @test "checkpoint and restore in external network namespace")
+fn checkpoint_and_restore_in_external_netns() -> TestResult {
+    let ns_name = format!("contest-{}", generate_uuid());
```
nayuta723, line 802: We need to check if the `external_net_ns` feature is enabled.

nayuta723 (commented):
There are quite a few changes, so for now I’ve reviewed it mainly from the perspectives of compatibility with runc and readability. Please take a look when you have time.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: I decided to use the FIFO ping-pong test for all basic tests because it acts as a much more robust active verification compared to a simple sleep delay. 

It guarantees that the container process isn't just marked as 'running' in state, but is actually alive, not deadlocked, and capable of handling I/O immediately after the restore completes. 

I think it's a great improvement over the original method, so I'd prefer to keep it here if that's alright!

nayuta723: Got it, that approach sounds good to me. My only concern is whether it’s the right choice to make everything pipe-based.

This might just be a gap in the current youki tests, but I think it would be good to also ensure that TTY-based flows are properly handled.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +307,1020 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    let work_dir = bundle.path().join("work-dir");
+    if let Err(e) = std::fs::create_dir_all(&parent_dir) {
+        return TestResult::Failed(anyhow!("failed to create parent dir: {e}"));
+    }
+    if let Err(e) = std::fs::create_dir_all(&image_dir) {
+        return TestResult::Failed(anyhow!("failed to create image dir: {e}"));
+    }
+    if let Err(e) = std::fs::create_dir_all(&work_dir) {
+        return TestResult::Failed(anyhow!("failed to create work dir: {e}"));
+    }
```
nayuta723: `setup_cr_test` already create these directories.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: I checked again, and I’d like to verify that checkpoint/restore works with the minimal definition from `youki spec`, so please align the implementation with `runc`.

* When running `run`, set the `-d` option and `--console-socket` to verify TTY behavior.
* Create the config based on the one generated by `youki spec`.

Please make it behave more closely to `runc`.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +307,1020 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    // Hardcoding a port (like 27277) causes "Address already in use" errors and subsequent
+    // test hangs if a previous test run failed/panicked and leaked the page server daemon.
+    let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
+    let port = listener.local_addr().unwrap().port().to_string();
+    drop(listener);
+    let status_fd_str = pipe_w.as_raw_fd().to_string();
+
+    // Spawn checkpoint command in background
+    let port_str = format!("0.0.0.0:{}", port);
+    let mut checkpoint_cmd = build_checkpoint_command(
```
nayuta723, line 732: Why is only this part not handled within `try_checkpoint_container`?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +307,1020 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    let bundle = &ctx.bundle;
+    let image_dir = &ctx.image_dir;
+    let work_dir = &ctx.work_dir;
+
+    let log_file_path = work_dir.join(custom_log_name);
+
+    if let Err(e) = try_checkpoint_container(bundle.path(), id, image_dir, Some(work_dir), &[], &[])
+    {
+        return TestResult::Failed(anyhow!("checkpoint failed: {e:?}"));
+    }
```
nayuta723:
```suggestion
    if let Err(e) = checkpoint_container(bundle.path(), id, image_dir, Some(work_dir), &[])
    {
        return TestResult::Failed(anyhow!("checkpoint failed: {e}"));
    }
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +307,1020 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+
+    let output1 =
+        try_checkpoint_container(bundle.path(), id, &parent_dir, None, &["--pre-dump"], &[])
+            .unwrap();
+
+    if !output1.status.success() {
+        let stderr = String::from_utf8_lossy(&output1.stderr);
+        return TestResult::Failed(anyhow!("pre-dump checkpoint failed: {stderr}"));
+    }
```
nayuta723:
```suggestion
    if let Err(e) = checkpoint_container(bundle.path(), id, &parent_dir, None,
    &["--pre-dump"]) {
        return TestResult::Failed(anyhow!("pre-dump checkpoint failed: {e}"));
    }
```

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -550,13 +535,68 @@ pub fn checkpoint_container(
...
+    for a in checkpoint_args {
+        command.arg(a);
+    }
+
+    command.arg(id);
+    command
+}
+
+// Execute checkpoint command and return the raw output instead of bailing on failure.
+pub fn try_checkpoint_container(
```
nayuta723, line 611: To clarify the distinction between `try_checkpoint_container` and `checkpoint_container`, please add a docstring to `try_checkpoint_container`.

nayuta723 (changes requested):
Sorry for the delay in the review. I’ve left some comments—please take a look.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: sharp eye

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: try_checkpoint_container uses .output(), which blocks until the command completes and returns the standard output/error. 

however, lazy-page checkpointing requires us to start the process in the background (.spawn()), read from the status-fd pipe to detect readiness, and then wait for the process to finish. .output() cannot do this.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: addressed

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: addressed

tests/contest/contest/src/utils/test_utils.rs
tommady: addressed

tommady:
https://github.com/youki-dev/youki/pull/3493#discussion_r3143708929

this is a very hard one... since the differences between how runc and youki do the integration tests...
I'll try

nayuta723:
I may be missing some of the difficulty here, but what parts do you think could be tricky?
Depending on the complexity, we can also consider splitting [this](https://github.com/youki-dev/youki/pull/3493#discussion_r3143708929) out into a separate PR.

tommady:
> I may be missing some of the difficulty here, but what parts do you think could be tricky? Depending on the complexity, we can also consider splitting [this](https://github.com/youki-dev/youki/pull/3493#discussion_r3143708929) out into a separate PR.

I will include it in this PR, but it might take a little more time to get right, since the checkpoint/restore integration test is trying to make it more like runc's.
Thanks

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -33,24 +38,184 @@ fn has_cgroupns() -> bool {
...
+            let _ = child.wait();
+        }
+        std::thread::sleep(Duration::from_millis(100));
+        if let Ok(mut child) = delete_container(&id, &bundle) {
+            let _ = child.wait();
+        }
+        return Err(TestResult::Failed(anyhow!(
+            "container did not reach running state: {e}"
+        )));
+    }
```
nayuta723: Isn’t it outside the responsibility to run a container inside the function that creates `CrTestContext`?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -33,24 +38,184 @@ fn has_cgroupns() -> bool {
...
+        .current_dir(&actual_bundle_path)
+        .output()
+        .map_err(|e| TestResult::Failed(anyhow!("failed to execute youki spec: {e}")))?;
+
+    if !output.status.success() {
+        return Err(TestResult::Failed(anyhow!(
+            "youki spec failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        )));
+    }
```
nayuta723, line 179: Please extract the spec generation logic into a utility module.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723:
```suggestion
/// Full checkpoint+restore test: prepares a bundle, calls `setup_cr_test` to allow the
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -67,56 +232,47 @@ fn make_cr_dirs(bundle_path: &Path) -> Result<(PathBuf, PathBuf), TestResult> {
 /// The first cycle verifies a normal container can be checkpointed and restored;
 /// the second verifies the restored container can itself be checkpointed and
 /// restored again.
+///
```
nayuta723, line 261: Please also clarify what exactly is being verified in `ping_container`.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -33,24 +38,184 @@ fn has_cgroupns() -> bool {
...
+    }
+}
+
+fn setup_cr_test(
+    mut setup: impl FnMut(&tempfile::TempDir, &mut oci_spec::runtime::Spec),
+) -> Result<CrTestContext, TestResult> {
+    let id = generate_uuid().to_string();
+    let bundle = prepare_bundle().unwrap();
+
+    let fifo_path = bundle.path().join("fifo");
```
nayuta723: Please add a comment clearly explaining why the FIFO path is being prepared.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -33,24 +38,184 @@ fn has_cgroupns() -> bool {
...
+    let bind_mount = MountBuilder::default()
+        .source(&fifo_path)
+        .destination(PathBuf::from("/fifo"))
+        .options(vec!["bind".to_string()])
+        .build()
+        .unwrap();
+
+    let mut mounts = spec.mounts().clone().unwrap_or_default();
+    mounts.push(bind_mount);
+    spec.set_mounts(Some(mounts));
```
nayuta723: Since the FIFO configuration is an addition on top of the default settings, please add a docstring to the function explaining the details. If possible, also include the rationale for choosing FIFOs instead of the pipe-based approach used by runc.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -33,24 +38,184 @@ fn has_cgroupns() -> bool {
...
+    Ok(CrTestContext {
+        id,
+        bundle,
+        image_dir,
+        work_dir,
+    })
+}
+
 /// Create image-dir and work-dir inside the bundle temp directory.
 fn make_cr_dirs(bundle_path: &Path) -> Result<(PathBuf, PathBuf), TestResult> {
```
nayuta723: Since `make_cr_dir` is only called in one place, it would be better to integrate it into `setup_cr_test`.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        Ok(c) => c,
+        Err(e) => return e,
+    };
+
+    let id = &ctx.id;
+    let bundle = &ctx.bundle;
+    let image_dir = &ctx.image_dir;
+    let work_dir = &ctx.work_dir;
+
+    let absolute_parent = bundle.path().join("parent-dir");
```
nayuta723, line 541: Please add a comment clarifying that failure in the case of an absolute path is the expected behavior.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: Please add this to the comment.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    // Cleanup read pipe
+    let _ = nix::unistd::close(pipe_r.as_raw_fd());
+
+    if !ready {
+        let _ = checkpoint_child.kill();
+        let _ = checkpoint_child.wait();
+        let mut stderr_msg = String::new();
+        if let Some(mut stderr) = checkpoint_child.stderr.take() {
+            let _ = std::io::Read::read_to_string(&mut stderr, &mut stderr_msg);
+        }
```
nayuta723:
```suggestion
        let stderr_msg = checkpoint_child
            .stderr
            .take()
            .map(|mut s| {
                let mut msg = String::new();
                let _ = std::io::Read::read_to_string(&mut s, &mut msg);
                msg
            })
            .unwrap_or_default();
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+            }
+            _ => std::thread::sleep(Duration::from_millis(100)),
         }
-    })
+    }
+
+    // Cleanup read pipe
+    let _ = nix::unistd::close(pipe_r.as_raw_fd());
+
+    if !ready {
```
nayuta723, line 805: Since `checkpoint_cmd.stderr(std::process::Stdio::null());` is set, wouldn’t this always be null here?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        &restore_id,
+        image_dir,
+        Some(image_dir),
+        &["--lazy-pages", "--manage-cgroups-mode=ignore"],
+        &[],
+    );
+
+    // Terminate criu daemon if it's still running, it might stay alive if restore failed
+    let mut criu_daemon = criu_daemon_child;
+    let _ = criu_daemon.kill();
```
nayuta723: Isn’t it a problem to always kill it?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+
+    if let Err(e) = wait_for_state(
+        &restore_id,
+        bundle,
+        WaitTarget::Status(LifecycleStatus::Running),
+        Duration::from_secs(10),
+        Duration::from_millis(100),
+    ) {
+        let _ = kill_container(&restore_id, bundle).map(|mut c| c.wait());
+        let _ = delete_container(&restore_id, bundle).map(|mut c| c.wait());
```
nayuta723: Since `CrTestContext` should handle the cleanup, isn’t this unnecessary?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        Ok(c) => c,
+        Err(e) => return e,
+    };
+    let id = &ctx.id;
+    let bundle = &ctx.bundle;
+    let image_dir = &ctx.image_dir;
+    let work_dir = &ctx.work_dir;
+
+    let (stdout, _) = get_state(id, bundle.path()).unwrap();
+    let state: crate::utils::State = serde_json::from_str(&stdout).unwrap();
```
nayuta723:
```suggestion
    let state: State = serde_json::from_str(&stdout).unwrap();
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +358,1016 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    let pid = state.pid.unwrap();
```
nayuta723: Please implement the `get_container_pid` function.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -33,24 +38,184 @@ fn has_cgroupns() -> bool {
...
+            .arg(&actual_bundle_path)
+            .arg("--console-socket")
+            .arg(&console_socket)
+            .arg(&id)
+            .current_dir(bundle.path())
+            .spawn()?;
+
+        let (stream, _) = listener.accept()?;
+
+        crate::utils::handle_console_socket(stream);
```
nayuta723:
```suggestion
        handle_console_socket(stream);
```

tests/contest/contest/src/utils/test_utils.rs
```diff
@@ -569,17 +638,57 @@ pub fn restore_container(
...
+    }
+
+    let config_path = bundle_path.join("bundle").join("config.json");
+    let config_path = if config_path.exists() {
+        config_path
+    } else {
+        bundle_path.join("config.json")
+    };
+
+    let is_terminal = if config_path.exists() {
```
nayuta723, line 665: Unlike runc, youki seems to have `terminal: false` by default. According to the runtime-spec, the default value is `terminal: false`, so please explicitly add a setting in `setup_cr_test` to set `terminal: true`.
ref: https://github.com/youki-dev/youki/issues/3522

nayuta723 (changes requested):

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -40,29 +42,94 @@ fn has_cgroupns() -> bool {
...
-            let _ = child.wait();
+    }
+}
+
+impl CrTestContext {
+    fn register_restore_id(&self, rid: String) {
+        *self.restore_id.borrow_mut() = Some(rid);
+    }
+
+    fn start(&self) -> Result<(), TestResult> {
```
nayuta723, line 79: Please move it to `test_utils` file as `start_container`.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: @tommady 
I don’t understand why `kill` is called except when `restore_result` results in an error.
I’d especially like to understand in detail what kinds of cases can lead to a hang.

nayuta723 (commented):
I left some nit comments.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: Regarding extracting the start() method into test_utils.rs as start_container: 
 
I initially looked into doing this, but found that there's a fundamental difference between what the checkpoint/restore suite needs and what test_utils::start_container does.

The existing test_utils::start_container executes the OCI start command, which is used for tests that require a discrete create -> start lifecycle. 

However, the checkpoint/restore suite specifically relies on run -d (along with --console-socket) because we need to intercept the terminal FIFO connection via handle_console_socket concurrently as the container spawns.

If we merge this logic into test_utils::start_container, it breaks the standard create -> start test cases. 

Keeping .start() encapsulated within CrTestContext ensures that this specialized -d spawn and socket-handling logic stays tightly coupled with the context that actually needs it, avoiding clutter and confusion in the global test_utils.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: The CRIU lazy-pages daemon is designed to stay alive indefinitely waiting for TCP connections to serve memory pages. 

When runc restore succeeds, it sends a specific PS_IOV_FORCE_CLOSE command to the CRIU daemon, which tells the daemon to naturally exit. 

However, if the runc restore process panics, times out, or fails for any reason before sending that close signal, the daemon will remain alive in the background indefinitely. 

Because Rust's test runner does not automatically kill orphaned Command::spawn() children when a test finishes, this leaked daemon will continue holding the TCP port open (e.g. 127.0.0.1:{port}). 

If the test suite is run again, the port binding will fail with Address already in use. 

I actually ran into this exact failure during implementation.
It permanently broke my subsequent test runs until I manually tracked down the orphaned process and killed it via kill -9. 

My initial workaround was to make the daemon listen on a random port every time, but ultimately, ensuring the daemon is explicitly killed after the test finishes is the safest best practice to prevent state leakage.

By unconditionally calling .kill() and .wait(), we guarantee the daemon process is always reaped:
- If the daemon already exited naturally (because the restore succeeded), .kill() safely returns a negligible ESRCH error which we just ignore. 
- If the daemon hung, .kill() forcefully terminates it, guaranteeing no leaked background processes.

Does this context help clarify why this unconditional cleanup is necessary?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: How about using `run_container_with_console`?

I’m thinking that we could modify the implementation here:
[https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/utils/test_utils.rs#L480-L498](https://github.com/youki-dev/youki/blob/main/tests/contest/contest/src/utils/test_utils.rs#L480-L498)

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: > However, if the runc restore process panics, times out, or fails for any reason before sending that close signal, the daemon will remain alive in the background indefinitely.

So hanging cases are considered abnormal cases, right? I think ignoring such abnormal cases while there are subsequent tests is not appropriate from a testing perspective.

> Does this context help clarify why this unconditional cleanup is necessary?

I understand, but based on the above, I’d like to propose treating hanging cases as abnormal and terminating them. What do you think about the following implementation approach?

https://github.com/youki-dev/youki/pull/3493/changes#r3178241365

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +389,1052 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    // if the daemon has already exited.
+    let mut criu_daemon = criu_daemon_child;
+    let _ = criu_daemon.kill();
+    let _ = criu_daemon.wait();
+
+    if let Err(e) = restore_result {
+        let _ = checkpoint_child.kill();
+        let _ = checkpoint_child.wait();
+        return TestResult::Failed(anyhow!("restore --lazy-pages failed: {e}"));
+    }
```
nayuta723, line 889:
```suggestion
    let mut criu_daemon = criu_daemon_child;
    let cleanup_err = match criu_daemon.try_wait() {                                                        
        Ok(Some(_)) => None,
        Ok(None) => {                                                                                       
            // daemon should have exited naturally after restore, but it is still running                   
            let _ = criu_daemon.kill();                                                                     
            let _ = criu_daemon.wait();
            Some(std::io::Error::new(
                std::io::ErrorKind::Other,
                "criu daemon did not exit naturally",
            ))
        }                                                                                                        
        Err(e) => {                                                                                         
            // try_wait failed, but still try to kill
            let _ = criu_daemon.kill();
            let _ = criu_daemon.wait();
            Some(e)
        }
    };
    
    if let Err(e) = restore_result {
        let _ = checkpoint_child.kill();
        let _ = checkpoint_child.wait();
        return TestResult::Failed(anyhow!("restore --lazy-pages failed: {e}"));
    }
  
    if let Some(e) = cleanup_err {
        return TestResult::Failed(anyhow!("failed to clean up criu daemon: {e}"));
    }
```

tommady: I completely agree that a hanging daemon is abnormal and indicates a failure. 

However, from my debugging, the daemon only hangs when runc restore itself fails, panics, or times out, because it never reaches the point where it sends the PS_IOV_FORCE_CLOSE signal to the daemon. 

Because we already explicitly check and fail the test if restore_result returns an error (if let Err(e) = restore_result { return TestResult::Failed(...) }), treating the daemon hang as a separate failure condition is redundant. 

In fact, if restore fails, the daemon will hang. 
If we intercept the hung daemon and fail early with "criu daemon did not exit naturally", we risk shadowing the actual root cause of the failure (the runc restore error output), which makes debugging much harder.
 
My goal with the unconditional .kill() isn't to ignore a failure, but simply to act as a robust teardown mechanism (like a Drop implementation or defer block) that guarantees the test environment is clean for the next test, allowing the actual restore_result to correctly report why it failed. 

Does that reasoning make sense? 

I'd prefer to keep it as an unconditional teardown to ensure we always get the true runc restore error logs when a failure occurs.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: I completely agree with the principle that a hanging daemon is an abnormal state we shouldn't just ignore. 

However, there is a technical risk with the proposed try_wait() implementation: it introduces a race condition that will likely cause flaky tests in CI. 

When runc restore successfully finishes and sends the close signal, it takes a short amount of time for the CRIU daemon process to actually receive the signal and exit at the OS level. 

Because try_wait() is non-blocking and executes immediately after restore_container returns, it will frequently evaluate to Ok(None) simply because the daemon is still in the middle of shutting down.

To implement your suggestion safely without test flakiness, we would need to add a polling loop with a timeout (e.g., waiting for up to a few seconds for the daemon to naturally exit) before falling back to treating it as an error. 

Given that our focus here is testing youki's integration rather than testing CRIU's internal TCP teardown timings, adding complex polling logic feels like it might overcomplicate the test and introduce stability risks.

Using an unconditional .kill() acts as a simple, race-free teardown mechanism. 
It guarantees the environment is clean for the next test without requiring us to manage polling timeouts.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: You're completely right

we absolutely could extract this logic into test_utils.rs as a new helper called run_container_with_console (or run_container_detached). 

The main reason I prefer keeping it as CrTestContext::start inside checkpoint_restore/mod.rs for now is encapsulation. 

The logic for binding a unix socket, launching the container with -d, accepting the connection, and calling handle_console_socket concurrently is highly specific to how the Checkpoint/Restore tests test the OCI terminal: true requirement. 

No other test suite in contest currently uses this pattern—they all rely on the standard create -> start -> kill sequence or basic foreground run commands. 

If we move run_container_with_console into the global test_utils.rs, we are exposing very specialized test logic to the rest of the testing framework where it isn't needed. 

Keeping .start() directly attached to CrTestContext ensures that this specialized -d spawn and socket-handling logic stays tightly coupled with the context that actually needs it.

That being said, if we do decide that it's valuable to generalize run_container_with_console globally for future test suites, I would strongly prefer to tackle that refactoring in a separate, dedicated PR. 

That would allow us to properly review the global test utility changes without blocking the core checkpoint/restore features being delivered here. What do you think?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: > Because we already explicitly check and fail the test if restore_result returns an error (if let Err(e) = restore_result { return TestResult::Failed(...) }), treating the daemon hang as a separate failure condition is redundant.

Do you see any issues with the below implementation?
```suggestion
    let mut criu_daemon = criu_daemon_child;

    if let Err(e) = restore_result {
        let _ = criu_daemon.kill();
        let _ = criu_daemon.wait();
        let _ = checkpoint_child.kill();
        let _ = checkpoint_child.wait();
        return TestResult::Failed(anyhow!("restore --lazy-pages failed: {e}"));
    }
    
    let _ = criu_daemon.wait();
```


> In fact, if restore fails, the daemon will hang.
If we intercept the hung daemon and fail early with "criu daemon did not exit naturally", we risk shadowing the actual root cause of the failure (the runc restore error output), which makes debugging much harder.

As for the changes above, since the structure modifies `cleanup_error` after handling errors from `restore_result`, I believe the concern mentioned earlier should not apply.

> My goal with the unconditional .kill() isn't to ignore a failure, but simply to act as a robust teardown mechanism (like a Drop implementation or defer block) that guarantees the test environment is clean for the next test, allowing the actual restore_result to correctly report why it failed.

I agree that eventually reaching the unconditional `.kill()` makes sense, so please let me know if there are any issues with the two approaches above that I may be missing.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
nayuta723: I got it. I agree this should be handled in a separate PR. Could you create an issue and link it in the code comments?

nayuta723 (approved):
Thanks! LGTM!

tommady:
Hi @saku3, @utam0k 
please help check and merge
thank you

utam0k:
@saku3 I'll leave it to you.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        });
+    }
+
+    let mut checkpoint_child = match checkpoint_cmd.spawn() {
+        Ok(c) => c,
+        Err(e) => return TestResult::Failed(anyhow!("failed to spawn checkpoint: {e}")),
+    };
+
+    // Close the write end in the parent immediately so read() will get EOF if CRIU exits
+    let _ = nix::unistd::close(pipe_w.as_raw_fd());
```
saku3: Since pipe returns `OwnedFd`s, I think we should close them by dropping the OwnedFd instead of calling `close()` on the raw fds. Calling `close()` on the raw fd while the OwnedFd is still alive may lead to a double close when it is dropped.

```suggestion
    drop(pipe_w);
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+            Err(nix::errno::Errno::EAGAIN) => {
+                std::thread::sleep(Duration::from_millis(100));
+            }
+            _ => std::thread::sleep(Duration::from_millis(100)),
         }
-    })
+    }
+
+    // Cleanup read pipe
+    let _ = nix::unistd::close(pipe_r.as_raw_fd());
```
saku3: As with pipe_w, explicitly closing this may lead to a double close.
I think this line can be removed.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+            Duration::from_millis(100),
+        ) {
+            return TestResult::Failed(anyhow!("not running after restore: {e}"));
+        }
+
+        if let Err(e) = ping_container(bundle.path()) {
+            return TestResult::Failed(anyhow!("ping container failed after restore: {e}"));
+        }
+
+        let (stdout, _) = get_state(id, bundle.path()).unwrap();
```
saku3: Could we use `get_container_pid` to get this PID?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        return e;
+    }
+
+    let id = &ctx.id;
+    let bundle = &ctx.bundle;
+    let image_dir = &ctx.image_dir;
+    let work_dir = &ctx.work_dir;
+
+    // Get the container PID and use it to find the real cgroup path on the host
+    let (stdout, _) = get_state(id, bundle.path()).unwrap();
```
saku3: Could we use `get_container_pid` to get this PID?

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    checkpoint_cmd.stdout(std::process::Stdio::null());
+    checkpoint_cmd.stderr(std::process::Stdio::piped());
+
+    let pipe_w_raw = pipe_w.as_raw_fd();
+
+    // Ensure the child process inherits the write end of the pipe
+    unsafe {
+        checkpoint_cmd.pre_exec(move || {
+            let flags = FdFlag::from_bits_truncate(
+                fcntl(pipe_w_raw, FcntlArg::F_GETFD).expect("from_bits_truncate failed"),
```
saku3:
```suggestion
                fcntl(pipe_w_raw, FcntlArg::F_GETFD).expect("fcntl failed"),
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    let _dev = match net::DummyDevice::create(dev_name.clone()) {
+        Ok(d) => d,
+        Err(e) => return TestResult::Failed(anyhow!("Failed to create dummy dev: {e}")),
+    };
+
+    let mtu = "1789";
+    let mac = "00:11:22:33:44:55";
+    let ip = "169.254.169.77/32";
+
+    let out = match Command::new("ip")
```
saku3, line 428: According to the original runc test, this is a test for `linux.netDevices`.

https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/tests/integration/checkpoint.bats#L154

I think we should let the runtime move the device via `linux.netDevices`, rather than moving it manually with `ip link set <dev_name> netns <ns_name>` in the test setup.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    }
+
+    TestResult::Passed
+}
+
+// Test: checkpoint then restore into a different cgroup (via --manage-cgroups-mode ignore)
+// (runc: @test "checkpoint then restore into a different cgroup (via --manage-cgroups-mode ignore)")
+fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
+    let mut initial_cgroup_val = String::new();
+    let ctx = match setup_cr_test(|_, spec| {
```
saku3, line 1178: The original runc test sets linux.resources.pids.limit.

https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/tests/integration/checkpoint.bats#L453

https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/tests/integration/helpers.bash#L415

It looks like the original runc test does not check whether `pids.limit` is set after restore, but since this test explicitly sets `linux.resources.pids.limit`, I personally think it would be better to verify it as well.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        return TestResult::Failed(anyhow!("ping container failed after restore: {e}"));
+    }
+
+    TestResult::Passed
+}
+
+// Test: checkpoint then restore into a different cgroup (via --manage-cgroups-mode ignore)
+// (runc: @test "checkpoint then restore into a different cgroup (via --manage-cgroups-mode ignore)")
+fn checkpoint_then_restore_into_a_different_cgroup() -> TestResult {
+    let mut initial_cgroup_val = String::new();
```
saku3: initial_cgroup_val does not seem to be used, so I think we can remove it.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -4,18 +4,25 @@
 // not yet implemented in youki.  They are also skipped when CRIU is not
 // installed on the host.
 
+use std::cell::RefCell;
```
saku3: Is `RefCell` needed here? It looks like `restore_id` is only updated once in the lazy-pages test. I think this could be a plain `Option<String>` if `register_restore_id` takes `&mut self` and the lazy-pages test binds the context as `mut ctx`.


Please also refer to the comment below.
https://github.com/youki-dev/youki/pull/3493/changes#r3206544289

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+    }
+
+    let ctx = match setup_cr_test(|_, _| {}) {
+        Ok(c) => c,
+        Err(e) => return e,
+    };
+    if let Err(e) = ctx.start() {
+        return e;
+    }
+
```
saku3, line 705: If we move the restore_id registration around line 860 before these immutable borrows, we no longer need RefCell.

```
let restore_id = format!("{}-restore", ctx.id);

ctx.register_restore_id(restore_id.clone());
```

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
```diff
@@ -185,18 +390,1053 @@ fn checkpoint_and_restore_cgroupns() -> TestResult {
...
+        Err(e) => {
+            let _ = checkpoint_child.kill();
+            return TestResult::Failed(anyhow!("failed to spawn criu lazy-pages: {e}"));
+        }
+    };
+
+    let restore_id = format!("{id}-restore");
+    ctx.register_restore_id(restore_id.clone());
+
+    let new_cgroup = format!("/runtime-test/cgroup-2-{restore_id}");
```
saku3, line 859: Please add a comment explaining why a new cgroup name is set.

https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/tests/integration/checkpoint.bats#L303

saku3 (commented):
Thank you for the PR, and for working on this!

I left a few comments. Please take a look when you have time.

tests/contest/contest/src/tests/checkpoint_restore/mod.rs
tommady: sharp eye !
I forgot to remove it after I changed the implementation to read from the process ID

tommady:
rrr... today is not my day to run the testing...

@saku3 could you help review again and help me trigger the k8s-tests when `https://kind.sigs.k8s.io` domain is back?

thanks 🙇

```bash
#9 [kind-fetch 1/1] RUN curl -sSLf https://kind.sigs.k8s.io/dl/v0.30.0/kind-linux-amd64 > /root/kind && chmod +x /root/kind
#9 0.180 curl: (6) Could not resolve host: kind.sigs.k8s.io
#9 ERROR: process "/bin/sh -c curl -sSLf [https://kind.sigs.k8s.io/dl/${KIND_VERSION}/kind-linux-${TARGETARCH}](https://kind.sigs.k8s.io/dl/$%7BKIND_VERSION%7D/kind-linux-$%7BTARGETARCH%7D) > /root/kind && chmod +x /root/kind" did not complete successfully: exit code: 6
------
 > [kind-fetch 1/1] RUN curl -sSLf https://kind.sigs.k8s.io/dl/v0.30.0/kind-linux-amd64 > /root/kind && chmod +x /root/kind:
0.180 curl: (6) Could not resolve host: kind.sigs.k8s.io
------
ERROR: failed to build: failed to solve: process "/bin/sh -c curl -sSLf [https://kind.sigs.k8s.io/dl/${KIND_VERSION}/kind-linux-${TARGETARCH}](https://kind.sigs.k8s.io/dl/$%7BKIND_VERSION%7D/kind-linux-$%7BTARGETARCH%7D) > /root/kind && chmod +x /root/kind" did not complete successfully: exit code: 6
error: Recipe `bin-kind` failed on line 110 with exit code 1
```

saku3 (approved):
Thanks!

PR #3495, opened 2026-04-13
feat(checkpoint): register external namespaces with CRIU

nayuta723:
## Description
When a container uses externally-created namespaces (e.g., a network namespace set up by the container runtime like Podman), CRIU must be told to treat them as external rather than trying to save and restore them itself. Without this, checkpoint fails with errors such as "Unknown peer net namespace" for network namespaces, and PID reassignment issues for PID namespaces.

This PR adds `handle_checkpointing_external_namespaces` in `container_criu.rs`, which reads the namespace path from the OCI spec and — if the path is set — opens the namespace file, stats its inode, and registers it with CRIU via `--external <TYPE>[<inode>]:<key>`. The function is called for both `LinuxNamespaceType::Network` and `LinuxNamespaceType::Pid` during checkpoint. This mirrors runc's `handleCheckpointingExternalNamespaces`.

The `rust-criu` dependency is switched from the crates.io release to the git `main` branch to pick up the `criu_ns_to_key` helper needed to generate the correct key strings (`extRootNetNS`, `extRootPidNS`).

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing
- [x] Added new integration tests (\`checkpoint_with_external_namespaces\` in the contest framework)
- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues
Fixes #

## Additional Context
Follows runc's `handleCheckpointingExternalNamespaces` and crun's `libcrun_criu_add_external_namespaces`.

### DIFF
--- a/crates/libcontainer/src/container/container_criu.rs
+++ b/crates/libcontainer/src/container/container_criu.rs
@@ -3,6 +3,13 @@
 //! This module provides shared functionality between container checkpoint and restore,
 //! following the patterns established by runc's CRIU integration.
 
+use std::fs::File;
+use std::os::unix::io::AsRawFd;
+
+use nix::sys::stat::fstat;
+use oci_spec::runtime::{LinuxNamespaceType, Spec};
+use rust_criu::{Criu, criu_ns_to_key};
+
 use crate::error::LibcontainerError;
 
 /// Minimum CRIU version required for checkpoint/restore functionality.
@@ -22,7 +29,7 @@ fn compare_criu_version(version: u32, min_version: u32) -> Result<(), Libcontain
 
 /// Check if CRIU version is greater than or equal to min_version.
 pub fn check_criu_version(min_version: u32) -> Result<(), LibcontainerError> {
-    let mut criu = rust_criu::Criu::new()
+    let mut criu = Criu::new()
         .map_err(|e| LibcontainerError::Other(format!("failed to create CRIU instance: {}", e)))?;
 
     let version = criu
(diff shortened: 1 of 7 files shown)


nayuta723:
This part depends on the rust-criu side releasing an update, but I’ve been able to open a PR for the rest.

saku3:
@donkomura 
If you don't mind, I would appreciate your help reviewing this PR.

donkomura:
Sure! I'm taking a look.

crates/libcontainer/src/container/container_criu.rs
```diff
@@ -32,6 +39,75 @@ pub fn check_criu_version(min_version: u32) -> Result<(), LibcontainerError> {
...
+        LinuxNamespaceType::Mount => "mnt",
+        LinuxNamespaceType::Ipc => "ipc",
+        LinuxNamespaceType::Uts => "uts",
+        LinuxNamespaceType::User => "user",
+        LinuxNamespaceType::Cgroup => "cgroup",
+        LinuxNamespaceType::Time => "time",
+    }
+}
+
+fn get_namespace_path(spec: &Spec, ns_type: LinuxNamespaceType) -> Option<String> {
```
donkomura, line 55: note. In my understanding, we don't need to check the criu version which supports ext ns because youki require criu >= 3.15.0 for minimum supported version.

https://github.com/opencontainers/runc/blob/89f46d647095d5d596fb5aa573a6d6588630a135/libcontainer/criu_linux.go#L174-L188

donkomura (commented):
Looks good to me.

crates/libcontainer/src/container/container_criu.rs
nayuta723: Thanks!

This may not affect typical youki usage environments, but adding it as a safeguard sounds like a good idea. I’ll take care of it.

nayuta723: I checked again and realized I was looking at the 3.15 check, so this doesn’t seem necessary.
https://github.com/youki-dev/youki/blob/main/crates/libcontainer/src/container/container_checkpoint.rs#L38-L39
Thanks for the comment and review!

donkomura: Yea, that's why I recognize that we do not have to check the version here.

saku3:
Could you please rebase this PR?

nayuta723:
I have rebased the branch.

It looks like the CI failed due to a flaky test, so I would appreciate it if you could rerun the CI and take another look afterward.

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
```diff
@@ -137,6 +204,63 @@ impl ContainerLifecycle {
         checkpoint::checkpoint_link_remap()
     }
 
+    /// Checkpoint a container that was started with an external network namespace
+    /// and an external PID namespace, then verify CRIU recorded both as external.
+    ///
+    /// A named netns is created with `ip netns add` and `/proc/self/ns/pid` is
+    /// used as the external PID namespace (the test runner's own PID namespace).
+    pub fn checkpoint_with_external_namespaces(&self) -> TestResult {
```
saku3: `self` does not seem to be used in this method.

I understand that this test creates a separate ContainerLifecycle because it needs a custom spec with external namespaces. In that case, would it make sense to make this an associated function, or add a short comment explaining why a separate lifecycle is created here?

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -432,3 +432,84 @@ pub fn checkpoint_link_remap() -> TestResult {
     cleanup();
     result
 }
+
+/// Check that the network namespace was treated as external by CRIU.
+/// Fails if netns-*.img is absent or lacks ext_key="extRootNetNS".
+/// CRIU img files embed protobuf strings as raw UTF-8, so a byte search suffices.
+pub fn check_external_netns(checkpoint_dir: &Path) -> Result<(), TestResult> {
```
saku3: `check_external_netns` and `check_external_pidns` look almost identical except for the namespace type. Could we factor them into a common helper?

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -432,3 +432,84 @@ pub fn checkpoint_link_remap() -> TestResult {
...
+    let img = netns_img.ok_or_else(|| {
+        TestResult::Failed(anyhow::anyhow!(
+            "netns-*.img not found in {:?}: network namespace image is missing",
+            checkpoint_dir,
+        ))
+    })?;
+
+    let bytes = std::fs::read(&img)
+        .map_err(|e| TestResult::Failed(anyhow::anyhow!("failed to read {:?}: {}", img, e)))?;
+    if !bytes.windows(12).any(|w| w == b"extRootNetNS") {
```
saku3: I think it would be clearer not to hardcode the length.

```suggestion
    let ext_key = b"extRootNetNS";
    if !bytes .windows(ext_key.len()) .any(|w| w == ext_key) {
```

saku3 (commented):
I left a few comments. Could you please take a look?

tests/contest/contest/src/tests/lifecycle/container_lifecycle.rs
nayuta723: Thank you!

I removed `self` and updated the implementation. I also created an issue for the future work and linked it in the comment.
#3641

saku3 (approved):
Thanks!

PR #3504, opened 2026-04-17
fix(channel): split intermediate and init readiness channels

uran0sH:
## Description

Under high concurrency, the init process can send InitReady before the
intermediate process sends IntermediateReady. The main process used to wait
for both messages on the same receiver, so wait_for_intermediate_ready()
could receive InitReady first and fail with an unexpected message error.

Add separate main-facing receivers for intermediate-owned and init-owned
readiness messages. IntermediateReady is read from the intermediate receiver,
while InitReady is forwarded through the init receiver and remains available
until wait_for_init_ready() consumes it.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite
- [x] Tested manually (please provide steps): 1. clone [boxlite](https://github.com/uran0sH/boxlite/tree/fix/test-rustflags) 2. apply this patch and change the youki's version in the boxlite 3. run `make test`.

## Related Issues

ContainerBuilder::build() fails intermittently with:
  received unexpected message: InitReady, expected: IntermediateReady(0)
  The error comes from libcontainer::process::channel::MainReceiver::wait_for_intermediate_ready().

Normal message flow
During container creation, libcontainer spawns two child processes:
  1. Intermediate process — sets up namespaces, cgroups, etc.
  2. Init process — the actual container payload.
  They communicate with the parent via a SOCK_SEQPACKET socket:
  1. Intermediate sends IntermediateReady(pid).
  2. Parent receives it in wait_for_intermediate_ready().
  3. Init sends InitReady.
  4. Parent receives it in wait_for_init_ready().
  
Root cause: parallel scheduling race
  The init process is forked by the intermediate process using clone3(CLONE_PARENT). The two processes run in parallel. Under CPU pressure, the scheduler may execute the init process faster than the intermediate process. Consequently, InitReady can arrive at the parent before IntermediateReady.
  wait_for_intermediate_ready() only calls recv() once. When it sees InitReady, it treats it as an unexpected message and aborts the build

### DIFF
--- a/crates/libcontainer/src/process/channel.rs
+++ b/crates/libcontainer/src/process/channel.rs
@@ -43,7 +43,13 @@ pub enum ChannelError {
 
 pub fn main_channel() -> Result<(MainSender, MainReceiver), ChannelError> {
     let (sender, receiver) = channel::<Message>()?;
-    Ok((MainSender { sender }, MainReceiver { receiver }))
+    Ok((
+        MainSender { sender },
+        MainReceiver {
+            receiver,
+            init_ready_cached: false,
+        },
+    ))
 }
 
 pub struct MainSender {
@@ -112,28 +118,40 @@ impl MainSender {
 
 pub struct MainReceiver {
     receiver: Receiver<Message>,
+    init_ready_cached: bool,
 }
 
 impl MainReceiver {
     /// Waits for associated intermediate process to send ready message
     /// and return the pid of init process which is forked by intermediate process
     pub fn wait_for_intermediate_ready(&mut self) -> Result<Pid, ChannelError> {
-        let msg = self
-            .receiver
-            .recv()
-            .map_err(|err| ChannelError::ReceiveError {
-                msg: "waiting for intermediate process".to_string(),
-                source: err,
-            })?;
+        loop {
+            let msg = self
+                .receiver
+                .recv()
+                .map_err(|err| ChannelError::ReceiveError {
+                    msg: "waiting for intermediate process".to_string(),
+                    source: err,
+                })?;
 
-        match msg {
-            Message::IntermediateReady(pid) => Ok(Pid::from_raw(pid)),
-            Message::ExecFailed(err) => Err(ChannelError::ExecError(err)),
-            Message::OtherError(err) => Err(ChannelError::OtherError(err)),
-            msg => Err(ChannelError::UnexpectedMessage {
-                expected: Message::IntermediateReady(0),
-                received: msg,
-            }),
+            match msg {
+                Message::IntermediateReady(pid) => return Ok(Pid::from_raw(pid)),
+                Message::ExecFailed(err) => return Err(ChannelError::ExecError(err)),
+                Message::OtherError(err) => return Err(ChannelError::OtherError(err)),
+                Message::InitReady => {
+                    tracing::debug!(
+                        "received InitReady before IntermediateReady, caching for later"
+                    );
+                    self.init_ready_cached = true;
+                    continue;
+                }
+                msg => {
+                    return Err(ChannelError::UnexpectedMessage {
+                        expected: Message::IntermediateReady(0),
+                        received: msg,
+                    });
+                }
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


saku3:
Thank you for the fix.

Would it be possible to share the actual reproduction steps?

I understand that the intermediate process and the init process run concurrently, but once the init process has been cloned, the intermediate process seems to have very little left to do before sending `IntermediateReady`, whereas the init process still appears to have a fair amount of work remaining, such as namespace setup and mount-related operations.

So I’d like to better understand under what kind of real-world scenario or environment this was observed.

As for the fix itself, I’d also like to see a unit test added for this scenario.

uran0sH:
> Thank you for the fix.
> 
> Would it be possible to share the actual reproduction steps?
> 
> I understand that the intermediate process and the init process run concurrently, but once the init process has been cloned, the intermediate process seems to have very little left to do before sending `IntermediateReady`, whereas the init process still appears to have a fair amount of work remaining, such as namespace setup and mount-related operations.
> 
> So I’d like to better understand under what kind of real-world scenario or environment this was observed.
> 
> As for the fix itself, I’d also like to see a unit test added for this scenario.

I run `make test:integration` in boxlite(https://github.com/boxlite-ai/boxlite). It will fail:
```
FAIL [  21.353s] boxlite::zygote_integration test_zygote_concurrent_stderr_isolation
FAIL [  20.259s] boxlite::zygote_integration test_zygote_concurrent_workdir_isolation
FAIL [  19.162s] boxlite::zygote_integration test_zygote_repeated_bursts
FAIL [  20.012s] boxlite::zygote_integration test_zygote_sequential_then_concurrent
```
And the reason is:
```
thread 'test_concurrent_exec_high_concurrency' (1656451) panicked at src/boxlite/tests/execution_shutdown.rs:1466:46:
    called `Result::unwrap()` on an `Err` value: Internal("spawn_failed: internal error: build failed: failed to create container: received unexpected message: InitReady, expected: IntermediateReady(0)")
```
I suspect it's because it's running inside a VM, which increases the probability of this happening.

nayuta723:
Is this an issue that can also be reproduced in `runc`?

nayuta723:
My understanding is that this issue does not occur in `runc` because it uses separate pipes for communication between the main process and the intermediate process, and between the main process and the init process.

In contrast, in youki’s implementation, the main receiver handles messages from both the intermediate and init processes over the same channel, which I believe is what leads to this issue.

saku3:
Thank you for the comment.

I think this PR may be acceptable as an immediate fix, assuming the tests pass.

However, I think the same ordering issue could also happen with messages other than `InitReady`, such as `HookRequest`, `SetupNetworkDeviceReady`, or `SeccompNotify`.

As nayuta723 mentioned, youki currently receives messages from both the intermediate process and the init process through the shared MainReceiver. To fix this more permanently, I think we should split the communication channels, for example by using separate intermediate-to-main and init-to-main channels.

uran0sH:
> Thank you for the comment.
> 
> I think this PR may be acceptable as an immediate fix, assuming the tests pass.
> 
> However, I think the same ordering issue could also happen with messages other than `InitReady`, such as `HookRequest`, `SetupNetworkDeviceReady`, or `SeccompNotify`.
> 
> As nayuta723 mentioned, youki currently receives messages from both the intermediate process and the init process through the shared MainReceiver. To fix this more permanently, I think we should split the communication channels, for example by using separate intermediate-to-main and init-to-main channels.

I agree with you. I can also provide a permanent fix.

uran0sH:
@saku3 @nayuta723 Hi, I've submitted a new version that splits the channels. I've verified it in Boxlite's tests, but I haven't added the corresponding tests to the Youki repository yet. I thought it is difficult to design a test to reproduce this problem. Do you have any good suggestions?

nayuta723:
Thank you! In my opinion, we don't need to add tests for asynchronous edge cases like CPU pressure, etc. I believe that for this PR, ensuring all existing tests pass and sharing the manual verification results on Boxlite under 'Tested manually (please provide steps)' should be sufficient.

@saku3 
What do you think about that?

saku3:
I agree with nayuta723’s suggestion.

saku3 (approved):
Thanks, that makes sense. If you have a chance, could you update both the PR title and the commit message to reflect the current implementation?

nayuta723 (approved):
LGTM

Thanks!

PR #3523, opened 2026-05-03
Update LinuxIntelRdt to align with runtime-spec

### LINKED ISSUE #3405: Update LinuxIntelRdt to align with runtime-spec
## Overview

We need to update the RDT (Resource DesAllocation Technology) implementation to align with **runtime-spec v1.3.0**. The core shift is moving away from specific flags (`enable_cmt`, `enable_mbm`) toward a more flexible `schemata` string and a unified `enable_monitoring` flag.

This requires changes not only in the OCI structs but also in the logic that interacts with the `resctrl` filesystem.

## Related Implementation
crun: https://github.com/containers/crun/pull/1866

## Reference
https://github.com/opencontainers/runtime-spec/pull/1287

tommady:
## Description
update the RDT (Resource DesAllocation Technology) implementation to align with runtime-spec v1.3.0

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3405


tommady:
Hi @nayuta723
Please help review while you have time.
Thank you.

nayuta723:
@tommady 
Thanks! I have some question for that.
- Would it be possible to release oci-spec-rs before raising this PR? There are some changes that are not related to intelRdt.
- Could you possibly share the execution results from a real environment?

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -328,6 +448,22 @@ fn is_same_schema(combined_schema: &str, existing_schema: &str) -> Result<bool>
     Ok(compare_lines(&combined, &existing))
 }
 
+/// Retrieves the schemata data to be written to the resctrl filesystem.
```
nayuta723, line 475: > If both l3CacheSchema and memBwSchema are set, runtimes MUST write the values to the schemata file in that sub-directory discussed in closID. The runtimes MUST write l3CacheSchema first and memBwSchema last.
If either l3CacheSchema or memBwSchema is set, runtimes MUST write the value to the schemata file in the that sub-directory discussed in closID.
If schemata field is set, runtimes MUST write the value to the schemata file in the that sub-directory discussed in closID. If also l3CacheSchema or memBwSchema is set the value of schemata field must be written last, after the values from l3CacheSchema and memBwSchema has been written.

Please fix this, as it behaves differently from what is described above.

tommady:
Hey @nayuta723 
Thanks for the review.

- Would it be possible to release oci-spec-rs before raising this PR? There are some changes that are not related to intelRdt.
 
Yes, absolutely. We should definitely get the `oci-spec-rs` changes released first since this PR relies on the updated struct definitions. 
I can hold off on merging this until that crate is published.

- Could you possibly share the execution results from a real environment?

I am sorry, I don't have L3 CAT resctrl supported hardware right now.
I can get some temporally (negotiating), ideally I'll test this under Intel, Amd and arch64 (MPAM) platforms.

crates/libcontainer/src/seccomp/mod.rs
nayuta723, line 1: This configuration is out of scope for this PR, so I'd like to handle it in a separate PR.

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -192,7 +194,7 @@ impl ContainerBuilderImpl {
             pid_file: self.pid_file.to_owned(),
         };
 
-        let (init_pid, need_to_clean_up_intel_rdt_dir) =
```
nayuta723, line 195: What is the motivation behind changing the return value here?

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -231,9 +234,16 @@ impl ContainerBuilderImpl {
...
-                    tracing::error!(id = ?container.id(), error = ?e, "failed to delete resctrl subdirectory");
+            if let Some(path) = container.intel_rdt_monitoring_dir() {
+                if let Err(e) = delete_resctrl_monitoring_subdirectory(path) {
+                    tracing::error!(path = ?path, error = ?e, "failed to delete resctrl monitoring subdirectory");
+                    errors.push(e.to_string());
+                }
+            }
+
+            if let Some(path) = container.intel_rdt_dir() {
+                if let Err(e) = delete_resctrl_subdirectory(path) {
```
nayuta723: Since it throws an error when the file doesn't exist, it might be better not to return an error for `ENOENT`.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -63,17 +63,25 @@ pub enum ParseLineError {
...
 }
 
 type Result<T> = std::result::Result<T, IntelRdtError>;
 
-pub fn delete_resctrl_subdirectory(id: &str) -> Result<()> {
+/// Removes the main resource control group (CLOS) directory for the container.
+/// This function includes a safety check (`parent == dir`) to ensure that only directories
+/// immediately under the resctrl mount point are deleted, preventing accidental deletion
+/// of nested kernel files or other sensitive directories.
+pub fn delete_resctrl_subdirectory(path: &Path) -> Result<()> {
```
nayuta723: There doesn't seem to be a strong motivation to switch to path here.
```suggestion
pub fn delete_resctrl_subdirectory(id: &str) -> Result<()> {
```

crates/libcontainer/src/container/state.rs
```diff
@@ -119,8 +119,14 @@ pub struct State {
     pub creator: Option<u32>,
     // Specifies if systemd should be used to manage cgroups
     pub use_systemd: bool,
-    // Specifies if the Intel RDT subdirectory needs be cleaned up.
-    pub clean_up_intel_rdt_subdirectory: Option<bool>,
+    // Specifies the Intel RDT subdirectory path that needs to be cleaned up.
```
nayuta723, line 127: The logic here implies that this path exists when a deletion is required. Could you rename it to something that makes this intent clearer? Also, what was the motivation behind using a path instead of a bool as the state's responsibility?

crates/libcontainer/src/container/container_delete.rs
```diff
@@ -69,41 +71,12 @@ impl Container {
...
 
-        if let Some(true) = &self.clean_up_intel_rdt_subdirectory() {
-            if let Err(err) = delete_resctrl_subdirectory(self.id()) {
-                tracing::warn!(
-                    "failed to delete resctrl subdirectory due to: {err:?}, continue to delete"
-                );
-            }
-        }
-
+        let mut config_opt = None;
```
nayuta723: What is the motivation behind changing `config_opt`? Personally, I don't see the necessity for this change.

tests/contest/contest/src/tests/intel_rdt/intel_rdt_test.rs
```diff
@@ -33,16 +35,46 @@ fn create_spec(
 
 pub fn test_intel_rdt() -> TestResult {
```
nayuta723, line 34: Since the scope of this PR is strictly for v1.3.0 support, could you please add a note to the PR context describing the other behavioral changes (outside of enable_monitoring) and what the ideal behavior for v1.3.0 should look like?

nayuta723: Looking at the code, it seems necessary to include support for `schemata`.

tests/contest/contest/src/tests/intel_rdt/intel_rdt_test.rs
```diff
@@ -10,11 +10,13 @@ fn create_spec(
     maybe_l3_cache: Option<&str>,
     maybe_mem_bw: Option<&str>,
     maybe_clos_id: Option<&str>,
```
nayuta723, line 12: Regarding `maybe_clos_id: Option<&str>`, do we actually need this arg in create_spec under this context?

nayuta723 (changes requested):
Since the scope of this PR seems quite broad, I’ve left comments on the high-level points for now. For intelRdt, even excluding enable_monitoring, we still need changes to align with v1.3.0 and additional tests. This will likely make the PR quite large, so I think it might be a good idea to split it.

Also, as an unreleased side effect, some changes to seccomp have slipped in. It might be better to cut a release before we proceed with reviewing this PR.

Lastly, since you mentioned that real-world testing hasn't been done yet, I'd like to see more rigorous tests added to the PR context. Also, it looks like the tests are being skipped in CI as well, so we might need to start by setting up a working environment. Could you check how runc handles this to see what they do?

nayuta723:
@utam0k 
I’d like to release `oci-spec-rs` version v0.10.0. Are there any blocking issues we need to address before that?

crates/libcontainer/src/process/intel_rdt.rs
tommady: Good catch! 
I misunderstood the specification regarding how the legacy fields should interact with the new schemata list. 

Currently, get_schemata_data returns early if the schemata list is present, completely ignoring l3CacheSchema and memBwSchema. 

I will update the logic to combine all of them in the strict order mandated by the spec (L3 first, then MB, and finally the generic schemata list elements). 

https://specs.opencontainers.org/runtime-spec/config-linux/?v=v1.3.0
```
If schemata field is set, runtimes MUST write the value to the schemata file in the that sub-directory discussed in closID. 
If also l3CacheSchema or memBwSchema is set the value of schemata field must be written last, after the values from l3CacheSchema and memBwSchema has been written.
```

crates/libcontainer/src/seccomp/mod.rs
tommady: Agreed, I will revert the changes.

crates/libcontainer/src/container/builder_impl.rs
tommady: My motivation for changing the return value to pass intel_rdt_dir and intel_rdt_monitoring_dir and storing these paths directly in state.json was to closely mirror how runc implements this. 

runc explicitly stores both IntelRdtPath and IntelRdtMonPath directly in its container state object, and relies on those saved paths when it needs to clean up the directories during teardown.

You can see runc storing the paths in the state here:
https://github.com/opencontainers/runc/blob/main/libcontainer/container_linux.go#L79

crates/libcontainer/src/container/builder_impl.rs
tommady: I will update the logic to handle the ErrorKind::NotFound case gracefully so that if the directory is already gone, it silently succeeds instead of throwing an error during teardown.

crates/libcontainer/src/process/intel_rdt.rs
tommady: Like the above.
https://github.com/youki-dev/youki/pull/3523#discussion_r3254564959

runc explicitly stores both the IntelRdtPath and IntelRdtMonPath directly in its linuxContainer state structure. 

Reconstructing the exact paths dynamically during container deletion is complex (it requires re-parsing the spec for closID, container_id, and enable_monitoring), and storing the literal paths ensures the teardown phase always knows exactly which directories it created and needs to remove, matching the reference implementation's behavior.

utam0k:
> @utam0k I’d like to release `oci-spec-rs` version v0.10.0. Are there any blocking issues we need to address before that?

There are no blockers at the moment. Could you please open the PR for the new release?

crates/libcontainer/src/container/state.rs
tommady: To answer the second part first regarding the motivation for using a path instead of a boolean flag:

This was done to match runc's approach. runc stores the explicit filesystem paths (IntelRdtPath and IntelRdtMonPath) directly in its linuxContainer state structure.

Since youki aims for compatibility, saving the explicit paths avoids complex dynamic reconstruction of those paths during container deletion (which requires re-evaluating the spec, closID, etc.).

Regarding the naming: I just followed how runc named them (intel_rdt_path and intel_rdt_mon_path).

crates/libcontainer/src/container/container_delete.rs
tommady: The motivation behind introducing config_opt was simply to decouple the Intel RDT cleanup logic from the YoukiConfig processing block. 

I wanted to group the RDT teardown operations clearly before the cgroup and hooks teardown, similar to runc's teardown order.

tests/contest/contest/src/tests/intel_rdt/intel_rdt_test.rs
tommady: Will do!

tests/contest/contest/src/tests/intel_rdt/intel_rdt_test.rs
tommady: You're right, I will remove the maybe_clos_id argument from create_spec to simplify the test setup.

tommady:
> Since the scope of this PR seems quite broad, I’ve left comments on the high-level points for now. For intelRdt, even excluding enable_monitoring, we still need changes to align with v1.3.0 and additional tests. This will likely make the PR quite large, so I think it might be a good idea to split it.
> 
> Also, as an unreleased side effect, some changes to seccomp have slipped in. It might be better to cut a release before we proceed with reviewing this PR.
> 
> Lastly, since you mentioned that real-world testing hasn't been done yet, I'd like to see more rigorous tests added to the PR context. Also, it looks like the tests are being skipped in CI as well, so we might need to start by setting up a working environment. Could you check how runc handles this to see what they do?

I understand the concern regarding the scope of the PR and the changes to ContainerState.

To address this and ensure existing deployments are unaffected, I will put all the new OCI v1.3.0 intel_rdt features (including enable_monitoring, the schemata list handling, and the path caching in ContainerState which aligns us with runc) behind a new Cargo feature flag (e.g., OCI_v1_3_0). WDYT?

Regarding the **real-world testing and CI environment**, I investigated how runc handles this since it's the reference implementation. 

It turns out that runc intentionally does not run any Intel RDT integration tests in CI or against a real kernel (you can check their tests/integration/ folder, there are zero RDT tests). 

Instead, runc solely relies on unit tests that mock the /sys/fs/resctrl filesystem by creating a temporary directory (t.TempDir()) and writing plain text files to simulate the kernel interface (e.g., in libcontainer/intelrdt/util_test.go).

This is because the Linux kernel strictly enforces physical hardware capabilities (via MSRs) when interacting with the real /sys/fs/resctrl. 

If the host doesn't physically have Xeon L3 CAT or MBA hardware (which GitHub Actions VMs and most consumer machines do not), the kernel outright rejects the hardcoded L3 or MB schema writes, causing integration tests to fail. 

That being said, I completely agree we need real-world validation for a change this large.

Once the review is complete and the test cases and code changes are all approved, I am willing to manually provision a supported bare-metal instance 

such as a GCP Intel C3, AMD C3D, or Google Axion ARM instance

to prove the implementation works end-to-end against a real kernel. 
I will post the successful test output here before we merge.

Alternatively, if you agree that we should follow runc's approach of mocking the filesystem entirely for unit testing instead of relying on integration tests, I am happy to implement that as well ( to save my money lol ).

nayuta723:
> There are no blockers at the moment. Could you please open the PR for the new release?

@utam0k 
Sure!
https://github.com/youki-dev/oci-spec-rs/pull/336

nayuta723:
@tommady 
The release has been completed.
https://github.com/youki-dev/oci-spec-rs/releases/tag/v0.10.0

tommady:
> @tommady The release has been completed. https://github.com/youki-dev/oci-spec-rs/releases/tag/v0.10.0

@nayuta723 Thanks you for that!!

Before I push the updates, I'd love to get your feeback on two architectural points from my previous comment:

1. Feature Toggle: Are you okay with putting the new intel_rdt v1.3.0 logic (and the associated ContainerState path caching) behind a Cargo feature flag (e.g., v1_3_0)? This avoids introducing any breaking changes to existing deployments and keeps the scope contained, which should address your concern about splitting the PR into smaller pieces.
2. Testing Approach: Would you prefer I keep the integration tests (which I will manually verify on bare-metal GCP Intel/AMD/ARM instances), or should I rewrite the testing to fully mock the /sys/fs/resctrl filesystem for unit testing? To double check, mocking the filesystem is exactly how runc handles intelRdt testing since they also don't run integration tests to avoid physical hardware dependencies. 

Let me know your preference and I'll get the PR updated with the fixes for your inline comments!

nayuta723:
Sorry for the delayed response.

> Feature Toggle: Are you okay with putting the new intel_rdt v1.3.0 logic (and the associated ContainerState path caching) behind a Cargo feature flag (e.g., v1_3_0)? This avoids introducing any breaking changes to existing deployments and keeps the scope contained, which should address your concern about splitting the PR into smaller pieces.

Sorry, I’m not able to make a decision on that yet. Could you help clarify which functionality is strictly required to support v1.3.0? Based on that, could you also explain which parts would be considered breaking changes and why they cannot be implemented incrementally?

> Testing Approach: Would you prefer I keep the integration tests (which I will manually verify on bare-metal GCP Intel/AMD/ARM instances), or should I rewrite the testing to fully mock the /sys/fs/resctrl filesystem for unit testing? To double check, mocking the filesystem is exactly how runc handles intelRdt testing since they also don't run integration tests to avoid physical hardware dependencies.

Following the same approach as runc sounds reasonable to me.
That said, I'm not sure how we should evaluate the `contest` functionality. My understanding is that we should primarily rely on unit tests with a mocked filesystem to validate the implementation, while verification of the actual Intel RDT behavior would need to be performed on hardware that supports Intel RDT.
Given that, I think it's acceptable to exclude the contest implementation from the scope of this PR and address it separately.

tommady:
> Sorry for the delayed response.
> 
> > Feature Toggle: Are you okay with putting the new intel_rdt v1.3.0 logic (and the associated ContainerState path caching) behind a Cargo feature flag (e.g., v1_3_0)? This avoids introducing any breaking changes to existing deployments and keeps the scope contained, which should address your concern about splitting the PR into smaller pieces.
> 
> Sorry, I’m not able to make a decision on that yet. Could you help clarify which functionality is strictly required to support v1.3.0? Based on that, could you also explain which parts would be considered breaking changes and why they cannot be implemented incrementally?
> 
> > Testing Approach: Would you prefer I keep the integration tests (which I will manually verify on bare-metal GCP Intel/AMD/ARM instances), or should I rewrite the testing to fully mock the /sys/fs/resctrl filesystem for unit testing? To double check, mocking the filesystem is exactly how runc handles intelRdt testing since they also don't run integration tests to avoid physical hardware dependencies.
> 
> Following the same approach as runc sounds reasonable to me. That said, I'm not sure how we should evaluate the `contest` functionality. My understanding is that we should primarily rely on unit tests with a mocked filesystem to validate the implementation, while verification of the actual Intel RDT behavior would need to be performed on hardware that supports Intel RDT. Given that, I think it's acceptable to exclude the contest implementation from the scope of this PR and address it separately.

Hi @nayuta723 

Regarding the feature flag and breaking changes: 
You make a great point.
To fix this and avoid the need for a  v1_3_0  feature flag, I will make the state change fully backwards compatible:
I'll keep the legacy **clean_up_intel_rdt_subdirectory** field in the state struct and fall back to the old deletion logic if the new path field isn't present. 
The OCI v1.3.0 spec additions ( schemata  and  enable_monitoring ) are already additive ( Option<T> ), so this whole feature can be implemented incrementally without breaking existing deployments.

WDYT?

Regarding testing: 
You are completely right. I double-checked how runc handles this, and they also restrict their resctrl  mocking strictly to unit tests, while relying on real hardware for E2E tests. 
I will remove the contest implementation from the scope of this PR so we can address it separately, and ensure the unit tests in intel_rdt.rs cover the new logic.

nayuta723:
I think it's fine to proceed with the same incremental approach for the fixes as well!

tommady:
Hi @nayuta723,
please review again.
thank you!

crates/libcontainer/src/container/state.rs
nayuta723: In runc, the internal [`State`](https://github.com/opencontainers/runc/blob/main/libcontainer/container_linux.go#L48-L80) and the exported [`State`](https://github.com/opencontainers/runc/blob/main/list.go#L22-L43) are separated. In crun, on the other hand, information such as the Intel RDT `closID` is reconstructed when needed.

Personally, I consider fields like `intel_rdt_dir` to be internal implementation details, so I'm fine with either approach if we want to expose them in `state.json`.

Which approach do you think is preferable, the runc approach or the crun approach?

If we decide to follow runc, I think we would first need a separate PR to introduce a dedicated structure for the `state.json` output, similar to how runc separates its internal and exported state representations.

crates/libcontainer/src/container/container_delete.rs
nayuta723: I took another look, but I still don't understand the motivation for reordering these changes.

If the goal is to match runc's ordering, does that mean we want to remove the cgroup before removing the Intel RDT resources? I don't think the current implementation is structured that way.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -138,16 +140,18 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
                 .container
                 .as_ref()
                 .map(|container| container.id());
-            need_to_clean_up_intel_rdt_subdirectory =
-                setup_intel_rdt(container_id, &init_pid, intel_rdt)?;
+            let (dir, mon_dir) = setup_intel_rdt(container_id, &init_pid, intel_rdt)?;
+            intel_rdt_dir = dir;
+            intel_rdt_monitoring_dir = mon_dir;
         }
     }
```
nayuta723: Why does this logic live in `container_main_process` in the first place?

It seems to me that this responsibility would fit more naturally in `builder_impl.rs`.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -371,48 +544,53 @@ fn write_resctrl_schemata(
...
+    write_resctrl_schemata(&mount_point, id, intel_rdt, clos_id_set, created_dir).inspect_err(
+        |_err| {
+            tracing::error!("failed to write schemata to resctrl schemata file");
+        },
+    )?;
+
+    let mut created_monitoring_dir = None;
+    if intel_rdt.enable_monitoring().unwrap_or(false) {
+        let mon_dir = container_dir.join("mon_groups").join(container_id);
+        if setup_monitoring_group(&mon_dir, *init_pid)? {
```
nayuta723: > The runtime MUST delete the MON group after the container is deleted.

I think it would be better to always set `monitoring_dir`, regardless of whether `setup_monitoring_group` is enabled or not.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -371,48 +544,53 @@ fn write_resctrl_schemata(
...
+        }
+    }
 
     // If closID is not set and the runtime has created the sub-directory,
     // the runtime MUST remove the sub-directory when the container is deleted.
-    let need_to_delete_directory = !clos_id_set && created_dir;
+    let mut need_to_delete_directory = None;
+    if !clos_id_set && created_dir {
+        need_to_delete_directory = Some(container_dir);
+    }
```
nayuta723:
```suggestion
    let mut need_to_delete_directory = (!clos_id_set && created_dir).then_some(container_dir);
    }
```

crates/libcontainer/src/container/container_delete.rs
```diff
@@ -121,7 +89,61 @@ impl Container {
...
+                    "failed to delete resctrl subdirectory due to: {err:?}, continue to delete"
+                );
+            }
+        } else if let Some(true) = self.clean_up_intel_rdt_subdirectory() {
+            if let Err(err) = delete_resctrl_subdirectory_by_id(self.id()) {
+                tracing::warn!(
+                    "failed to delete resctrl subdirectory by id due to: {err:?}, continue to delete"
+                );
+            }
+        }
```
nayuta723: I think applying the Intel RDT cleanup logic directly in `container_delete.rs` is outside the scope of its responsibility.

I'd prefer to define a `cleanup_intel_rdt()` function in `intel_rdt.rs` and have `container_delete.rs` simply invoke it.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -118,50 +173,85 @@ pub fn find_resctrl_mount_point() -> Result<PathBuf> {
...
+        .open(tasks)
+        .map_err(|err| {
+            tracing::error!("failed to open resctrl monitoring tasks file: {}", err);
+            IntelRdtError::OpenClosIDDirectory(err)
+        })?;
+
+    write!(file, "{init_pid}").map_err(|err| {
+        tracing::error!("failed to write to resctrl monitoring tasks file: {}", err);
+        IntelRdtError::WriteClosIDDirectory(err)
+    })?;
```
nayuta723: I recommend extracting this into a `write_pid_to_tasks()` function to avoid duplicating the logic.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -371,48 +544,53 @@ fn write_resctrl_schemata(
...
-        clos_id_set,
-        created_dir,
-    )
-    .inspect_err(|_err| {
-        tracing::error!("failed to write schemata to resctrl schemata file");
-    })?;
+    let container_id = maybe_container_id.ok_or(IntelRdtError::ResctrlIdNotFound)?;
+    let clos_id_set = intel_rdt.clos_id().is_some();
+    let id = intel_rdt.clos_id().as_deref().unwrap_or(container_id);
+    let only_clos_id_set = clos_id_set && get_schemata_data(intel_rdt).is_none();
```
nayuta723: `get_schemata_data()` allocates a `String`, so I think it's fine to check each field directly with `is_none()` instead.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -118,50 +173,85 @@ pub fn find_resctrl_mount_point() -> Result<PathBuf> {
...
+/// memory bandwidth) of the container independently from the parent allocation group (CLOS).
+///
+/// Returns `true` if the runtime created the directory, or `false` if it already existed.
+fn setup_monitoring_group(mon_dir: &Path, init_pid: Pid) -> Result<bool> {
+    let mut created_dir = false;
+
+    if !mon_dir.exists() {
+        fs::create_dir_all(mon_dir).map_err(|err| {
+            tracing::error!("failed to create resctrl monitoring subdirectory: {}", err);
+            IntelRdtError::CreateClosIDDirectory(err)
```
nayuta723: Since this is a monitoring directory, `CreateClosIDDirectory` is not an appropriate error name.

I think we should either introduce a dedicated error variant or rename it to something more generic.

nayuta723 (changes requested):

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -594,29 +895,40 @@ mod test {
...
+        let written_data =
+            fs::read_to_string(tmp.path().join("foobar_modern").join("schemata")).unwrap();
+        assert_eq!(
+            written_data,
+            "L3:0=f;1=f0\nL3:2=f\nMB:0=70;1=20\nL2:0=f;1=f0\n"
+        );
+
         Ok(())
     }
 }
```
nayuta723, line 967: Could you also add a test case for container deletion?

crates/libcontainer/src/container/state.rs
tommady: I prefer to follow the runc approach of explicitly storing intel_rdt_dir and intel_rdt_monitoring_dir in our internal state, as it avoids complex and potentially error-prone path reconstruction during container deletion.

Regarding your concern about leaking these internal implementation details: you are completely right that adding these fields will expose them when a user runs the youki state CLI command, because that command currently prints our internal State struct directly.

However, we actually already have the foundation to fix this! 
We had TryFrom<&State> for OciState (used to pass strict OCI state to lifecycle hooks), which perfectly filters out our internal fields (like use_systemd, clean_up_intel_rdt_subdirectory, and these new RDT paths). 

As you suggested, fully migrating the youki state command (and potentially the state.json structure on disk) to separate the internal and exported representations is out of scope for this PR. We can handle that migration in a separate, dedicated PR. 

For now, I'll proceed with storing the paths explicitly to match runc's teardown behavior.

crates/libcontainer/src/container/container_delete.rs
tommady: Ah, you are completely right.

I mistakenly thought I needed to reorder the logic to ensure Intel RDT cleanup happened before cgroup cleanup (like runc does).

Extracting config_opt was completely unnecessary and just convoluted the code.

As you mentioned in your other comment, I will also define a cleanup_intel_rdt() function in intel_rdt.rs and have container_delete.rs simply invoke it.

crates/libcontainer/src/process/intel_rdt.rs
tommady: added~

crates/libcontainer/src/container/state.rs
nayuta723: Let's follow runc's implementation and use this struct here.

crates/libcontainer/src/container/builder_impl.rs
```diff
@@ -192,21 +192,36 @@ impl ContainerBuilderImpl {
...
                 .set_status(ContainerStatus::Created)
                 .set_creator(nix::unistd::geteuid().as_raw())
                 .set_pid(init_pid.as_raw())
-                .set_clean_up_intel_rdt_directory(need_to_clean_up_intel_rdt_dir)
+                .set_intel_rdt_dir(intel_rdt_dir)
+                .set_intel_rdt_monitoring_dir(intel_rdt_monitoring_dir)
+                // DEPRECATED: We populate the legacy boolean field here so that if a user creates
+                // a container with this version of youki and then downgrades their binary to an
+                // older version, the older binary will still know to clean up the resctrl directory.
+                .set_clean_up_intel_rdt_directory(needs_cleanup)
```
nayuta723: Isn't the legacy fallback in `clean_up_intel_rdt_subdirectory()` only necessary on the cleanup?

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -63,34 +71,128 @@ pub enum ParseLineError {
...
+    GenericToken,
 }
 
 type Result<T> = std::result::Result<T, IntelRdtError>;
 
-pub fn delete_resctrl_subdirectory(id: &str) -> Result<()> {
+/// Removes the main resource control group (CLOS) directory for the container using its ID.
+/// This is kept for backwards compatibility with older state.json files where only
+/// `clean_up_intel_rdt_subdirectory` was stored instead of the explicit path.
+pub fn delete_resctrl_subdirectory_by_id(id: &str) -> Result<()> {
```
nayuta723, line 89: There seem to be too many variants of the delete helpers (`delete_resctrl_subdirectory_by_id()`, `delete_resctrl_subdirectory_with_dir()`, `delete_resctrl_subdirectory()`, and `delete_resctrl_monitoring_subdirectory()`). Can we consolidate them and keep the API surface minimal?I'd prefer to keep only `delete_resctrl_subdirectory()` and `delete_resctrl_monitoring_subdirectory()`.

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -118,50 +220,88 @@ pub fn find_resctrl_mount_point() -> Result<PathBuf> {
...
+
+/// Creates a dedicated monitoring group (`mon_groups/<container_id>`) inside the container's
+/// Intel RDT resource control directory and adds the container's PID to its `tasks` file.
+///
+/// As per OCI runtime-spec v1.3.0, if `enable_monitoring` is set to true, the runtime MUST
+/// create this subdirectory. This allows users to track the resource utilization (e.g., L3 cache,
+/// memory bandwidth) of the container independently from the parent allocation group (CLOS).
+///
+/// Returns `true` if the runtime created the directory, or `false` if it already existed.
+fn setup_monitoring_group(mon_dir: &Path, init_pid: Pid) -> Result<bool> {
```
nayuta723:
```suggestion
fn setup_monitoring_group(mon_dir: &Path, init_pid: Pid) -> Result<()> {
```

nayuta723: Do we need to factor this out into a separate function at all?

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -118,50 +220,88 @@ pub fn find_resctrl_mount_point() -> Result<PathBuf> {
...
+{
+    let tasks = dir.join("tasks");
+    // TODO(ipuustin): File doesn't need to be created, but it's easier
+    // to test this way. Fix the tests so that the fake resctrl
+    // filesystem is pre-populated.
+    let mut file = OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(tasks)
+        .map_err(on_open_err)?;
```
nayuta723: I think it would be better to remove the test-specific code. Would it be difficult to address this TODO as part of this change?

crates/libcontainer/src/process/intel_rdt.rs
```diff
@@ -63,34 +71,113 @@ pub enum ParseLineError {
...
+/// directory is already gone.
+pub fn delete_resctrl_subdirectory(path: &Path) -> Result<()> {
+    if let Err(err) = fs::remove_dir(path) {
+        if err.kind() != ErrorKind::NotFound {
+            tracing::error!(?path, "failed to remove resctrl subdirectory: {err}");
+            return Err(IntelRdtError::RemoveSubdirectory(err));
+        }
+    }
+    Ok(())
+}
```
nayuta723: Is it necessary to make it a function?

nayuta723 (commented):
I left one small nit. Thanks for your contribution!

nayuta723 (approved):
LGTM

Please add the items that are out of scope for this PR to this issue, or create a new issue for them.
https://github.com/youki-dev/youki/issues/3405

tommady:
> LGTM
> 
> Please add the items that are out of scope for this PR to this issue, or create a new issue for them. #3405

I found out we had a similar issue there. 
I created a sub-issue under it, and I'm going to take it on. 
Thank you.

utam0k (approved):

PR #3528, opened 2026-05-04
fix: validate process.terminal field against --console-socket option

### LINKED ISSUE #3522: bug: terminal field in OCI spec is ignored when handling --console-socket
## Description

youki does not validate or respect the `process.terminal` field from the OCI Runtime Spec when handling `--console-socket`. This causes incorrect behavior that deviates from runc's implementation.

## Expected Behavior (runc)

runc performs the following validation in `checkTerminal()` ([utils_linux.go](https://github.com/opencontainers/runc/blob/main/utils_linux.go#L342-L352)):

```go
func (r *runner) checkTerminal(config *specs.Process) error {
    detach := r.detach || (r.action == CT_ACT_CREATE)
    if detach && config.Terminal && r.consoleSocket == "" {
        return errors.New("cannot allocate tty if runc will detach without setting console socket")
    }
    if (!detach || !config.Terminal) && r.consoleSocket != "" {
        return errors.New("cannot use console socket if runc will not detach or allocate tty")
    }
    return nil
}
```

## Actual Behavior (youki)

youki ignores the `process.terminal` field entirely. The PTY setup is determined solely by whether `--console-socket` is provided:

```rust
// crates/libcontainer/src/container/init_builder.rs
let csocketfd = if let Some(console_socket) = &self.base.console_socket {
    Some(tty::setup_console_socket(...))  // terminal field never checked
} else {
    None
};
```

## Behavior Comparison

| Case | runc | youki |
|---|---|---|
| `terminal=true` + `--console-socket` provided | OK | OK |
| `terminal=true` + `--console-socket` missing | Error | Not validated |
| `terminal=false` + `--console-socket` provided | Error | Silently creates PTY |
| `terminal=false` + `--console-socket` missing | OK | OK |

## Additional Context

`youki spec` generates `terminal: false` via `Spec::default()` (from oci-spec-rs), whereas `runc spec` generates `terminal: true` explicitly. This means:

- In the contest integration tests (e.g., `setup_cr_test`), containers are started with `--console-socket` despite `terminal: false` in `config.json`. This works only because youki ignores the `terminal` field — runc would reject the same invocation with an error.
- The `restore_container` helper checks `is_terminal` from `config.json` to decide whether to pass `--console-socket` during restore. Since `terminal` is `false` in the generated spec, the console socket is never passed on restore, meaning the restore-with-TTY path is never actually tested.

## Steps to Reproduce

1. Create a bundle with `process.terminal: false` in `config.json`
2. Run `youki run -d --console-socket /tmp/console.sock <id>`
3. youki accepts this and creates a PTY — runc would return an error

<details>

```
#!/usr/bin/env python3
import array
import os
import socket
import subprocess
import time
import select

CONSOLE_SOCK = "/tmp/console.sock"
CONTAINER_ID = "3"
BUNDLE_DIR = "/home/ubuntu/youki/tutorial"


def recv_fd(conn: socket.socket) -> int:
    fds = array.array("i")
    msg, ancdata, flags, addr = conn.recvmsg(
        1024,
        socket.CMSG_SPACE(fds.itemsize),
    )

    for level, ctype, data in ancdata:
        if level == socket.SOL_SOCKET and ctype == socket.SCM_RIGHTS:
            fds.frombytes(data[: fds.itemsize])
            return fds[0]

    raise RuntimeError("no fd received via SCM_RIGHTS")


def cleanup():
    subprocess.run(
        ["sudo", "youki", "kill", CONTAINER_ID, "SIGKILL"],
        cwd=BUNDLE_DIR,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
    )
    subprocess.run(
        ["sudo", "youki", "delete", CONTAINER_ID],
        cwd=BUNDLE_DIR,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
    )
    try:
        os.unlink(CONSOLE_SOCK)
    except FileNotFoundError:
        pass


def main():
    try:
        # prepare socket
        try:
            os.unlink(CONSOLE_SOCK)
        except FileNotFoundError:
            pass

        listener = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
        listener.bind(CONSOLE_SOCK)
        listener.listen(1)

        # start container
        cmd = [
            "sudo",
            "youki",
            "run",
            "-d",
            "--console-socket",
            CONSOLE_SOCK,
            CONTAINER_ID,
        ]

        proc = subprocess.Popen(
            cmd,
            cwd=BUNDLE_DIR,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            text=True,
        )

        # accept console connection
        listener.settimeout(10)
        conn, _ = listener.accept()

        # receive PTY fd
        console_fd = recv_fd(conn)
        print(f"received console fd: {console_fd}")

        # verify console works
        os.write(console_fd, b"echo console-ready\n")

        ready, _, _ = select.select([console_fd], [], [], 3)
        if ready:
            output = os.read(console_fd, 4096)
            print("console output:")
            print(output.decode(errors="replace"))
        else:
            raise RuntimeError("console fd did not produce any output")

        # verify container state
        time.sleep(1)
        status = subprocess.run(
            ["sudo", "youki", "state", CONTAINER_ID],
            cwd=BUNDLE_DIR,
            text=True,
            capture_output=True,
        )

        if status.returncode != 0:
            raise RuntimeError(status.stderr)

        print("container state:")
        print(status.stdout)

        proc.wait(timeout=5)

    finally:
        cleanup()


if __name__ == "__main__":
    main()
```

</details>

## References

- OCI Runtime Spec: https://github.com/opencontainers/runtime-spec/blob/main/config.md#process
- runc `checkTerminal`: https://github.com/opencontainers/runc/blob/main/utils_linux.go

nayuta723:
## Description

Fixes the issue where youki ignored the `process.terminal` field from the OCI Runtime Spec when handling `--console-socket`, causing incorrect behavior compared to runc.

Adds `check_terminal()` validation that mirrors runc's `checkTerminal()` in `utils_linux.go`: it rejects invalid combinations of `process.terminal` and `--console-socket` **before** any container state is created on disk.

| Case | Before | After |
|------|--------|-------|
| `terminal=true` + `--console-socket` provided | OK | OK |
| `terminal=true` + `--console-socket` missing | Not validated | Error |
| `terminal=false` + `--console-socket` provided | Silently creates PTY | Error |
| `terminal=false` + `--console-socket` missing | OK | OK |

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3522

## Additional Context

The validation is intentionally placed before `create_container_dir()` so that no container state is written to disk when validation fails — matching the ordering of runc's `checkTerminal()` call at the top of `runner.run()`.

### DIFF
--- a/crates/libcontainer/src/container/init_builder.rs
+++ b/crates/libcontainer/src/container/init_builder.rs
@@ -73,6 +73,10 @@ impl InitContainerBuilder {
     /// Creates a new container
     pub fn build(self) -> Result<Container, LibcontainerError> {
         let spec = self.load_spec()?;
+        // validate terminal field against console socket presence before any side effects
+        // (mirrors runc's checkTerminal called at the top of runner.run())
+        self.base.check_terminal(&spec, self.detached)?;
+
         let container_dir = self.create_container_dir()?;
 
         let mut container = self.create_container_state(&container_dir)?;
--- a/crates/libcontainer/src/container/builder.rs
+++ b/crates/libcontainer/src/container/builder.rs
@@ -1,9 +1,11 @@
 use std::os::fd::OwnedFd;
 use std::path::PathBuf;
 
+use oci_spec::runtime::Spec;
+
 use super::init_builder::InitContainerBuilder;
 use super::tenant_builder::TenantContainerBuilder;
-use crate::error::{ErrInvalidID, LibcontainerError};
+use crate::error::{ErrInvalidID, ErrInvalidSpec, LibcontainerError};
 use crate::syscall::syscall::SyscallType;
 use crate::utils::PathBufExt;
 use crate::workload::{self, Executor};
@@ -336,6 +338,28 @@ impl ContainerBuilder {
         self.stderr = Some(stderr.into());
         self
     }
+
+    pub(super) fn check_terminal(
+        &self,
+        spec: &Spec,
+        detached: bool,
+    ) -> Result<(), LibcontainerError> {
+        let terminal = spec
+            .process()
+            .as_ref()
+            .and_then(|p| p.terminal())
+            .unwrap_or(false);
+        let has_console_socket = self.console_socket.is_some();
+        let requires_console_socket = detached && terminal;
+
+        if requires_console_socket && !has_console_socket {
+            Err(ErrInvalidSpec::ConsoleSocketRequired)?;
+        }
+        if !requires_console_socket && has_console_socket {
+            Err(ErrInvalidSpec::InvalidConsoleSocket)?;
+        }
+        Ok(())
+    }
 }
 
 #[cfg(test)]
@@ -345,8 +369,10 @@ mod tests {
 
     use anyhow::{Context, Result};
     use nix::unistd::pipe;
+    use oci_spec::runtime::{ProcessBuilder, SpecBuilder};
 
     use crate::container::builder::ContainerBuilder;
+    use crate::error::{ErrInvalidSpec, LibcontainerError};
     use crate::syscall::syscall::SyscallType;
 
     #[test]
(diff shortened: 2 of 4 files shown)


crates/libcontainer/src/container/init_builder.rs
```diff
@@ -136,6 +140,31 @@ impl InitContainerBuilder {
...
 
+    fn check_terminal(&self, spec: &Spec) -> Result<(), LibcontainerError> {
+        let terminal = spec
+            .process()
+            .as_ref()
+            .and_then(|p| p.terminal())
+            .unwrap_or(false);
+        let has_console_socket = self.base.console_socket.is_some();
+
+        if self.detached && terminal && !has_console_socket {
```
saku3: Personally, I think the condition would be easier to read if we introduced a variable like this:

let requires_console_socket = self.detached && terminal;

Then the checks could be written as:

```
if requires_console_socket && !has_console_socket { # if self.detached && terminal && !has_console_socket {

if !requires_console_socket && has_console_socket { #  if (!self.detached || !terminal) && has_console_socket {
```
This makes the second condition easier to understand than `(!self.detached || !terminal) && has_console_socket`.

crates/libcontainer/src/container/init_builder.rs
```diff
@@ -136,6 +140,31 @@ impl InitContainerBuilder {
...
+            tracing::error!(
```
saku3: I think we can just return the error here without logging.

The same error is already reported by the top-level command handler, so logging it in `check_terminal()`seems to make the CLI output duplicated and noisy.

```
ERROR libcontainer::container::init_builder: console socket provided but container will not detach or allocate tty detached=true terminal=false
ERROR youki: error in executing command: invalid runtime spec

Caused by:
    cannot use console socket if youki will not detach or allocate tty
```

saku3 (commented):
Thanks for PR.

I think the same validation is also needed on the tenant_builder.rs side.

I would expect the following case to return an error:

```sh
youki run container # starts with terminal: false
youki exec -d --console-socket /tmp/exec.sock container sh # without -t, this means terminal: false
```

saku3 (approved):

PR #3530, opened 2026-05-05
add aarch64 bundle for integration test

### LINKED ISSUE #3311: Integration test failure: kill_start_container_test reports "cannot kill container due to incorrect state" on Linux ARM64
Hello, I'm interested in this project, so I tried following the steps in "Implement Integration Tests".
But when I ran some tests, I encountered an error, for example:

```sh
just test-contest kill_container
```
```sh
#test.log
# Start group kill_container
1 / 5 : kill_created_container_test : ok
2 / 5 : kill_non_existed_container : ok
3 / 5 : kill_start_container_test : not ok
    Error :
stdout : 
stderr : ERROR libcontainer::container::container_kill: cannot kill container due to incorrect state id="e58dd6ae-121c-2ced-3eb7-0c38e08e826c" status=Stopped
ERROR youki: error in executing command: container not running

Caused by:
    failed to perform operation due to incorrect container status
error in executing command: container not running

Caused by:
    failed to perform operation due to incorrect container status
Error: container not running

Caused by:
    failed to perform operation due to incorrect container status

4 / 5 : kill_stopped_container_test : ok
5 / 5 : kill_with_empty_id_test : ok
# End group kill_container
```
I tested on Linux arm64 (Vagrant + VMware Fusion on an M1 MacBook).
I tested using the code from main (commit 890b3ed560fe23ecca83bb19d9a04db20a92d02b).
I'm not sure if I did something wrong. I apologize in advance if I’m misunderstanding something here. 
:bow:

saku3:
## Description

This adds an `aarch64` bundle.

I would like to run youki on Lima on macOS and also run the integration tests on `aarch64`.

On Lima, `uname -m` returns `aarch64`, and youki is built as an `aarch64` binary as well.

I created the bundle as follows:

```
mkdir -p bundle/rootfs
CID=$(docker create --platform=linux/arm64 busybox:musl)
docker export "$CID" | tar -C bundle/rootfs -xf -
docker rm "$CID"
tar --uid 0 --gid 0 --uname root --gname root -czf bundle-aarch64.tar.gz bundle/
```

Also, the personality test does not work correctly unless it runs on x86_64, so I will skip it.
This is because PER_LINUX32 makes uname(2) return a 32-bit architecture name, such as i686 instead of x86_64, and this behavior is architecture-dependent.
See: https://man7.org/linux/man-pages/man2/personality.2.html

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [x] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3311


tests/contest/contest/src/tests/personality/mod.rs
```diff
@@ -70,14 +70,27 @@ fn personality_for_linux64() -> TestResult {
     personality_for_linux(LinuxPersonalityDomain::PerLinux, "x86_64")
 }
```
nayuta723, line 71: Rather than failing because it’s `aarch`, I suspect the failure is caused by the expected value being different here.
I think it should work if we adjust the expected values for each architecture. What do you think?

nayuta723 (commented):
It looks like the `aarch64` tests may not have been added to CI. Could you please check?

tests/contest/contest/src/tests/personality/mod.rs
saku3: That may be true, but I do not intend to address it in this PR.

runc also runs this test only on x86_64:

https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/tests/integration/personality.bats#L6

Personally, I do not think we need to handle this here, but if it is necessary, I can create a separate issue for it.

saku3:
> It looks like the `aarch64` tests may not have been added to CI. Could you please check?

I did not intend to add aarch64 tests to CI in this PR.

The purpose of this PR is only to make the integration tests work in a Lima environment.

If adding aarch64 tests to CI is necessary, I can open a separate issue

tests/contest/contest/src/tests/personality/mod.rs
nayuta723: I see, that makes sense!
For now, it seems reasonable to follow `runc` and support only `x86_64`.

It would also be helpful if you could leave a comment in the code explaining that supporting other architectures is technically possible, but intentionally not implemented at this time.

nayuta723 (approved):
I only left one comment — once that’s addressed, the rest looks LGTM!

> If adding aarch64 tests to CI is necessary, I can open a separate issue

I think we should support `aarch64` as well.
Could you create a separate issue for that?

saku3:
@nayuta723 

Thanks.

>I think we should support aarch64 as well.
>Could you create a separate issue for that?

Added.

https://github.com/youki-dev/youki/issues/3533

>It would also be helpful if you could leave a comment in the code explaining that supporting other architectures is technically possible, but intentionally not implemented at this time.

I added a comment.

nayuta723:
LGTM

scripts/contest.sh
```diff
@@ -18,8 +18,18 @@ fi
 
 LOGFILE="${ROOT}/test.log"
 
+# Pick arch-specific bundle when available.
+# Fall back to bundle.tar.gz (x86_64) if there is no
+# arch-specific tarball.
```
utam0k, line 23: Let's add a warning log for the fallback case.

saku3: added.

utam0k (approved):
LGTM

PR #3541, opened 2026-05-17
fix: parse info parameters exactly

immanuwell:
## Description
Fixes a tiny parser footgun in `youki info`.

`find_parameter` used prefix matching and split on the last `=`, so `NAME_SUFFIX=wrong` could beat `NAME=right`, and `PRETTY_NAME="Foo=Bar"` became `Bar"`. Not great.

Repro with the old code:
```text
NAME_SUFFIX=wrong
NAME=right
```
asking for `NAME` returns `wrong`.

```text
PRETTY_NAME="Foo=Bar"
```
asking for `PRETTY_NAME` returns `Bar"`.

Now it matches the exact key and keeps the full value after the first `=`.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing
- [x] Added new unit tests
- [x] Ran existing test suite

Ran:
```text
cargo test -p youki commands::info::tests
cargo test -p youki commands::info
cargo test -p youki
cargo fmt --check
```

## Related Issues
No direct issue found, afaict.

## Additional Context
`/etc/os-release` is environment-like key/value data, so exact key matching is the boring-but-right move here imo.

### DIFF
--- a/crates/youki/src/commands/info.rs
+++ b/crates/youki/src/commands/info.rs
@@ -115,8 +115,8 @@ fn try_read_os_from<P: AsRef<Path>>(path: P) -> Option<String> {
 fn find_parameter<'a>(content: &'a str, param_name: &str) -> Option<&'a str> {
     content
         .lines()
-        .find(|l| l.starts_with(param_name))
-        .and_then(|l| l.split_terminator('=').next_back())
+        .filter_map(|l| l.split_once('='))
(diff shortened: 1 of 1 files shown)


saku3 (approved):
Thanks!

nayuta723 (approved):
LGTM!

PR #3546, opened 2026-05-20
fix(libcgroups): return typed error when systemd is not available

### LINKED ISSUE #3545: [Bug]: libcgroups can panic if systemd is not available
### Bug Description

I'm using a crate which itself depends on libcgroups. On some systems the whole process crashes down via panic:

```
2026-05-19T22:01:31.943476Z ERROR orchestrator: execution for a deliverable panicked! task 80 panicked with message "systemd cgroup flag passed, but systemd support for managing cgroups is not available"
```

### Steps to Reproduce

Drive systemd cgroups in a sandbox or non-systemd system

### Expectation

An error instead of a panic.

The issue is that this code:

https://github.com/youki-dev/youki/blob/6ab57db163275d82a2b16cd0f0735e2bfce55a6e/crates/libcgroups/src/common.rs#L417-L421

Panics instead of just returning a `Result::Err` representing that the system didnt boot systemd.

### System and Setup Info

_No response_

### Additional Context

_No response_

SAY-5:
## Description
`create_systemd_cgroup_manager` in `crates/libcgroups/src/common.rs` aborted the entire process with `panic!` whenever `systemd::booted()` returned false, e.g. when a caller passed the systemd cgroup flag inside a sandbox or on a host that wasn't booted with systemd. The crash shows up in any binary that links libcgroups, so it's a host-environment denial-of-service rather than a programmer error.

This change replaces the panic with a typed `SystemdManagerError::SystemdNotAvailable` variant and returns it through the existing `SystemdManagerError -> CreateCgroupSetupError::Systemd` conversion, so callers can handle it like every other cgroup-manager error.

## Type of Change

## Testing
- [x] Added new unit tests

Added `systemd_not_available_is_a_typed_error` regression test in `crates/libcgroups/src/systemd/manager.rs::tests` that locks in the new variant, its message, and the `From` conversion into `CreateCgroupSetupError::Systemd`.

Locally verified with `cargo check -p libcgroups --tests --target x86_64-unknown-linux-gnu` and `cargo clippy -p libcgroups --tests --target x86_64-unknown-linux-gnu --no-deps -- -D warnings` (the crate is Linux-only so the full `cargo test` runs on CI).

## Related Issues
Fixes #3545

### DIFF
--- a/crates/libcgroups/src/systemd/manager.rs
+++ b/crates/libcgroups/src/systemd/manager.rs
@@ -176,6 +176,10 @@ pub enum SystemdManagerError {
     Pids(Infallible),
     #[error("in pids unified controller: {0}")]
     Unified(#[from] super::unified::SystemdUnifiedError),
+    #[error(
+        "systemd cgroup flag passed, but systemd support for managing cgroups is not available"
+    )]
+    SystemdNotAvailable,
 }
 
 impl Manager {
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -415,9 +415,7 @@ fn create_systemd_cgroup_manager(
     use crate::systemd::manager::PROCESS_IN_CGROUP_TIMEOUT_DURATION;
 
     if !systemd::booted() {
-        panic!(
-            "systemd cgroup flag passed, but systemd support for managing cgroups is not available"
-        );
+        return Err(systemd::manager::SystemdManagerError::SystemdNotAvailable);
     }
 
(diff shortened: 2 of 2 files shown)


crates/libcgroups/src/systemd/manager.rs
```diff
@@ -642,4 +646,26 @@ mod tests {
         ));
         Ok(())
     }
+
```
saku3: This test case does not exercise the target `create_systemd_cgroup_manager` path, so I don't think it really works as a regression test for the original issue.

I think we can remove it.

saku3 (approved):
Thanks.

PR #3265, opened 2025-09-30
Time Namespace

CheatCodeSam:
## Description

Adding the time namespace.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [ ] Added new unit tests
- [x] Added new integration tests
- [x] Ran existing test suite
- [x] Tested manually (please provide steps)

## Related Issues

Fixes #3259

PR #3550, opened 2026-05-23
support time namespace

### LINKED ISSUE #3259: support time namespace
As far as I can confirm, there’s no implementation for the time namespace. Please refer to the following.

runtime spec
* https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#namespaces
* https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#offset-for-time-namespace

runc implementation

* https://github.com/opencontainers/runc/pull/3876

### LINKED ISSUE #3538: youki fails to start any container with Docker 29.5 (time namespace rejected)
## Description

Starting with Docker 29.5 (after [moby/moby#52326](https://github.com/moby/moby/pull/52326)), the Docker daemon adds `{"type": "time"}` to the default OCI runtime spec. youki rejects this namespace type, so `docker run --runtime=youki` invocation fails on a freshly installed Docker 29.5.

## Steps to Reproduce

```
$ docker --version
Docker version 29.5.0, build 98f1464

$ docker run --runtime youki hello-world
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: /usr/local/sbin/youki did not terminate successfully: exit status 1

Run 'docker run --help' for more information

$ sudo journalctl SYSLOG_IDENTIFIER=youki -r
error in executing command: invalid spec for new user namespace container
Caused by:
Namespace type not supported: time
```

## Expected behaviour

`docker run --runtime=youki hello-world` succeeds.

## Root cause

As the error indicates, this is because time namespaces are not implemented in youki yet.

https://github.com/youki-dev/youki/issues/3259

I think the recent `just test-dind` failures are also caused by using the latest Docker version.

https://github.com/youki-dev/youki/actions/runs/25983703182/job/76377158770?pr=3530

https://github.com/youki-dev/youki/blob/main/tests/dind/run.sh#L10

saku3:
## Description

This continues the work from the following PR.

https://github.com/youki-dev/youki/pull/3265

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3259
Fixes #3538


crates/libcontainer/src/utils.rs
```diff
@@ -419,6 +419,31 @@ fn dev_valid_name(name: &str) -> bool {
     true
 }
 
+// Validates time namespace configuration. Rejects two invalid combinations
+// when time_offsets is set:
+//   (1) no time namespace in the spec -- nothing for the offsets to apply to
+//   (2) time namespace has a path -- joining an existing NS, whose offsets are
+//       already fixed and cannot be rewritten
+pub fn validate_time_namespace(spec: &Spec) -> Result<(), LibcontainerError> {
```
saku3: I addressed the review comment below.

https://github.com/youki-dev/youki/pull/3265#discussion_r2619111878

In addition, I added validation to reject the invalid case where `timeOffsets` is specified together with a `path` for the time namespace.

see: 
https://github.com/opencontainers/runc/commit/09822c3da8ad8fa91b8796c5abf27ef06814a0c3

saku3:
The `rootless-podman-test` failure is fixed by the following PR:

https://github.com/youki-dev/youki/pull/3551

It looks like `youki exec` was attempting to `setns` into namespaces that were not specified for the init process. Those namespaces are inherited from the parent executor, so joining them is unnecessary and can require extra privileges in rootless environments.

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -130,6 +130,13 @@ pub fn container_intermediate_process(
         }
     }
 
+    if let Some(time_namespace) = namespaces.get(LinuxNamespaceType::Time)? {
+        namespaces.unshare_or_setns(time_namespace)?;
+        if time_namespace.path().is_none() && linux.time_offsets().is_some() {
+            setup_time_offsets(intermediate_main_sender, inter_receiver)?;
+        }
+    }
```
utam0k, line 143: Do we need to execute it in this order? I'm asking because I'm wondering if it's necessary to increase the communication via pipe.

crates/libcontainer/src/utils.rs
```diff
@@ -419,6 +419,31 @@ fn dev_valid_name(name: &str) -> bool {
...
+    if let Some(linux) = spec.linux() {
```
utam0k:
```suggestion
    if let Some(linux) = spec.linux() && linux.time_offsets().is_some() {
```

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -128,6 +130,16 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
         inter_sender.mapping_written()?;
     }
 
+    // If creating a container with time namespace and offsets are specified,
+    // the intermediate process will ask the main process to write the time offsets.
+    let time_offsets = extract_time_offsets(container_args.spec.linux().as_ref())?;
+    // This must be done from the parent process since CAP_SYS_TIME is required.
+    if let Some(time_offsets) = time_offsets {
+        intermediate_main_receiver.wait_for_time_offset_request()?;
```
utam0k:
```suggestion
    // If creating a container with time namespace and offsets are specified,
    // the intermediate process will ask the main process to write the time offsets.
    extract_time_offsets(container_args.spec.linux().as_ref())?.map(|time_offset| {
        // This must be done from the parent process since CAP_SYS_TIME is required.
        intermediate_main_receiver.wait_for_time_offset_request()?;
    });
```

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -317,6 +329,49 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
...
+
+    let time_ns = linux
+        .namespaces()
+        .as_ref()
+        .and_then(|nss| {
+            nss.iter()
+                .find(|ns| ns.typ() == oci_spec::runtime::LinuxNamespaceType::Time)
+        })
+        .ok_or_else(|| {
+            ProcessError::InvalidTimeNamespace(
```
utam0k: Hasn't this already been checked in the validation?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -317,6 +329,49 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
     Ok(())
 }
 
+fn extract_time_offsets(linux: Option<&oci_spec::runtime::Linux>) -> Result<Option<String>> {
```
utam0k: Rather than "extract," isn't this more of a conversion for procfs? What exactly are you extracting?

utam0k:
It might be better to stay in step with this PR.
https://github.com/youki-dev/youki/pull/3556/

crates/libcontainer/src/process/container_intermediate_process.rs
saku3: It follows runc’s implementation: even if we switch user namespaces, the owner of /proc/self/timens_offsets does not change, so the intermediate process cannot write to it and the write is done in the main process.

Ref: https://github.com/opencontainers/runc/commit/e6fb7fe51525f73ce848ef2c3d1612a531403b16

crates/libcontainer/src/utils.rs
saku3: Done

crates/libcontainer/src/process/container_main_process.rs
saku3: Done

crates/libcontainer/src/process/container_main_process.rs
saku3: Since validation is already performed, I removed the validation here.

crates/libcontainer/src/process/container_main_process.rs
saku3: The intention was to extract the `timeOffsets` settings from the OCI spec and convert them into a string that can be written to `/proc/<pid>/timens_offsets`.

I changed the function name to `build_timens_offsets`.

I thought this was concise and good, but we could also use one of the following:

* build_timens_offsets_payload
* format_time_offsets_for_procfs

saku3:
@utam0k 
Thank you for the review.

I’ve fixed it, so please take another look.

crates/libcontainer/src/process/channel.rs
```diff
@@ -73,6 +73,13 @@ impl MainSender {
         Ok(())
     }
 
+    pub fn time_offset_request(&mut self) -> Result<(), ChannelError> {
```
nayuta723, line 76: Could you add these unit tests?

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -274,6 +286,36 @@ fn setup_userns(
     Ok(())
 }
 
+fn setup_time_offsets(sender: &mut MainSender, receiver: &mut IntermediateReceiver) -> Result<()> {
+    tracing::debug!("requesting parent to write time namespace offsets");
+
+    prctl::set_dumpable(true).map_err(|e| {
```
nayuta723, line 292: Why do we need to set `dumpable`? It looks like `runc` doesn’t set it.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -381,6 +425,16 @@ fn move_network_devices_to_container(
             .collect::<Result<HashMap<String, Vec<crate::network::cidr::CidrAddress>>>>()?;
         init_sender.move_network_device(addrs_map)?;
     }
+    Ok(())
+}
+
+fn setup_time_offsets(offsets: &str, pid: Pid) -> Result<()> {
```
nayuta723: Is it necessary to make it a function?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -319,6 +331,38 @@ fn setup_mapping(config: &UserNamespaceConfig, pid: Pid) -> Result<()> {
     Ok(())
 }
 
+/// Builds the payload to write to the kernel's `/proc/<pid>/timens_offsets`
+/// from the spec's `timeOffsets`.
+fn build_timens_offsets(linux: Option<&Linux>) -> Option<String> {
```
nayuta723: Is it necessary to make it a function?
Instead of passing `Linux`, it might be more consistent with the existing implementation to pass `linux.time_offsets()` and have the function convert it to `Option<String>`.

crates/libcontainer/src/validator.rs
```diff
@@ -834,4 +856,79 @@ mod tests {
             ErrInvalidSpec::InvalidIntelRdtClosId
         ));
     }
+
+    #[test]
```
nayuta723: Please split this into two tests, one for the error case and one for the success case, to keep it consistent with the existing test style.

tests/contest/contest/src/tests/time_ns/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+use test_framework::{Test, TestGroup, TestResult, test_result};
+
+mod util;
+use crate::tests::time_ns::util::{TimeOffsets, create_time_offset};
+use crate::utils::test_utils::CreateOptions;
+use crate::utils::{
+    exec_container, start_container, test_inside_container, test_outside_container,
+};
+
+fn create_spec_for_set_times() -> Result<Spec> {
```
nayuta723: There are many `create_spec_for_xxx` functions, so it might be better to consolidate them into one. This would reduce the amount of code and centralize the shared logic, and it also seems to match the style used elsewhere in youki.

tests/contest/runtimetest/src/tests.rs
```diff
@@ -1412,3 +1412,60 @@ pub fn validate_net_devices(spec: &Spec) {
...
+        Err(e) => {
+            eprintln!("Failed to read /proc/self/timens_offsets: {}", e);
+            return;
+        }
+    };
+
+    for line in actual_offsets.lines() {
+        let parts: Vec<&str> = line.split_whitespace().collect();
+        if parts.len() != 3 {
+            continue;
```
nayuta723, line 1452: This won’t catch cases where the format is invalid.

nayuta723 (changes requested):
I left some comments. Please take a look, when you have a moment.

crates/libcontainer/src/process/channel.rs
saku3: Done

crates/libcontainer/src/process/container_intermediate_process.rs
saku3: After investigating this, I found that runc and youki handle the dumpable setting for the container process differently.

Please refer to the following issue for more details:
https://github.com/youki-dev/youki/issues/3649

With youki's current implementation, the process needs to be made non-dumpable at this point.

Since user namespaces are also involved, I think the dumpable-related issue should be handled in the issue above.

crates/libcontainer/src/process/container_main_process.rs
saku3: Deleted.

crates/libcontainer/src/process/container_main_process.rs
saku3: Done.

Since this contains a self-contained set of logic related to time namespaces, I think it would be reasonable to extract it into a separate function.

crates/libcontainer/src/validator.rs
saku3: Done.

tests/contest/contest/src/tests/time_ns/mod.rs
saku3: I consolidated the common logic.

tests/contest/runtimetest/src/tests.rs
saku3: Done.

I changed it so that it returns an error.

saku3:
@nayuta723 
Thank you for the review.

I’ve made the requested changes. Could you please take another look?

nayuta723 (approved):
LGTM, Thanks!

saku3:
@utam0k 
PTAL 🙏

utam0k (approved):

PR #3556, opened 2026-05-24
fix: validation differences between youki and runc

### LINKED ISSUE #3553: Investigate validation differences between youki and runc
## Description

There is a validation difference between youki and runc.

For example, if `hostname` is specified but the UTS namespace is not configured, youki currently succeeds and the hostname is effectively ignored.

runc explicitly returns an error in this case.

Investigate these kinds of validation differences.

## Steps to Reproduce

Create a config.json that specifies hostname but does not include the UTS namespace.

For example:

config.json
```

{

...

  "hostname": "youki",
  "linux": {
    "namespaces": [
      {
        "type": "pid"
      },
      {
        "type": "ipc"
      },
      {
        "type": "mount"
      },
      {
        "type": "cgroup"
      }
    ]
  }

...

}
```

Then run the container with youki and runc.

```
$ youki run container 
# Succeeds but hostname not applyed

$ runc run container
ERRO[0000] runc run failed: unable to set hostname without a private UTS namespace 
```

## Expected Behavior

youki should return an error, matching runc’s behavior, when hostname is specified without an UTS namespace.

tommady:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3553

## Additional Context

case 1: UTS Namespace (description one)
```bash
# before
./target/debug/youki run container
# Succeeds but hostname not applyed

# after
./target/debug/youki run container
ERROR youki: error in executing command: unable to set hostname without a private UTS namespace
```

case 2: User Namespace
```bash
{
...
# no type user but both "uidMappings" and "gidMappings" are populated
"linux": {
    "namespaces": [
      {
        "type": "pid"
      },
      {
        "type": "ipc"
      },
      {
        "type": "uts"
      },
      {
        "type": "mount"
      }
    ],
    "uidMappings": [
      {
        "containerID": 0,
        "hostID": 1000,
        "size": 1
      }
    ],
    "gidMappings": [
      {
        "containerID": 0,
        "hostID": 1000,
        "size": 1
      }
    ]
  }
...
}

./target/debug/youki run container
ERROR youki: error in executing command: user namespace mappings specified, but user namespace isn't enabled in the config
```

case 3: Mount Namespace
```bash
{
...
# no type mount
"linux": {
    "namespaces": [
      {
        "type": "pid"
      },
      {
        "type": "ipc"
      },
      {
        "type": "uts"
      },
      {
        "type": "cgroup"
      }
    ],
    "maskedPaths": [
      "/proc/acpi",
      ...
    ],
    "readonlyPaths": [
      "/proc/bus",
      ...
    ]
  }
...
}

./target/debug/youki run container
ERROR youki: error in executing command: unable to restrict sys entries without a private MNT namespace
```

case 4: Sysctl Namespaces
```bash
{
...
# no type ipc but `sysctl: fs.mqueue.msg_max` is populated
"linux": {
    "sysctl": {
      "fs.mqueue.msg_max": "100"
    },
    "namespaces": [
      {
        "type": "pid"
      },
      {
        "type": "uts"
      },
      {
        "type": "cgroup"
      }
    ],
    "maskedPaths": [
      "/proc/acpi",
      ...
    ],
    "readonlyPaths": [
      "/proc/bus",
      ...
    ]
  }
...
}

./target/debug/youki run container
ERROR youki: error in executing command: sysctl fs.mqueue.msg_max is not allowed in the hosts ipc namespace
```

case 5: Cgroups
youki uses the oci-spec crate, which automatically deserializes the config.json into a strictly-typed Rust struct.
```rust
#[derive(Default, PartialEq, Eq, Clone, Debug, Deserialize, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct Linux {
    // ... other fields ...
    #[serde(skip_serializing_if = "Option::is_none")]
    cgroups_path: Option<PathBuf>,
    // ...
}
```
Because of this rigid deserialization, youki simply accesses it as a single Option<PathBuf> like this:
```rust
let linux = self.spec.linux().as_ref().unwrap();
let cgroups_path = utils::get_cgroup_path(linux.cgroups_path(), &self.container_id);
```
Since the Rust parser strictly enforces that the input is just a single string path, it is completely impossible for youki to be passed an ambiguous state where "both Path and Name/Parent are set." The types don't even exist to allow it.

case 6: Memory Policy
youki performs this logic right before applying the policy in crates/libcontainer/src/process/memory_policy.rs during the validate_memory_policy function

### DIFF
--- a/crates/libcontainer/src/validator.rs
+++ b/crates/libcontainer/src/validator.rs
@@ -16,6 +16,7 @@ impl Validator {
         Self::validate_spec_for_sysctl(spec)?;
         Self::validate_spec_for_scheduler(spec)?;
         Self::validate_spec_for_io_priority(spec)?;
+        Self::validate_spec_for_intel_rdt(spec)?;
 
         Ok(())
     }
@@ -262,13 +263,30 @@ impl Validator {
 
         Ok(())
     }
+
+    fn validate_spec_for_intel_rdt(spec: &Spec) -> Result<(), ErrInvalidSpec> {
+        if let Some(linux) = spec.linux() {
+            if let Some(intel_rdt) = linux.intel_rdt() {
+                if let Some(clos_id) = intel_rdt.clos_id() {
+                    if clos_id == "."
+                        || clos_id == ".."
+                        || (clos_id.len() > 1 && clos_id.contains('/'))
+                    {
+                        return Err(ErrInvalidSpec::InvalidIntelRdtClosId);
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use oci_spec::runtime::{
         IOPriorityClass, LinuxBuilder, LinuxIOPriorityBuilder, LinuxIdMappingBuilder,
-        LinuxNamespaceBuilder, ProcessBuilder, SchedulerBuilder, SpecBuilder,
+        LinuxIntelRdtBuilder, LinuxNamespaceBuilder, ProcessBuilder, SchedulerBuilder, SpecBuilder,
     };
 
     use super::*;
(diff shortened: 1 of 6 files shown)


saku3:
@tommady cc: @utam0k @YamasouA 

I would like to merge the skeleton of this implementation first.

This PR wires `Validator::validate_spec(spec)?` into the init/tenant builder paths and adds the validation entry point. Some validation rules can be refined in follow-up PRs.

We are also trying to add other validation rules in separate PRs, such as #3550 and #3572.

Landing this structure first should help avoid duplicating changes around the validation entry point.

tommady:
> @tommady cc: @utam0k @YamasouA
> 
> I would like to merge the skeleton of this implementation first.
> 
> This PR wires `Validator::validate_spec(spec)?` into the init/tenant builder paths and adds the validation entry point. Some validation rules can be refined in follow-up PRs.
> 
> We are also trying to add other validation rules in separate PRs, such as #3550 and #3572.
> 
> Landing this structure first should help avoid duplicating changes around the validation entry point.

sure, I am totally ok~

crates/libcontainer/src/validator.rs
```diff
@@ -0,0 +1,830 @@
...
+    use oci_spec::runtime::{
+        IOPriorityClass, LinuxBuilder, LinuxIOPriorityBuilder, LinuxIdMappingBuilder,
+        LinuxIntelRdtBuilder, LinuxNamespaceBuilder, ProcessBuilder, SchedulerBuilder, SpecBuilder,
+    };
+
+    use super::*;
+
+    #[test]
+    fn test_validate_spec_for_uts_namespace() {
+        let sepc_no_uts_with_hostname = SpecBuilder::default()
```
saku3:
```suggestion
        let spec_no_uts_with_hostname = SpecBuilder::default() 
```

crates/libcontainer/src/validator.rs
```diff
@@ -0,0 +1,830 @@
...
+            Ok(host_stat.st_dev == target_stat.st_dev && host_stat.st_ino == target_stat.st_ino)
+        }
+
+        if let Some(linux) = spec.linux() {
+            if let Some(sysctls) = linux.sysctl() {
+                let has_ipc = linux
+                    .namespaces()
+                    .as_ref()
+                    .is_some_and(|ns| ns.iter().any(|n| n.typ() == LinuxNamespaceType::Ipc));
+                let is_host_net = match linux
```
saku3: I think the `is_host_net` / `is_host_net_ns` check can be evaluated lazily, only when `s.starts_with("net.")` is true, rather than performing it eagerly.

If we want to avoid calling it multiple times for multiple `net.*` sysctls, we could cache the result locally, for example by keeping something like `Option<Result<bool, ErrInvalidSpec>>` and computing it only the first time we encounter a `net.*` key.

saku3 (commented):
Thank you for the PR.

I left a few comments. Could you please take a look?

crates/libcontainer/src/validator.rs
tommady: sorry I did not understand this change suggestion.

tommady:
Hi @saku3,
sorry for the late reply (I did not receive github email lol)
please help review again, thanks!

crates/libcontainer/src/validator.rs
saku3: Please fix the typo :smile:

I think it should be `spec`, not `sepc`.

tommady: lol, ya you are right, ahah thanks dude!
I need to check my eye ahah

tommady: this just remind me a rust typos check
```bash
https://github.com/crate-ci/typos
```

maybe we could add this into CI check ahah.

anyway, thanks for your sharpppppp eye!!!

saku3 (approved):
Thanks!!

PR #3557, opened 2026-05-24
fix(libcgroups): add /run/systemd/private fallback for dbus system connection

### LINKED ISSUE #3524: Support /run/systemd/private fallback for systemd cgroup D-Bus connection
## Description

This appears to be a missing piece for runc compatibility.

`youki`'s native systemd D-Bus implementation does not appear to fall back to `/run/systemd/private` when the regular system D-Bus socket is unavailable.

This causes `SystemdCgroup = true` to fail in minimal systemd environments such as kind nodes, where systemd is running and `/run/systemd/private` exists, but `/run/dbus/system_bus_socket` does not exist.

## Steps to Reproduce 1

Start a kind cluster in the following environment and deploy a pod.

- Kubernetes environment: kind
- kind node image: `kindest/node:v1.34.0`
- youki build features:

```sh
cargo build -p youki --features "v1 v2 systemd"
```

containerd runtime configuration:

```
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.youki]
  runtime_type = "io.containerd.runc.v2"

[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.youki.options]
  BinaryName = "/usr/local/bin/youki"
  SystemdCgroup = true
```

## Steps to Reproduce 2

This is not intended to represent a real-world scenario, but it provides a simple way to reproduce the issue.

```
sudo unshare -m bash
mount -t tmpfs tmpfs /var/run/dbus

ls /var/run/dbus/  # No such file or directory

ls /run/systemd/private # /run/systemd/private

youki --systemd-cgroup run -b tutorial container # NG
runc --systemd-cgroup run -b tutorial container # OK
```

## Expected Behavior

youki should be able to use systemd cgroups in environments where the regular system D-Bus socket is unavailable but /run/systemd/private is available, similar to runc.

## Additional Context

As described above, the cause is that there is no fallback to `/run/systemd/private`.

* https://github.com/youki-dev/youki/blob/main/crates/libcgroups/src/systemd/dbus_native/dbus.rs#L90

runc performs this fallback as shown below.

* https://github.com/opencontainers/cgroups/blob/main/systemd/dbus.go#L69
* https://github.com/coreos/go-systemd/blob/main/dbus/dbus.go#L122
* https://github.com/coreos/go-systemd/blob/main/dbus/dbus.go#L165

nayuta723:
## Description

When systemd is PID 1 but `dbus-daemon` is not installed (e.g. kind nodes, minimal container images), nginx pods using `runtimeClassName: youki` with `SystemdCgroup = true` would hang for ~4 minutes before failing. There were two independent root causes:

### Root cause 1 — standard bus unavailable

The standard system bus socket `/run/dbus/system_bus_socket` does not exist on kind nodes. The previous `new_system()` unconditionally tried this path and failed immediately.

**Fix:** `new_system()` now falls back to `/run/systemd/private` when:
1. The standard system bus connection fails, AND
2. The process is root, AND
3. `/run/systemd/private` exists

This matches the behavior of [go-systemd's `NewSystemdConnectionContext()`](https://github.com/coreos/go-systemd/blob/main/dbus/dbus.go).

`new_direct()` is added to connect via the private socket, skipping the `Hello` method call that systemd does not implement on `/run/systemd/private`.

### Root cause 2 — indefinite recv() block under concurrent load

`/run/systemd/private` accepts connections but can be slow to respond when many containers start simultaneously (e.g. a Deployment with 5+ replicas). Without a timeout, `recv()` blocks indefinitely on the socket, causing youki to hang until the kubelet kills the pod sandbox creation.

**Fix:** `SO_RCVTIMEO=10s` is set on every dbus socket immediately after connect, covering both the AUTH handshake and all subsequent method calls. When `SO_RCVTIMEO` fires, `recv()` returns `EAGAIN`.

A unified retry loop in `Manager::new()` (7 retries, exponential backoff ~100ms×2ⁿ + jitter, ~15s max) catches `EAGAIN` from either the connection phase or the initial method calls (`control_cgroup_root()`), drops the socket, and reconnects from scratch.

> **Note:** runc uses `context.TODO()` for dbus calls via go-systemd, leaving the same indefinite-block risk open ([runc#3904](https://github.com/opencontainers/runc/issues/3904)). The `SO_RCVTIMEO` approach is youki's equivalent fix.

### Additional improvements

- **`read_one_message()`** replaces `receive_complete_response()`: the old implementation used a heuristic ("if received bytes < buffer size, we're done") that could truncate large messages. The new implementation reads the fixed 16-byte dbus header, extracts `body_len` and `header_array_len`, then reads the exact remainder.
- **`new_system()`** logs the standard bus error via `tracing::debug!` before falling back, improving diagnostics.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [x] CI/CD related changes

## Testing
- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3524

### DIFF
--- a/crates/libcgroups/src/systemd/dbus_native/dbus.rs
+++ b/crates/libcgroups/src/systemd/dbus_native/dbus.rs
@@ -15,6 +15,11 @@ use crate::systemd::dbus_native::serialize::{DbusSerialize, Structure, Variant};
 
 const REPLY_BUF_SIZE: usize = 128; // seems good enough tradeoff between extra size and repeated calls
 
+// systemd exposes a private socket for direct communication without a dbus daemon.
+// Used as fallback when the standard system bus socket is unavailable, matching runc behavior.
+// See: https://github.com/coreos/go-systemd/blob/main/dbus/dbus.go (NewSystemdConnectionContext)
+const SYSTEMD_PRIVATE_SOCKET: &str = "/run/systemd/private";
+
 /// NOTE that this is meant for a single-threaded use, and concurrent
 /// usage can cause errors, primarily because then the message received over
 /// socket can be out of order and we need to manager buffer and check with message counter
@@ -127,32 +132,38 @@ fn get_actual_uid() -> Result<u32> {
 }
 
 impl DbusConnection {
-    /// Open a new dbus connection to given address
-    /// authenticating as user with given uid
+    /// Open a new dbus connection to the given address, authenticate, and register with the daemon.
     pub fn new(addr: &str, uid: u32, system: bool) -> Result<Self> {
-        // Use ManuallyDrop to keep the socket open.
-        let socket = std::mem::ManuallyDrop::new(socket::socket(
-            socket::AddressFamily::Unix,
-            socket::SockType::Stream,
-            socket::SockFlag::empty(),
-            None,
-        )?);
-
-        let addr = socket::UnixAddr::new(addr)?;
-        socket::connect(socket.as_raw_fd(), &addr)?;
-        let mut dbus = Self {
-            socket: socket.as_raw_fd(),
-            msg_ctr: AtomicU32::new(0),
-            id: None,
-            system,
-        };
-        dbus.authenticate(uid)?;
-        Ok(dbus)
+        let mut conn = Self::open(addr, uid, system)?;
+        conn.register()?;
+        Ok(conn)
     }
 
     pub fn new_system() -> Result<Self> {
-        let addr = get_system_bus_address()?;
-        Self::new(&addr, 0, true)
+        // Try the standard system bus first (DBUS_SYSTEM_BUS_ADDRESS or well-known path).
+        // and_then chains address lookup + connection so either failure falls through.
+        let standard = get_system_bus_address().and_then(|addr| Self::new(&addr, 0, true));
+        if let Ok(conn) = standard {
+            return Ok(conn);
+        }
+
+        // Fallback: connect directly to systemd without a dbus daemon (root only).
+        // /run/systemd/private is created by systemd itself and exists even when
+        // dbus-daemon is not installed, e.g. in kind nodes.
+        // register() must be skipped because systemd does not implement org.freedesktop.DBus.Hello.
+        if nix::unistd::getuid().is_root() && std::path::Path::new(SYSTEMD_PRIVATE_SOCKET).exists()
+        {
+            tracing::debug!(
+                "system bus unavailable, falling back to systemd private socket {}",
+                SYSTEMD_PRIVATE_SOCKET
+            );
+            return Self::new_direct(SYSTEMD_PRIVATE_SOCKET, true);
+        }
+
+        Err(DbusError::BusAddressError(
+            "could not connect to system bus or systemd private socket".into(),
+        )
+        .into())
     }
 
     pub fn new_session() -> Result<Self> {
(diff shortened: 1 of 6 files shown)


saku3:
Thank you for PR.

I have one suggestion: since this behavior can be covered as a regression test by enabling `systemdCgroup` in the k8s test, would it be possible to add such a test?

saku3:
@nayuta723 
Sorry for the delayed response.

Is this PR ready?

nayuta723:
Not yet. The tests are failing in local, so I’m currently fixing the issues.

nayuta723:
@saku3 
I've fixed the issues. When you have a moment, could you please take a look?

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -190,31 +194,99 @@ impl Manager {
...
+            let client = match try_connect() {
+                Ok(c) => c,
+                Err(e) if !is_eagain(&e) => return Err(e),
+                Err(e) => {
+                    tracing::debug!(
+                        "dbus connect EAGAIN on attempt {}/{}, retrying after backoff",
+                        retry + 1,
+                        CONNECT_MAX_RETRIES
+                    );
+                    std::thread::sleep(backoff(retry));
```
saku3: I think we should skip `sleep(backoff(retry))` when `retry + 1 == CONNECT_MAX_RETRIES`.

At that point there is no subsequent retry, so sleeping only delays returning the final error.

The debug log should also avoid saying "retrying" on the last attempt.

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -190,31 +194,99 @@ impl Manager {
...
+                        destructured_path,
+                        client,
+                        fs_manager,
+                        delegation_boundary,
+                        cgroup_wait_timeout_duration,
+                    });
+                }
+                Err(e) if !is_eagain(&e) => return Err(e),
+                Err(e) => {
+                    tracing::debug!(
```
saku3: Could you also check the behavior when `retry + 1 == CONNECT_MAX_RETRIES` here?

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -190,31 +194,99 @@ impl Manager {
...
+                    std::thread::sleep(backoff(retry));
+                    last_err = Some(e);
+                }
+            }
+        }
+
+        Err(SystemdClientError::DBus(DbusError::ConnectionError(format!(
+            "dbus connection failed after {} retries: {}",
+            CONNECT_MAX_RETRIES,
+            last_err.as_ref().unwrap()
```
saku3: Since `CONNECT_MAX_RETRIES` is guaranteed to be at least 1, `last_err.as_ref().unwrap()` is probably safe with the current control flow. However, I would still prefer to avoid unwrap() here and handle the None case explicitly.

Would this be better?

```
let err_msg = last_err
    .as_ref()
    .map(|e| e.to_string())
    .unwrap_or_else(|| "unknown error".to_string());
```

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -127,43 +144,83 @@ fn get_actual_uid() -> Result<u32> {
...
-        let addr = socket::UnixAddr::new(addr)?;
-        socket::connect(socket.as_raw_fd(), &addr)?;
-        let mut dbus = Self {
+        let unix_addr = socket::UnixAddr::new(addr)?;
+        socket::connect(socket.as_raw_fd(), &unix_addr)?;
+
+        // Set SO_RCVTIMEO so that every recv() on this socket times out instead
+        // of blocking indefinitely.  See RECV_TIMEOUT_SECS for rationale.
+        // SAFETY: socket is a valid open fd for the lifetime of this call.
+        let borrowed = unsafe { BorrowedFd::borrow_raw(socket.as_raw_fd()) };
```
saku3: Could we use `socket.as_fd()` here instead?

crates/libcgroups/src/systemd/dbus_native/dbus.rs
```diff
@@ -127,43 +144,83 @@ fn get_actual_uid() -> Result<u32> {
...
+
+    pub fn new_session() -> Result<Self> {
+        let addr = get_session_bus_address()?;
+        let uid = get_actual_uid()?;
+        Self::new(&addr, uid, false)
+    }
+
+    /// Create a Unix socket and connect it to the given dbus address.
+    fn connect(addr: &str, system: bool) -> Result<Self> {
         let socket = std::mem::ManuallyDrop::new(socket::socket(
```
saku3, line 196: Note: It might be better to let DbusConnection own an OwnedFd instead of using ManuallyDrop + raw fd, so that fd ownership is explicit.

Since this is an issue in the existing implementation, it probably doesn’t need to be addressed in this PR. Just leaving it as a note.

saku3 (commented):
Thank you for the PR.

I left a few comments. Please take a look.

saku3 (approved):

PR #3558, opened 2026-05-24
fix(libcontainer): inherit init env for startContainer hooks

### LINKED ISSUE #3380: runc compatibility: startContainer hook should inherit process environment
Part of [runc compatibilities](https://github.com/youki-dev/youki/issues/3178) 

## Description

In runc, for StartContainer hooks, if no `env` is explicitly specified, the hook inherits the container init process’s environment.

See also:

* https://github.com/opencontainers/runc/blob/main/libcontainer/standard_init_linux.go#L211
* https://github.com/opencontainers/runc/blob/main/tests/integration/hooks.bats#L44

## Steps to Reproduce

Place the following script in the container’s rootfs.

```
cat >"rootfs/check-env.sh" <<-'EOF'
	#!/bin/sh -ue
	test $ONE = two
	test $FOO = bar
	echo $HOME 
EOF
chmod +x "rootfs/check-env.sh"
```

update config.json

```
  "args": [
    "/bin/true"
  ],
  "env": [
    "ONE=two", 
    "FOO=bar"
  ],
...
  "hooks": {
    "startContainer": [
      { "path": "/check-env.sh" }
    ]
  },
``` 

then run

In youki

```
$ youki run -b tutorial/ a
/check-env.sh: line 2: ONE: parameter not set
ERROR libcontainer::process::init::process: failed to run start container hooks err=NonZeroExitCode(2)
ERROR libcontainer::process::container_intermediate_process: failed to initialize container process: failed to run hooks
ERROR libcontainer::process::container_intermediate_process: failed sending error to main sender err=BaseChannelError(Nix(EBADF))
```

In runc (no error)

```
$  runc run -b tutorial/ a
```

## Expected behavior

StartContainer hooks without explicit env should be executed with the same environment as the container init process (runc-compatible behavior), so the hook can see ONE, FOO, and HOME.

immanuwell:
## Description
startContainer hooks now inherit the init process env when the hook does not set `env`, same as `runc`.

Before this, the hook was spawned with an empty env, so stuff like `ONE`, `FOO`, and `HOME` just vanished. Kinda sneaky bug, but very real.

## Type of Change
- [x] Test updates

## Testing
- [x] Added new unit tests
- [x] Ran existing test suite
`cargo check -p libcontainer --no-default-features --features v2`

`cargo test -p libcontainer --no-default-features --features v2 hooks::test::test_run_hook -- --nocapture`

`cargo test -p libcontainer --no-default-features --features v2 hooks::test::test_run_hook_timeout -- --nocapture`

Manual repro from the bug report:

```sh
cat >"rootfs/check-env.sh" <<'EOF2'
#!/bin/sh -ue
test "$ONE" = two
test "$FOO" = bar
echo "$HOME"
EOF2
chmod +x rootfs/check-env.sh
```

```json
"process": {
  "args": ["/bin/true"],
  "env": [
    "ONE=two",
    "FOO=bar"
  ]
},
"hooks": {
  "startContainer": [
    { "path": "/check-env.sh" }
  ]
}
```

Before this patch, `youki run -b tutorial a` blows up with `ONE: parameter not set`.
After this patch, the hook sees the init env like `runc` does.

## Related Issues
Fixes #3380
Related to #3178

## Additional Context
This keeps explicit hook `env` behavior as-is. The inheritance only kicks in when the hook does not define `env`.

### DIFF
--- a/crates/libcontainer/src/container/container_delete.rs
+++ b/crates/libcontainer/src/container/container_delete.rs
@@ -98,11 +98,17 @@ impl Container {
                     })?;
 
                     if let Some(hooks) = config.hooks.as_ref() {
-                        hooks::run_hooks(hooks.poststop().as_ref(), Some(&self.state), None, None)
-                            .map_err(|err| {
-                                tracing::error!(err = ?err, "failed to run post stop hooks");
-                                err
-                            })?;
+                        hooks::run_hooks(
+                            hooks.poststop().as_ref(),
+                            Some(&self.state),
+                            None,
+                            None,
+                            None,
+                        )
+                        .map_err(|err| {
+                            tracing::error!(err = ?err, "failed to run post stop hooks");
+                            err
+                        })?;
                     }
                 }
                 Err(err) => {
--- a/crates/libcontainer/src/container/container_start.rs
+++ b/crates/libcontainer/src/container/container_start.rs
@@ -59,6 +59,7 @@ impl Container {
                 Some(&self.state),
                 Some(&self.root),
                 None,
+                None,
             )
             .map_err(|err| {
                 tracing::error!("failed to run post start hooks: {}", err);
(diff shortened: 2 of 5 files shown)


nayuta723:
Thank you for the contribution. We are currently working on a similar change in [https://github.com/youki-dev/youki/pull/3470](https://github.com/youki-dev/youki/pull/3470). From what I can see, this appears to be mostly the same, but please let us know if there are any differences.

saku3:
Thank you for the PR.

Since this issue has already been resolved by #3470 , I’ll close this PR.

PR #3562, opened 2026-05-25
fix(libcontainer): pin target by fd when applying readonly_paths and masked_paths (TOCTOU)

### LINKED ISSUE #3561: Readonly mounts in a container may be skipped due to a race condition
## Summary

youki enforces the OCI `linux.readonlyPaths` and `linux.maskedPaths` settings in the init process by issuing mount syscalls that name the target as a path string. Each syscall re-resolves that path. A process inside the container can win a race and swap a component of the target path (via a symlink or a bind mount it controls) between the steps, so the mount that is supposed to enforce read-only or masking lands on a different target than intended. The result is that a path the runtime believes it has made read-only or masked is left writable or unmasked, or the operation fails (DoS).

This was reported privately and accepted as a bug under advisory GHSA-3q8j-pf29-vwv8, which is now closed/public, so it is being filed here in the open to track the fix.

To be precise about scope: this is an isolation and enforcement-correctness bug. The reproducer did not overwrite host files. The demonstrated impact is that `readonlyPaths` is not enforced (the target stays writable) plus a denial-of-service variant where the mount lands on the wrong target. It does not by itself grant arbitrary host file overwrite.

## Affected code

`crates/libcontainer/src/process/init/process.rs`

`readonly_path` (around lines 514-553 on the affected revision):

```rust
fn readonly_path(path: &Path, syscall: &dyn Syscall) -> Result<()> {
    if let Err(err) = syscall.mount(
        Some(path), path, None,
        MsFlags::MS_BIND | MsFlags::MS_REC,
        None,
    ) { ... }

    syscall.mount(
        Some(path), path, None,
        MsFlags::MS_NOSUID | MsFlags::MS_NODEV | MsFlags::MS_NOEXEC
            | MsFlags::MS_BIND | MsFlags::MS_REMOUNT | MsFlags::MS_RDONLY,
        None,
    )?;
    ...
}
```

`masked_paths`, directory branch (around lines 576-593 on the affected revision):

```rust
if path.is_dir() {
    syscall.mount(
        Some(Path::new("tmpfs")), path, Some("tmpfs"),
        MsFlags::MS_RDONLY, Some(label.as_str()),
    )?;
}
```

The masked-paths file branch already uses the fd-based `mount_from_fd` (the earlier hardening for that case), so the file branch is not affected. The directory branch and the whole of `readonly_path` are.

## Mechanism

`readonly_path` does a bind mount of `path` onto itself, then a separate remount of `path` to add `MS_RDONLY`. Both syscalls take `path` as a string, and the kernel resolves the string each time. The two resolutions are not guaranteed to point at the same vfsmount/inode. A peer process sharing the mount namespace (or reaching it through shared mount propagation, for example a bidirectional volume or a hostile createContainer hook) can, in the window between the bind and the remount, replace a component of `path` so that:

- the read-only remount applies to a mount the attacker controls, while the writable bind from the first syscall remains reachable, or
- the resolution fails and the call errors out (DoS).

The masked-paths directory branch has the same shape: the tmpfs is mounted onto `path` resolved as a string, so the same swap redirects or breaks the masking mount.

## Reproduction outline

1. Start a container whose config sets `linux.readonlyPaths` to include a directory target inside the container.
2. From a process that shares the mount namespace (or reaches the target through shared mount propagation), run a tight loop that binds and unbinds a writable mount over the target path, racing youki's two-syscall sequence.
3. After init finishes applying readonly paths, attempt to write under the target.

Observed result on the affected revision: in a meaningful fraction of trials the target ends up with a writable mount reachable on top of (or instead of) the read-only one, and a write that should return `EROFS` succeeds. The inter-syscall window between the bind and the remount is on the order of a few microseconds, which is wide enough to win repeatedly with a busy attacker loop.

The same race against the masked-paths directory branch redirects or breaks the masking tmpfs.

## Proposed fix

Stop re-resolving the target path between check and use. Resolve the target exactly once to a file descriptor and perform all subsequent steps against that fd, using the fd-based mount API:

- For `readonly_path`: `open_tree(target, OPEN_TREE_CLONE)` to pin the target's mount tree to an fd, `mount_setattr` on that fd to set `MOUNT_ATTR_RDONLY | NOSUID | NODEV | NOEXEC`, then `move_mount` it back over the destination pinned by an `O_PATH` fd on the parent plus the leaf name. This is the same approach crun uses in its `do_masked_or_readonly_path` handling.
- For the `masked_paths` directory branch: build the tmpfs with `fsopen` / `fsconfig` / `fsmount`, set `MOUNT_ATTR_RDONLY` via `mount_setattr` on the resulting fd, then `move_mount` over the fd-pinned destination.

youki already has all of these syscall wrappers (`open_tree`, `mount_setattr`, `move_mount`, `fsopen`, `fsconfig`, `fsmount`) and already uses the `open_tree -> mount_setattr -> move_mount` pattern in `crates/libcontainer/src/rootfs/mount.rs`, and already uses fd-pinning via `mount_from_fd` for the masked-file branch. The fix extends the same pattern to the two remaining path-resolved sites. A PR follows.

## Credit

Reported by Berkant Koc (me@berkoc.com), accepted under GHSA-3q8j-pf29-vwv8.

berkant-koc:
Fixes #3561 (advisory GHSA-3q8j-pf29-vwv8).

## Problem

`readonly_path` and the directory branch of `masked_paths` in
`crates/libcontainer/src/process/init/process.rs` applied the OCI
`readonlyPaths` / `maskedPaths` enforcement by issuing mount syscalls that name
the target as a path string. The kernel re-resolves that string on every
syscall, so the bind and the read-only remount (for readonly paths), or the
tmpfs mount (for masked directories), are not guaranteed to resolve to the same
target. A process sharing the mount namespace, or reaching it through shared
mount propagation, can swap a path component between the steps and redirect the
enforcing mount to a different target. The effect is that a path the runtime
believes is read-only or masked can stay writable or unmasked, or the operation
fails (DoS).

This is an enforcement-correctness / isolation bug. It does not by itself allow
overwriting host files; the impact is readonly-not-enforced plus a DoS variant.

The masked-paths file branch was already hardened to use the fd-based
`mount_from_fd`, so it is not affected. This PR extends the same fd-pinning idea
to the two remaining path-resolved sites.

## Fix

Resolve the target exactly once to a file descriptor and perform all subsequent
steps against that fd, so there is no path re-resolution window:

- New `Syscall::readonly_path_fd`: `open_tree(target, OPEN_TREE_CLONE |
  OPEN_TREE_CLOEXEC | AT_RECURSIVE)` pins the target's mount tree to an fd,
  `mount_setattr` on that fd sets `MOUNT_ATTR_RDONLY | NOSUID | NODEV | NOEXEC`,
  then `move_mount` re-attaches it over the destination pinned by an `O_PATH` fd
  on the parent directory plus the leaf name. `ENOENT` from `open_tree` is
  treated as "path does not exist" and ignored, preserving the previous
  behavior.
- New `Syscall::mount_masked_tmpfs`: builds the masking tmpfs with `fsopen` /
  `fsconfig` / `fsmount`, sets `MOUNT_ATTR_RDONLY` via `mount_setattr` on the
  resulting fd, then `move_mount` over the fd-pinned destination. The SELinux
  mount label is passed through `fsconfig` as the `context` option.
- `readonly_path` and the `masked_paths` directory branch now call these
  helpers instead of `Syscall::mount` with a path string.

This mirrors the `open_tree -> mount_setattr -> move_mount` pattern youki
already uses in `crates/libcontainer/src/rootfs/mount.rs`, and the fd-pinning
approach already used by `mount_from_fd` for the masked-file branch. It is the
same approach crun takes in its `do_masked_or_readonly_path` handling.

## Files changed

- `crates/libcontainer/src/syscall/syscall.rs` — two new trait methods.
- `crates/libcontainer/src/syscall/linux.rs` — Linux implementations plus two
  small helpers (`open_target_parent`, `leaf_name`); add `AsFd` to imports.
- `crates/libcontainer/src/syscall/test.rs` — mock recording for the two new
  methods (`ReadonlyPathFdArgs`, `MaskedTmpfsArgs`) and their getters.
- `crates/libcontainer/src/process/init/process.rs` — `readonly_path` and the
  masked-paths directory branch call the new fd-based helpers; unit tests
  updated to assert the fd-based calls.

## Testing

- `cargo check -p libcontainer`: clean.
- `cargo test -p libcontainer --no-default-features --features "systemd v2 v1"`
  for the `process::init::process` module: all tests pass (17 passed, 0 failed),
  including the updated `test_readonly_path`, `test_masked_path_is_dir_with_no_label`,
  `test_masked_path_is_dir_with_label`, `test_masked_path_does_not_exist`,
  `test_masked_path_with_unknown_error`, and the existing
  `test_masked_path_mounts_via_fd`.
- `libseccomp` is disabled for the test run only because the build host lacks
  `libseccomp.so` at link time; it does not affect the changed code paths.

Behavioral notes for reviewers:

- `readonly_path` previously layered an in-place recursive bind plus remount on
  the existing mount at the target. The fd-based path clones the target's mount
  tree, applies the read-only attributes, and moves the clone over the target.
  The end state for a consumer (a recursive read-only mount with
  nosuid/nodev/noexec) is the same; the underlying topology differs slightly. I
  checked the cases that differ on kernel 6.12 in an unprivileged user+mount
  namespace: a plain-directory target and a plain-file target both end up
  enforced read-only (writes return `EROFS`), and under shared/slave propagation
  the moved-in clone re-acquires the destination parent's propagation, matching
  what the old bind+remount produced, with read-only still enforced.
- The read-only attribute now applies recursively to submounts of the target
  (`AT_RECURSIVE` + `mount_setattr`), which is more restrictive than the old
  per-path remount and matches the `MS_REC` intent for readonlyPaths.
- This change relies on the same mount-API baseline that the normal container
  mount path in `rootfs/mount.rs` already requires (`mount_setattr` is Linux
  5.12; `open_tree`/`move_mount`/`fsopen` are 5.1-5.2), so it does not raise
  youki's effective kernel floor.

### DIFF
--- a/crates/libcontainer/src/process/init/process.rs
+++ b/crates/libcontainer/src/process/init/process.rs
@@ -508,45 +508,20 @@ fn sysctl(kernel_params: &HashMap<String, String>) -> Result<()> {
 }
 
 // make a read only path
-// The first time we bind mount, other flags are ignored,
-// so we need to mount it once and then remount it with the necessary flags specified.
-// https://man7.org/linux/man-pages/man2/mount.2.html
+//
+// The OCI readonlyPaths entries are enforced by making the target a read-only
+// recursive bind mount. The path must be resolved exactly once: a path-based
+// bind followed by a path-based remount re-resolves the target on each
+// syscall, which lets a process inside the container win a race and swap a
+// path component (symlink/bind) so the remount lands on a different target,
+// bypassing the read-only enforcement (TOCTOU). To close that window we pin
+// the target to an fd via open_tree(2) and apply the read-only attribute and
+// re-attachment by fd, the same way the masked-file branch uses mount_from_fd.
 fn readonly_path(path: &Path, syscall: &dyn Syscall) -> Result<()> {
-    if let Err(err) = syscall.mount(
-        Some(path),
-        path,
-        None,
-        MsFlags::MS_BIND | MsFlags::MS_REC,
-        None,
-    ) {
-        if let SyscallError::Nix(errno) = err {
-            // ignore error if path is not exist.
-            if matches!(errno, nix::errno::Errno::ENOENT) {
-                return Ok(());
-            }
-        }
-
+    syscall.readonly_path_fd(path).map_err(|err| {
         tracing::error!(?path, ?err, "failed to mount path as readonly");
-        return Err(InitProcessError::MountPathReadonly(err));
-    }
-
-    syscall
-        .mount(
-            Some(path),
-            path,
-            None,
-            MsFlags::MS_NOSUID
-                | MsFlags::MS_NODEV
-                | MsFlags::MS_NOEXEC
-                | MsFlags::MS_BIND
-                | MsFlags::MS_REMOUNT
-                | MsFlags::MS_RDONLY,
-            None,
-        )
-        .map_err(|err| {
-            tracing::error!(?path, ?err, "failed to remount path as readonly");
-            InitProcessError::MountPathReadonly(err)
-        })?;
+        InitProcessError::MountPathReadonly(err)
+    })?;
 
     tracing::debug!("readonly path {:?} mounted", path);
     Ok(())
@@ -574,23 +549,16 @@ fn masked_paths(
         }
 
         if path.is_dir() {
-            // Destination is a directory, mount a read-only tmpfs over the top of it.
-            let label = match mount_label {
-                Some(l) => format!("context=\"{l}\""),
-                None => "".to_string(),
-            };
-            syscall
-                .mount(
-                    Some(Path::new("tmpfs")),
-                    path,
-                    Some("tmpfs"),
-                    MsFlags::MS_RDONLY,
-                    Some(label.as_str()),
-                )
-                .map_err(|err| {
-                    tracing::error!(?path, ?err, "failed to mount path as masked using tempfs");
-                    InitProcessError::MountPathMasked(err)
-                })?;
+            // Destination is a directory, mount a read-only tmpfs over the top
+            // of it. The target is pinned by fd (fsmount + move_mount) instead
+            // of being re-resolved as a path string, so a racing process cannot
+            // swap a path component to redirect the masking tmpfs to a
+            // different target (TOCTOU).
+            let label = mount_label.as_deref();
+            syscall.mount_masked_tmpfs(path, label).map_err(|err| {
+                tracing::error!(?path, ?err, "failed to mount path as masked using tempfs");
+                InitProcessError::MountPathMasked(err)
+            })?;
         } else {
             // Destination is a file, bind mount /dev/null over the top of it.
             syscall.mount_from_fd(&dev_null_fd, path).map_err(|err| {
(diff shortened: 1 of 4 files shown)


utam0k:
This is not a correction to the advisory.

utam0k:
Please follow the template.

utam0k:
Are you compliant with the LF guidelines on generative AI?
https://www.linuxfoundation.org/legal/generative-ai

berkant-koc:
Withdrawing this PR. You're right that it's a bug rather than a bypass, and I overstated it, and it didn't follow your contribution process. I'll leave the fix approach here in case it's ever useful: pin the target by fd with open_tree/move_mount instead of re-resolving the path between check and use. Thanks for taking the time to look it over.

PR #3347, opened 2026-01-03
Try to join the cgroup of the init process of the landlord container when apply_cgroup for a tenant container fails due to a "Device or resource busy" error

logica0419:
## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes
- [ ] Other (please describe):

## Testing

- [x] Added new unit tests
- [ ] Added new integration tests
- [x] Ran existing test suite
- [x] Tested manually (please provide steps)
  - Followed the `Steps to Reproduce` of #3342 and got the expected result

## Related Issues

Fixes #3342

Issue #3342, opened 2025-12-30
[Bug]: Docker(moby) + youki cannot launch Dev Container with DinD


logica0419:
### Bug Description

I ran into an error when trying to create a Dev Container using the `ghcr.io/devcontainers/features/docker-in-docker:2` feature on youki.

### Steps to Reproduce

1. Set up [Docker on youki](https://docs.docker.com/engine/daemon/alternative-runtimes/#youki)
2. Create a Dev Container definition file with the Docker-in-Docker feature
   ```json
   {
     "name": "Docker in Docker",
     "image": "mcr.microsoft.com/devcontainers/base:bullseye",
     "features": {
       "ghcr.io/devcontainers/features/docker-in-docker:2": {
         "version": "latest",
         "enableNonRootDocker": "true",
         "moby": "true"
       }
     }
   }
   ```
3. Launch Dev Container

### Expectation

The Dev Container should launch successfully.

### System and Setup Info

I reproduced this bug in two environments.

--- **WSL Environment** ---

```console
$ uname -a
Linux {machine_name} 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

```console
$ youki info
youki version: 0.5.7
commit: 0.5.7-bd54457ba9de1629f21fe678687f433aec63e7c7
Kernel-Release    6.6.87.2-microsoft-standard-WSL2
Kernel-Version    #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025
Architecture      x86_64
Operating System  Ubuntu 24.04.3 LTS
Cores             24
Total Memory      15952
Cgroup setup      unified
Cgroup mounts
Capabilities
CAP_BPF           available
CAP_PERFMON       available
CAP_CHECKPOINT_RESTORE available
```

youki is running on Docker
```console
$ docker info | grep Runtime
 Runtimes: io.containerd.runc.v2 runc youki
 Default Runtime: youki
```

--- **Ubuntu Server Environment** ---

```console
$ uname -a
Linux {machine_name} 6.8.0-90-generic #91-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 18 14:14:30 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

```console
$ youki info
youki version: 0.5.7
[...]

PR #3563, opened 2026-05-27
feat(libcgroups): Support sub-cgroups for unified systemd cgroup manager

### LINKED ISSUE #3565: [FEATURE]: sub-cgroup is not supported for unified systemd (D-Bus) cgroup manager
### Background

As #3210 was merged, `youki` now supports the `exec` command in sub-cgroups.
However, the unified systemd (D-Bus) cgroup manager still does not support this.

The `AttachProcessesToUnit` D-Bus message can take a sub-cgroup path as its second argument, but it is currently left empty, as shown below:
https://github.com/youki-dev/youki/blob/e84ed5b69b7e522948bf29fe50e37bf964747b58/crates/libcgroups/src/systemd/manager.rs#L399-L400

### Feature Request

Support sub-cgroups for unified systemd (D-Bus) cgroup manager.

_No response_

### Proposed Solution

_No response_

### Considerations

_No response_

### Additional Context

refs https://github.com/youki-dev/youki/pull/3342
refs https://github.com/youki-dev/youki/pull/3347
dependent #3564

logica0419:
## Description

Adds support for sub-cgroups in the unified systemd cgroup manager.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

fixes #3565 
refs #3342
refs #3347
dependent #3564 

## Additional Context

https://github.com/youki-dev/youki/pull/3347#issuecomment-4557200770

### DIFF
--- a/crates/libcgroups/src/systemd/manager.rs
+++ b/crates/libcgroups/src/systemd/manager.rs
@@ -201,7 +201,8 @@ impl Manager {
         let mut destructured_path: CgroupsPath = cgroups_path.as_path().try_into()?;
         ensure_parent_unit(&mut destructured_path, use_system);
 
-        let sub_cgroup = Self::extract_sub_cgroup(&mut destructured_path.name);
+        let (name, sub_cgroup) = Self::extract_sub_cgroup(&destructured_path.name);
+        destructured_path.name = name;
 
         // EAGAIN from a recv() that timed out via SO_RCVTIMEO (set in dbus::connect()).
         let is_eagain = |e: &SystemdManagerError| {
@@ -316,16 +317,20 @@ impl Manager {
     }
 
     // If the provided unit `name` contains a sub-cgroup suffix, split it off and
-    // return that suffix as a separate cgroup path.
-    // The original `name` is mutated in-place to keep only the base unit name.
+    // return it as a separate cgroup path along with the base unit name.
     //
     // Example: "{id}/sub/init" becomes:
-    //   - name: "{id}"
-    //   - returned sub-cgroup: "/sub/init"
-    fn extract_sub_cgroup(name: &mut String) -> String {
-        name.find("/")
-            .map(|separator_index| name.split_off(separator_index))
-            .unwrap_or_default()
+    //   - base name: "{id}"
+    //   - sub-cgroup: "/sub/init"
+    fn extract_sub_cgroup(name: &str) -> (String, String) {
+        name.find('/')
+            .map(|separator_index| {
+                (
+                    name[..separator_index].to_owned(),
+                    name[separator_index..].to_owned(),
+                )
+            })
+            .unwrap_or_else(|| (name.to_owned(), String::new()))
     }
 
     /// get_unit_name returns the unit (scope) name from the path provided by the user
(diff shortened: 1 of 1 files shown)


logica0419:
Although I worked on this as part of #3564, I’ve opened a separate issue for this PR since the enhancement is necessary regardless of the approach we take.

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -260,7 +262,8 @@ impl Manager {
 
             // Step 2 — initial method calls to discover the cgroup path.
             // On EAGAIN, client is dropped here and the next iteration reconnects.
-            match Self::construct_cgroups_path(&destructured_path, &client) {
+            let sub_cgroup = Self::extract_sub_cgroup(&mut destructured_path.name);
```
utam0k: Passing a mutable value will cause issues during retries. It should be static. Or it'd be better to move it out of the loop.

utam0k (changes requested):

crates/libcgroups/src/systemd/manager.rs
logica0419: Thx!
I moved the extraction out of the loop,

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -310,6 +315,19 @@ impl Manager {
...
     }
 
+    // If the provided unit `name` contains a sub-cgroup suffix, split it off and
+    // return that suffix as a separate cgroup path.
+    // The original `name` is mutated in-place to keep only the base unit name.
+    //
+    // Example: "{id}/sub/init" becomes:
+    //   - name: "{id}"
+    //   - returned sub-cgroup: "/sub/init"
+    fn extract_sub_cgroup(name: &mut String) -> String {
```
utam0k: Is `&mut` still needed?

logica0419: I guess yes, since the "original `name` is mutated in-place to keep only the base unit name," as I wrote in the comment.
Is it unnecessary in this case?

utam0k: It is surprising that there are side effects. I'd like to avoid it as much as possible. Is it possible to receive it as a `str`?

logica0419: I got it. I agree that it's reasonable to avoid side effects.
You mean taking the name as a `str` and returning two strings (the name and the sub-cgroup), right?

utam0k: > You mean taking the name as a str and returning two strings (the name and the sub-cgroup), right?

No, I'm talking about types.
https://doc.rust-lang.org/std/primitive.str.html

logica0419: I’ve fixed the argument type based on your suggestion.
Could you confirm that I understood your suggestion correctly?
https://github.com/youki-dev/youki/commit/0598fa65b7375d0c46fbcf07bfeb954e1f814c89

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -310,6 +316,23 @@ impl Manager {
...
+    fn extract_sub_cgroup(name: &str) -> (String, String) {
+        name.find('/')
+            .map(|separator_index| {
+                (
+                    name[..separator_index].to_owned(),
+                    name[separator_index..].to_owned(),
+                )
+            })
+            .unwrap_or_else(|| (name.to_owned(), String::new()))
+    }
```
utam0k: Will this be sufficient?
```suggestion
fn split_sub_cgroup(name: &str) -> (&str, &str) {
    match name.find('/') {
        Some(i) => name.split_at(i),
        None => (name, ""),
    }
}
```

logica0419: Yes! The suggestion is pretty neat!
Thanks a lot.

utam0k (approved):
Kudos

PR #3568, opened 2026-05-28
refactor(devcontainer): Devcontainer files enhancement

### LINKED ISSUE #3567: [REFACTOR]: enhance Devcontainer definition files
Currently, youki’s Devcontainer configuration has several slight issues:

* The Docker-in-Docker feature version is pinned to `2` instead of a specific patch version
  * Also, a newer major version is already available
* After creating the container, running `just youki-*` still requires `rustup toolchain install`.
  * It would be better to run this during the Devcontainer setup phase
* With the default Docker setup, we cannot inspect youki’s debug logs
  * This is because systemd is not available inside the container
* The `Dockerfile` has some hadolint warnings

Though these are not critical to youki’s functionality, improving them would enhance the developer experience, especially on macOS.

logica0419:
## Description

Refactored and improved the Devcontainer configuration:

* Upgraded the `docker-in-docker` feature to 3.0.1
* Formatted files with `oxfmt` and fixed hadolint warnings
* Reorganized commands in the `Dockerfile`
* Enabled youki debug logging when testing with Docker
* Added a `postStartCommand` to install the declared Rust toolchain version and build youki

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3567

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,17 +1,6 @@
 ARG VARIANT="bullseye"
 FROM mcr.microsoft.com/vscode/devcontainers/rust:1-${VARIANT}
 
-# Install docker with youki
-COPY <<EOF /etc/docker/daemon.json
-  {
-    "runtimes": {
-      "youki": {
-        "path": "/workspaces/youki/youki"
-      }
-    }
-  }
-EOF
-
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 RUN <<'EOF'
@@ -44,4 +33,15 @@ rustup component add clippy
 # Systemd is not available in this Dev Container, but Podman `event_logger` defaults to journald.
 # Switch the logger to file so container startup does not fail.
 sed -i 's/^# events_logger = "journald"/events_logger = "file"/' /usr/share/containers/containers.conf
+EOF
+
+# Install docker with youki
+COPY <<'EOF' /etc/docker/daemon.json
+  {
+    "runtimes": {
+      "youki": {
+        "path": "/workspaces/youki/youki"
+      }
+    }
+  }
 EOF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -35,13 +35,33 @@ rustup component add clippy
 sed -i 's/^# events_logger = "journald"/events_logger = "file"/' /usr/share/containers/containers.conf
 EOF
 
-# Install docker with youki
+# Configure Docker to use youki as the default runtime.
+# Docker invokes the wrapper at /usr/local/bin/youki which forwards to /workspaces/youki/youki.
 COPY <<'EOF' /etc/docker/daemon.json
-  {
-    "runtimes": {
-      "youki": {
-        "path": "/workspaces/youki/youki"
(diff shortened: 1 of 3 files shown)


logica0419:
Concerns:
* I’m not sure whether building the youki binary on startup is appropriate. Do you think it’s reasonable?
* I assume some extensions in `devcontainer.json` (`mutantdino.resourcemonitor`, `fill-labs.dependi`) are not necessarily needed in **every** environment. Would it be a problem to remove them?
  * `LLDB`, `rust-analyzer`, and `even-better-toml` seem reasonable to keep, though
  * Ultimately, I feel we may not need to specify any extensions here, since VS Code can copy all local extensions into the container with a single action anyway

saku3:
@youki-dev/youki-reviewers @bells17 
I’d like to hear from people who actually use the devcontainer.

utam0k:
> @youki-dev/youki-reviewers @bells17
> I’d like to hear from people who actually use the devcontainer.

+1

.devcontainer/Dockerfile
```diff
@@ -1,54 +1,67 @@
...
+  "runtimes": {
+    "youki": {
+      "path": "/usr/local/bin/youki",
+      "runtimeArgs": ["--debug"]
+    }
+  }
+}
+EOF
+
+# Wrapper script that forces logs to /var/log/youki.log.
```
bells17, line 52: Could you remind me why we explicitly set `--log=/var/log/youki.log` here?

.devcontainer/Dockerfile
```diff
@@ -1,54 +1,67 @@
...
+
+# Configure Docker to use youki as the default runtime.
+# Docker invokes the wrapper at /usr/local/bin/youki which forwards to /workspaces/youki/youki.
+COPY <<'EOF' /etc/docker/daemon.json
+{
+  "default-runtime": "youki",
+  "runtimes": {
+    "youki": {
+      "path": "/usr/local/bin/youki",
+      "runtimeArgs": ["--debug"]
```
bells17, line 46: Do we need to enable `--debug` all the time? Would it be sufficient to enable it only when we need to inspect the debug logs?

bells17 (commented):
Sorry for the delayed review.

> I’m not sure whether building the youki binary on startup is appropriate. Do you think it’s reasonable?

Since developers will usually build youki after making local changes, I personally don’t think it needs to be included in postStartCommand. In what situations would it be useful to have the youki binary built automatically when the Dev Container starts?

Also, while all of these changes are related to updating the Dev Container, I feel that several different types of changes are mixed into this PR.

For example, it might be easier to review and discuss the changes if items such as upgrading the docker-in-docker feature to 3.0.1 were split into separate PRs based on their purpose.

bells17:
> I assume some extensions in `devcontainer.json` (`mutantdino.resourcemonitor`, `fill-labs.dependi`) are not necessarily needed in every environment. Would it be a problem to remove them?

Yes, I think it would be fine to remove them for now.

PR #3572, opened 2026-05-30
[ID-Mapped Mount] add idmapped mount spec validation

YamasouA:
## Description

This PR only recognizes and validates idmapped mount configuration from the OCI spec.
It does not enable the actual idmapped mount setup flow yet.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Related to #2307 

## Additional Context

The following PR has too many changes, so it will be split and submitted for review.
https://github.com/youki-dev/youki/pull/3384


crates/libcontainer/src/container/mount_validation.rs
```diff
@@ -0,0 +1,317 @@
...
+        namespaces
+            .iter()
+            .any(|ns| ns.typ() == LinuxNamespaceType::User)
+    });
+    if !has_userns {
+        return false;
+    }
+
+    has_non_empty_mappings(linux.uid_mappings().as_deref())
+        && has_non_empty_mappings(linux.gid_mappings().as_deref())
```
nayuta723:
```suggestion
    let Some(namespaces) = linux.namespaces().as_deref() else { return false };
    match namespaces.iter().find(|ns| ns.typ() == LinuxNamespaceType::User) {
        None => false,
        Some(ns) if ns.path().is_some() => true,
        Some(_) => {
          has_non_empty_mappings(linux.uid_mappings().as_deref().unwrap_or(&[]))
              && has_non_empty_mappings(linux.gid_mappings().as_deref().unwrap_or(&[]))
        }
    }
```

crates/libcontainer/src/container/mount_validation.rs
```diff
@@ -0,0 +1,317 @@
...
+            // idmap/ridmap can use the container user namespace when mount-specific mappings are absent.
+            (None, None) => false,
+            _ => {
+                tracing::error!(
+                    destination = ?mount.destination(),
+                    "mount uid/gid mappings must be non-empty and specified together"
+                );
+                return Err(ErrInvalidSpec::MountIdmapInvalidConfig);
+            }
+        };
```
nayuta723: How do you feel about creating a function like this in terms of separation of responsibilities?
```
fn has_valid_mount_mappings(mount: &SpecMount) -> Result<bool, ErrInvalidSpec> {
    match (mount.uid_mappings().as_deref(), mount.gid_mappings().as_deref()) {
        (Some(uid), Some(gid)) if has_non_empty_mappings(uid) && has_non_empty_mappings(gid) => Ok(true),
        (None, None) => Ok(false),
        _ => Err(ErrInvalidSpec::MountIdmapInvalidConfig),
    }
}
```

crates/libcontainer/src/rootfs/utils.rs
```diff
@@ -18,6 +21,12 @@ pub struct MountOptionConfig {
 
     /// RecAttr represents mount properties to be applied recursively.
     pub rec_attr: Option<linux::MountAttr>,
+
+    /// Indicates whether ID-mapped mount semantics are requested.
+    pub apply_idmap: bool,
+
+    /// Indicates whether ID-mapping should be applied recursively.
+    pub apply_idmap_recursively: bool,
```
nayuta723: I’m not sure how to validate whether these flags are appropriate. Do you think this set of flags is sufficient?

Since the main focus of this change is the validation logic, I was wondering if this might be outside the scope of this PR. Do you have any thoughts on that?

crates/libcontainer/src/container/mount_validation.rs
```diff
@@ -0,0 +1,317 @@
...
+    #[test]
+    fn validate_idmapped_mounts_requires_both_mappings() {
+        let mount = serde_json::from_value(serde_json::json!({
+            "destination": "/mnt",
+            "type": "bind",
+            "source": "/src",
+            "options": ["bind", "idmap"],
+            "uidMappings": [{"containerID": 0, "hostID": 0, "size": 1}]
+        }))
+        .unwrap();
```
nayuta723, line 275:
```suggestion
      let mount = base_mount()
          .options(vec!["bind".to_string(), "idmap".to_string()])
          .uid_mappings(vec![make_mapping()])
          .build()
          .unwrap();
```

crates/libcontainer/src/container/mount_validation.rs
```diff
@@ -0,0 +1,317 @@
...
+        let res = validate_idmapped_mounts(&[mount], None);
+        assert!(matches!(res, Err(ErrInvalidSpec::MountIdmapInvalidConfig)));
+    }
+
```
nayuta723, line 283: It'd be great to add coverage for the following cases as well:
* `is_bind` path when `type=bind` is specified without any bind options
* multiple mounts
* successful ridmap resolution
* non-empty `uid` with an empty `gid` array

nayuta723 (commented):
Thanks for the PR! I've left some comments.

crates/libcontainer/src/container/mount_validation.rs
YamasouA: @nayuta723 
Thanks for the suggestion! Unfortunately, MountBuilder itself validates that uid_mappings and gid_mappings must be specified together.

crates/libcontainer/src/rootfs/utils.rs
YamasouA: @nayuta723 
You're right. We shouldn't have included that in this update.

crates/libcontainer/src/container/mount_validation.rs
nayuta723: Ah, I see. I overlooked that.

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -1002,6 +1006,16 @@ impl Mount {
     }
 }
 
+fn requests_idmapped_mount(mount: &SpecMount) -> bool {
```
nayuta723: Could you add unit tests for this change?

crates/libcontainer/src/container/mount_validation.rs
```diff
@@ -0,0 +1,399 @@
...
+            tracing::error!(
+                destination = ?mount.destination(),
+                "mount uid/gid mappings must be non-empty and specified together"
+            );
+            Err(ErrInvalidSpec::MountIdmapInvalidConfig)
+        }
+    }
+}
+
+pub(crate) fn validate_idmapped_mounts(
```
nayuta723, line 53: Comparing this with runc, it seems that rootless support is not adequately tested. Please add test coverage for the rootless case.

crates/libcontainer/src/rootfs/mount.rs
nayuta723: Also, to catch invalid configurations earlier, could we move the `requests_idmapped_mount` validation into the validation phase?

nayuta723 (commented):
I've left some nits. Please take a look.

crates/libcontainer/src/rootfs/mount.rs
YamasouA: @nayuta723 
Thank you for your review.
Moved the validation into the validation phase as suggested.
As a result, many tests now assert `MountIdmapUnsupported` rather than `Ok`, since the unsupported guard fires after the structural checks.
These assertions will be updated to the appropriate errors when the `MountIdmapUnsupported` guard is removed in a later PR.

[a894044](https://github.com/youki-dev/youki/pull/3572/commits/a894044937e494efa99c9e65db49b3bb7ab3d898)

crates/libcontainer/src/container/mount_validation.rs
YamasouA: I miss this validation...
Thank you!
[07b206d](https://github.com/youki-dev/youki/pull/3572/commits/07b206d1c773a933c838aba5e02c44c84dea037d)

nayuta723 (approved):
LGTM, Thanks!

@youki-dev/youki 
Could you take a look when you get a chance?

crates/libcontainer/src/container/mount_validation.rs
```diff
@@ -0,0 +1,486 @@
...
+            );
+            return Err(ErrInvalidSpec::MountIdmapMissingMappings);
+        }
+
+        if requests_idmapped_mount && !is_bind {
+            tracing::error!(
+                destination = ?mount.destination(),
+                "mount specifies idmap option for non-bind mount"
+            );
+            return Err(ErrInvalidSpec::MountIdmapNonBind);
```
utam0k, line 78: MountIdmapNonBind should be returned before MountIdmapMissingMappings because it has higher priority. WDYT?

YamasouA: @utam0k 
You are right. I cleaned up the validation order in a few places, so could you please take another look?

utam0k (approved):
💯

PR #3573, opened 2026-05-31
fix: reject hostname without a private UTS namespace

### LINKED ISSUE #3553: Investigate validation differences between youki and runc

immanuwell:
## Description
Reject a non-empty `hostname` unless the spec creates a private UTS namespace. Before this, youki would accept the config and then quietly ignore the hostname, which is kinda misleading. Empty hostname still stays valid, same as `runc`.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing
- [x] Added new unit tests
- [ ] Ran existing test suite

Checked with `cargo check -p libcontainer` and `cargo check -p libcontainer --tests`.
Local `cargo test` is blocked here because `libseccomp` is missing, rip.

## Related Issues
Fixes #3553

## Additional Context
Repro:
1. Set `"hostname": "youki"` in `config.json`, but do not add a private `uts` namespace under `linux.namespaces`.
2. Run `youki run ctr`.
3. Before this patch it succeeds, but the hostname is ignored.
4. After this patch it fails with `unable to set hostname without a private UTS namespace`.

Related: #3556 is doing a broader validator pass too.

### DIFF
--- a/crates/libcontainer/src/container/init_builder.rs
+++ b/crates/libcontainer/src/container/init_builder.rs
@@ -228,6 +228,7 @@ impl InitContainerBuilder {
 
         let syscall = create_syscall();
         utils::validate_spec_for_new_user_ns(spec, &*syscall)?;
+        utils::validate_spec_for_hostname(spec)?;
         utils::validate_spec_for_net_devices(spec, &*syscall)
             .map_err(LibcontainerError::NetDevicesError)?;
 
--- a/crates/libcontainer/src/container/tenant_builder.rs
+++ b/crates/libcontainer/src/container/tenant_builder.rs
@@ -444,6 +444,7 @@ impl TenantContainerBuilder {
 
         let syscall = create_syscall();
         utils::validate_spec_for_new_user_ns(spec, &*syscall)?;
+        utils::validate_spec_for_hostname(spec)?;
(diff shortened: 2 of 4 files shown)


saku3:
Closing this as a duplicate.

https://github.com/youki-dev/youki/pull/3556

PR #3577, opened 2026-06-03
[ID-Mapped Mount] handle init sync messages in event loop

### LINKED ISSUE #2307: Support idmapped mount

YamasouA:
## Description

`container_main_process` currently processes init-to-main setup messages in a fixed order. For idmapped mount support, however, mount fd requests can arrive at any point during rootfs setup and may occur multiple times, making the fixed-order model unworkable. This PR changes the implementation to a message-driven event loop.
Removing the now-unnecessary message handling is intentionally left out to keep this PR's diff small; it will be addressed in a separate PR.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #2307


YamasouA:
@nayuta723 @utam0k 
This is a draft implementation. Could you share your thoughts on the overall approach?
<details>
# Idmapped Mount Sync Design Notes

## Background

Idmapped mount support needs a way for the container init process to ask the
main process to create a mount fd and send it back. This request can happen
while the init process is setting up the rootfs, and it can happen more than
once.

A sample implementation is available at:

- https://github.com/youki-dev/youki/pull/3384

The existing main/init synchronization model is mostly a fixed-order handshake:

- wait for hook request, if hooks are configured
- wait for network setup ready, if network devices are configured
- wait for seccomp notify fd, if seccomp notify is configured
- wait for init ready

Those messages are effectively "required at most once" setup messages. A mount
fd request is different: it is a request/response service call that may happen
multiple times.

## What runc Does

runc handles mount fd requests through the same parent-side sync loop that also
handles other init-to-parent setup messages.

In `libcontainer/rootfs_linux.go`, `prepareRootfs` iterates over configured
mounts and decides per mount whether it needs a mount fd from the parent:

```go
wantSourceFile := m.IsIDMapped()
if m.IsBind() && !config.RootlessEUID {
    if _, err := os.Stat(m.Source); err != nil {
        wantSourceFile = true
    }
}
if wantSourceFile {
    writeSyncArg(pipe, procMountPlease, m)
    readSyncFull(pipe, procMountFd)
}
```

For idmapped mounts, a mount fd request is always needed. runc also uses this
path for some non-idmapped bind mounts when the init side cannot resolve the
source.

In `libcontainer/process_linux.go`, the parent side uses `parseSync` and handles
`procMountPlease` as one of the possible sync messages:

```go
case procMountPlease:
    ...
    mnt, err := mountRequest(m)
    ...
    doWriteSync(... procMountFd ...)
case procSeccomp:
    ...
case procHooks:
    ...
case procReady:
    ...
```

So runc does not pre-count expected mount fd requests. It processes mount fd
requests when the init process sends them.

## What the Current youki Prototype Does

In the `ft/id-mapped-mount` prototype, rootfs setup passes mount channels into
the mount code. `setup_mount` parses mount options and sends a mount fd request
only when idmapped mount semantics are requested:

```rust
let mount_option_config = parse_mount(mount)?;
if mount_option_config.apply_idmap {
    ...
    self.mount_idmapped(...)?;
    return Ok(());
}
```

Inside `mount_idmapped`, the request is sent after several rootfs-side
preparation steps:

```rust
let source = mount.source().as_ref().ok_or(MountError::NoSource)?;
let src = self.prepare_bind_source(source, &root, container_dest)?;

comm.main.request_mount_fd(msg)?;
let mount_fd = comm.init.wait_for_mount_fd_reply()?;
```

This means request emission is decided during rootfs setup, after the per-mount
preparation has reached the request point.

## Design Options

### Option A: Full Main-Side Event Loop

Change the main process to receive init-to-main setup messages in a loop until
the init process reports ready. Existing hook, network, seccomp, and ready
messages are handled in the same loop. Mount fd requests can be added as another
message arm.

The receiver/channel API can mostly stay as-is. The loop lives in the main
process control flow.

Pros:

- Matches runc's request-driven sync model.
- Handles multiple mount fd requests naturally.
- Does not require pre-counting requests from config.
- Error messages from init can be handled while waiting for any request.
- Future init-to-main request types can be added in one place.

Cons:

- Changes the main/init synchronization model from fixed-order waits to a loop.
- In an incremental implementation, existing `MainReceiver::wait_for_xxx` APIs
  may remain for tests or compatibility even if the production main-process path
  moves to the generic receive loop.
- Validation previously implicit in `wait_for_xxx` must be represented in the
  loop.
- The change is design-level, not just a small local refactor.

#### Future Cleanup: MainReceiver/API Redesign

If Option A is adopted, `MainReceiver` can be cleaned up later around typed
request receiving instead of fixed `wait_for_xxx` methods. For example, it
could return a domain-specific request enum containing hook, network, seccomp,
mount fd, ready, and error messages.

This is not an alternative to the main-side loop. It is a later API cleanup that
can be done after the request-driven synchronization model is in place.

Potential benefits:

- Cleaner long-term API.
- Moves fd ownership and message decoding into the receiver abstraction.
- Avoids mixing old sequential wait APIs with the new event-loop style.
- Makes future request types easier to model.

Why it should be separate:

- It requires broader channel API and test updates.
- It is not required to make idmapped mount support work.
- It could distract from the mount fd feature itself if done in the same change.

### Option B: Rootfs-Phase Scoped Loop

Limit mount fd requests to the rootfs setup phase. During that phase, the parent
handles mount fd requests, errors, and the next phase transition. After rootfs
setup is complete, the code returns to the existing fixed-order handshake for
hooks, network, seccomp, and ready.

#### Can We Count Requests from the Config?

Yes, for the no-error path.

For the current prototype, the count is:

```text
number of spec.mounts where parse_mount(mount).apply_idmap == true
```

However, using this as the parent-side synchronization contract is risky:

- The init side may fail before sending a request for a mount.
- The parent side would need to keep receiving error messages while waiting for
  the expected number of mount fd requests. Without that, the parent can wait
  for a request that init will never send.
- The request-counting logic would duplicate rootfs mount decision logic.
- Future changes to mount handling could make the count drift from actual
  request behavior.
- runc avoids this by using a request-driven sync loop.

So config-based counting is possible and exact for the planned no-error path,
but it is not a robust synchronization boundary by itself.

Pros:

- Keeps the existing fixed-order handshake for most setup messages.
- Makes the mount fd request phase explicit.
- Could be smaller than a full event loop if the phase boundary is clean.

Cons:

- The parent still needs a loop during the rootfs phase.
- If the phase is count-driven, request counting duplicates mount logic and is
  fragile on error paths. Without also handling init-side errors, the parent can
  wait for a request that will never be sent.
- If the phase is sentinel-driven, the protocol needs an explicit rootfs-phase
  boundary. This does not require buffering in the current ordering, where
  rootfs setup happens before hooks, network setup, and seccomp, but it is less
  flexible if future request types interleave across phases.
- Future multi-request sync messages would require more phase-specific logic.

## My Recommendation

The most robust direction appears to be Option A: a request-driven main-side
loop for setup messages, while keeping `MainReceiver` API redesign out of the
initial change.

This keeps the immediate change focused on the main process synchronization
model and leaves channel API cleanup for a later, separate change. It also
matches runc's approach more closely than a count-driven design.

If maintainers prefer to keep the fixed-order handshake as much as possible,
Option B is the main alternative. In that case, the rootfs phase should still be
request-driven rather than purely count-driven, because init-side errors before
the request point must be handled without deadlocking the parent.

</details>

nayuta723:
@YamasouA 
Thanks! 
I think Option A looks good to me as well.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +273,190 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+        container_args,
+        init_pid,
+        init_sender,
+        |hooks, state, pid| {
+            hooks::run_hooks(hooks, Some(state), None, Some(pid), None).map_err(Into::into)
+        },
+    )
+}
+
+fn handle_hook_request_with(
```
YamasouA: Each *_with function takes its side-effecting dependency (run hooks / send to the seccomp listener / move the network device) as an injected parameter so unit tests can pass a mock and cover the success and error paths without the real side effect. The plain handle_* functions just delegate with the real implementation.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +273,187 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
     Ok((init_pid, need_to_clean_up_intel_rdt_subdirectory))
 }
 
+const EXPECTED_INIT_MESSAGE: &str = "InitReady or a pending init setup request";
+
+/// Init-side setup requests that must complete before `InitReady`.
+struct PendingInitRequests<'a> {
```
nayuta723: Since `PendingInitRequests` is really just tracking whether each expected request has been processed, I think a set of booleans is sufficient here.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -158,84 +165,56 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+                    return Err(unexpected_init_message(
+                        "pending init setup request",
+                        Message::InitReady,
+                    ));
+                }
+                break;
+            }
+            Message::HookRequest => {
+                let hooks = pending.hooks.take().ok_or_else(|| {
+                    unexpected_init_message(EXPECTED_INIT_MESSAGE, Message::HookRequest)
```
nayuta723:
```suggestion
                    ProcessError::Channel(channel::ChannelError::unexpected("xxx", Message::HookRequest))
```

ref: https://github.com/youki-dev/youki/pull/3577/changes#r3460454466

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +273,187 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+            seccomp,
+        }
+    }
+
+    fn has_pending(&self) -> bool {
+        self.hooks.is_some() || self.net_linux.is_some() || self.seccomp.is_some()
+    }
+}
+
+fn unexpected_init_message(expected: &'static str, received: Message) -> ProcessError {
```
nayuta723: Do we really need this function?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +273,187 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
     Ok((init_pid, need_to_clean_up_intel_rdt_subdirectory))
 }
 
+const EXPECTED_INIT_MESSAGE: &str = "InitReady or a pending init setup request";
```
nayuta723: Would it make sense to define a dedicated error, such as `PrematureInitReady`, rather than introducing `EXPECTED_INIT_MESSAGE`?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +273,187 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+
+        run_hooks(hooks.create_runtime().as_ref(), &container.state, init_pid).map_err(|err| {
+            tracing::error!("failed to run create runtime hooks: {}", err);
+            err
+        })?;
+    }
+
+    init_sender.hook_done()?;
+    Ok(())
+}
```
nayuta723, line 443: I'm not sure adding an extra function solely for testing is worth it.

nayuta723 (changes requested):
Thank you for your contribution. Please take a look when you have some time.

crates/libcontainer/src/process/container_main_process.rs
YamasouA: @nayuta723 
Thank you for your review!
I agree that this is a bit verbose. Since this area was lightly tested before, I’d prefer to keep the coverage. Would you suggest a different way to structure these tests without introducing test-only helper functions, or would it be reasonable to leave this behavior to integration test coverage instead?

crates/libcontainer/src/process/container_main_process.rs
YamasouA: @nayuta723 
I addressed your review comments around the init message error handling
[0fcd52a](https://github.com/youki-dev/youki/pull/3577/commits/0fcd52af77edaaad02232b5bcce2375d3c4f7a7b)

crates/libcontainer/src/process/container_main_process.rs
YamasouA: I originally implemented it this way to avoid checking the expected requests separately from retrieving the corresponding values from container_args.
However, that ended up making the state tracking unnecessarily complex, so I changed PendingInitRequests to track only booleans.

1016323

crates/libcontainer/src/process/container_main_process.rs
nayuta723: That makes sense. This test mainly verifies that hook_done is not sent when hook execution fails. While that behavior is useful to cover, I agree it may not justify adding an extra helper only for testing. I’ll remove the test-only helper and either cover this through the existing handle path or leave it to integration coverage.

YamasouA: @nayuta723 
I delete unit test.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +298,148 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+        self.hooks || self.network_device || self.seccomp
+    }
+}
+
+fn handle_hook_request(
+    hooks: &oci_spec::runtime::Hooks,
+    container_args: &ContainerArgs,
+    init_pid: Pid,
+    init_sender: &mut channel::InitSender,
+) -> Result<()> {
```
nayuta723, line 414:
```suggestion
fn handle_hook_request(
    hooks: &oci_spec::runtime::Hooks,
    container: Option<&Container>,
    init_pid: Pid,
    init_sender: &mut channel::InitSender,
) -> Result<()> {
```

We currently mix two formats: passing `container_args` directly and including values such as `hooks` inside `container_args`. We should standardize on one of them.

nayuta723: The same issue also applies to `build_container_process_state` and `handle_seccomp_notify`.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +298,148 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+    crate::process::seccomp_listener::sync_seccomp_send_msg(
+        listener_path,
+        &encoded_state,
+        seccomp_fd.as_raw_fd(),
+    )?;
+    init_sender.seccomp_notify_done()?;
+    // `seccomp_fd` is dropped here, closing the duplicated fd. The SCM_RIGHTS
+    // msg already duplicated the fd to the process behind the listener.
+    Ok(())
+}
```
nayuta723: Is there a reason this is exposed from `container_main_process.rs`?

From a responsibility perspective, I wonder if `seccomp_listener.rs` might be a more appropriate place for it.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -299,6 +298,148 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<(Pid, bo
...
+
+    oci_spec::runtime::ContainerProcessStateBuilder::default()
+        .version(OCI_VERSION)
+        .fds(vec![SECCOMP_FD_NAME.to_string()])
+        .pid(init_pid.as_raw())
+        .metadata(seccomp.listener_metadata().clone().unwrap_or_default())
+        .state(oci_state)
+        .build()
+        .map_err(|e| ProcessError::OciStateBuild(e.to_string()))
+}
```
nayuta723: Please add a unit test for `build_container_process_state`.

nayuta723 (commented):
I left some nit comments. Please take a look, when you have a moment.

YamasouA:
@nayuta723 
We have addressed your comment.
Thankyou.

nayuta723 (approved):
LGTM! Thanks!

@youki-dev/youki 
If there are no additional comments on this, I will merge it.

utam0k (commented):
The event-loop approach makes sense to me, particularly since mount-fd requests can happen multiple times and shouldn't be limited by the spec.

I do have one concern about how `PendingInitRequests` handles the existing one-shot requests as an unordered set. The previous implementation enforced a specific sequence where hooks came before network, then seccomp, and finally ready. With this change, however, the system would also accept something like seccomp before hooks. Was it your intention to relax that ordering?

If we only need mount-fd requests to be repeatable, we could instead model the protocol as an explicit state transition. This would let us allow mount-fd requests as many times as needed during the rootfs phase while keeping the original order for hooks, network, seccomp, and ready.

Moving that transition logic away from the handlers would also make the main behavior much easier to unit test, as we wouldn't need to introduce special abstractions just for testing.

The receiver and API cleanup doesn't have to be part of this PR, but I'd like to make sure we're on the same page about whether this broader change in ordering is intentional.

YamasouA:
@utam0k 
Thanks for your comment!

I think an explicit state transition would be a good approach if container_main_process is responsible for enforcing the order.

However, I’d first like to confirm whether it needs to enforce the order at all.
The init process already sends these requests in a fixed order and waits for each reply before moving to the next step. Given that, would it be acceptable for the main process to handle them in the order they arrive without enforcing the order itself?

utam0k:
The readability of the code in the main process concerns me. My concern is that it is impossible to determine the processing order from the code alone without comparing it against the init process code. What do you think?

YamasouA:
@utam0k 
I committed an implementation to show what I had in mind.
Does this match your suggestion?
[6993f64](https://github.com/youki-dev/youki/pull/3577/commits/6993f6492a562fe464e2a5c8f72f53f4dfe0fcba)

YamasouA:
The above created an ordering dependency between HookRequest and SetupNetworkDeviceReady as well, which　doesn't actually exist, so I fixed it. My understanding is that Seccomp must come after those, and InitReady comes last.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -286,6 +254,175 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
 }
 
+/// One-shot init-side setup requests.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum InitRequest {
+    Hooks,
+    Network,
+    Seccomp,
+    Ready,
+}
```
utam0k: Could you please write the description for each field?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -286,6 +254,175 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+        &mut self,
+        message: &Message,
+    ) -> std::result::Result<InitRequest, channel::ChannelError> {
+        let received = match message {
+            Message::HookRequest => InitRequest::Hooks,
+            Message::SetupNetworkDeviceReady => InitRequest::Network,
+            Message::SeccompNotify => InitRequest::Seccomp,
+            Message::InitReady => InitRequest::Ready,
+            _ => return Err(unexpected_init_message(message)),
+        };
```
utam0k: Please use the From trait to translate that.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -286,6 +254,175 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+                self.pending_setups.remove(position);
+            }
+            InitRequest::Seccomp => {
+                if !self.pending_setups.is_empty() || !self.pending_seccomp {
+                    return Err(unexpected_init_message(message));
+                }
+                self.pending_seccomp = false;
+            }
+            InitRequest::Ready => {
+                if !self.pending_setups.is_empty() || self.pending_seccomp {
```
utam0k: Should we check whether all of these conditional expressions are being called? What exactly is being checked here?

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -286,6 +254,175 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+    Seccomp,
+    Ready,
+}
+
+struct InitRequestSequence {
+    pending_setups: Vec<InitRequest>,
+    // Seccomp must be the last setup request: once the init process applies
+    // the filter, its own syscalls are subject to the container's seccomp
+    // policy, so all other setup must be finished first.
+    pending_seccomp: bool,
```
utam0k: How about controlling the status instead of having a boolean for seccomp only?

utam0k (commented):
Overall looks good.

YamasouA:
@utam0k 
Thanks for the suggestion! I reworked the implementation to control the flow using Status, as you suggested.
I think it turned out quite nicely.
Could you take another look when you have a chance?

utam0k (approved):
Kudos!

Issue #3576, opened 2026-06-01
Add update command integration tests to contest


saku3:
### Description

Add update integration tests to `tests/contest` equivalent to [`tests/integration/update.bats`](https://github.com/opencontainers/runc/blob/main/tests/integration/update.bats) in runc, then make youki pass them.

### Motivation

The `contest` framework (Rust-based OCI runtime integration tests) currently has no coverage for update functionality. The existing BATS-based tests in runc's `update.bats` cover this area but are not available through `contest`.

### Implement contest tests
Add the following tests to tests/contest.

These tests are skipped when running with youki, as the required functionality is not fully implemented yet.

- [x] update cgroup v1/v2 common limits (#3580)
- [ ] update cgroup cpu limits
- [x] update pids.limit (#3595 )
- [x] cpu burst (#3592)
- [x] set cpu period with no quota (#3592)
- [x] set cpu period with no quota (invalid period) (#3592)
- [x] set cpu quota with no period (#3592)
- [x] update cpu period with no previous period/quota set (#3614)
- [x] update cpu quota with no previous period/quota set  (#3614)
- [x] update cpu period in a pod cgroup with pod limit set  (#3614)
- [x] update cgroup cpu.idle  (#3614)
- [ ] update cgroup cpu.idle via systemd v252+  (#3614)
- [ ] update cgroup v2 resources via unified map 
- [ ] update cpuset parameters via resources.CPU
- [ ] update cpuset parameters via v2 unified map
- [ ] update cpuset cpus range via v2 unified map
- [ ] update rt period and runtime
- [ ] update devices [minimal transition rules]
- [ ] update paused container
- [ ] update memory vs CheckBeforeUpdate
- [ ] update per-device iops/bps values

saku3:
@hayama17 @moz-sec 

Would you be interested in taking a look at this issue?

We would appreciate your help.

moz-sec:
I'm interested.
I'll look into it and start with what I can.

hayama17:
Thanks @saku3 !
I'm interested in taking this on. I'll go ahead and break it down into sub-issues based on the task list and start working through them.

hayama17:
@saku3 
#3580 
I've created a draft scaffold for the update tests. 

@moz-sec
The test logic isn't implemented yet — Would you like to build on top of this with me?

hayama17:
@saku3 
Open PR #3580, which adds e2e tests for cgroup resource update (memory, pids, cpuset).

[...]

PR #3592, opened 2026-06-15
test(contest): add cpu update integration tests for cgroup v2

hayama17:
This PR adds CPU update integration tests for cgroup v2 in `contest`, as part of #3576.
Ported from runc's `update.bats`, scoped to cgroup v2 + cgroupfs driver only.

## Description

Adds `tests/contest/contest/src/tests/update/cpu.rs` with 4 test cases:

- **cpu burst**: verify `cpu.max.burst` is updated and reset correctly
- **set cpu period without quota**: verify `cpu.max` defaults to `max <period>` when no quota is set
- **set cpu period without quota (invalid period)**: verify container creation fails with an invalid period value
- **set cpu quota without period**: verify `cpu.max` uses the default period (100000) when only quota is set

Tests that use `update_container` are skipped for youki, as youki's update command currently only supports `--pids-limit` and `--resources` (JSON).

Each test uses a unique `cgroupsPath` to avoid conflicts during parallel test execution.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

#3576 

### youki version 0.6.0
```
$ sudo RUNTIME_KIND=youki ./scripts/contest.sh target/release/youki update
  # Start group update
  1 / 5 : cpu_burst_test : skipped
          test cannot run in the current environment (run_all, parallel)
  2 / 5 : set_cpu_period_without_quota_invalid_test : ok
  3 / 5 : set_cpu_period_without_quota_test : ok
  4 / 5 : set_cpu_quota_without_period_test : ok
  5 / 5 : update_cgroup_v2_common_limits_test : skipped
          test cannot run in the current environment (run_all, parallel)
  # End group update

  Validation successful for runtime target/release/youki
```
### runc version 1.5.0-rc.2
```
$ sudo RUNTIME_KIND=runc ./scripts/contest.sh runc update
  # Start group update
  1 / 5 : cpu_burst_test : ok
  2 / 5 : set_cpu_period_without_quota_invalid_test : ok
  3 / 5 : set_cpu_period_without_quota_test : ok
  4 / 5 : set_cpu_quota_without_period_test : ok
  5 / 5 : update_cgroup_v2_common_limits_test : ok
  # End group update

  Validation successful for runtime runc
```

### DIFF
--- a/tests/contest/contest/src/tests/update/cpu.rs
+++ b/tests/contest/contest/src/tests/update/cpu.rs
@@ -1,0 +1,183 @@
+use std::path::Path;
+
+use anyhow::{Context, Result, anyhow};
+use oci_spec::runtime::{
+    LinuxCpuBuilder, LinuxResources, LinuxResourcesBuilder, ProcessBuilder, Spec, SpecBuilder,
+};
+use test_framework::{TestResult, test_result};
+
+use super::check_cgroup_value;
+use crate::utils::test_utils::check_container_created;
+use crate::utils::{start_container, test_outside_container, update_container};
+
+fn create_spec(cgroup_name: &str, resources: Option<LinuxResources>) -> Result<Spec> {
+    let mut spec = SpecBuilder::default()
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["sleep".to_string(), "1000".to_string()])
+                .build()?,
+        )
+        .build()
+        .context("failed to build spec")?;
+
+    if let Some(linux) = spec.linux_mut() {
+        linux.set_cgroups_path(Some(Path::new("/runtime-test").join(cgroup_name)));
+        linux.set_resources(resources);
+    }
+
+    Ok(spec)
+}
+
+pub(crate) fn cpu_burst_test() -> TestResult {
+    const CGROUP_NAME: &str = "cpu_burst";
+    let spec = test_result!(create_spec(CGROUP_NAME, None));
+
+    test_outside_container(&spec, &|data| {
+        test_result!(check_container_created(&data));
+
+        let id = &data.id;
+        let dir = &data.bundle;
+
+        let start_result = start_container(id, dir).unwrap().wait().unwrap();
+
+        if !start_result.success() {
+            return TestResult::Failed(anyhow!("container start failed"));
+        }
+        let cgroup_path = Path::new("/sys/fs/cgroup/runtime-test").join(CGROUP_NAME);
+
+        test_result!(check_cgroup_value(&cgroup_path, "cpu.max.burst", "0"));
+
+        update_container(
+            id,
+            dir,
+            &["--cpu-period", "900000", "--cpu-burst", "500000"],
+        )
+        .unwrap()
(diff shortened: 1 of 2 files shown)


tests/contest/contest/src/tests/update/cpu.rs
```diff
@@ -0,0 +1,183 @@
...
+            &["--cpu-period", "900000", "--cpu-burst", "500000"],
+        )
+        .unwrap()
+        .wait()
+        .unwrap();
+        test_result!(check_cgroup_value(&cgroup_path, "cpu.max.burst", "500000"));
+
+        // issue: https://github.com/opencontainers/runc/issues/4210
+        // for systemd, cpu-burst value will be cleared, it's a known issue.
+        update_container(id, dir, &["--memory", "100M"])
```
saku3: We should check whether `update_container` succeeded. Otherwise, even if `update_container` fails, the following cgroup value check may still pass if the value was already set before the update.

tests/contest/contest/src/tests/update/cpu.rs
```diff
@@ -0,0 +1,183 @@
...
+            dir,
+            &["--cpu-period", "900000", "--cpu-burst", "500000"],
+        )
+        .unwrap()
+        .wait()
+        .unwrap();
+        test_result!(check_cgroup_value(&cgroup_path, "cpu.max.burst", "500000"));
+
+        // issue: https://github.com/opencontainers/runc/issues/4210
+        // for systemd, cpu-burst value will be cleared, it's a known issue.
```
saku3: The original runc test had a systemd-related condition: `if [ ! -v RUNC_USE_SYSTEMD ];`.

As far as I understand, this condition is specific to runc's test setup and does not apply to youki, so this comment may not be necessary here.

saku3 (commented):
Thank you for the PR.

I left some comments, so please take a look.

tests/contest/contest/src/tests/update/cpu.rs
hayama17: Thanks, fixed. I added an `update_container_and_wait` helper in `update/mod.rs` that checks the exit status and fails the test if the update command exits unsuccessfully before checking the cgroup value.

tests/contest/contest/src/tests/update/cpu.rs
hayama17: I agree, removed the runc/systemd-specific comment. I replaced it with a comment describing the intent of this test in youki: ensuring a memory-only update does not reset `cpu.max.burst`.

tests/contest/contest/src/tests/update/cpu.rs
hayama17: `update/common.rs` has the same unchecked `update` calls, so I'll address those in a separate PR.

saku3 (approved):
Thanks!

Issue #3594, opened 2026-06-16
[FEATURE]: handling of -1 and 0 with update --pids-limit differs from runc


moz-sec:
# Summary

related to #3576.
I discovered this while testing `pids.limit`.

The behavior of `youki update --pids-limit` does not match the expected behavior from the OCI runtime-spec or the behavior of runc.

- `--pids-limit -1` is not accepted as a CLI argument
- `--pids-limit 0` is specified, `pids.max` becomes `max`

In runc, `--pids-limit -1` results in `pids.max = max`, and `--pids-limit 0` results in `pids.max = 1`.

# Reproduce

Create a `config.json` using `youki spec`, then add the following fields:

```json
{
  "process": {
    "args": [
      "sleep",
      "1000"
    ]
  },
  "linux": {
    "cgroupsPath": "/runtime-test/update_common_limits",
    "resources": {
      "pids": {
        "limit": 20
      }
    }
  }
}
```

## youki

```bash
$ sudo ./youki create runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
20

$ sudo ./youki update --pids-limit 12345 runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
12345

$ sudo ./youki update --pids-limit -1 runtime-test
error: unexpected argument '-1' found

  tip: to pass '-1' as a value, use '-- -1'

Usage: youki update [OPTIONS] <CONTAINER_ID>

For more information, try '--help'.

$ sudo ./youki update --pids-limit 0 runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
max
```

## runc

```bash
$ sudo runc create runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
20

$ sudo runc update --pids-limit 12345 runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
12345

$ sudo runc update --pids-limit -1 runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
max

$ sudo runc update --pids-limit 0 runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
1
```

# Expected and Actual Behavior

The results above can be summarized as follows.
The behavior when the value is 0 has been fixed in #3634.

| `--pids-limit` | youki | runc | Result |
| --- | --- | --- | --- |
[...]

PR #3595, opened 2026-06-16
test(contest): add update pids limit integration test

moz-sec:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Imprement #3576 

## Additional Context

As noted in #3594, the test for `--update pids-limit` fails, so skipped it.

I have confirmed that the test passes in runc.
```
$ just validate-contest-runc update
sudo RUNTIME_KIND="runc" /home/user/youki/scripts/contest.sh runc update
/usr/sbin/runc
# Start group update
1 / 2 : update_cgroup_v2_common_limits_test : ok
2 / 2 : update_pids_limit_test : ok
# End group update
```

This integration test was created based on the following description in [updates.bats](https://github.com/opencontainers/runc/blob/main/tests/integration/update.bats).
https://github.com/opencontainers/runc/blob/da3d022b06dd96a898f6d0cc1df2773df91ea035/tests/integration/update.bats#L330-L359

### DIFF
--- a/tests/contest/contest/src/tests/update/pids_limit.rs
+++ b/tests/contest/contest/src/tests/update/pids_limit.rs
@@ -1,0 +1,84 @@
+use std::path::Path;
+
+use anyhow::{Context, Result, anyhow};
+use oci_spec::runtime::{
+    LinuxPidsBuilder, LinuxResourcesBuilder, ProcessBuilder, Spec, SpecBuilder,
+};
+use test_framework::{TestResult, test_result};
+
+use super::check_cgroup_value;
+use crate::utils::test_utils::check_container_created;
+use crate::utils::{start_container, test_outside_container, update_container};
+
+fn create_spec(cgroup_name: &str) -> Result<Spec> {
+    let pids = LinuxPidsBuilder::default()
+        .limit(20)
+        .build()
+        .context("failed to build pids spec")?;
+
+    let resources = LinuxResourcesBuilder::default()
+        .pids(pids)
+        .build()
+        .context("failed to build resources spec")?;
+
+    let mut spec = SpecBuilder::default()
+        .process(
+            ProcessBuilder::default()
+                .args(vec!["sleep".to_string(), "1000".to_string()])
+                .build()?,
+        )
+        .build()
+        .context("failed to build spec")?;
+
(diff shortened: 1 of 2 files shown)


moz-sec:
@saku3 
Sorry.
The update test is failing during parallel execution due to duplicate cgroup names.
Also, it appears that some of these changes overlap with the changes in #3592.
Therefore, would it be okay to wait until this PR is merged and then modify the program to match that implementation?

moz-sec:
@saku3 
It looks like #3592 has been merged into main, so I merged it and then extracted it into pids_limit.rs.
Please review it.

tests/contest/contest/src/tests/update/pids_limit.rs
```diff
@@ -0,0 +1,84 @@
...
+        test_result!(check_cgroup_value(&cgroup_path, "pids.max", "12345"));
+
+        // update pids.limit to -1, pids.max will become `max`
+        update_container(id, dir, &["--pids-limit", "-1"])
+            .unwrap()
+            .wait()
+            .unwrap();
+        test_result!(check_cgroup_value(&cgroup_path, "pids.max", "max"));
+
+        // update pids.limit to 0, pids.max will become 1
```
saku3, line 75: Could we explain why pids-limit 0 maps to pids.max = 1?

It seems to follow runc’s behavior, but the reason is not obvious from the current comment.

saku3 (commented):
Thank you for the PR.

I left one comment. Please take a look when you have a chance.

tests/contest/contest/src/tests/update/pids_limit.rs
moz-sec: Set `pids.max = 0` in a cgroup, new processes will no longer be created, causing the container to fail to start.
For this reason, `runc` converts the value of `--pids-limit` to 1 if it is 0.
This test is designed to match that behavior of `runc`.

I added this explanation as a comment in pids_limit.rs.

saku3 (approved):
Thanks!

PR #3608, opened 2026-06-21
fix(utils): reject network device names exceeding the kernel limit according to the kernel reference

### LINKED ISSUE #3607: [Bug]: Current dev_valid_name in `libcontainer/src/utils.rs` deviates from the kernel reference
### Bug Description

In this [reference](https://elixir.bootlin.com/linux/v6.12/source/net/core/dev.c#L1066), `dev_valid_name` accepts network device names that are 16 bytes long, but the Linux kernel rejects them.

And I understand that the kernel allows at most 15 bytes (IFNAMSIZ - 1), reserving the null byte for the NUL terminator (16th byte).

But in `dev_valid_name` function, it allows the 16 bytes of characters (network device name) by using below logic.
```rust
if name.is_empty() || name.len() > IFNAMSIZ {
    return false;
}
```

`strnlen(name, IFNAMSIZ)` function in kernel implementation, it counts bytes up to the first NUL byte, but never more than IFNAMSIZ. So if the name has no NUL within IFNAMSIZ bytes, it returns IFNAMSIZ (16) that making the condition true and the function returns `false` to reject it.

To respect the kernel implementation, I would like to propose a solution like below.
```rust
if name.is_empty() || name.len() >= IFNAMSIZ {
    return false;
}
```

*tests*
```rust
let long_name = "a".repeat(IFNAMSIZ + 1);
assert!(!dev_valid_name(&long_name));

let over_length_name = "a".repeat(IFNAMSIZ);
assert!(!dev_valid_name(&over_length_name));

let valid_name = "a".repeat(IFNAMSIZ - 1);
assert!(dev_valid_name(&valid_name));
```

### Steps to Reproduce

Current 

### Expectation

_No response_

### System and Setup Info

_No response_

### Additional Context

_No response_

Scanf-s:
## Description

- Fixed dev_vaild_name function to follow the linux kernel implementation 
- Added few test cases which validates the dev_valid_name behavior

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

- Fixes #3607
- #279

## Additional Context

PR #3163

### DIFF
--- a/crates/libcontainer/src/utils.rs
+++ b/crates/libcontainer/src/utils.rs
@@ -403,7 +403,7 @@ pub fn validate_mount_options(
 
 // https://elixir.bootlin.com/linux/v6.12/source/net/core/dev.c#L1066
 fn dev_valid_name(name: &str) -> bool {
-    if name.is_empty() || name.len() > IFNAMSIZ {
+    if name.is_empty() || name.len() >= IFNAMSIZ {
         return false;
     }
     if name.eq(".") || name.eq("..") {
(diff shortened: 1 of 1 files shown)


saku3 (approved):
Thanks.

For reference, this change also improves the error surface for `linux.netDevices`.

For example, with the following configuration:

```json
"linux": {
  "netDevices": {
    "dummy0": {
      "name": "abcdabcdabcdabcd"
    }
  }
}
```

Previously, a 16-byte interface name passed validation and failed later while moving the device into the network namespace:

```
 failed to dev_change_net_namespace: Netlink error: -34
```

On the other hand, a 17-byte name already failed during validation:

```
invalid network device name: abcdabcdabcdabcda
```

With this fix, 16-byte names are rejected consistently during validation as well.

PR #3609, opened 2026-06-21
feat(libcgroups): rootless cgroup v2 support

### LINKED ISSUE #3597: bug: rootless containers inside of docker error with `cgroup.subtree_controll`
We use rootless containers via libcontainer in our binary. However, when running our binary inside of a standard docker container, we cannot get libcontainer to work.

When creating a container using the `oci_spec::runtime::Linux::rootless()` which is roughly

```rust
    let container = ContainerBuilder::new(container_id.clone(), SyscallType::Linux)
        .validate_id()?
        .with_root_path(&state_dir)?
        .with_executor(crostini::Crostini)
        .as_init(bundle_dir)
        // false in docker
        .with_systemd(false)
        .with_detach(false)
        .with_no_pivot(true)
        .build()?;
```

we get this error: 

```
    Failed to create init container 'rico-01KV99C15V0RH8FCHJSB5Z6458':
      caused by: failed to create container: received unexpected message: OtherError("cgroup error: io error: failed to open /sys/fs/cgroup/cgroup.subtree_control: Read-only file system (os error 30)"), expected: WriteMapping
```
Note that we're not setting any cgroup limits or `LinuxResources`. 

When using `--privileged` we were able to build the rootless runtime. But, ideally, that wouldn't be necessary. 

After debugging this for a few hours with the help of claude i think, this is related to the v2 manager in libcgroups: 

https://github.com/youki-dev/youki/blob/749353b3fbe692daa0c43aa1b39fb817b7bdddaa/crates/libcgroups/src/v2/manager.rs#L96-L129
  <details><summary>AI Summary</summary>

  The failing write is the **first line** of `create_unified_cgroup`:

  ```rust
  let controllers = util::get_available_controllers(&self.root_path)?…;
  Self::write_controllers(&self.root_path, &controllers)?;   // writes ROOT subtree_control

  Before walking down to the requested cgroup_path, it enables every available
  controller on self.root_path (the cgroup v2 mount root) by writing
  {root}/cgroup.subtree_control.

  Inside an unprivileged Docker container, the container is delegated only its own
  scope (…/docker-<id>.scope), not the mount root. So this root write hits a
  read-only file → EROFS (os error 30). The subsequent path-walk loop would
  write subtree_control on each ancestor it owns, which is fine — it's only the
  unconditional root write that fails.

  This happens regardless of requested resources (LinuxResources is unset) and
  regardless of cgroups_path / --cgroupns (host or private), which is
  consistent with the root write being unconditional rather than driven by config.

  Possible fix direction: skip enabling controllers on the mount root when the
  process doesn't own it / they're already enabled in the delegated parent, or
  anchor controller-enabling at the delegated root rather than the fs mount root.

  (libcgroups 0.6.0, cgroup v2, Docker default config, Debian trixie)

  </details>
  ```

JosiahParry:
## Description

This PR closes https://github.com/youki-dev/youki/issues/3597 by implementing the suggestion from @saku3 and @cschanhniem. 

When running rootless containers via `libcontainer` inside of a standard, unprivileged Docker container, cgroups are read only resulting in errors related to `cgroups.subtree_control`. 

This PR adds a new enum `CgroupOwnership` with variants `Full` and `Delegated` which determines how `CgroupConfig` behaves. 

When `CgroupOwnership::Delegated` permission errors are logged and reported matching runc's fs2 behavior. Note that `Delegated` is inferred from the user-namespace config.

## Running rootless in Docker

While this PR resolves issues related to cgroups, running a rootless libcontainer inside of docker may still require more work. For example, we use [crostini](https://github.com/ricochet-rs/crostini) as our executor which requires PID namespace. In order to get Docker working we have to pass `--security-opt systempaths=unconfined` which is still better than `--privileged`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #

## Additional Context

Closes #3597

### DIFF
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -123,10 +123,13 @@ impl CgroupManager for AnyCgroupManager {
 }
 
 impl AnyCgroupManager {
-    /// Set cgroup ownership. Only affects the v2 manager; ownership is a no-op
-    /// for v1 and systemd.
+    /// Set cgroup ownership.
+    ///
+    /// Only affects the v2 manager for v1 and systemd.
+    #[cfg_attr(not(feature = "v2"), allow(unused_variables))]
     pub fn with_ownership(self, ownership: CgroupOwnership) -> Self {
         match self {
+            #[cfg(feature = "v2")]
             AnyCgroupManager::V2(m) => AnyCgroupManager::V2(m.with_ownership(ownership)),
             other => other,
         }
--- a/crates/libcgroups/src/common.rs
+++ b/crates/libcgroups/src/common.rs
@@ -326,11 +326,55 @@ pub enum CreateCgroupSetupError {
     Systemd(#[from] systemd::manager::SystemdManagerError),
 }
 
+/// Describes how much of the cgroup hierarchy the process owns.
+///
+/// Under cgroup v2 delegation (e.g. a rootless container, or a read-only
+/// `/sys/fs/cgroup` mount) the process owns at most its own subtree; the mount
+/// root and the ancestors above its delegation boundary are owned upstream and
+/// read-only. `Delegated` makes cgroup creation best-effort so the container
+/// can still run without its own cgroup, mirroring runc's `rootless_cgroups`.
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum CgroupOwnership {
+    /// The process fully owns the cgroup hierarchy (privileged / root). Failures
+    /// to create or join a cgroup are fatal.
+    #[default]
+    Full,
+    /// The cgroup hierarchy is delegated and may be partly read-only. Failures
+    /// to create or join a cgroup are non-fatal; the process stays in its parent
+    /// cgroup. A requested limit that cannot be applied still errors later.
+    Delegated,
+}
+
 #[derive(Clone)]
+#[non_exhaustive]
 pub struct CgroupConfig {
     pub cgroup_path: PathBuf,
     pub systemd_cgroup: bool,
     pub container_name: String,
+    pub ownership: CgroupOwnership,
+}
+
+impl CgroupConfig {
+    /// Creates a config with `Full` ownership. Use [`Self::with_ownership`] to
+    /// opt into `Delegated` for rootless / delegated environments.
+    pub fn new(
+        cgroup_path: impl Into<PathBuf>,
+        systemd_cgroup: bool,
+        container_name: impl Into<String>,
+    ) -> Self {
+        Self {
+            cgroup_path: cgroup_path.into(),
+            systemd_cgroup,
+            container_name: container_name.into(),
+            ownership: CgroupOwnership::Full,
+        }
+    }
+
+    pub fn with_ownership(mut self, ownership: CgroupOwnership) -> Self {
+        self.ownership = ownership;
+        self
+    }
 }
 
 // Create any cgroup manager with customize root path. If root_path provided
@@ -358,7 +402,7 @@ pub fn create_cgroup_manager_with_root(
         CgroupSetup::Unified => {
             // ref https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#cgroups-path
             if cgroup_path.is_absolute() || !config.systemd_cgroup {
-                return Ok(create_v2_cgroup_manager(root, cgroup_path)?.any());
+                return Ok(create_v2_cgroup_manager(root, cgroup_path, config.ownership)?.any());
             }
             Ok(
                 create_systemd_cgroup_manager(root, cgroup_path, config.container_name.as_str())?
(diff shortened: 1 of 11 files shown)


JosiahParry:
tried to do the DCO fixes and broke commit history... will work on that.

JosiahParry:
CI is failing because I lack the ability to add labels. :)

JosiahParry:
Hi team! CI is now passing, (thanks, saku3)! Please let me know if any changes are needed here :)

saku3:
I’ll take a look at it within the next few days.

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -51,8 +51,15 @@ pub fn container_intermediate_process(
     let spec = &args.spec;
     let linux = spec.linux().as_ref().ok_or(MissingSpecError::Linux)?;
     let namespaces = Namespaces::try_from(linux.namespaces().as_ref())?;
+    // If in a rootless container, use best effort
+    let cgroup_ownership = if args.user_ns_config.is_some() {
```
saku3: Ideally, this should follow runc’s `shouldUseRootlessCgroupManager` logic. Since ownership only affects the cgroup v2 implementation, we may be able to omit the systemd-specific OwnerUID handling and decide this based on the euid and whether the process is running in a user namespace, rather than using args.user_ns_config.is_some().

https://github.com/opencontainers/runc/blob/89f46d647095d5d596fb5aa573a6d6588630a135/rootless_linux.go#L13

crates/libcgroups/src/v2/manager.rs
```diff
@@ -110,31 +138,97 @@ impl Manager {
...
+                );
+                Ok(())
+            }
+            Err(err) => Err(err.into()),
         }
+    }
 
-        Ok(())
+    /// Reads the controllers already enabled in `{path}/cgroup.subtree_control`
+    fn get_enabled_controllers(path: &Path) -> Result<Vec<String>, WrappedIoError> {
```
saku3: I think `missing_controllers` and `get_enabled_controllers` can be dropped. Writing an already-enabled controller is idempotent in cgroup v2, and controller enabling is already best-effort here. So we can simply attempt to write all requested controllers and ignore expected errors on the write path.

Adding these functions makes the implementation more complex.

crates/libcgroups/src/common.rs
```diff
@@ -326,6 +340,19 @@ pub enum CreateCgroupSetupError {
...
 }
 
+/// Whether cgroups are delegated or owned by the process
+///
+/// In delegated environments such as container-in-container,
+/// the cgroup hierarchy is likely read-only. When using
+/// [CgroupOwnership::Delegated], errors become non-fatal.
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum CgroupOwnership {
```
saku3: `CgroupOwnership::Delegated` may be a bit misleading. In cgroup v2 terminology, delegation usually means that the process has been given a writable subtree, while this flag seems to mean that certain cgroup errors should be tolerated in rootless/read-only environments. If this is intended to model rootless-style cgroup handling, a name like Rootless may be clearer, similar to runc’s Cgroup.Rootless.

https://docs.kernel.org/6.0/admin-guide/cgroup-v2.html#delegation

I think the real question here is whether this is running rootless or not. So I think introducing a boolean variable named rootless should be enough.

crates/libcgroups/src/v2/manager.rs
```diff
@@ -77,29 +78,56 @@ pub struct Manager {
     root_path: PathBuf,
     cgroup_path: PathBuf,
     full_path: PathBuf,
+    ownership: CgroupOwnership,
```
saku3: I left a comment in crates/libcgroups/src/common.rs, but I think we should use a boolean variable named rootless here.

crates/libcontainer/src/process/container_intermediate_process.rs
```diff
@@ -51,8 +52,10 @@ pub fn container_intermediate_process(
     let spec = &args.spec;
     let linux = spec.linux().as_ref().ok_or(MissingSpecError::Linux)?;
     let namespaces = Namespaces::try_from(linux.namespaces().as_ref())?;
+    let rootless = rootless_required(command.as_ref()).unwrap_or(false);
```
JosiahParry, line 55: @saku3 this function appears to have all of the logic of `shouldUseRootlessCgroupManager()` except the systemd branch—which I believe is what you asked for!

crates/libcgroups/src/v2/manager.rs
JosiahParry: deleted

crates/libcgroups/src/v2/manager.rs
JosiahParry: changed! I tend to try and avoid boolean arguments. I've been indoctrinated into the "a bool is an enum with two variants" camp http://teamten.com/lawrence/programming/prefer-enums-over-booleans.html

JosiahParry (commented):
Thanks for the thorough review, @saku3! I believe I've faithfully made the requested changes

crates/libcontainer/src/process/container_intermediate_process.rs
JosiahParry: It now uses the existing `rootless_required()` utility function which mimics `shouldUseRootlessCgroupManager`

saku3:
@JosiahParry 
Could you please rebase this PR when you have a chance?

It looks like this branch includes commits from other contributors.

JosiahParry:
Yes! I think I did the DCO fix wrong!

JosiahParry:
@saku3 rebased and the upstream changes aren't reflectted in here

saku3 (approved):
Thanks.

JosiahParry:
Thank you!

Issue #3604, opened 2026-06-20
Implement  OCI per-mount propagation options such as rshared and rslave


saku3:
## Description

youki does not appear to correctly apply OCI per-mount propagation options such as shared, rshared, slave, rslave, private, rprivate, unbindable, and runbindable.

This issue is about propagation options specified in mounts[].options, not linux.rootfsPropagation.

For example:

```
    {
      "destination": "/mnt/shared",
      "type": "none",
      "source": "/host/shared",
      "options": ["rbind", "shared"]
    }
```

In this case, /host/shared should be recursively bind-mounted to /mnt/shared, and the resulting mount should be recursively shared.

## Expected behavior

The resulting mount should have the requested propagation type.

For example, with rbind and rshared, /proc/self/mountinfo inside the container should contain a shared:<id> optional field for /mnt/shared

## Actual behavior

```
$ youki run -b tutorial a
$ cat /proc/self/mountinfo | grep /shared
1348 1340 253:1 /host/shared /mnt/shared rw,relatime master:1 - ext4 /dev/vda1 rw,discard,errors=remount-ro,commit=30
```

## Additional Context

I traced the implementation back to some extent, and it does not seem that this stopped working due to a regression. Rather, it appears that this functionality was never implemented in the first place.

PR #3621, opened 2026-06-27
fix(libcontainer): scope pivot_root rslave to the old root only

saku3:
## Description

This PR fixes the mount propagation handling in `pivot_rootfs`.

In `pivot_rootfs`, after `pivot_root` is executed, the old root should be made `rslave`. However, the previous implementation made the new root, i.e. `/`, `rslave` instead.

Making the new root `rslave` can affect the mount propagation settings that were configured before `pivot_rootfs`.

This PR changes the logic so that the old root is correctly made `rslave` after `pivot_root`.

This fix is based on the `pivotRoot` implementation in runc.

https://github.com/opencontainers/runc/blob/1aff3ed358ea555c22ad75cc8f085c025f048b48/libcontainer/rootfs_linux.go#L1146

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

### DIFF
--- a/crates/libcontainer/src/syscall/linux.rs
+++ b/crates/libcontainer/src/syscall/linux.rs
@@ -18,7 +18,7 @@ use nix::fcntl::{OFlag, open};
 use nix::mount::{MntFlags, MsFlags, mount, umount2};
 use nix::sched::{CloneFlags, unshare};
 use nix::sys::stat::{Mode, SFlag, mknod};
-use nix::unistd::{Gid, Uid, chown, chroot, close, fchdir, pivot_root, sethostname};
+use nix::unistd::{Gid, Uid, chdir, chown, chroot, fchdir, pivot_root, sethostname};
 use oci_spec::runtime::PosixRlimit;
 use pathrs::flags::OpenFlags;
 use pathrs::procfs::{ProcfsBase, ProcfsHandle};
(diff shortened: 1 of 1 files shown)


crates/libcontainer/src/syscall/linux.rs
```diff
@@ -378,6 +378,16 @@ impl Syscall for LinuxSyscall {
 
     /// Function to set given path as root path inside process
     fn pivot_rootfs(&self, path: &Path) -> Result<()> {
+        // open oldroot as directory and read only
+        let oldroot = open(
```
nayuta723: How about using `OwnedFd` to avoid leaking the `oldroot` fd on early returns?

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -378,6 +378,16 @@ impl Syscall for LinuxSyscall {
...
+            "/",
+            OFlag::O_DIRECTORY | OFlag::O_RDONLY | OFlag::O_CLOEXEC,
```
nayuta723:
```suggestion
            OFlag::O_DIRECTORY | OFlag::O_PATH | OFlag::O_CLOEXEC,
```
If we don’t need to read or write it, O_PATH might be better.

crates/libcontainer/src/syscall/linux.rs
```diff
@@ -417,18 +439,22 @@ impl Syscall for LinuxSyscall {
...
         close(newroot).inspect_err(|errno| {
             tracing::error!(?errno, ?newroot, "failed to close new root directory");
         })?;
 
+        close(oldroot).inspect_err(|errno| {
+            tracing::error!(?errno, ?oldroot, "failed to close old root directory");
+        })?;
+
         Ok(())
     }
```
nayuta723, line 460: Please add a test for rbind with mount propagation.
https://github.com/youki-dev/youki/issues/3604

nayuta723 (commented):
I left some comments, please take a look.

crates/libcontainer/src/syscall/linux.rs
saku3: fixed.

crates/libcontainer/src/syscall/linux.rs
saku3: fixed.

crates/libcontainer/src/syscall/linux.rs
saku3: Sorry for the confusion.(It became confusing because I initially linked #3604 in the PR description.)

This change is specifically a fix for the `pivot_root` handling. It is not intended to address mount propagation itself.

I noticed this issue while working on the mount propagation implementation, so I decided to fix it separately. If this fix were included together with the mount propagation changes in a single PR, I think it would make the PR harder to review and understand.

Mount propagation support is not implemented in youki yet, and I plan to work on it next. I also plan to add tests as part of that mount propagation work.

nayuta723: I see, I misunderstood.

nayuta723 (approved):
LGTM

PR #3624, opened 2026-06-28
fix(libcontainer): consider gid_mappings when looking up id-map binaries

### LINKED ISSUE #3623: [Bug]: lookup_map_binaries ignores gid_mappings, can panic with one uid mapping and multiple gid mappings
### Bug Description

While reading through the rootless user-namespace mapping code I noticed that `lookup_map_binaries` in `crates/libcontainer/src/user_ns.rs` checks the length of `uid_mappings` twice and never looks at `gid_mappings`:

```rust
pub fn lookup_map_binaries(
    spec: &Linux,
) -> std::result::Result<Option<(PathBuf, PathBuf)>, MappingError> {
    if let Some(uid_mappings) = spec.uid_mappings() {
        if uid_mappings.len() == 1 && uid_mappings.len() == 1 {
            return Ok(None);
        }
        ...
```

The function decides whether the `newuidmap`/`newgidmap` helper binaries are needed. They're only required when more than one mapping line has to be written; a single mapping can be written directly to `/proc/<pid>/{uid,gid}_map`. Because the right-hand side of the `&&` is a copy of the left-hand side, the gid mapping count is never considered.

The practical consequence is a panic. If a spec has exactly one uid mapping but more than one gid mapping, the function returns `Ok(None)`, so `newgidmap` is never looked up and `UserNamespaceConfig::newgidmap` stays `None`. Later, `write_gid_mapping` -> `write_id_mapping` falls into the multiple-mapping branch:

```rust
_ => {
    ...
    Command::new(map_binary.unwrap())   // map_binary is None -> panic
```

and unwraps a `None`.

### Steps to Reproduce

Run a rootless container whose config has a single uid mapping and two or more gid mappings, e.g.:

```json
"uidMappings": [
  { "containerID": 0, "hostID": 1000, "size": 1 }
],
"gidMappings": [
  { "containerID": 0, "hostID": 1000, "size": 1 },
  { "containerID": 1, "hostID": 100000, "size": 65536 }
]
```

together with a user namespace. youki ends up calling `map_binary.unwrap()` on a `None` while writing the gid map.

It is also visible by reading the code alone: the `&&` has the same operand on both sides, so the condition is equivalent to `uid_mappings.len() == 1`.

### Expectation

The helper-binary lookup should take both the uid and gid mapping counts into account. With one uid mapping and multiple gid mappings, youki should resolve `newgidmap` (and fail cleanly with `BinaryNotFound` if it is not installed) instead of reporting that no binary is needed and panicking later.

### System and Setup Info

Found by code inspection on current `main`. The duplicated condition has been there since the function was first added, so every version with this code path is affected.

### Additional Context

I have a fix ready and will open a PR. The change computes both mapping lengths and only short-circuits when uid and gid both have at most one mapping.

tkshsbcue:
## Description

`lookup_map_binaries` in `crates/libcontainer/src/user_ns.rs` decides whether the `newuidmap`/`newgidmap` helper binaries are needed to write the user-namespace id mappings. The guard that short-circuits the lookup checked `uid_mappings.len()` on both sides of the `&&`:

```rust
if uid_mappings.len() == 1 && uid_mappings.len() == 1 {
    return Ok(None);
}
```

so the gid mapping count was never taken into account. When a spec has a single uid mapping but more than one gid mapping, the function returns `Ok(None)`, `newgidmap` is left unset, and `write_id_mapping` then panics on `map_binary.unwrap()` in the multiple-mapping branch while writing the gid map.

This changes the guard to compute both the uid and gid mapping lengths and only skip the lookup when both have at most one mapping (the case that can be written directly to `/proc/<pid>/{uid,gid}_map`). The `if let` is replaced with `map_or(0, ..)` over the two getters so the empty/`None` cases are handled the same way as before.

## Type of Change

## Testing
- [x] Added new unit tests
- [x] Ran existing test suite

Added two unit tests next to the existing `user_ns` tests: one for the single uid + single gid case (no helper binary needed) and a regression test for the single uid + multiple gid case, which must never return `Ok(None)` regardless of whether the helper binaries are installed on the host.

## Related Issues
Fixes #3623

## Additional Context
The duplicated condition has been present since the function was introduced, so the panic affects any rootless container configured with one uid mapping and multiple gid mappings.

### DIFF
--- a/crates/libcontainer/src/user_ns.rs
+++ b/crates/libcontainer/src/user_ns.rs
@@ -373,20 +373,22 @@ fn is_id_mapped(id: u32, mappings: &[LinuxIdMapping]) -> bool {
 pub fn lookup_map_binaries(
     spec: &Linux,
 ) -> std::result::Result<Option<(PathBuf, PathBuf)>, MappingError> {
-    if let Some(uid_mappings) = spec.uid_mappings() {
-        if uid_mappings.len() == 1 && uid_mappings.len() == 1 {
-            return Ok(None);
-        }
+    let uid_mappings_len = spec.uid_mappings().as_ref().map_or(0, |m| m.len());
+    let gid_mappings_len = spec.gid_mappings().as_ref().map_or(0, |m| m.len());
 
-        let uidmap = lookup_map_binary("newuidmap")?;
-        let gidmap = lookup_map_binary("newgidmap")?;
+    // The newuidmap/newgidmap helper binaries are only required to write more
+    // than one mapping. A single mapping (or none) for both uid and gid can be
+    // written directly to /proc/<pid>/{uid,gid}_map without them.
+    if uid_mappings_len <= 1 && gid_mappings_len <= 1 {
+        return Ok(None);
+    }
 
-        match (uidmap, gidmap) {
-            (Some(newuidmap), Some(newgidmap)) => Ok(Some((newuidmap, newgidmap))),
-            _ => Err(MappingError::BinaryNotFound),
-        }
-    } else {
-        Ok(None)
+    let uidmap = lookup_map_binary("newuidmap")?;
+    let gidmap = lookup_map_binary("newgidmap")?;
+
+    match (uidmap, gidmap) {
+        (Some(newuidmap), Some(newgidmap)) => Ok(Some((newuidmap, newgidmap))),
+        _ => Err(MappingError::BinaryNotFound),
     }
 }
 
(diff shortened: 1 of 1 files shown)


saku3:
Thank you for PR!
Please check DCO.
https://github.com/youki-dev/youki/pull/3624/checks?check_run_id=83878641437

tkshsbcue:
hey @saku3, thanks, added the sign-off DCO is green now.

saku3 (approved):
Thanks!

PR #3627, opened 2026-06-29
refactor(contest): extract detached console-socket spawn into test_utils

### LINKED ISSUE #3529: [Refactor]: Extract Checkpoint/Restore terminal spawn logic into test_utils.rs
### Background

In the Checkpoint/Restore test suite (tests/contest/contest/src/tests/checkpoint_restore/mod.rs), we heavily rely on the OCI terminal: true requirement to test persistent IPC file descriptors (FIFOs). 

Because terminal: true requires an active console socket, the C/R tests use a specialized CrTestContext::start method. 

This method concurrently spawns the container in detached mode (run -d --console-socket), accepts the socket connection, and hands it off to handle_console_socket(stream). 

Currently, this -d spawn and concurrent socket-handling logic is encapsulated exclusively within the C/R test context. 

### Feature Request

During the review of PR #3493, see link to PR [comment](https://github.com/youki-dev/youki/pull/3493#discussion_r3179099731), it was pointed out that this logic could be highly useful for other test suites that might need to test console sockets or detached runs in the future. 

We agreed to extract this out of CrTestContext and into tests/contest/contest/src/utils/test_utils.rs as a global helper, but deferred it to this separate PR to avoid delaying the core C/R features and to properly review the global test utility changes.

tkshsbcue:
## Description

This is the follow-up refactor agreed during the review of #3493.

The checkpoint/restore suite needs to start a container with `terminal: true`,
which requires running it detached (`run -d --console-socket`) and then servicing
the console socket so the PTY master fd stays open for the container's lifetime.
That whole dance lived inside `CrTestContext::start`, so it was only reachable from
the C/R tests even though it is generally useful for any suite that wants to test
console sockets or detached runs.

This PR extracts it into a reusable `run_container_with_console(id, dir)` helper in
`tests/contest/contest/src/utils/test_utils.rs`:

- The helper binds `<dir>/console.sock`, spawns `run -d --console-socket`, accepts the
  connection and hands the stream to `handle_console_socket`, waits on the child, and
  blocks until the container is `Running`.
- `CrTestContext::start` now just calls the helper and keeps the C/R-specific FIFO
  `ping_container` check layered on top.
- The `// TODO: Consider extracting this into test_utils.rs ...` note that pointed at
  this issue is removed, and the now-unused imports (`Stdio`, `handle_console_socket`,
  `wait_container_running`) are dropped from the C/R module.

The behaviour of the existing tests is unchanged — this is a pure extraction so the
logic can be shared by future suites.

## Type of Change

## Testing
- [x] Ran existing test suite

The C/R integration tests (`checkpoint_restore`) exercise this path through
`CrTestContext::start`; they run only on non-youki runtimes with CRIU installed, and
their behaviour is unchanged. `cargo +nightly fmt` is clean.

## Related Issues
Closes #3529

## Additional Context
Extraction was deferred out of #3493 to keep that PR focused on the core C/R features
and to review the shared test-utility change on its own, as discussed in
https://github.com/youki-dev/youki/pull/3493#discussion_r3179099731.

### DIFF
--- a/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
+++ b/tests/contest/contest/src/tests/checkpoint_restore/mod.rs
@@ -8,7 +8,7 @@ use std::os::unix::fs::symlink;
 use std::os::unix::io::AsRawFd;
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::process::Command;
 use std::time::Duration;
 
 use anyhow::{Result, anyhow, bail};
@@ -19,9 +19,8 @@ use test_framework::{ConditionalTest, TestGroup, TestResult};
 use crate::utils::{
     LifecycleStatus, WaitTarget, build_checkpoint_command, checkpoint_container, criu_has_feature,
     criu_installed, delete_container, exec_container, generate_uuid, get_container_pid,
-    get_runtime_path, handle_console_socket, is_runtime_youki, kill_container, net, prepare_bundle,
-    restore_container, set_config, try_checkpoint_container, wait_container_running,
-    wait_for_state,
+    get_runtime_path, is_runtime_youki, kill_container, net, prepare_bundle, restore_container,
+    run_container_with_console, set_config, try_checkpoint_container, wait_for_state,
 };
 
 /// Used as check_fn for all `ConditionalTests` in this module:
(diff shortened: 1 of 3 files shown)


utam0k:
May I ask you to check the CI?

saku3:
This has been resolved.
https://github.com/youki-dev/youki/pull/3693

PR #3633, opened 2026-07-01
feat(validator): reject null bytes in mount fields during spec validation

### LINKED ISSUE #3591: [FEATURE]: Validate mount options for null bytes (\u0000) early to provide clearer error messages
### Background

I noticed that while `youki` (v0.6.0) correctly rejects mount options containing null bytes (`\u0000`), the failure occurs very late in the lifecycle—at the syscall level.

### Feature Request

Add early validation during the configuration parsing phase to explicitly check for and reject null bytes (`\u0000`) in string parameters like mount options, rather than passing them down to the syscall level.

Yes. When injecting a null byte into mount options (e.g., `nosuid\u0000`) for filesystems like `cgroup`, `mqueue`, `sysfs`, and `tmpfs`, the unvalidated string is passed down to the `mount(2)` syscall, causing the kernel to return an `EINVAL` (Invalid Argument) error. This results in a somewhat generic and cryptic error.
**Steps to reproduce:**

1. Use the following `config.json` snippet where `\u0000` is injected into the mount options:

config.json snippet

```json
"mounts": [
    {
        "destination": "/proc",
        "type": "proc",
        "source": "proc"
    },
    {
        "destination": "/dev",
        "type": "tmpfs",
        "source": "tmpfs",
        "options": [
            "nosuid",
            "strictatime",
            "mode=755",
            "size=65536k"
        ]
    },
    {
        "destination": "/dev/pts",
        "type": "devpts",
        "source": "devpts",
        "options": [
            "nosuid",
            "noexec",
            "newinstance",
            "ptmxmode=0666",
            "mode=0620",
            "gid=5"
        ]
    },
    {
        "destination": "/dev/shm",
        "type": "tmpfs",
        "source": "shm",
        "options": [
            "nosuid",
            "noexec",
            "nodev",
            "mode=1777",
            "size=65536k"
        ]
    },
    {
        "destination": "/dev/mqueue",
        "type": "mqueue",
        "source": "mqueue",
        "options": [
            "nosuid",
            "noexec",
            "nodev"
        ]
    },
    {
        "destination": "/sys",
        "type": "sysfs",
        "source": "sysfs",
        "options": [
            "nosuid",
            "noexec",
            "nodev",
            "ro"
        ]
    },
    {
        "type": "cgroup2",
        "source": "cgroup2",
        "options": [
            "nosuid\u0000",
            "noexec",
            "nodev",
            "relatime",
            "ro",
            "nsdelegate"
        ]
    }
]
```

1. Run the container creation command (using `sudo`):

Bash

```
sudo youki create mycontainer
```

**Current Logs and Outputs:**

Currently, the runtime passes the string to the kernel and crashes with a syscall error.

Plaintext

```
ERROR libcontainer::rootfs::mount: failed to mount Mount { destination: "/sys/fs/cgroup", typ: Some("cgroup2"), source: Some("cgroup2"), options: Some(["nosuid\0", "noexec", "nodev", "relatime", "ro", "nsdelegate"]), uid_mappings: None, gid_mappings: None }: syscall
ERROR libcontainer::process::init::process: failed to prepare rootfs err=Mount(Syscall(Nix(EINVAL)))
ERROR libcontainer::process::container_intermediate_process: failed to initialize container process: failed to prepare rootfs
```

### Proposed Solution

Ideally, the runtime should catch the null byte during the spec validation phase (before executing syscalls) and print a much more actionable and user-friendly error message, such as: 
`Error: mount option 'nosuid\u0000' contains illegal null byte`.

### Considerations

Relying on the kernel to catch it (current behavior) is secure but provides poor UX. 
*(For context, `runc` v1.4.2 exhibits similar behavior by crashing at the syscall level, whereas `crun` v1.28 silently strips everything after the null byte and proceeds, which poses a configuration fidelity risk).* Implementing early validation would make `youki` both secure and more user-friendly than the alternatives.

### Additional Context

_No response_

tkshsbcue:
## Summary

Fixes #3591.

When a mount string field (`destination`, `source`, `type`, or an entry in `options`) contains an interior null byte (`\0`), the value was passed unchanged all the way down to the `mount(2)` syscall. The kernel then rejects it with a generic `EINVAL`, producing a cryptic error that gives no indication of which mount or field is at fault.

## Changes

- Add `Validator::validate_spec_for_mounts`, wired into `Validator::validate_spec`, which scans every mount's `destination`, `source`, `type`, and `options` for an interior null byte and returns a clear error early during spec validation.
- Add `ErrInvalidSpec::MountContainsNullByte { field, value }` so the error names the offending field and value.

Because `validate_spec` runs on both the create and exec (tenant) paths, invalid mounts are now rejected up front with an actionable message instead of failing late at the syscall level.

## Tests

Added unit tests covering a null byte in each of `destination`, `source`, `type`, and `options`, plus the happy path (clean mounts) and the no-mounts case.

### DIFF
--- a/crates/libcontainer/src/error.rs
+++ b/crates/libcontainer/src/error.rs
@@ -138,6 +138,8 @@ pub enum ErrInvalidSpec {
     SysctlNotInSeparateNamespace(String),
     #[error("invalid intelRdt.closID (must not contain '.', '..', or '/')")]
     InvalidIntelRdtClosId,
+    #[error("mount {field} contains a null byte: {value:?}")]
+    MountContainsNullByte { field: String, value: String },
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/crates/libcontainer/src/validator.rs
+++ b/crates/libcontainer/src/validator.rs
@@ -1,4 +1,6 @@
 use std::collections::HashSet;
+use std::ffi::OsStr;
+use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 
 use nix::sys::stat::stat;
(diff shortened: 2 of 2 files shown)


nayuta723:
@tkshsbcue 
Please check DCO.

saku3:
I do not think validation needs to be this strict.

If we make any changes here, I feel that improving the error message would be sufficient.

There is also a comment like the following, so I think it is reasonable to be cautious about adding more validation:

https://github.com/opencontainers/runc/issues/5320#issuecomment-4707773285

tkshsbcue:
alright i will have a look into this!

Issue #719, opened 2022-02-19
Unable to start rootless container under Podman when Youki used as runtime


ondra05:
When I try to start a rootless podman container with Youki used as runtime, it tries to create a `/run/youki` directory, which, as not running as root, fails with permission denied.

Youki version: 0.0.2 (commit: 0.0.2-0-73dc75c)

```console
% podman create --runtime /home/ondra/.cargo/bin/youki --name fedora fedora
128fed7131b2a59030b9691108977dd0b4fc4c1fcc6b064c6f4c16d615637b96

% podman start fedora        
Error: failed to create directory /run/youki

Caused by:
    Permission denied (os error 13)
ERRO[0000] Error removing container 128fed7131b2a59030b9691108977dd0b4fc4c1fcc6b064c6f4c16d615637b96 from runtime after creation failed 
Error: unable to start container "128fed7131b2a59030b9691108977dd0b4fc4c1fcc6b064c6f4c16d615637b96":     Permission denied (os error 13): OCI permission denied
```

utam0k:
@ondra05 Thanks for your report. But, I couldn't reproduce this error in my environment.
```console
$ podman create --runtime /home/utam0k/ghq/github.com/utam0k/youki/youki --name fedora fedora
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers
Resolved "fedora" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.fedoraproject.org/fedora:latest...
Getting image source signatures
Copying blob c6183d119aa8 done
Copying config e417cd49a8 done
Writing manifest to image destination
Storing signatures
e60cafa77762fe924e97d1d9e9ce2cdf75315517ed511b1eb53e0058ba0bfbe0
```
Please tell me the result of `youki info`?

ondra05:
Sure thing!
```
Version           0.0.2
Commit            73dc75c
Kernel-Release    5.16.8-200.fc35.x86_64
Kernel-Version    #1 SMP PREEMPT Tue Feb 8 20:58:59 UTC 2022
Architecture      x86_64
Operating System  Fedora Linux 35 (Workstation Edition)
Cores             8
Total Memory      15699
Cgroup setup      unified
Cgroup mounts
  unified         /sys/fs/cgroup
[WARN crates/libcgroups/src/v2/util.rs:41] 2022-02-20T10:19:40.888719702+01:00 Controller misc is not yet implemented.
[...]

PR #3671, opened 2026-08-01
docs: fix outdated steps in WebAssembly guide

jtakakura:
## Description
Update the WebAssembly user guide, which had several outdated or incorrect steps that prevented the walkthrough from working:

- Replace the removed `wasm32-wasi` Rust target with `wasm32-wasip1` (rustup, cargo build, and the Dockerfile COPY path)
- Use an absolute entrypoint path (`/wasm-module.wasm`) to match the sample and work across all wasm runtimes
- Pass `--annotation "run.oci.handler=wasm"` on `podman run` so the wasm handler is triggered even when the build-time annotation is not propagated
- Drop `sudo` and the stale `#719` footnote (rootless youki now uses `/run/user/{uid}/youki`), and add a note about running buildah/podman in the same rootless/rootful context
- Fix typos: `runc.oci.handler` → `run.oci.handler`, stray quote in `variant=compat"`, `WHARE` → `WHERE`
- Describe `.wat` correctly as "webassembly text format" (was "test") and generalize the feature-flag wording
- Apply the same `wasm32-wasip1` fix to `tools/wasm-sample/README.md`

## Type of Change
- [x] Documentation update

## Testing

Verified `cargo build --target wasm32-wasip1` builds `tools/wasm-sample` successfully on Rust 1.96 (the old `wasm32-wasi` target no longer exists). Annotation keys and rootless path handling were cross-checked against the youki source (`crates/youki/src/workload/*.rs`, `crates/youki/src/rootpath.rs`).

## Additional Context
`wasm32-wasi` was renamed to `wasm32-wasip1` and removed from recent Rust toolchains, so the previous instructions failed at the very first step.

### DIFF
--- a/docs/src/user/webassembly.md
+++ b/docs/src/user/webassembly.md
@@ -2,7 +2,7 @@
 
 There are 3 things you need to do to run a WebAssembly module with youki.
 
-1. Build youki with wasm-wasmer feature flag enabled
+1. Build youki with one of the wasm feature flags enabled
 2. Build a container image with the WebAssembly module
 3. Run the container with youki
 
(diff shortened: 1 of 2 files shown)


jtakakura:
Verified the full walkthrough end-to-end on popos: build, buildah image, and rootless podman run all succeed.

saku3 (approved):
Thanks!

PR #3682, opened 2026-08-08
refactor(contest): make netns setup an explicit checkpoint test step

### LINKED ISSUE #3658: refactor(test): consolidate duplicated `setup_network_namespace` calls in checkpoint test
### Background

In the checkpoint test suite (tests/contest/contest/src/tests/lifecycle/checkpoint.rs),
`setup_network_namespace` brings `lo` up inside the container's network namespace via `nsenter`.
This is because CRIU requires a minimal network setup in the netns before it can dump the container.

This function is currently called from two places:

- `checkpoint()`: this function brings `lo` up right before invoking the checkpoint command,
  so every checkpoint test implicitly gets the netns setup as a side effect of calling `checkpoint()`.
- `wait_in_flight()`: the `tcp-skip-in-flight` test needs `lo` up before it starts polling `ss`,
  since an in-flight connection only accumulates in the accept queue once clients inside the
  container can reach 127.0.0.1. The container itself cannot bring `lo` up because the default
  contest spec grants no `NET_ADMIN`, so the host has to do it.

The two call sites need the setup at different times, so `ip link set up dev lo` runs twice in the
`tcp-skip-in-flight` test. There is no functional impact: `ip link set up` is idempotent, so bringing
up an already-up link leaves the interface state unchanged.

### Feature Request

This improvement was discussed during the review of PR #3650, see link to PR
[comment](https://github.com/youki-dev/youki/pull/3650#discussion_r3596326337). It was not done in that PR because it requires changes outside its scope.

Two things are worth sorting out here:

- **The preconditions of `checkpoint()`.** Right now `checkpoint()` sets up the netns as a side
  effect, so it is unclear whether it should assume the netns is already configured or configure it
  itself. This decision should be made explicitly.
- **Consistency across test cases.** Other checkpoint tests rely on the hidden setup inside
  `checkpoint()`, while `wait_in_flight()` calls it directly.

Proposal: separate the netns setup from `checkpoint()` and make it an explicit step, then drop the
call in `wait_in_flight()`.

_No response_

### Proposed Solution

_No response_

### Considerations

_No response_

### Additional Context

_No response_

donkomura:
## Description

setup_network_namespace was called from checkpoint(), so every checkpoint
test got the netns setup as a hidden side effect, and wait_in_flight() called it again for the tcp-skip-in-flight test. It was unclear whether checkpoint() should assume a configured netns or configure it itself.

Move the setup out of checkpoint() into prepare_checkpoint(), which each checkpoint test now calls before dumping. checkpoint() no longer touches the netns, so its precondition is explicit and the duplicated call in wait_in_flight() is gone.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes https://github.com/youki-dev/youki/issues/3658

### DIFF
--- a/tests/contest/contest/src/tests/lifecycle/checkpoint.rs
+++ b/tests/contest/contest/src/tests/lifecycle/checkpoint.rs
@@ -58,9 +58,11 @@ fn get_container_pid(project_path: &Path, id: &str) -> Result<i32, TestResult> {
 fn bring_up_loopback(project_path: &Path, id: &str) -> Result<(), TestResult> {
     let pid = get_container_pid(project_path, id)?;
 
-    // CRIU requires a minimal network setup in the network namespace, so this
-    // brings `lo` up. The container cannot do that itself because the default
-    // contest spec grants no NET_ADMIN, hence entering the netns from the host.
+    // Brings `lo` up so the container can talk to itself over 127.0.0.1. Only
+    // tests that actually use the loopback need this; the CRIU dump itself does
+    // not require `lo` to be up. The container cannot bring it up itself because
+    // the default contest spec grants no NET_ADMIN, hence entering the netns
+    // from the host.
     let output = match Command::new("nsenter")
         .stdout(Stdio::piped())
         .stderr(Stdio::piped())
@@ -232,10 +234,6 @@ fn create_checkpoint_image_dir() -> Result<(tempfile::TempDir, std::path::PathBu
 }
 
 pub fn checkpoint_leave_running_work_path_tmp(project_path: &Path, id: &str) -> TestResult {
-    if let Err(e) = bring_up_loopback(project_path, id) {
-        return e;
-    }
-
     let (_temp_dir, image_path) = match create_checkpoint_image_dir() {
         Ok(v) => v,
         Err(e) => return e,
@@ -251,10 +249,6 @@ pub fn checkpoint_leave_running_work_path_tmp(project_path: &Path, id: &str) -> 
 }
 
 pub fn checkpoint_leave_running(project_path: &Path, id: &str) -> TestResult {
-    if let Err(e) = bring_up_loopback(project_path, id) {
-        return e;
-    }
-
     let (_temp_dir, image_path) = match create_checkpoint_image_dir() {
         Ok(v) => v,
         Err(e) => return e,
@@ -264,10 +258,6 @@ pub fn checkpoint_leave_running(project_path: &Path, id: &str) -> TestResult {
 }
 
 pub fn checkpoint_manage_cgroups_mode_ignore(project_path: &Path, id: &str) -> TestResult {
-    if let Err(e) = bring_up_loopback(project_path, id) {
-        return e;
-    }
-
     let (_temp_dir, image_path) = match create_checkpoint_image_dir() {
         Ok(v) => v,
         Err(e) => return e,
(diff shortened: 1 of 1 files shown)


tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -55,10 +55,12 @@ fn get_container_pid(project_path: &Path, id: &str) -> Result<i32, TestResult> {
     Ok(state.pid.unwrap_or(-1))
 }
 
-// CRIU requires a minimal network setup in the network namespace
-fn setup_network_namespace(project_path: &Path, id: &str) -> Result<(), TestResult> {
+fn prepare_checkpoint(project_path: &Path, id: &str) -> Result<(), TestResult> {
```
nayuta723:
```suggestion
fn bring_up_loopback(project_path: &Path, id: &str) -> Result<(), TestResult> {
```

nayuta723: It might be better to introduce a name like `prepare_checkpoint` once the process becomes more complex.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
nayuta723: It would be better to check the status of the output here.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -435,20 +454,14 @@ pub fn checkpoint_link_remap() -> TestResult {
...
     id: &str,
     port: u16,
     timeout: std::time::Duration,
 ) -> Result<(), TestResult> {
-    // The in-flight connection only exists once lo is up. Until then the clients
-    // running inside the container cannot reach 127.0.0.1, so no connection is
-    // established and nothing accumulates in the accept queue for `ss` to report.
-    // The container itself cannot bring lo up because the default spec grants no
-    // NET_ADMIN, so we do it here from the host before polling. `checkpoint` brings
-    // it up again later, but `ip link set up` is idempotent.
```
nayuta723, line 449: The comment here has been removed.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
```diff
@@ -98,10 +100,6 @@ fn checkpoint(
         Err(e) => return e,
     };
 
-    if let Err(e) = setup_network_namespace(project_path, id) {
-        return e;
-    }
```
nayuta723, line 103: This has been added to all tests, but is it really necessary to set it up every time?

Since it only brings the `lo` link up, I don't think it is needed for tests unrelated to networking.

nayuta723 (commented):
I left some comments, please take a look.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
donkomura: Since the original placement is no longer appropriate, I've moved the call to `wait_in_flight` to its proper location.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
donkomura: I've add a status check.

tests/contest/contest/src/tests/lifecycle/checkpoint.rs
donkomura: Initialy, I thought that `lo` link up was needed to dump in `checkpoint()` based on the comments in the main branch, but upon reviewing CRIU's code, it requests the link up state for lo within netns only.
cf. https://github.com/checkpoint-restore/criu/blob/f4142635867ebc10b6f72d3ec4911e0dd67cb12c/criu/net.c#L912-L913

Based on my research, as you say, we can reduce `lo` link up calls.

nayuta723 (approved):
LGTM, Thanks!

PR #3687, opened 2026-08-13
fix(libcontainer): postpone cgroup limits application to main process

### LINKED ISSUE #3644: [Bug]: Container creation fails when pids.limit is set to 1
### Bug Description

When `linux.resources.pids.limit` is set to `0` in config.json at container **creation** time, `youki create` fails with a fork error in the intermediate process. runc handles this case by writing `pids.max=1` and successfully creating the container, but youki fails outright.

This was discovered in #3634.

### Steps to Reproduce

1. Run `youki spec` to generate the default `config.json`.
2. Edit `config.json` so that it runs a long-lived process and sets `pids.limit` to `0`:
   ```json
   {
     "process": {
       "args": [
         "sleep",
         "1000"
       ]
     },
     "linux": {
       "cgroupsPath": "/runtime-test/update_common_limits",
       "resources": {
         "pids": {
           "limit": 0
         }
       }
     }
   }
   ```
   all other fields left at the `youki spec` default.
3. Run `youki create runtime-test`.
4. Observe that container creation fails.

For comparison, runc succeeds under the same config and ends up with `pids.max=1`:

```sh
$ runc create runtime-test
$ cat /sys/fs/cgroup/runtime-test/update_common_limits/pids.max
1
```

youki fails with:

```sh
$ youki create runtime-test
ERROR libcontainer::process::container_intermediate_process: failed to fork init process: failed to clone process
ERROR libcontainer::process::container_main_process: failed to run intermediate process failed to launch init process
ERROR libcontainer::container::builder_impl: failed to run container process intermediate process error failed to launch init process
ERROR youki: error in executing command: failed to create container: intermediate process error failed to launch init process
Error: failed to create container: intermediate process error failed to launch init process
```

### Expectation

youki should succeed when `pids.limit` is `0`, matching runc's behavior of setting `pids.max=1`

### System and Setup Info

_No response_

### Additional Context

- Related PR: youki-dev/youki#3634
- Related issue: youki-dev/youki#3594

TotyLabs:
Description
Problem
When pids.limit is set to 1 in config.json, container creation fails with a fork/clone error:

ERROR libcontainer::process::container_intermediate_process: failed to fork init process: failed to clone process
This happens because the container_intermediate_process applies the cgroup resource limits (cgroup_manager.apply()) before cloning the init process. Since the intermediate process is already inside the target cgroup, the limit of 1 is enforced immediately. When it attempts to clone the init process, the clone syscall fails because it would temporarily bring the process count in the cgroup to 2, which exceeds the limit of 1.

Solution
To align with runc behavior and allow successful container creation with a pids.limit of 1, we postpone the application of cgroup resource limits:

The intermediate process still joins the cgroup (cmanager.add_task(pid)) before entering namespaces and cloning, ensuring the init process is created inside the correct cgroup. However, it no longer calls .apply().
Once the main process receives the init_pid from the intermediate process (confirming the init process was cloned successfully), the main process instantiates the cgroup manager and calls .apply(...) to write the resource limits.
Linux cgroups allow writing a limit (like pids.max = 1) that is lower than the current count of tasks (which is temporarily 2: intermediate process + init process). The kernel accepts the write without error, and once the intermediate process exits shortly after, the task count drops to 1, correctly enforcing the limit.
Changes
crates/libcontainer/src/process/container_intermediate_process.rs:
Simplified apply_cgroups to only add the task to the cgroup. Removed resources and init parameters.
Cleaned up the unused import of LinuxResources.
Updated and simplified unit tests in the module.
crates/libcontainer/src/process/container_main_process.rs:
Added the Cgroup(String) variant to ProcessError.
Implemented cgroup resource limits application in container_main_process after receiving init_pid for InitContainer.
Testing
Verified the fix with pids.limit = 1.
Ran the libcontainer test suite: 264 tests passed.
Two TTY-related tests fail in the WSL environment because /dev/ptmx is not backed by a real devpts mount; the same failures occur independently of this change.
Related Issues
Closes #3644 
Refs #3594

### DIFF
--- a/crates/liboci-cli/src/update.rs
+++ b/crates/liboci-cli/src/update.rs
@@ -55,7 +55,7 @@ pub struct Update {
     pub memory_swap: Option<i64>,
 
     /// Set the maximum number of processes allowed in the container
-    #[arg(long)]
+    #[arg(long, allow_hyphen_values = true)]
     pub pids_limit: Option<i64>,
 
     /// Set the value for Intel RDT/CAT L3 cache schema.


crates/liboci-cli/src/update.rs
```diff
@@ -55,7 +55,7 @@ pub struct Update {
     pub memory_swap: Option<i64>,
 
     /// Set the maximum number of processes allowed in the container
-    #[arg(long)]
+    #[arg(long, allow_hyphen_values = true)]
```
saku3, line 58: Please don't include any changes unrelated to this issue.

crates/liboci-cli/src/update.rs
```diff
@@ -70,3 +70,25 @@ pub struct Update {
     #[arg(value_parser = clap::builder::NonEmptyStringValueParser::new(), required = true)]
     pub container_id: String,
 }
+
+#[cfg(test)]
```
saku3: This test is also unrelated to the current change, so please don't add it.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -164,6 +166,36 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
     // process.  The intermediate process should exit after this point.
     let init_pid = intermediate_main_receiver.wait_for_intermediate_ready()?;
 
+    if matches!(container_args.container_type, ContainerType::InitContainer) {
```
saku3: I think cgroup_manager.apply should be called at a well-defined synchronization point, preferably immediately before running the prestart hooks / create_runtime, similar to where runc applies its cgroup-related setup.

https://github.com/opencontainers/runc/blob/main/libcontainer/process_linux.go#L1016

At the very least, I don't think it is a good idea to apply the cgroup settings while the init process is running concurrently, as in the current implementation.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -164,6 +166,36 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+        if let Some(linux) = container_args.spec.linux() {
+            if let Some(resources) = linux.resources() {
+                let command = container_args.syscall.create_syscall();
```
saku3: I think this is doing the same thing as `intermediate_process`, so could we extract the common logic into a helper function or something similar?

crates/libcontainer/src/process/container_intermediate_process.rs
saku3, line 322: Since we are not applying the cgroup settings here, could we rename this to something like `add_to_cgroup` or `join_cgroup`?

crates/libcontainer/src/process/container_main_process.rs
saku3: If possible, it would be good to make this cgroup application logic testable, especially for `apply_cgroup_tenant` and `apply_cgroup_no_resources`.

For example:

* Extract this logic into a separate function.
* Make the cgroup manager injectable so that it can be mocked with `TestManager` in unit tests.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -164,6 +166,36 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+                .with_rootless(rootless);
+
+                let controller_opt = libcgroups::common::ControllerOpt {
+                    resources,
+                    freezer_state: None,
+                    oom_score_adj: None,
+                    disable_oom_killer: false,
+                };
+
+                use libcgroups::common::CgroupManager;
```
saku3: Please put the use statements at the top of the file.

crates/libcontainer/src/process/container_main_process.rs
```diff
@@ -164,6 +166,36 @@ pub fn container_main_process(container_args: &ContainerArgs) -> Result<Pid> {
...
+                cgroup_manager.apply(&controller_opt).map_err(|err| {
+                    tracing::error!("failed to apply cgroup in main process: {}", err);
+                    ProcessError::Cgroup(err.to_string())
+                })?;
```
saku3: If cgroup_manager.apply() fails here, do we need to clean up the init process, cgroup or etc?

saku3 (commented):
Thank you for the PR. I left some comments, so please take a look.

PR #3697, opened 2026-08-19
test: port OCI state validation

ITSMERNB:
## Summary

- port the OCI runtime-tools `state` validation into the Rust contest suite
- verify `state` fails without a container ID
- verify `state` fails for a non-existent container
- verify a created container returns parseable OCI state with the expected ID and status

Part of #361.

## Testing

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo check -p contest` cannot complete on native Windows because the contest suite depends on Linux-only crates; Linux CI should validate the build.

Behavioral reference: `opencontainers/runtime-tools/validation/state`

### DIFF
--- a/tests/contest/contest/src/tests/state/state_test.rs
+++ b/tests/contest/contest/src/tests/state/state_test.rs
@@ -1,0 +1,98 @@
+use std::process::{Command, Output};
+
+use anyhow::anyhow;
+use test_framework::{Test, TestGroup, TestResult};
+
+use crate::tests::lifecycle::ContainerLifecycle;
+use crate::utils::{State, get_runtime_path};
+
+fn run_state(container: &ContainerLifecycle, id: Option<&str>) -> std::io::Result<Output> {
+    let mut command = Command::new(get_runtime_path());
+    command
+        .arg("--root")
+        .arg(container.get_project_path().join("runtime"))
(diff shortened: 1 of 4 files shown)


tests/contest/contest/src/tests/state/state_test.rs
```diff
@@ -0,0 +1,98 @@
...
+        Err(err) => TestResult::Failed(anyhow!("failed to execute state command: {err}")),
+    }
+}
+
+fn state_created_container() -> TestResult {
+    let container = ContainerLifecycle::new();
+
+    if !matches!(container.create(), TestResult::Passed) {
+        return TestResult::Failed(anyhow!("failed to create container"));
+    }
```
saku3: I think it would be better to preserve the original error and add context instead of replacing it with a generic error.

```suggestion
    if let TestResult::Failed(err) = container.create() {
        return TestResult::Failed(err.context("failed to create container"));
    }
```

ITSMERNB: Thanks, good point. Updated in 19a707f2 to preserve the original create error and add the state-test context on top. `cargo fmt --all -- --check` and `git diff --check` are both clean.

saku3 (approved):

PR #3698, opened 2026-08-21
feat: add support for `--blkio-weight` in `youki update`

### LINKED ISSUE #3679: [FEATURE]: support --blkio-weight in youki update
### Background

Sub-issue of #147, split per resource as proposed in https://github.com/youki-dev/youki/issues/147#issuecomment-5183952337. This one covers the blkio group.

### Feature Request

Make `--blkio-weight` apply to the running container.

`youki update --blkio-weight 500 <id>` exits 0 without changing `io.weight`. The flag is declared in `crates/liboci-cli/src/update.rs` but never read.

### Proposed Solution

Handle this arg in `crates/youki/src/commands/update.rs`, where the options are turned into `LinuxResources`. The io/blkio controller in `libcgroups` is already there.

Note the flag is `Option<u64>` while `LinuxBlockIo::weight` is `u16`.

### Suggestion
* Please verify that this works as expected.

* If an E2E test for this feature already exists but is currently skipped, please update it so that the test is executed.
  * See https://github.com/youki-dev/youki/issues/3576

* There is no need to make any changes for cgroup v1.

aayush0325:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

Fixes #3679

## Additional Context

I've written the test by reading [crates/libcgroups/src/v2/io.rs#L139](https://github.com/youki-dev/youki/blob/main/crates/libcgroups/src/v2/io.rs#L140) please correct me if my understanding is wrong here.

### DIFF
--- a/crates/libcgroups/src/v2/io.rs
+++ b/crates/libcgroups/src/v2/io.rs
@@ -112,7 +112,8 @@ impl Io {
         if v == 0 {
             return 0;
         }
-        1 + (v.saturating_sub(10)) * 9999 / 990
+        // Widen before multiplying because the intermediate value may exceed u16.
+        (1 + (u32::from(v) - 10) * 9999 / 990) as u16
     }
 
     fn io_max_path(path: &Path) -> PathBuf {
--- a/crates/libcgroups/src/v2/io.rs
+++ b/crates/libcgroups/src/v2/io.rs
@@ -21,6 +21,8 @@ pub enum V2IoControllerError {
     WrappedIo(#[from] WrappedIoError),
     #[error("cannot set leaf_weight with cgroupv2")]
     LeafWeight,
+    #[error("invalid weight {0}: must be in range [10, 1000]")]
+    InvalidWeight(u16),
 }
 
 pub struct Io {}
@@ -138,9 +140,12 @@ impl Io {
             }
         }
         if let Some(io_weight) = blkio.weight() {
-            // be aligned with what runc does
-            // See also: https://github.com/opencontainers/runc/blob/81044ad7c902f3fc153cb8ffadaf4da62855193f/libcontainer/cgroups/fs2/io.go#L57-L69
-            if io_weight > 0 {
+            if io_weight != 0 {
+                if !(10..=1000).contains(&io_weight) {
+                    return Err(V2IoControllerError::InvalidWeight(io_weight));
+                }
+                // be aligned with what runc does
+                // See also: https://github.com/opencontainers/runc/blob/81044ad7c902f3fc153cb8ffadaf4da62855193f/libcontainer/cgroups/fs2/io.go#L57-L69
                 let cgroup_file = root_path.join(CGROUP_BFQ_IO_WEIGHT);
                 if cgroup_file.exists() {
                     common::write_cgroup_file(cgroup_file, io_weight)?;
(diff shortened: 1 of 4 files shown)


crates/youki/src/commands/update.rs
```diff
@@ -23,6 +25,17 @@ pub fn update(args: Update, root_path: PathBuf) -> Result<()> {
         };
     } else {
         let mut builder = LinuxResourcesBuilder::default();
+        if let Some(blkio_weight) = args.blkio_weight {
+            if !(10..=1000).contains(&blkio_weight) {
```
aayush0325, line 29: please let me know if we should fail here or let it pass, runc doesn't fail here: [runc/update.go#L183](https://github.com/opencontainers/runc/blob/main/update.go#L183)

crates/libcgroups/src/v2/io.rs
```diff
@@ -112,7 +112,8 @@ impl Io {
         if v == 0 {
             return 0;
         }
-        1 + (v.saturating_sub(10)) * 9999 / 990
+        // Widen before multiplying because the intermediate value may exceed u16.
```
aayush0325, line 115: This was causing the CI failures. updated to match [runc/vendor/github.com/opencontainers/cgroups/utils.go#L478](https://github.com/opencontainers/runc/blob/main/vendor/github.com/opencontainers/cgroups/utils.go#L478)

aayush0325: please let me know if this is in scope for this PR or should I raise a separate PR for this!

saku3: Thank you for the fix. It’s fine to include this change in this PR.

It would be better to add an explicit range check like this:

```rust
if !(10..=1000).contains(&v) {
    return Err(...);
}
```

Please also refer to this comment.

https://github.com/youki-dev/youki/pull/3698/changes#r3994939252

crates/youki/src/commands/update.rs
saku3: I checked the behavior and found that the following command fails with youki:

```bash
youki update container --blkio-weight 0
```

while the equivalent runc command succeeds:

```bash
runc update container --blkio-weight 0
```

Considering compatibility with runc, I think it would be better to add the validation in `crates/libcgroups/src/v2/io.rs` rather than in the `update` command.

Also, this range validation is needed not only for the `update` command, but for the `run / start` command as well.

So, I think the following approach would be better.


* `0` → accepted by the `update` command(no-op)
* `10..=1000` → treated as a valid weight and converted/applied
* `1..=9` / `1001..` → explicitly rejected with an error in `libcgroups`

tests/contest/contest/src/tests/update/blkio.rs
```diff
@@ -0,0 +1,139 @@
...
+};
+use test_framework::{TestResult, test_result};
+
+use super::update_container_and_wait;
+use crate::utils::test_utils::check_container_created;
+use crate::utils::{
+    start_container, test_outside_container, update_container, update_container_with_stdin,
+};
+
+const INITIAL_WEIGHT: u16 = 100;
```
saku3: Since the default value of `io.bfq.weight` is `100`, I think it would be better to use a different value, such as `200`, for the initial weight. Otherwise, the test could pass even if the weight was not actually applied.

tests/contest/contest/src/tests/update/blkio.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    if status.success() {
+        bail!("expected --blkio-weight {weight} to fail, but it succeeded");
+    }
+
+    Ok(())
+}
+
+// "update cgroup blkio weight"
+pub(crate) fn update_blkio_weight_test() -> TestResult {
```
saku3, line 67: Please also add a test to verify that updating the weight to `0` succeeds and leaves the current value unchanged.

saku3 (commented):
Sorry for the delayed review.

I left a few comments. I’d appreciate it if you could take a look.

tests/contest/contest/src/tests/update/blkio.rs
```diff
@@ -0,0 +1,150 @@
...
+        test_result!(check_blkio_weight(&cgroup_path, INITIAL_WEIGHT));
+
+        // Updating to 0 must succeed and leave the value unchanged.
+        test_result!(update_container_and_wait(id, dir, &["--blkio-weight", "0"]));
+        test_result!(check_blkio_weight(&cgroup_path, INITIAL_WEIGHT));
+
+        // Out of range values must be rejected and leave the weight unchanged.
+        // youki enforces OCI spec range [10, 1000] in libcgroups; runc
+        // delegates to the kernel (io.bfq.weight allows [1, 10000]).
+        if is_runtime_youki() {
```
saku3: Note for maintainers:

runc accepts `--blkio-weight` values in the range 1–9 when BFQ is enabled.

However, Docker validates `--blkio-weight` to the range [10, 1000], so I don't think there is much value in accepting values below 10. Therefore, I think the current validation of [10, 1000] is reasonable.

Without this validation, users may receive a less clear error from the kernel or cgroup interface, so validating the value in youki also provides a more understandable error message.

https://docs.docker.com/reference/cli/docker/container/run/#options
https://docs.docker.com/reference/cli/docker/container/update/#options

runc itself also expects the value to be in the range [10, 1000].

https://github.com/opencontainers/runc/blob/930f199c7fcaebb65ec3a913917fd428fc3f7742/update.go#L64

If you think this behavior is a problem, please open a separate issue to discuss it.

ref: https://github.com/opencontainers/runtime-spec/pull/780

saku3: I don't think lines 141–146 need to be covered by the contest tests.

Since this is validation logic specific to youki, I think unit test coverage should be sufficient.

aayush0325: Removing.

saku3 (approved):
Thanks

PR #3699, opened 2026-08-22
feat: implement OCI per-mount propagation options

### LINKED ISSUE #3604: Implement  OCI per-mount propagation options such as rshared and rslave

saku3:
## Description

implement OCI per-mount propagation options

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Fixes #3604

### DIFF
--- a/crates/libcontainer/src/rootfs/mount.rs
+++ b/crates/libcontainer/src/rootfs/mount.rs
@@ -392,6 +392,7 @@ impl Mount {
             flags: MsFlags::MS_NOEXEC | MsFlags::MS_NOSUID | MsFlags::MS_NODEV,
             data: vec![data.into_owned()],
             rec_attr: None,
+            propagation_flags: vec![],
         };
 
         self.mount_into_container(
@@ -800,6 +801,48 @@ impl Mount {
                 }
                 Err(e) => return Err(e.into()),
             }
+        }
+
+        self.apply_mount_propagation(
+            rootfs,
+            container_dest,
+            &mount_option_config.propagation_flags,
+        )?;
+
+        Ok(())
+    }
+
+    fn apply_mount_propagation(
+        &self,
+        rootfs: &Path,
+        container_dest: &Path,
+        propagation_flags: &[MsFlags],
+    ) -> Result<()> {
+        if propagation_flags.is_empty() {
+            return Ok(());
+        }
+
+        let root = Root::open(rootfs)?;
+        let dest: OwnedFd = root.resolve(container_dest)?.into();
+
+        for flags in propagation_flags {
+            let mut setattr_flags = linux::AT_EMPTY_PATH;
+            if flags.contains(MsFlags::MS_REC) {
+                setattr_flags |= linux::AT_RECURSIVE;
+            }
+            let mount_attr = linux::MountAttr {
+                attr_set: 0,
+                attr_clr: 0,
+                propagation: (*flags & !MsFlags::MS_REC).bits(),
+                userns_fd: 0,
+            };
+            self.syscall.mount_setattr(
+                dest.as_fd(),
+                Path::new(""),
+                setattr_flags,
+                &mount_attr,
+                mem::size_of::<linux::MountAttr>(),
+            )?;
         }
 
         Ok(())
@@ -1422,6 +1465,7 @@ mod tests {
             flags,
             data: vec![],
             rec_attr: None,
+            propagation_flags: vec![],
         };
         mounter
             .mount_cgroup_v2(&spec_cgroup_mount, &mount_opts, &mount_option_config)
--- a/crates/libcontainer/src/rootfs/utils.rs
+++ b/crates/libcontainer/src/rootfs/utils.rs
@@ -21,6 +21,10 @@ pub struct MountOptionConfig {
 
     /// RecAttr represents mount properties to be applied recursively.
     pub rec_attr: Option<linux::MountAttr>,
+
+    /// Mount propagation flags, kept separate from regular mount flags because
+    /// they are applied after the mount is attached.
+    pub propagation_flags: Vec<MsFlags>,
 }
 
 pub fn default_devices() -> Vec<LinuxDevice> {
@@ -87,6 +91,7 @@ pub fn to_sflag(dev_type: LinuxDeviceType) -> SFlag {
 
 pub fn parse_mount(m: &Mount) -> std::result::Result<MountOptionConfig, MountError> {
     let mut flags = MsFlags::empty();
+    let mut propagation_flags = Vec::new();
     let mut data = Vec::new();
     let mut mount_attr: Option<linux::MountAttr> = None;
 
@@ -167,14 +172,17 @@ pub fn parse_mount(m: &Mount) -> std::result::Result<MountOptionConfig, MountErr
                     MountOption::Nodiratime(is_clear, flag) => Some((is_clear, flag)),
                     MountOption::Bind(is_clear, flag) => Some((is_clear, flag)),
                     MountOption::Rbind(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Unbindable(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Runbindable(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Private(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Rprivate(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Shared(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Rshared(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Slave(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Rslave(is_clear, flag) => Some((is_clear, flag)),
+                    MountOption::Unbindable(_, flag)
+                    | MountOption::Runbindable(_, flag)
+                    | MountOption::Private(_, flag)
+                    | MountOption::Rprivate(_, flag)
+                    | MountOption::Shared(_, flag)
+                    | MountOption::Rshared(_, flag)
+                    | MountOption::Slave(_, flag)
+                    | MountOption::Rslave(_, flag) => {
+                        propagation_flags.push(flag);
+                        continue;
+                    }
                     MountOption::Relatime(is_clear, flag) => Some((is_clear, flag)),
                     MountOption::Norelatime(is_clear, flag) => Some((is_clear, flag)),
                     MountOption::Strictatime(is_clear, flag) => Some((is_clear, flag)),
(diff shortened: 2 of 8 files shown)


crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -802,6 +803,58 @@ impl Mount {
...
+        self.apply_mount_propagation(
+            rootfs,
+            container_dest,
+            &mount_option_config.propagation_flags,
+        )?;
+
+        Ok(())
+    }
+
+    fn apply_mount_propagation(
```
nayuta723, line 815: `apply_mount_propagation` uses `mount_setattr(2)` for propagation flags (MS_SHARED, MS_SLAVE, etc.). However, `mount_setattr(2)` was added in Linux 5.12 (https://man7.org/linux/man-pages/man2/mount_setattr.2.html), and is not available on kernels still widely in use — notably Ubuntu 20.04 LTS (5.4), Debian 11 (5.10), Amazon Linux 2 (5.10), and RHEL 8 (4.18).

Neither runc nor crun uses `mount_setattr(2)` for propagation:

- **runc** uses `mount_setattr(2)` exclusively for `RecAttr` ([setRecAttr, rootfs_linux.go:1475](https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/libcontainer/rootfs_linux.go#L1475)) — attributes that cannot be applied recursively and atomically via `mount(2)`. All other mount attributes (MS_RDONLY, MS_NOSUID, etc.) are set via `mount(2)` flags. Propagation flags are also applied via `mount(2)` in `mountPropagate()` ([rootfs_linux.go:1458](https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/libcontainer/rootfs_linux.go#L1458)):
  ```go
  for _, pflag := range m.PropagationFlags {
      mountViaFds("", nil, m.Destination, dstFd, "", uintptr(pflag), "")
  }
  ```

- **crun** similarly retains a `mount(2)` path for propagation inside `do_mount()` ([linux.c:1317-1322](https://github.com/containers/crun/blob/dcd0f4055b1116dedeae9d501678982fa6980e7a/src/libcrun/linux.c#L1317-L1322)):
  ```c
  if (mountflags & ALL_PROPAGATIONS_NO_REC)
      mount(NULL, real_target, NULL, mountflags & ALL_PROPAGATIONS, NULL);
  ```

Note that youki already uses `mount_setattr(2)` in other parts of the mount flow. For mount attribute setting ([L671](https://github.com/youki-dev/youki/blob/be6fbe6cb830dd34a483ba0ffd79e7a60a25aa49/crates/libcontainer/src/rootfs/mount.rs#L671), [L760](https://github.com/youki-dev/youki/blob/be6fbe6cb830dd34a483ba0ffd79e7a60a25aa49/crates/libcontainer/src/rootfs/mount.rs#L760)), youki uses `mount_setattr` as part of its new mount API (`open_tree → mount_setattr → move_mount`), which has no equivalent in runc (runc uses `mount(2)` flags for attributes). For RecAttr ([L681](https://github.com/youki-dev/youki/blob/be6fbe6cb830dd34a483ba0ffd79e7a60a25aa49/crates/libcontainer/src/rootfs/mount.rs#L681), [L770](https://github.com/youki-dev/youki/blob/be6fbe6cb830dd34a483ba0ffd79e7a60a25aa49/crates/libcontainer/src/rootfs/mount.rs#L770)), the `AT_RECURSIVE` usage is consistent with runc's `setRecAttr`. The issue is specifically that PR #3699 extends `mount_setattr(2)` to propagation flags, whereas both runc and crun use `mount(2)` for that purpose.

Since `MS_REC` can be passed directly to `mount(2)`, using `mount(NULL, dest, NULL, propagation_flag, NULL)` would be sufficient and avoids the Linux 5.12 requirement for this step.

saku3: In terms of avoiding the Linux 5.12 requirement, doesn’t that have little practical meaning for youki as a whole, since `mount_setattr` is already used in other parts of the runtime?

If the only reason is simply to align the implementation with runc and crun, I’m not particularly convinced that this is worth doing.

If we switch this back to mount(2), how do we preserve the fd-based path resolution guarantees we get from mount_setattr? As far as I can tell, youki does not currently have a helper equivalent to runc's handling for safely applying a legacy mount operation to an already-resolved fd-backed target.

Even if we decide to implement this using `mount`, I think it should be treated as a fallback path and addressed in a separate issue.

nayuta723: Thanks, this is a good point. Sorry, I realize that my comment may not have made my intention clear.

My concern was not simply to align with `runc` or `crun`. I think we should separately discuss which kernel versions `youki` itself should support. Since `youki` already uses `mount_setattr(2)` in other parts of the runtime, switching this path to `mount(2)` would not meaningfully change the minimum kernel version for `youki` as a whole.

For now, I think it is reasonable to support kernel 5.12 and later. My main goal was to make sure we were aligned on the support policy.

For this PR, I’m fine with keeping `mount_setattr(2)` and considering the legacy `mount(2)` fallback, including safe fd-based path resolution, out of scope. We can discuss that separately in another issue.

crates/libcontainer/src/rootfs/mount.rs
```diff
@@ -802,6 +803,58 @@ impl Mount {
...
+        for flags in propagation_flags {
+            let propagation = *flags
+                & (MsFlags::MS_SHARED
+                    | MsFlags::MS_SLAVE
+                    | MsFlags::MS_PRIVATE
+                    | MsFlags::MS_UNBINDABLE
+                    | MsFlags::MS_REC);
+            if propagation.is_empty() {
+                continue;
+            }
```
nayuta723: As far as I understand, this operation is a no-op. Are there any cases where flags other than this one can be set?

crates/libcontainer/src/rootfs/utils.rs
```diff
@@ -167,14 +172,38 @@ pub fn parse_mount(m: &Mount) -> std::result::Result<MountOptionConfig, MountErr
...
-                    MountOption::Private(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Rprivate(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Shared(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Rshared(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Slave(is_clear, flag) => Some((is_clear, flag)),
-                    MountOption::Rslave(is_clear, flag) => Some((is_clear, flag)),
+                    MountOption::Unbindable(_, _) => {
+                        propagation_flags.push(MsFlags::MS_UNBINDABLE);
+                        continue;
+                    }
```
nayuta723: Is there any reason why we can’t write it this way?
```suggestion
                    MountOption::Unbindable(_, flag) => {
                        propagation_flags.push(flag);
                        continue;
                    }
```

crates/libcontainer/src/rootfs/utils.rs
```diff
@@ -448,9 +485,19 @@ mod tests {
                     | MsFlags::MS_NOATIME
                     | MsFlags::MS_NODIRATIME
                     | MsFlags::MS_BIND
-                    | MsFlags::MS_UNBINDABLE,
+                    | MsFlags::MS_REC,
```
nayuta723, line 467: Does changing this mean that the existing behavior will also change? Is this change acceptable?

tests/contest/contest/src/tests/mount_propagation/mount_propagation_test.rs
```diff
@@ -0,0 +1,143 @@
...
+        Box::new(unbindable_test),
+    );
+    let runbindable_test = Test::new(
+        "mount_propagation_runbindable_test",
+        Box::new(runbindable_test),
+    );
+
+    test_group.add(vec![
+        Box::new(shared_test),
+        Box::new(rshared_test),
```
nayuta723, line 139: Please add the test case for slave.

nayuta723 (changes requested):
I left some comments. Please take a look.

crates/libcontainer/src/rootfs/mount.rs
saku3: Thanks. Fixed.

crates/libcontainer/src/rootfs/utils.rs
saku3: fixed.

tests/contest/contest/src/tests/mount_propagation/mount_propagation_test.rs
saku3: Added.

crates/libcontainer/src/rootfs/utils.rs
saku3: `MsFlags::MS_REC` is included here because of `rbind`.

`rbind` reads the flags directly, so there is no issue:
https://github.com/youki-dev/youki/blob/a46e6a1f9e41417c84d6511ca6ac085fec0591d3/crates/libcontainer/src/rootfs/mount.rs#L640

`MS_REC` is supposed to be included here. Previously, however, propagation flags were not handled correctly, so it was missing.

The `mount_recursive` tests ensure that this change does not alter the existing behavior.

https://github.com/youki-dev/youki/tree/main/tests/contest/contest/src/tests/mounts_recursive

crates/libcontainer/src/rootfs/mount.rs
saku3: I created an issue here.

https://github.com/youki-dev/youki/issues/3726

nayuta723 (approved):
Sorry for the delay. LGTM!

PR #3700, opened 2026-08-22
feat: Support --empty-ns checkpoint/restore option

donkomura:
## Description

Pass --empty-ns from the CLI to CRIU so the network namespace is
created empty on restore instead of being dumped. Youki does not
manage network devices, so their external dependencies cannot be
described to CRIU. Runc sets this unconditionally for the same reason.
To verifty the functionality, contest check the existence of
netdev-*.img which holds the network devices of a dumped namespace.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [ ] Ran existing test suite

## Related Issues

### DIFF
--- a/crates/liboci-cli/src/checkpoint.rs
+++ b/crates/liboci-cli/src/checkpoint.rs
@@ -51,9 +51,9 @@ pub struct Checkpoint {
     // pub pre_dump: bool,
     #[arg(long, default_value = "soft", value_parser = clap::builder::PossibleValuesParser::new(["ignore", "full", "strict", "soft"]))]
     pub manage_cgroups_mode: String,
-    // TODO: Checkpoint a namespace, but don't save its properties
-    // #[arg(long)]
-    // pub empty_ns: bool,
+    // Checkpoint a namespace, but don't save its properties
+    #[arg(long, default_value = "network", value_parser = clap::builder::PossibleValuesParser::new(["network"]))]
+    pub empty_ns: String,
     // TODO: Enable auto-deduplication
     // #[arg(long)]
     // pub auto_dedup: bool,
--- a/crates/liboci-cli/src/checkpoint.rs
+++ b/crates/liboci-cli/src/checkpoint.rs
@@ -51,8 +51,11 @@ pub struct Checkpoint {
     // pub pre_dump: bool,
     #[arg(long, default_value = "soft", value_parser = clap::builder::PossibleValuesParser::new(["ignore", "full", "strict", "soft"]))]
     pub manage_cgroups_mode: String,
-    // Checkpoint a namespace, but don't save its properties
-    #[arg(long, default_value = "network", value_parser = clap::builder::PossibleValuesParser::new(["network"]))]
+    /// Checkpoint a namespace, but don't save its properties
+    ///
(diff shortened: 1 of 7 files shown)


crates/youki/src/commands/checkpoint.rs
```diff
@@ -20,6 +20,7 @@ pub fn checkpoint(args: Checkpoint, root_path: PathBuf) -> Result<()> {
         work_path: args.work_path,
         manage_cgroups_mode: parse_cgroups_mode(&args.manage_cgroups_mode)?,
         link_remap: args.link_remap,
+        empty_net_ns: args.empty_ns == "network",
```
nayuta723, line 23: Please define `"network"` as a constant.

crates/liboci-cli/src/checkpoint.rs
```diff
@@ -51,9 +51,9 @@ pub struct Checkpoint {
     // pub pre_dump: bool,
     #[arg(long, default_value = "soft", value_parser = clap::builder::PossibleValuesParser::new(["ignore", "full", "strict", "soft"]))]
     pub manage_cgroups_mode: String,
-    // TODO: Checkpoint a namespace, but don't save its properties
-    // #[arg(long)]
-    // pub empty_ns: bool,
+    // Checkpoint a namespace, but don't save its properties
```
nayuta723, line 54:
```suggestion
    /// Checkpoint a namespace, but don't save its properties
```

crates/liboci-cli/src/checkpoint.rs
```diff
@@ -51,9 +51,9 @@ pub struct Checkpoint {
...
+    #[arg(long, default_value = "network", value_parser = clap::builder::PossibleValuesParser::new(["network"]))]
+    pub empty_ns: String,
```
nayuta723, line 56: As with `runc`, please implement this so that `StringValue` is rejected with `not supported` for namespaces other than `network`.

crates/liboci-cli/src/checkpoint.rs
nayuta723, line 54: Please also document what can be configured for this flag. Currently, `network` is always specified by default, so please mention that it is applied even when the flag is not explicitly set.

nayuta723 (changes requested):
Sorry for the delayed response. I’ve left some comments. Please take a look when you have a chance.

crates/liboci-cli/src/checkpoint.rs
donkomura: Fixed.

crates/liboci-cli/src/checkpoint.rs
donkomura: I added a validation in youki to check the empty-ns value, and reject the other excluding `network`.

crates/liboci-cli/src/checkpoint.rs
donkomura: I leave a comment to mention it.

nayuta723 (approved):
LGTM, Thanks!

PR #3707, opened 2026-08-24
test(utils): Add normalize method unit test in libcontainer/utils.rs

Scanf-s:
## Description

- The normalize() method in utils.rs has no unit tests at all.
- To improve test coverage and prevent malfunctioning behavior when using the normalize method, add a normalize method unit test with several path cases.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite

## Related Issues

Refer to #279

### DIFF
--- a/crates/libcontainer/src/utils.rs
+++ b/crates/libcontainer/src/utils.rs
@@ -95,7 +95,6 @@ impl PathBufExt for Path {
     // adapted from https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
     fn normalize(&self) -> PathBuf {
         let mut components = self.components().peekable();
-
         let mut ret = if let Some(c @ Component::Prefix(..)) = components.peek().cloned() {
             components.next();
             PathBuf::from(c.as_os_str())
--- a/crates/libcontainer/src/utils.rs
+++ b/crates/libcontainer/src/utils.rs
@@ -705,21 +705,19 @@ mod tests {
             ("/a/b/c", "/a/b/c"),
             ("/a/./b", "/a/b"),
             ("/a/b/../c", "/a/c"),
-            ("/proc/self/../1", "/proc/1"),
-            
+            ("/asd/dfg/../gjoi", "/asd/gjoi"),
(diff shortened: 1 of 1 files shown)


crates/libcontainer/src/utils.rs
```diff
@@ -95,6 +95,7 @@ impl PathBufExt for Path {
     // adapted from https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
     fn normalize(&self) -> PathBuf {
         let mut components = self.components().peekable();
+
```
saku3: This change is not necessary.

crates/libcontainer/src/utils.rs
```diff
@@ -693,4 +694,36 @@ mod tests {
...
+            ("/a/b/../c", "/a/c"),
+            ("/asd/dfg/../gjoi", "/asd/gjoi"),
+            // paths processed first by components()
+            ("/a/b/", "/a/b"),
+            ("/a//b", "/a/b"),
+            // relative paths
+            ("", ""),
+            (".", ""),
+            ("..", ""),
+            // Leading '..' is just dropped under current logic, leaving only 'a'
```
saku3: What is the intention behind this comment?

What does "under current logic" mean here? Is this behavior intentional, or are you suggesting that the logic may be changed in the future

Scanf-s: Hi, good afternoon.

I added this comment to clarify the function's existing behavior. When the leading `..` (ParentDir) is processed in this function, the PathBuf is still empty, so pop() has no effect and as the result, ../a becomes just `a` (cuz `Normal` would just push an element into the PathBuf).

And I wasn't suggesting a future change or something make a claim about it. 
If you think the comment is unnecessary or it makes misunderstanding, I'll reword or remove the comment to make it clear!

saku3: Please explain why leading `..` components are dropped, or remove this comment.

saku3 (approved):

PR #3721, opened 2026-09-06
feat(systemd): implement DeviceAllow/DevicePolicy cgroup controller

### LINKED ISSUE #3589: [FEATURE]:add missing support for devices in systemd
### Background

Quick map of how devices are wired in youki:

- **cgroup v1** — `crates/libcgroups/src/v1/devices.rs` (writes `devices.allow`/`devices.deny`)
- **cgroup v2 (fs driver)** — `crates/libcgroups/src/v2/devices/` (controller.rs / bpf.rs / program.rs / emulator.rs); attaches a BPF program to the cgroup
- **systemd cgroup driver** — no devices module here. In `systemd/manager.rs::apply()`, the controller match handles Cpu / CpuSet / Pids / Memory / Io and falls through to `Unified::apply`, but **there is no Devices arm**. So youki does not currently translate device rules into systemd properties (`DeviceAllow=` / `DevicePolicy=`) the way runc does in `cgroups/devices/systemd.go::findDeviceGroup`.

This has been already been discussed in #3001

### Feature Request

Add the missing support for devices in systemd so that it support device allow and deny while using systemD

_No response_

### Proposed Solution

_No response_

### Considerations

_No response_

### Additional Context

_No response_

nayuta723:
## Description

The systemd cgroup manager applied no device restrictions, so a container started with `--systemd-cgroup` could reach any host device through `mknod`.

The spec's rules now become `DevicePolicy` and `DeviceAllow`, with the eBPF device filter attached over the program systemd derives from that list. Both layers are needed, the same way runc needs them: `DeviceAllow` is a whitelist, so a rule that takes access away cannot go in it. When the rules contain one, the unit gets a deny-all list and the filter carries the rules. The D-Bus changes are what that needs.

`cgroupsv2_devices` was not enabled by any default, so this turns the eBPF filter on for the cgroupfs v2 path as well. Can be split out if you would rather land it separately.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test updates
- [ ] CI/CD related changes

## Testing

- [x] Added new unit tests
- [x] Ran existing test suite
- [ ] Tested manually

## Related Issues

Closes #3589


crates/libcgroups/src/systemd/dbus_native/serialize.rs
```diff
@@ -462,6 +467,14 @@ impl DbusSerialize for Variant {
                 buf.push(0);
                 s.serialize(buf);
             }
+            Self::ArrayStructSS(s) => {
+                let sub_type = <Vec<Structure<String>>>::get_signature();
```
saku3, line 471: Not necessary for this PR, but since the logic is identical except for the type parameter, it might be worth extracting it into a helper function.

crates/libcgroups/src/systemd/manager.rs
```diff
@@ -570,6 +580,16 @@ impl CgroupManager for Manager {
                 .set_unit_properties(&self.unit_name, &properties)?;
         }
 
+        // systemd derives a device filter from DeviceAllow and attaches it every time
+        // unit properties are set, so this has to run afterwards to replace it. A rule
+        // DeviceAllow cannot express, a deny rule above all, is only honoured by the
+        // program built here.
+        #[cfg(feature = "cgroupsv2_devices")]
+        crate::v2::devices::Devices::apply_devices(
```
saku3, line 588: This error needs to be ignored in rootless mode. Please refer to the following:

https://github.com/opencontainers/cgroups/blob/30a2293c0b0325280ea70a86fe991ec1e87b5d00/fs2/fs2.go#L241

crates/libcgroups/src/v2/devices/controller.rs
```diff
@@ -44,10 +44,57 @@ impl Controller for Devices {
...
+            .all(grants_full_access)
+}
+
 impl Devices {
     pub fn apply_devices(
         cgroup_root: &Path,
         linux_devices: &Option<Vec<LinuxDeviceCgroup>>,
+    ) -> Result<(), DevicesControllerError> {
+        match Self::attach_device_filter(cgroup_root, linux_devices) {
+            Err(err) if can_skip_ebpf_error(linux_devices) => {
```
saku3, line 87: When I tried to start a rootless container, I encountered the following error:

```
ERROR libcontainer::process::container_intermediate_process: failed to apply cgroup pid=Pid(207844) err=Systemd(EbpfDevices(Bpf(FailedToIncreaseRLimit))) init=true
ERROR libcontainer::process::container_main_process: failed to run intermediate process cgroup error: in ebpf devices controller: bpf error: Failed to increase rlimit
ERROR libcontainer::container::builder_impl: failed to run container process received unexpected message: OtherError("cgroup error: in ebpf devices controller: bpf error: Failed to increase rlimit"), expected: WriteMapping
```

The FailedToIncreaseRLimit error originates from the call to bpf_prog::bump_memlock_rlimit()? on line 126.

However, runc ignores errors from the equivalent operation.
https://github.com/opencontainers/cgroups/blob/30a2293c0b0325280ea70a86fe991ec1e87b5d00/devices/ebpf_linux.go#L347

Also, since cgroup `apply` is performed before the intermediate process enters the user namespace, `in_user_namespace()` always returns `false` in `can_skip_ebpf_error()`. (However, `can_skip_ebpf_error` appears to use the same logic as runc, so it seems that this check may not be functioning as intended in runc either.)

I think errors from `apply_devices()` also need to be ignored in rootless mode by the caller.
Please refer to the comment below.
https://github.com/youki-dev/youki/pull/3721/changes#r4062246162

tests/contest/contest/src/tests/cgroups/systemd_devices.rs
```diff
@@ -0,0 +1,344 @@
...
+/// The systemd cgroup manager needs a unified cgroup hierarchy, a running systemd to talk
+/// to, and root to place units in system.slice.
+fn systemd_cgroups_available() -> bool {
+    nix::unistd::geteuid().is_root() && is_cgroup_v2() && Path::new("/run/systemd/system").is_dir()
+}
+
+fn can_run() -> bool {
+    // DeviceAllow is youki specific enough (property names and defaults) that this is not
+    // run against other runtimes.
+    is_runtime_youki() && systemd_cgroups_available()
```
saku3, line 312: I'm not sure why the tests are restricted to running only with youki. I think these tests should generally also be run in `validate-contest-runc`.

In fact, when I removed the guard and ran `validate-contest-runc`, 3 of the 4 tests passed. The only failure was `deny_rule_on_allow_all_is_enforced`, which failed due to a warning logged by runc, rather than a device enforcement issue.


[NOTE]
Implementation differences may arise with the following rule combination, which is still marked as a FIXME in the youki code:
`deny-all + allow X + deny Y`
In this case, the explicit `deny Y` rule is redundant, so I think runc would allow X through `DeviceAllow`, whereas youki uses eBPF.

tests/contest/contest/src/tests/cgroups/systemd_devices.rs
```diff
@@ -0,0 +1,344 @@
+//! Tests for the systemd cgroup manager's device controller.
```
saku3, line 1: Personally, I think it would be better to organize the directory structure as follows. What do you think?
I think separating the directories, as in `libcgroups`, would make the structure easier to understand.

I think we could move the common systemd-related functions, such as `unit_name` and `show_property`, into `mod.rs`.

I'll move the existing `cpu` directory directly under the `cgroups` directory.

```
tests/cgroups/
├── mod.rs   # v2 test
│
├── cpu.rs
├── cpuset.rs
├── memory.rs
├── pids.rs
├── io.rs
├── hugetlb.rs
├── devices.rs
├── unified.rs
│
└── systemd/
    ├── mod.rs  # systemd test
```

saku3 (commented):
Thanks for the careful implementation! 
I've left some review comments. Could you please take a look?

tests/contest/contest/src/tests/cgroups/systemd_devices.rs
nayuta723: Sorry, I overlooked it. I’ll check it.

PR #3739, opened 2026-09-20
fix(contest): skip memory_policy tests that need NUMA

### LINKED ISSUE #3735: [Bug]: contest memory_policy tests fail on kernels without NUMA
On a kernel built without NUMA, `/sys/devices/system/node` does not exist and `set_mempolicy` returns ENOSYS. The `memory_policy` tests set a policy and expect it to work, so four of them fail: `bind_static`, `default_with_missing_nodes_ok`, `interleave_without_flags`, `preferred_relative`.

Seen on Ubuntu 26.04 arm64 (OrbStack), kernel 7.0.14:

```
container stderr was not empty, found : ERROR libcontainer::process::memory_policy: failed to set memory policy err=Nix(ENOSYS)
```

youki looks right here, it only calls the syscall when the spec asks for one. The group just needs a condition like the tlb tests have, so it skips when the kernel has no NUMA support.

Happy to send the PR.

r14dd:
Closes #3735.

The `memory_policy` group registers every test with `Test::new`, so the four
cases that expect `set_mempolicy` to succeed also run on kernels built without
NUMA. There the syscall returns ENOSYS and they fail, when they should skip.

This gates those four on `/sys/devices/system/node` existing, the same shape as
`check_hugetlb` in the tlb tests. The five argument validation cases expect the
call to fail either way, so they stay unconditional.

Checked the directory rather than a node count on purpose: a single node kernel
still implements the syscall and should run these tests. What is missing here is
NUMA support itself.

On Ubuntu 26.04 arm64 (OrbStack), kernel 7.0.14, no `/sys/devices/system/node`:

```
# Start group memory_policy
1 / 9 : bind_static : skipped
2 / 9 : bind_way_too_large_node_number : ok
3 / 9 : default_with_missing_nodes_ok : skipped
4 / 9 : interleave_without_flags : skipped
5 / 9 : invalid_flag_string : ok
6 / 9 : invalid_mode_string : ok
7 / 9 : missing_mode_but_nodes_present : ok
8 / 9 : preferred_relative : skipped
9 / 9 : syscall_invalid_arguments : ok
# End group memory_policy
```

Before the change all four skipped ones failed.


r14dd:
CI is red on three jobs, none of them from this change.

- `label`: the PR has no `kind/*` label and I cannot add one. Could a maintainer label it?
- `validate`: `exec/terminal_exec_no_console_socket` on runc. The expected `TTY` line is in the captured output, buried in runc's debug log, so the bridging assertion looks flaky rather than broken.
- `oci-validation-rust (x86_64, musl)`: `checkpoint with tcp-skip-in-flight` and `kill`.

All three pass on other recent PRs and none of them touch memory_policy.

The `memory_policy` group itself is green. On `oci-validation-rust (x86_64, musl)` all 9 tests ran and passed, which is the point of the change: the condition only skips where `/sys/devices/system/node` is missing.

Happy to rebase or for someone to re-run the failed jobs.

saku3 (approved):
Thanks.

PR #3741, opened 2026-09-21
test(contest): add cgroup v2 pids integration tests

ARMeeru:
## Description

Adds `cgroup_v2_pids` to the contest suite, following the layout and gating of the cgroup v2 memory tests: each test creates a container with a pids limit in the spec and asserts the value written to `pids.max` under the container's cgroup. Gating uses the shared `is_cgroup_v2_with_controller(ControllerType::Pids)` helper.

Two mappings are covered: a positive limit is written as-is, and a negative limit becomes `"max"`. A limit of `0` is written as `1` (matching runc), but creation currently fails under that cap (#3644), so that case is not asserted and is left to the fix.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test updates
- [ ] CI/CD related changes

## Testing

- [ ] Added new unit tests
- [x] Ran existing test suite

Full contest suite against a release youki build, aarch64 Fedora 44, cgroup v2: 52 ok, 6 skipped, 1 not ok. The one failure is `exec/mount_test` ("expected exactly 1 mountinfo entry for /proc/acpi, got 0"), environmental: this host has no ACPI, and the assertion is unrelated to this change. The `cgroup_v2_pids` group passes 2/2 in the same run. fmt and clippy (`-D warnings`) are clean, and the value assertion was mutation-checked: with the expected value deliberately wrong, the test fails on `unexpected pids.max` and nothing else.

## Related Issues

Refs #3605

## Additional Context

The zero-limit case reproduces #3644 exactly (creation fails with `failed to fork init process`); a confirmation with a runc 1.5.1 comparison is on that issue.


ARMeeru:
The one failing check is unrelated to this change: `k8s-tests-systemd-cgroup` (gnu and musl) failed before any test ran. The docker buildx step could not pull `docker/dockerfile:1.4` from Docker Hub (`failed to fetch oauth token ... connection reset by peer`), which is a transient registry error. The sibling `k8s-tests` jobs passed in the same run, and main's recent e2e runs are green, so re-running the two failed jobs should clear them.

Everything that runs the new tests passed, including `tests (x86_64, gnu)` and `tests (x86_64, musl)`, which exercise `cgroup_v2_pids`.

